PR #50, opened 2021-11-14
radiswap example: fix add_liquidity

kroggen:


iamyulong:
Thank you @kroggen for creating this PR.

The original `a_share` and `b_share` are the additional percentage the LP provider is contributing. 

For example, if the pool has `100` A token, `50` B token, and `666` LP token, the corresponding `B` and `LP` amount, when `x` A token is added, is calculated as follows:
```
Pool A: 100   + x
Pool B: 50    + 50 * x / 100
Pool LP: 666  + 666 * x / 100
```

If we change the calculation to what you've proposed in the PR, we need to update the LP token minting  as well.

That said, I do realise the original code has a bug - Line `80` should be `b_share`.

kroggen:
Indeed, I realized yesterday that the changes should update the LP token generation too.

But now I see that the original logic is correct. The `share` in the variable name confused me, because if the pool has 100 A tokens and the user adds more 100, then its "share" is 50% not 100%.

Maybe renaming the variable or adding some comments may make things clear

kroggen:
This is just a simple example, right? It does not handle when everyone remove their liquidity.
It would lead to a division by zero in `add_liquidity()`

Also, the use of floating point leads to rounding issues when calculating financial quantities. Using fractions would be a better approach  (multiplying by 0.333333333 is not the same as dividing by 3)

iamyulong:
100% agree that the variable names are confusing. I will go ahead and rename them. Cheers!

And yes, the example doesn't handle correctly the case where all liquidity is pulled out (after which no liquidity can be added).

We found that using fixed-point decimal (`10^-18`) made the code much simpler, but introduces small precision loss as you described.

kroggen:
Thanks for your feedback! I will close this one

kroggen:
@iamyulong 

Maybe some day Scrypto could use variables that represent fractions (rational numbers) with numerator and denominator, like [num-rational](https://github.com/rust-num/num-rational/blob/master/src/lib.rs)

In this way we could use the simple syntax (easier to understand) without concerning about rounding issues

I don't know if there is any smart contract platform that supports this, and as Radix is focused on DeFi it would be a good fit

In this case the radiswap code could be kept as it is, as long as the returned type in math operations would be the rational/fraction class, example:

```rust
let share = lp_tokens.amount() / self.lp_resource_def.supply();
```

In this case `share` would be of faction/rational class, and when used in multiplications or divisions, it would avoid losses

iamyulong:
Thanks for the suggestion @kroggen! Could you please move your suggestion to the "Issues" tab so it won't lose track.

PR #58, opened 2021-11-23
radiswap: Fix add_liquidity into empty pool bug

tesslerc:
A cleaner solution would be to define initial lp supply as a fixed variable and not an input to the init and when adding liquidity into an empty pool mint the on creation defined initial lp supply.

However, I assumed there may be a usecase where based on token supplies and to prevent overflow or other errors the user may want to define the intial ratio.

I am not sure one would need to update the LP based on the current updated tokens in the pool. If not, the logic can be removed from the swap method.


iamyulong:
Thank you @tesslerc for providing the fix for the case where all liquidity is pulled out! It looks very good to me.

PR #72, opened 2021-11-30
Add deposit_all_buckets... to new-package template

### LINKED ISSUE #71: new-package template missing drop_all_bucket_refs and deposit_all_buckets
I think the template should default to including these instructions for the `new()` function call, as it already does for the example `from_token` method call.

Usage of the `TransactionBuilder` functions  `drop_all_bucket_refs()` and `deposit_all_buckets(...)` is inconsistent  throughout the core code as well as all of the example code.  This is understandable I think but leads to confusing errors when developing new blueprints based off of the template.

In terms of the user experience if this is not fixed we have this example.   Create a new package, and modify the `new()` function to return `(Component, Bucket)` (and the other correct changes) but then the tests will fail with `Err(ResourceCheckFailure)` due to the returned `Bucket` still left "dangling" even though the blueprint is correct.  In effect, the test code is wrong because `Instruction::DepositAllBuckets` is missing.

If you make the same `new()` call using `resim` it works properly because it does both:

```
...
├─ DropAllBucketRefs
├─ DepositAllBuckets { account: ... }
...
```

Fix goes just before this line:  https://github.com/radixdlt/radixdlt-scrypto/blob/a553e2497a64531a70db828899520aaabc980b90/assets/template/tests/lib.rs#L17

And looks like this:

```diff
// Test the `new` function.
let transaction1 = TransactionBuilder::new(&executor)
    .call_function(package, "Hello", "new", vec![], None)
+   .drop_all_bucket_refs()
+   .deposit_all_buckets(account)
    .build(vec![key])
    .unwrap();
let receipt1 = executor.run(transaction1, false).unwrap();
println!("{:?}\n", receipt1);
assert!(receipt1.success);
```

I'm happy to submit a small PR for this if you'd like.

devmannic:
Fixes the `TransactionBuilder` to use both `drop_all_bucket_refs` and
`deposit_all_buckets(account)` in the `Hello::new` function call of
`test_hello` unit test.  This is now consistent for both transactions
in the unit test and matches how `resim call-function` works.

Fixes #71

### DIFF
--- a/assets/template/tests/lib.rs
+++ b/assets/template/tests/lib.rs
@@ -14,6 +14,8 @@ fn test_hello() {
     // Test the `new` function.
     let transaction1 = TransactionBuilder::new(&executor)
         .call_function(package, "Hello", "new", vec![], None)
+        .drop_all_bucket_refs()
+        .deposit_all_buckets(account)
         .build(vec![key])
         .unwrap();
     let receipt1 = executor.run(transaction1, false).unwrap();


iamyulong:
Thanks for providing the fix!

PR #75, opened 2021-12-01
Allow to specify XRD instead of 030000..004 address in resim

cbisaillon:
This might not be the best way to do it but I think it is useful to only have to specify "XRD" when sending buckets to functions and methods instead of the full address.

### DIFF
--- a/radix-engine/src/transaction/builder.rs
+++ b/radix-engine/src/transaction/builder.rs
@@ -451,7 +451,14 @@ impl<'a, A: AbiProvider> TransactionBuilder<'a, A> {
             SCRYPTO_NAME_BID | SCRYPTO_NAME_BUCKET => {
                 let mut split = arg.split(',');
                 let amount = split.next().and_then(|v| v.trim().parse::<Decimal>().ok());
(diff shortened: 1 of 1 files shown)


cbisaillon:
@iamyulong that would be complicated because the symbol is not necessarily unique ?

cbisaillon:
Other idea: When we run `cargo install --path ./simulator` setup environment variables like $XRD=0300.. (if possible)

iamyulong:
Yes, there might be duplicate symbols (less likely in a simulator used by one individual developer). In this case, the tooling should complain "sorry, I'm unable to resolve `xrd` because there are multiple resource definitions with the given symbol.".

Don't want to run into situation where we need to answer "why can we use `xrd` as resource address, but not `doge`?".

iamyulong:
I do like the idea that `resim` can persist environment variables (symbol to address mapping in this case), so we don't need to export the addresses again and again when opening a new terminal.

cbisaillon:
I'll close this PR but continue thinking about a way to achieve this. Maybe bring something like revup inside resim ?

PR #79, opened 2021-12-04
Use a closure in the auth macro for compatibility

### LINKED ISSUE #76: Return keyword caused BucketRef not dropping
**From @MrAir on Discord:**

---

Hey all, so I'm now in the process of going through the two hour scrypto video and I have a weird error that does not make sense to me.

I'm now at the point in the video where the team is talking about badges and how they can be made. For the withdraw method, I have written the following:
```
#[auth(admin_badge)]
pub fn withdraw(&mut self) -> Bucket {
  return self.xrd_vault.take_all();
}
```

For some reason, calling this method fails with the following error:
```
Transaction Status: FAILURE
Execution Time: 11 ms
Instructions:
├─ DeclareTempBucketRef
├─ CallMethod { component: 02526629b90e1142492e934fbe807b446935407064db3ea2fcf856, method: "withdraw", args: [1, 03c71ddd6b7967d6fd7d0e4a437c45f4ceac132d74d948a07af561] }
├─ BorrowFromContext { amount: 1, resource_def: 03c71ddd6b7967d6fd7d0e4a437c45f4ceac132d74d948a07af561, to: Rid(0) }
├─ CallMethod { component: 029849b6426bf644066da5b7dd8b0c119cf75ce225523046cb23d5, method: "withdraw", args: [Rid(0)] }
├─ DropAllBucketRefs
├─ DepositAllBuckets { account: 02526629b90e1142492e934fbe807b446935407064db3ea2fcf856 }
└─ End { signers: [04005feceb66ffc86f38d952786c6d696c79c2dbc239dd4e91b46729d73a27fb57e9] }
Results:
├─ Ok(None)
├─ Ok(Some(Bid(1)))
├─ Ok(None)
└─ Err(ResourceCheckFailure)
Logs: 0
New Entities: 0
Error: TransactionFailed
```

However, if I change this method to be the following:
```
#[auth(admin_badge)]
pub fn withdraw(&mut self) -> Bucket {
  self.xrd_vault.take_all()
}
```
then I now longer get an error when calling the withdraw method with the badge. 

Why is it that removing the explicit return statement fix this issue? Isn't `self.xrd_vault.take_all()` and return `self.xrd_vault.take_all();` equivalent?

devmannic:
Instead of generating a block, generate a closure with the
same body.  This way, the body can include a return statement.

Fixes #76

### DIFF
--- a/scrypto-derive/src/auth.rs
+++ b/scrypto-derive/src/auth.rs
@@ -107,9 +107,10 @@ mod tests {
                     if !(auth.contains(self.foo.clone()) || auth.contains(self.bar.clone())) {
                         panic!("Auth check failure")
                     }
-                    let output = {
+                    let func = || {
                         self.a
                     };
+                    let output = func();
                     auth.drop();
                     output
                 }
@@ -130,10 +131,11 @@ mod tests {
                     if !(auth.contains(self.foo.clone())) {
                         panic!("Auth check failure")
                     }
-                    let output = {
+                    let func = || {
                         auth.drop();
                         self.a
                     };
+                    let output = func();
                     output
                 }
             },
--- a/scrypto-derive/src/auth.rs
+++ b/scrypto-derive/src/auth.rs
@@ -68,7 +68,8 @@ pub fn handle_auth(attr: TokenStream, item: TokenStream) -> Result<TokenStream> 
                 panic!("Auth check failure")
             }
 
-            let output = #f_body;
+            let func = || #f_body ;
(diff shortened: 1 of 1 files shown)


devmannic:
I'll fix those unit tests to match the new implementation too.  Oops:

* [ ]      auth::tests::test_auth
* [ ]   auth::tests::test_keep_auth

iamyulong:
Thank you @devmannic for providing the fix!

PR #85, opened 2021-12-08
Fix auth macro failure when capturing &mut self

### LINKED ISSUE #84: auth macro fails on &mut self methods
My closure implementation in #79  to fix #76 is incorrect.  PR to fix on the way

Error:
```
cannot borrow as mutable
help: consider changing this to mutable: `mut func`
```

Thanks for raising this @cbisaillon

devmannic:
Don't use a let binding at all so the closure can be called regardless
of Fn or FnMut type.  Also drops the closure as soon as possible which
leads to better inlining and less chance of lifetime/borrow issues later

Fixes #84

### DIFF
--- a/scrypto-derive/src/auth.rs
+++ b/scrypto-derive/src/auth.rs
@@ -68,8 +68,7 @@ pub fn handle_auth(attr: TokenStream, item: TokenStream) -> Result<TokenStream> 
                 panic!("Auth check failure")
             }
 
-            let func = || #f_body ;
-            let output = func();
+            let output = (|| #f_body )();
             #drop
             output
         }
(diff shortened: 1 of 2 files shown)


croquelois:
Can you add a test for this usecase ?

devmannic:
Ok.  I had asked for where to put a test case for this in my PR, and asked in the previous issue but I think the question got overlooked.  It would seem the only place to test this that is not in an example is: in `scrypto-tests/tests/auth.rs`  -- I'll add some functionality there that covers both the return statement case and the *mut self call.  -- The fact the macro spits out the right thing is already covered in unit tests, and the "everything" tests don't seem to use the macros.  So this is my best guess at the right place. Will update the PR with another commit shortly

devmannic:
CI should pass.  I ran the relevant part of `tests.sh` locally:
```
(cd scrypto-tests; cargo test --no-default-features --features alloc)
```
Ran it both with my fix reverted and reproduced the reported  "cannot borrow as mutable" error, and then with the fix so the newly added test acts as a regression test.

iamyulong:
Thanks for providing this fix!

PR #88, opened 2021-12-09
Feature/mut umute

cseidman:
Initial branch creation. Added mutability for clarity on a few method calls


iamyulong:
This was merged by accident. I've manually reverted it and turned on branch protection on `develop` and `main`.

PR #95, opened 2021-12-12
Added an option to scrypto new-package to create minimal files

cbisaillon:
I find it annoying to have to remove all the Hello code when making examples. Am I lazy ? 😂

### DIFF
--- a/assets/template/src/lib-minimal.rs
+++ b/assets/template/src/lib-minimal.rs
@@ -1,0 +1,11 @@
+use scrypto::prelude::*;
+
+blueprint! {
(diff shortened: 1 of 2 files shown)


devmannic:
I like this idea.  I'm lazy too. 😀

But I think adding a minimal integration test is worth it since there's some boiler plate there

cbisaillon:
@devmannic Good idea

cbisaillon:
You are right. I think it is better to have an easy example as default for the new people coming to Scrypto. I don't see a way to make the current template less verbose while keeping it useful. I'll think about it.

PR #114, opened 2022-01-18
Feature/transaction manifest

iamyulong:
This PR introduces transaction manifest, which enables developers to compose arbitrary transaction in a human-friendly format.


radix-engine/src/transaction/executor.rs
```diff
@@ -160,48 +160,45 @@ impl<'l, L: Ledger> TransactionExecutor<'l, L> {
...
         );
         let mut proc = track.start_process(trace);
 
         let mut results = vec![];
         let mut success = true;
-        for inst in &transaction.instructions {
+        for inst in &check_tx.instructions {
             let res = match inst {
-                Instruction::DeclareTempBucket => {
+                CheckedInstruction::DeclareTempBucket => {
```
talekhinezh, line 179: Perhaps this should be `Instruction` and the current `Instruction` be called `UncheckedTransaction`

talekhinezh (approved):

radix-engine/src/transaction/executor.rs
iamyulong: `UncheckedTransaction` sounds a bit negative. I'd propose we keep them for now and if we really dislike it, we change them in a separate PR.

The background here is to introduce three phases of transaction processing:
1. Parsing
2. Static checking
3. Executing

`CheckedTransaction(CheckedInstruction)` is after phase 2.

PR #117, opened 2022-01-24
Feature/virtual badge

talekhinezh:
* Add ECDSA_TOKEN which represents transaction signers
* Add mechanism to map transaction signers to a virtual bucket ref
* Remove `get_transaction_signers` system call
* Update Account blueprint to check for ECDSA_TOKEN authorization


russellharvey (approved):

radix-engine/src/transaction/builder.rs
```diff
@@ -486,6 +487,7 @@ impl<'a, A: AbiProvider> TransactionBuilder<'a, A> {
                 args: vec![
                     SmartValue::from(ids.clone()),
                     SmartValue::from(*resource_address),
+                    SmartValue::from(Rid(0))
```
iamyulong, line 490: As we're reserving `Rid(0)` and `Bid(1)` for signature, we should update the `IDAllocator` employed by `TransactionBuilder` to start counting from `2`.

scrypto/src/types/address.rs
```diff
@@ -63,6 +67,19 @@ impl Address {
...
+    pub fn to_u128(&self) -> u128 {
+        let mut bytes: [u8; 16] = [0; 16];
+        match self {
+            Self::Package(d) => bytes[..].copy_from_slice(&d[..16]),
+            Self::Component(d) => bytes[..].copy_from_slice(&d[..16]),
+            Self::ResourceDef(d) => bytes[..].copy_from_slice(&d[..16]),
+            Self::PublicKey(d) => bytes[..].copy_from_slice(&d[..16]),
+        }
+        u128::from_be_bytes(bytes)
+    }
```
iamyulong, line 81: I think we should move this to the `Account` blueprint before we finalise the NFT ID schema.

iamyulong (commented):
Congrats on the first major PR! The solution is clear and natural.

iamyulong:
Also, looks like a wrong branch has been targeted and CI is failing.

assets/account/src/lib.rs
```diff
@@ -44,10 +44,8 @@ blueprint! {
         }
 
         /// Withdraws resource from this account.
-        pub fn withdraw(&mut self, amount: Decimal, resource_address: Address) -> Bucket {
-            if !Context::transaction_signers().contains(&self.key) {
-                panic!("Not authorized! Make sure you sign transaction with the correct keys.",)
-            }
+        pub fn withdraw(&mut self, amount: Decimal, resource_address: Address, account_auth: BucketRef) -> Bucket {
+            account_auth.check_nft_id(ECDSA_TOKEN, |id| id == &self.key.to_u128());
```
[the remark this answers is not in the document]
talekhinezh: Right, but in this case, the key is really the primary key. It shouldnt really be seen as data, especially as that would make the mapping to ids awkward (e.g. treat them all as id 0? or what to do if there is id collision).

radix-engine/src/transaction/builder.rs
talekhinezh: Ah yes, was wondering about this.

assets/account/src/lib.rs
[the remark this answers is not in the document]
iamyulong: That is indeed a bit awkward. If we were to follow that, I would use the signature offset as the ID (with no intention to maintain global uniqueness given it's a virtual resource) and operations on the signature bucket would be based on signature offset (like to take the 1st signature out and pass it some component).

scrypto/src/types/address.rs
devmannic: Can I state the obvious... The overall system security for any transaction that would ever use an Account  is now degraded by truncating the public key by half.  This is almost certainly a bad idea that I wont elaborate on here in this comment, but it's an unnecessarily weak choice.  There are ways to make this translation cryptographically more secure if we absolutely must do comparisons of only 16 bytes everywhere else, but that also just seems like the "easy" choice and not a well designed choice.  These are security relevant decision points and should have some discussion.  -- Overall I like the design of this PR, but because it is extremely relevant it needs a more through review.

talekhinezh: Yes, we have certainly not missed this point. The fix to this will be allowing NFT ids be of greater length so that one may use 20 byte hash (ala bitcoin/eth) or the whole compressed public key for verification. This will be in a soon upcoming PR.

devmannic: Ok great.  Thanks for clarifying. I'll keep an eye out for the update.

simulator/src/resim/cmd_publish.rs
```diff
@@ -75,6 +75,7 @@ pub fn handle_publish(matches: &ArgMatches) -> Result<(), Error> {
             TransactionExecutor::new(&mut ledger, configs.current_epoch, configs.nonce);
         let transaction = TransactionBuilder::new(&executor)
             .publish_package(&code)
+            .drop_all_bucket_refs()
```
iamyulong, line 78: This is an interesting side effect. If a transaction is signed but the virtual badge is never passed, users will be responsible for consuming the badge.

The good news is we're introducing auto dropping.

talekhinezh: Yes, I dont mind this actually as it prevents some level of transaction mutability and also helps express whether or not a transaction actually needs to be signed.

PR #121, opened 2022-01-27
Scry 38 scrypto objects to mut or not to mut

cseidman:
The first commit contains the changes I made to the base methods inside of vault, bucket, resource, and auth. The rest are changes made to address the expêcted compiler errors arising from changes made to the mutability of these methods


cseidman:
I was told there were some pending changes I need to look at .. I'll resend once I find them

PR #122, opened 2022-01-27
Scry 38 scrypto objects to mut or not to mut

cseidman:
The first commit contains the changes I made to the base methods inside of vault, bucket, resource, and auth. The rest are changes made to address the expêcted compiler errors arising from changes made to the mutability of these methods


assets/account/src/lib.rs
```diff
@@ -33,7 +33,7 @@ blueprint! {
         pub fn deposit(&mut self, bucket: Bucket) {
             let address = bucket.resource_address();
             match self.vaults.get(&address) {
-                Some(v) => {
+                Some(mut v) => {
```
iamyulong, line 36: This might slightly go beyond the scope of the original task, but I think it would be nice if we could take the chance to refactor the implementation of lazy map to be consistent to `HashMap` from std.

Specifically, `get()` should return an immutable reference to the value and `get_mut` should return a mutable reference.

Because of the current implementation, I've seen some bad coding pattern (get-update-insert) from myself and the community.

examples/core/gumball-machine/src/lib.rs
```diff
@@ -30,7 +30,7 @@ blueprint! {
             self.price
         }
 
-        pub fn buy_gumball(&mut self, payment: Bucket) -> (Bucket, Bucket) {
+        pub fn buy_gumball(&mut self, mut payment: Bucket) -> (Bucket, Bucket) {
```
iamyulong, line 33: This one is fun!

iamyulong (commented):
Nice work on introducing mutability enforcement!

iamyulong (approved):

PR #123, opened 2022-01-27
Feature/clean up data processing

iamyulong:
This PR cleans up data handling between RE and Scrypto by introducing `CheckedData`.

|               | Bucket | BucketRef | Vault             | LazyMap           |
|---------------|--------|-----------|-------------------|-------------------|
| CallData      | Move   | Move      | No Interpretation | No Interpretation |
| ComponentData | Reject | Reject    | No Interpretation | No Interpretation |
| NftData       | Reject | Reject    | Reject            | Reject            |
| MapEntryData  | Reject | Reject    | No Interpretation | No Interpretation |

* "No Interpretation" means the system does not limit referencing corresponding entity, like component data may contain a reference to a vault.


radix-engine/src/model/transaction.rs
```diff
@@ -0,0 +1,68 @@
...
+use scrypto::rust::vec::Vec;
+use scrypto::types::*;
+
+/// A transaction consists a sequence of instructions.
+#[derive(Debug, Clone, TypeId, Encode, Decode, PartialEq, Eq)]
+pub struct Transaction {
+    pub instructions: Vec<Instruction>,
+}
+
+/// Represents an instruction in transaction
```
talekhinezh, line 12: Add "unvalidated instruction"

radix-engine/src/transaction/executor.rs
```diff
@@ -156,85 +150,84 @@ impl<'l, L: Ledger> TransactionExecutor<'l, L> {
         &mut self,
         transaction: Transaction,
         trace: bool,
-    ) -> Result<Receipt, TransactionExecutionError> {
+    ) -> Result<Receipt, TransactionValidationError> {
```
talekhinezh, line 153: Seems to me that we should split the validation and the execution into two separate methods and have `run` accept `ValidatedTransaction`. This will be important when we need to pipeline transactions. For example, transaction validation can be parallelized and occur outside of the execution hotspots.

talekhinezh (approved):

PR #125, opened 2022-01-31
Enhance transaction model

iamyulong:
This PR introduces a few changes to the transaction model and fixes an issue with component state dumping.
- Add `TakeAllFromWorktop`
- Add `ReturnToWorktop`
- Add `AssertWorktopContains`
- Add `CreateBucketRef`, `CloneBucketRef` and `DropBucketRef`
- Update `CallMethodWithAllResources` to include every resource and automatically drop references
- Changed how transaction object IDs are derived to support static analysis

For the full specs on transaction manifest, please see https://radixdlt.atlassian.net/wiki/spaces/S/pages/2572484635/Transaction+Manifest+v0.3


radix-engine/src/engine/id_allocator.rs
```diff
@@ -1,99 +1,98 @@
...
+    OutOfID,
+}
 
 /// An ID allocator defines how identities are generated.
 pub struct IdAllocator {
-    count: u32,
-    rid_count: u32,
-    bid_count: u32,
+    next: u32,
+    end: u32,
```
talekhinezh, line 19: Specify exclusive either in name or as a comment

radix-engine/src/engine/id_allocator.rs
```diff
@@ -1,99 +1,98 @@
...
-const USER_BID_START: u32 = 1024;
+pub const ECDSA_TOKEN_RID: Rid = Rid(1);
+
+pub const SYSTEM_OBJECT_ID_RANGE: (u32, u32) = (0, 512);
+pub const TRANSACTION_OBJECT_ID_RANGE: (u32, u32) = (512, 1024);
+pub const USER_OBJECT_ID_RANGE: (u32, u32) = (1024, u32::MAX);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum IdAllocatorError {
+    OutOfID,
```
talekhinezh, line 13: `TooManyOpenIds` is perhaps a better error? I think we should be able to support id reuse at some point right?

radix-engine/src/engine/id_allocator.rs
```diff
@@ -1,99 +1,98 @@
...
 
-pub const ECDSA_TOKEN_RID: Rid = Rid(0);
 pub const ECDSA_TOKEN_BID: Bid = Bid(0);
-const USER_RID_START: u32 = 1024;
-const USER_BID_START: u32 = 1024;
+pub const ECDSA_TOKEN_RID: Rid = Rid(1);
+
+pub const SYSTEM_OBJECT_ID_RANGE: (u32, u32) = (0, 512);
+pub const TRANSACTION_OBJECT_ID_RANGE: (u32, u32) = (512, 1024);
+pub const USER_OBJECT_ID_RANGE: (u32, u32) = (1024, u32::MAX);
```
talekhinezh, line 9: Seems like this would be better implemented as an enum as there should really never be an instance where an IdAllocator is used with user defined range.

iamyulong: Good point. Will apply

radix-engine/src/engine/id_allocator.rs
iamyulong: I didn't consider reusing ID as they're "sufficiently" large per transaction, but we should be able to support it in future I think.

radix-engine/src/engine/id_allocator.rs
iamyulong: Yeah, I will fix it.

radix-engine/src/engine/id_allocator.rs
```diff
@@ -1,99 +1,107 @@
...
+
+pub const SYSTEM_ID_SPACE: IdSpace = IdSpace::System(0..512);
+pub const TRANSACTION_ID_SPACE: IdSpace = IdSpace::Transaction(512..1024);
+pub const APPLICATION_ID_SPACE: IdSpace = IdSpace::Application(1024..u32::MAX);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum IdSpace {
+    System(Range<u32>),
+    Transaction(Range<u32>),
+    Application(Range<u32>),
```
talekhinezh, line 16: Doesnt seem right to have `Range` in the enums as it is exposed to caller. Rather, better to just match below on line 33 and set the range there.

talekhinezh (approved):

PR #126, opened 2022-02-01
Variable Length Nft Keys

talekhinezh:


scrypto/src/buffer/constants.rs
```diff
@@ -6,6 +6,7 @@ pub const SCRYPTO_TYPE_BID: u8 = 0x84;
 pub const SCRYPTO_TYPE_RID: u8 = 0x85;
 pub const SCRYPTO_TYPE_MID: u8 = 0x86;
 pub const SCRYPTO_TYPE_VID: u8 = 0x87;
+pub const SCRYPTO_TYPE_NFT_KEY: u8 = 0x88;
```
iamyulong, line 9: There are a few more places we need to update to fully support this new type. I'd suggest you do a global search `SCRYPTO_TYPE_VID`  and add corresponding handling of NftKeys.

simulator/src/ledger/dumper.rs
```diff
@@ -119,16 +119,16 @@ fn dump_resources<T: Ledger>(vaults: &HashSet<Vid>, ledger: &T) -> Result<(), Di
...
                 let mutable_data = validate_data(&nft.mutable_data()).unwrap();
                 println!(
                     "{}  {} NFT {{ id: {}, immutable_data: {}, mutable_data: {} }}",
                     if last { " " } else { "│" },
                     list_item_prefix(inner_last),
-                    id,
+                    key,
                     immutable_data,
                     mutable_data
                 );
```
iamyulong, line 134: We need to update resim and transaction builder to handle passing of NFT buckets (ideally to support all NFT keys including `u128`).

assets/account/src/lib.rs
```diff
@@ -2,24 +2,24 @@ use scrypto::prelude::*;
 
 blueprint! {
     struct Account {
-        key: Address,
+        address: Address,
```
iamyulong, line 5: I'd suggest we rename this to `public_key`.

scrypto/src/types/nft_key.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub fn new(v: Vec<u8>) -> Self {
+        NftKey(v)
+    }
+}
+
+impl From<u128> for NftKey {
+    fn from(u: u128) -> Self {
+        NftKey(u.to_be_bytes().to_vec())
+    }
+}
```
iamyulong, line 22: I'd suggest we add a conversion for any type with `sbor::Encode` trait.

scrypto/src/buffer/constants.rs
iamyulong: May also need to update blueprint ABI to handle this type more elegantly.  For now, it's encoded and decoded as `Struct` which is not very convenient.

scrypto/src/types/nft_key.rs
talekhinezh: This interface should only support conversions to bytes array of bounded length. `sbor::Encode` would expose too much.

iamyulong (approved):

PR #127, opened 2022-02-01
Feature/flagscheck

gianlucardx:
Disallow misuse of flags while still leaving them as flags.

This PR creates a unique flag space of 64bit, with the lowest 32bit used by resource flags and the higher 32 bit used by the resource permissions.

This makes it easy for the engine to check whether the wrong flags have been supplied.


radix-engine/tests/integration.rs
```diff
@@ -190,6 +190,51 @@ fn test_resource_def() {
...
+            "create_fungible_wrong_resource_permissions_should_fail",
+            vec![],
+            Some(account),
+        )
+        .call_method_with_all_resources(account, "deposit_batch")
+        .build(vec![key])
+        .unwrap();
+    let receipt = executor.run(transaction).unwrap();
+    println!("{:?}", receipt);
+    assert!(!receipt.result.is_ok());
```
talekhinezh, line 237: Perhaps these should be three different tests

scrypto/src/resource/resource_flags.rs
```diff
@@ -1,23 +1,36 @@
+const RESOURCE_FLAGS_SHIFT: u8 = 0;
+const RESOURCE_FLAGS_MASK: u64 = 0xffff_ffff;
+
+macro_rules! resource_flags {
+    ( $f:expr ) => {
+        ($f as u64) << RESOURCE_FLAGS_SHIFT
+    }
+}
+
+pub fn resource_flags_are_valid(flags: u64) -> bool {
```
talekhinezh, line 10: perhaps should be inlined?

scrypto/src/resource/resource_permissions.rs
```diff
@@ -1,23 +1,36 @@
+const RESOURCE_PERMISSIONS_SHIFT: u8 = 32;
+const RESOURCE_PERMISSIONS_MASK: u64 = 0xfff_ffff_0000_0000;
+
+macro_rules! resource_permissions {
+    ( $f:expr ) => {
+        ($f as u64) << RESOURCE_PERMISSIONS_SHIFT
+    }
+}
+
+pub fn resource_permissions_are_valid(flags: u64) -> bool {
```
talekhinezh, line 10: Perhaps should be inlined?

radix-engine/tests/integration.rs
gianlucardx: I agree this test is becoming a bit long. OTOH I am just adding new tests for creating resources, which is what test_resource_def does.

scrypto/src/resource/resource_flags.rs
gianlucardx: Yes, agree. Simple inline hint would be good.

scrypto/src/resource/resource_permissions.rs
gianlucardx: Likewise.

scrypto/src/resource/resource_flags.rs
```diff
@@ -1,23 +1,37 @@
...
+
+macro_rules! resource_flags {
+    ( $f:expr ) => {
+        ($f as u64) << RESOURCE_FLAGS_SHIFT
+    }
+}
+
+#[inline]
+pub fn resource_flags_are_valid(flags: u64) -> bool {
+    (flags & RESOURCE_FLAGS_MASK) == flags
```
iamyulong, line 12: This flag check seems wrong to me. Should be `flags & !RESOURCE_FLAGS_MASK == 0`.

scrypto/src/resource/resource_permissions.rs
```diff
@@ -1,23 +1,37 @@
+const RESOURCE_PERMISSIONS_SHIFT: u8 = 32;
+const RESOURCE_PERMISSIONS_MASK: u64 = 0xfff_ffff_0000_0000;
```
iamyulong, line 2: Missed a `f` here.

scrypto/src/resource/resource_permissions.rs
```diff
@@ -1,23 +1,37 @@
...
+
+macro_rules! resource_permissions {
+    ( $f:expr ) => {
+        ($f as u64) << RESOURCE_PERMISSIONS_SHIFT
+    }
+}
+
+#[inline]
+pub fn resource_permissions_are_valid(flags: u64) -> bool {
+    (flags & RESOURCE_PERMISSIONS_MASK) == flags
```
iamyulong, line 12: Same

scrypto/src/resource/resource_flags.rs
iamyulong: Never mind. It's the same.

radix-engine/tests/integration.rs
iamyulong: I'd suggest we keep them as-is for now. At a not-too-far point in time, we need to split the `integration.rs` into multiple files, each containing a set of tests covering one aspect of the system.  This can be done as an individual task.

iamyulong (approved):

PR #128, opened 2022-02-02
Replace FileSystem Ledger with Rocks DB

talekhinezh:


radix-engine/src/engine/track.rs
```diff
@@ -40,7 +40,7 @@ pub struct Track<'l, L: Ledger> {
     code_cache: LruCache<Address, Module>, // TODO: move to ledger level
 }
 
-impl<'l, L: Ledger> Track<'l, L> {
+impl<'l, L: SubstateStore> Track<'l, L> {
     pub fn new(
         ledger: &'l mut L,
```
gianlucardx, line 45: Since we're renaming Ledger to SubstateStore, should the various L and ledger reference be removed?

gianlucardx (commented):
very minor comments

radix-engine/src/ledger/memory.rs
```diff
@@ -40,7 +40,7 @@ impl Default for InMemoryLedger {
     }
 }
 
-impl Ledger for InMemoryLedger {
+impl SubstateStore for InMemoryLedger {
```
iamyulong, line 43: I'd suggest we rename `InMemoryLedger` too.

simulator/src/ledger/radix_engine_db.rs
```diff
@@ -0,0 +1,109 @@
...
+use radix_engine::model::*;
+use scrypto::types::*;
+
+pub struct RadixEngineDB {
+    db: DBWithThreadMode<SingleThreaded>,
+}
+
+impl RadixEngineDB {
+    pub fn new(root: PathBuf) -> Self {
+        let db = DB::open_default(root.as_path()).unwrap();
```
iamyulong, line 14: I wonder what happens if the folder already exists but not a database. Most developers will likely have this setup. Need to advise them how to migrate.

talekhinezh: Default settings creates a new database if empty

iamyulong (approved):

PR #133, opened 2022-02-03
Macro to simplify Decimal creation

r001:
Created the dec! macro to simplify Decimal creation.
```rust
dec!(5).to_string() = "5"
dec!(1.23).to_string() = won't compile
dec!("1.23").to_string() = "1.23"
dec!(123, -2).to_string() = "1.23"
dec!(123, 2).to_string() = "12300"
dec!(170_141_183_460_469_231_732, 0) = won't compile
dec!(1, 4_294_967_296) = panic!("Overflow of arg2.")
```

### DIFF
--- a/scrypto/src/types/decimal.rs
+++ b/scrypto/src/types/decimal.rs
@@ -108,6 +108,44 @@ from_int!(i64);
 from_int!(i128);
 from_int!(isize);
 
+impl From<&str> for Decimal {
+    fn from(val: &str) -> Self {
+        Self::from_str(&val).unwrap()
+    }
+}
+
+impl From<String> for Decimal {
+    fn from(val: String) -> Self {
+        Self::from_str(&val).unwrap()
+    }
+}
+
+impl From<bool> for Decimal {
+    fn from(val: bool) -> Self {
+        if val {
+            Self::from(1)
+        } else {
+            Self::from(0)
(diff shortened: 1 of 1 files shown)


scrypto/src/types/decimal.rs
```diff
@@ -99,6 +99,46 @@ from_int!(i64);
...
+       Decimal::from($x) 
+    };
+    
+    ($int:literal, $exponent:literal) => {
+        if ($exponent) < 0 {
+            Decimal(($int) * PRECISION)
+                .div(10i128.pow((-1i128 * ($exponent)) as u32))
+        } else {
+            Decimal(($int) * PRECISION)
+                .mul(10i128.pow((1i128 * ($exponent)) as u32))
```
iamyulong, line 137: `as u32` will silence underflow/overflow instead of panic. I'd suggest we force a check before doing the calculation and add a test case to cover this edge case.

iamyulong (commented):
Good work! Left one comment below. Please also update the PR description :)

scrypto/src/types/decimal.rs
r001: @iamyulong I have updated the code according to your remarks. Please review again.

iamyulong (approved):

PR #137, opened 2022-02-04
Remove PublicKey from Address

talekhinezh:


radix-engine/src/transaction/builder.rs
```diff
@@ -18,6 +18,47 @@ use crate::engine::*;
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+#[cfg(not(feature = "alloc"))]
+impl std::error::Error for ParsePublicKeyError {}
+
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct PublicKey(pub EcdsaPublicKey);
```
iamyulong, line 38: I see you have an intention to support generic signature schema, but I think we have the right abstraction here. 

I'd suggest we either:
- Fix `PublicKey` to ECDSA and use `PublicKey` everywhere without conversion
- Introduce trait `PublicKey` and ECDSA implementation `EcdsaPublicKey`, and make every function that takes any public key as input  use `<T: PublicKey>`.

iamyulong (approved):
LGTM

PR #143, opened 2022-02-08
Rename kernel to radix engine

iamyulong:
Two sets of terms have been used across the codebase:
1. Kernel and user
2. Radix Engine and Scrypto

This PR renames kernel to Radix Engine (or engine if there is no ambiguity) to align code with the products.

Also formatted the code by running `./format.sh`.


gianlucardx (approved):
LGTM, thanks for reformatting my changes. :)

PR #144, opened 2022-02-08
Feature/finer grained stack frame

iamyulong:
This PR associates an actor, either a blueprint or a component, to each call stack frame. This allows finer-grained access control based on the running entity.

From Scrypto, we can retrieve actor info via the `Context::actor()` API.

In addition, we now require both `BURNABLE` and `FREELY_BURNABLE` to burn resources freely.

_Note that after merging this PR, no runtime object auth check will be in place. We will have follow-up PRs to address this._


scrypto/src/engine/api.rs
```diff
@@ -117,6 +117,8 @@ pub const GET_CURRENT_EPOCH: u32 = 0xf3;
 pub const GET_TRANSACTION_HASH: u32 = 0xf4;
 /// Generate an UUID
 pub const GENERATE_UUID: u32 = 0xf5;
+/// Retrieve running entitty
```
gianlucardx, line 120: typo.

gianlucardx (approved):
LGTM, although it's a bit over my head currently on architectural changes.

PR #145, opened 2022-02-08
Tracked Vaults and LazyMaps

talekhinezh:


radix-engine/src/engine/process.rs
```diff
@@ -983,8 +1018,13 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
             .get_component(input.component_address)
             .ok_or(RuntimeError::ComponentNotFound(input.component_address))?;
 
+        let state = component.state();
+        let updating_component_data = Self::process_component_data(state).unwrap();
+        let current = self.updating_components.insert(input.component_address, updating_component_data);
+        assert!(current.is_none());
```
iamyulong, line 1024: I think we should throw a runtime error here. Can't make the assumption that Scrypto code (WASM) will never call `get_component_state`  twice. The `updating_components` is a really good addition, with that we can enforce additional constraint.

iamyulong (commented):
Really good implementation!

radix-engine/src/engine/process.rs
talekhinezh: good point

iamyulong (approved):

PR #149, opened 2022-02-10
Add an example for a DNS on Radix

backyard-coder:


examples/core/radix-name-service/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+
+            (component, admin_badge)
+        }
+
+        /// Lookup the address for a given `name`.
+        /// Panics if that name is not registered.
+        pub fn lookup_address(&self, name: String) -> Address {
+            let hash = Self::hash_name(name);
+            let name_data: DomainName = self.name_resource.get_nft_data(hash);
+            name_data.address
```
iamyulong, line 70: This will benefit from the to-be-implemented NFT existence check feature.

examples/core/radix-name-service/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+use sha2::{Digest, Sha256};
+
+#[derive(NftData)]
+struct DomainName {
+    #[scrypto(mutable)]
+    address: Address,
+    #[scrypto(mutable)]
+    last_valid_epoch: u64,
+    #[scrypto(mutable)]
+    deposit_amount: Decimal,
```
iamyulong, line 11: Is mutable `deposit_amount` intentional? I didn't find anywhere updating this critical field.

backyard-coder: No it isn't. I was originally going to increase the deposit amount on name renewal, but then decided to make it a non-refundable fee.

PR #151, opened 2022-02-12
Refactor process state

talekhinezh:


radix-engine/src/engine/component_objects.rs
```diff
@@ -0,0 +1,154 @@
...
+                .remove(&mid)
+                .ok_or(RuntimeError::LazyMapNotFound(mid))?;
+            lazy_maps.insert(mid, lazy_map);
+        }
+
+        Ok(ComponentObjectsSet { vaults, lazy_maps })
+    }
+
+    pub fn insert_objects_into_map(&mut self, new_objects: ComponentObjectsSet, mid: &Mid) {
+        let unclaimed_map = self.lazy_maps.get_mut(mid).unwrap();
```
iamyulong, line 118: RE crashes here when there is a cyclic reference graph in user space, as the root is taken from the process.

Test code:
```rust
use scrypto::prelude::*;
use scrypto::engine::*;

blueprint! {
    struct Hello {
        maps: LazyMap<u32, LazyMap<u32, u32>>
    }

    impl Hello {
        pub fn new() -> Component {
            let map0 = LazyMap::new();
            let map1 = LazyMap::new();
            map0.insert(1u32, map1);

            let input = PutLazyMapEntryInput {
                mid: Mid(Context::transaction_hash(), 1025),
                key: scrypto_encode(&0u32),
                value: scrypto_encode(&Mid(Context::transaction_hash(), 1024)),
            };
            let _: PutLazyMapEntryOutput = call_engine(PUT_LAZY_MAP_ENTRY, input);
            
            Hello {
                maps: map0
            }.instantiate()
        }
    }
}
```

radix-engine/src/engine/process.rs
```diff
@@ -51,38 +51,103 @@ macro_rules! re_warn {
...
+enum InterpreterState {
+    Blueprint,
+    ComponentEmpty {
+        component_address: Address,
+    },
+    ComponentLoaded {
         component_address: Address,
-        component_data: ValidatedData,
+        initial_loaded_object_refs: ComponentObjectsSetRef,
+        additional_object_refs: ComponentObjectsSetRef,
```
iamyulong, line 80: I see the direction you're moving towards. Component objects are grouped into a component, but access pattern must follow the reference chain. 

But I think this might deserve a proper design documentation before implementation, as this introduces a slightly more complex system. It would cover 
- How objects **within a component** are stored and accessed (complexity for read and write - which impacts metering)?
- How the reference structure is inferred?
- What assumptions have been made in user space (what if it's not Rust)?

iamyulong (commented):
Good job modularizing object tracking logic!

iamyulong (approved):
LGTM!

PR #155, opened 2022-02-15
Changed constructor function names

cbisaillon:
- Renamed all constructor functions so that devs don't think they have to call them `new`.
- Some tests were not updated for v0.3
- Fixed synthetics examples. It failed when the user had 0 debt and tried to unstake


iamyulong (approved):
LGTM. Thanks for cleaning up the examples. Will merge as soon as CI passes.

PR #156, opened 2022-02-15
Feature/clean up scrypto types

iamyulong:
- Cleaned up Scrypto types encoding and decoding in both binary and text forms (see [confluence page](https://radixdlt.atlassian.net/wiki/spaces/S/pages/2625896460/Scrypto+Types) for summary);
- Organized types into modules;
- Replaced `Address` with `PackageId`, `ComponentId` and `ResourceDefId`;
- Updated blueprint ABI with simplified names;
- Refactored custom types validation and formatting;
- Removed support of `LazyMap` and `Vault` from manifest;
- Renamed `BucketRef` to `Proof`.


assets/account/src/lib.rs
```diff
@@ -3,21 +3,21 @@ use scrypto::prelude::*;
 blueprint! {
     struct Account {
         public_key: EcdsaPublicKey,
-        vaults: LazyMap<Address, Vault>,
+        vaults: LazyMap<ResourceDefRef, Vault>,
     }
 
     impl Account {
-        pub fn new(public_key: EcdsaPublicKey) -> Component {
+        pub fn new(public_key: EcdsaPublicKey) -> ComponentRef {
```
talekhinezh, line 10: ComponentRef on the other hand feels entirely natural

examples/core/cross-blueprint-call/README.md
```diff
@@ -9,15 +9,15 @@ In this example, we demonstrate two ways of calling a function or method defined
...
+    resim export-abi <PACKAGE_REF> Airdrop
     ```
 2. Import the ABI into our package, e.g.,
     ```rust
     import! {
     r#"
     {
-        "package": "01bda8686d6c2fa45dce04fac71a09b54efbc8028c23aac74bc00e",
-        "name": "Airdrop",
+        "package_ref": "01bda8686d6c2fa45dce04fac71a09b54efbc8028c23aac74bc00e",
```
talekhinezh, line 19: hrm, feel like this should also just be package, package_ref feels weird, as its not a reference at this point but an actual value.

.github/workflows/ci.yml
```diff
@@ -93,40 +93,22 @@ jobs:
       run: cargo build --target wasm32-unknown-unknown --release
       working-directory: assets/system
     - name: Build hello-nft
```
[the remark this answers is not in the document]
iamyulong: Yes, will do!

README.md
```diff
@@ -63,23 +63,23 @@ scrypto test
...
-| List all entities in simulator     | ``` resim show-ledger  ```                                                       |
-| Reset simulator state              | ``` resim reset ```                                                              |
+| Action                             | Command                                                                                           |
+|------------------------------------|---------------------------------------------------------------------------------------------------|
+| Create an account                  | ``` resim new-account ```                                                                         |
+| Change the default account         | ``` resim set-default-account <account_component_ref> <account_public_key>```                     |
+| Create a token with fixed supply   | ``` resim new-token-fixed <amount> ```                                                            |
+| Create a token with mutable supply | ``` resim new-token-mutable <minter_resource_def_ref> ```                                         |
```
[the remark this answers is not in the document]
iamyulong: That's a unfortunately long name. ;-)

assets/account/src/lib.rs
```diff
@@ -3,21 +3,21 @@ use scrypto::prelude::*;
 blueprint! {
     struct Account {
         public_key: EcdsaPublicKey,
-        vaults: LazyMap<Address, Vault>,
+        vaults: LazyMap<ResourceDefRef, Vault>,
```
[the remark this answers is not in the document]
iamyulong: This dates back to our original discussion about **resource** and **resource definition**. I'm worried that `ResourceRef` will cause even more confusion.

I agree that the **`ResourceDefRef`** is a bit overwhelming, but if we can come up with _a simpler name for resource definition_, it will start to make sense.

examples/core/cross-blueprint-call/README.md
iamyulong: This right hand side is a package address, not a package.

Now I view `XxxxRef` in two ways:
- It's a reference to the substate object, so I can invoke it.
- It's an identifier of the substate object, so I can pass it around.

Another approach is to need two types for the two use cases, like before we have `Component` and `Address`.

iamyulong: Or some types like `PackageId` (identifer) and `PackageRef` (reference)

iamyulong: To make it even worse is mutable semantics. Ideally, there should be `PackageRef` and `PackageMutRef`, unless we have `Package`, `&Package` and `&mut Package`.

examples/defi/auto-lend/tests/lib.rs
```diff
@@ -6,28 +6,30 @@ use auto_lend::User;
 
 struct TestEnv<'a, L: SubstateStore> {
     executor: TransactionExecutor<'a, L>,
-    key: EcdsaPublicKey,
-    account: Address,
-    usd: Address,
-    lending_pool: Address,
+    public_key: EcdsaPublicKey,
+    account: ComponentId,
+    usd: ResourceDefId,
```
talekhinezh, line 11: `ResourceId` feels much better

talekhinezh (approved):

examples/defi/auto-lend/tests/lib.rs
iamyulong: Haha, I do like `ResourceId`. Let's revisit the terminologies in LA; we have a dedicated session for that :)

PR #157, opened 2022-02-16
Externalise dec and bdec macros

iamyulong:
This is to make `dec!` and `bdec!` work both inside the crate and outside.


gianlucardx (approved):
Probably a comment on why tests were removed would be useful.

PR #163, opened 2022-02-20
Preparatory Refactor for Substate Store

talekhinezh:


radix-engine/src/engine/track.rs
```diff
@@ -1,6 +1,10 @@
 use lru::LruCache;
+use sbor::Decode;
+use sbor::Encode;
+use scrypto::buffer::scrypto_decode;
 use scrypto::constants::*;
 use scrypto::engine::types::*;
+use scrypto::prelude::scrypto_encode;
```
iamyulong, line 7: Minor: can we use `scrypto::buffer::scrypto_encode`?  The prelude is indented for developers.

radix-engine/src/engine/track.rs
```diff
@@ -147,13 +142,46 @@ impl<'s, S: SubstateStore> Track<'s, S> {
         }
     }
 
+    fn get_substate<A: Encode, T: Decode>(&self, address: &A) -> Option<T> {
+        self.substate_store
+            .get_substate(address)
+            .and_then(|v| scrypto_decode(&v).map(|r| Some(r)).unwrap_or(None))
```
iamyulong, line 148: The `unwrap_or(None)` looks dangerous to me. Should panic instead of swallowing this illegal state. There are a few more places like this.

radix-engine/src/engine/track.rs
```diff
@@ -441,50 +449,68 @@ impl<'s, S: SubstateStore> Track<'s, S> {
...
-            self.ledger
-                .put_package(package_id, self.packages.get(&package_id).unwrap().clone());
+        let package_ids: Vec<PackageId> = self
+            .packages
+            .iter()
+            .map(|(address, _)| address.clone())
+            .collect();
+        for package_id in package_ids {
+            let package = self.packages.remove(&package_id).unwrap();
+            self.put_substate(&package_id, &package);
```
iamyulong, line 461: Minor: There must be a better way of doing with. Otherwise, I would be disappointed with Rust. ;-)

radix-engine/src/ledger/memory.rs
```diff
@@ -1,33 +1,24 @@
-use scrypto::buffer::*;
-use scrypto::engine::types::*;
+use sbor::Encode;
+use scrypto::prelude::scrypto_encode;
```
iamyulong, line 2: Minor: `scrypto::buffer`

radix-engine/src/ledger/traits.rs
```diff
@@ -33,110 +33,78 @@ pub trait QueryableSubstateStore {
...
-        resource_def_id: ResourceDefId,
-        key: &NonFungibleKey,
-    ) -> Option<NonFungible>;
-    fn put_non_fungible(
-        &mut self,
-        resource_def_id: ResourceDefId,
-        key: &NonFungibleKey,
-        non_fungible: NonFungible,
-    );
+    fn get_substate<T: Encode>(&self, address: &T) -> Option<Vec<u8>>;
```
iamyulong, line 36: Maybe we can introduce a trait for the three "address" types?

radix-engine/src/ledger/traits.rs
```diff
@@ -33,110 +33,78 @@ pub trait QueryableSubstateStore {
...
+    fn get_child_substate<T: Encode>(&self, address: &T, key: &[u8]) -> Option<Vec<u8>>;
+
+    fn put_substate<T: Encode>(&mut self, address: &T, substate: &[u8]);
+    fn put_child_substate<T: Encode>(&mut self, address: &T, key: &[u8], substate: &[u8]);
 
     fn bootstrap(&mut self) {
-        if self.get_package(SYSTEM_PACKAGE).is_none() {
+        let package: Option<Package> = self
+            .get_substate(&SYSTEM_PACKAGE)
+            .map(|v| scrypto_decode(&v).unwrap());
```
iamyulong, line 45: `scrypto_decode` and `scrypto_encode` everywhere are really annoying me. Can we have **default methods** in the `SubstateStore` trait that accepts rust structs and automatically decodes/encodes?

iamyulong (commented):
Nice work!

radix-engine/src/engine/track.rs
talekhinezh: Ah yes, this made more sense when there was only a single type of Address and wasn't sure the type you'd get back.

radix-engine/src/engine/track.rs
talekhinezh: A quick search from me didn't show any good way :) Iteration + Removal always tricky

radix-engine/src/ledger/traits.rs
talekhinezh: Would prefer not to overdesign this part now since it'll be subject to a lot of refactoring soon.

radix-engine/src/ledger/traits.rs
talekhinezh: This works for now.

radix-engine/src/ledger/traits.rs
iamyulong: Sounds good!

radix-engine/src/ledger/traits.rs
iamyulong: Agreed. This is why I proposed an independent layer (on top of substates) earlier. Didn't want to bring in too much extra work into this PR, so proposed default methods.

iamyulong (approved):
LGTM

PR #166, opened 2022-02-21
Add compile package macro

iamyulong:
- Added `compile_package` macro
- Renamed `include_code` to `include_package`


gianlucardx (approved):
What is the reason behind the different naming convention between path ( using '-') and packages (using '_')?

Other than that, LGTM.

PR #167, opened 2022-02-22
Add Unique top level substates

talekhinezh:


iamyulong (commented):
From my iPhone.

1. Are we always spinning down a substate if only READ is needed?

2. Could we not introduce manual encoding/decoding for struct? Suggest using SBOR drives.

iamyulong (approved):
Awesome! The code is nice and clean.

PR #168, opened 2022-02-24
Added a payment-splitter example

0xOmarA:
I’ve added a payment splitter example which demonstrates how profits / funds acquired may be shared across different shareholders in a project. This example comes with a number of example radix transaction manifest (rtm) files that users can use to test out this blueprint.


iamyulong (approved):
Good example!

PR #172, opened 2022-03-07
Add auth worktop

iamyulong:
Auth worktop is the place that collects proofs returned by other components/blueprints.

All proofs on auth worktop are used for determining eligibility of accessing components and resource managers by system auth checker.

**New Instructions:**
- `TAKE_FROM_AUTH_WORKTOP`
- `PUT_ON_AUTH_WORKTOP`


radix-engine/src/engine/id_validator.rs
```diff
@@ -15,13 +15,13 @@ pub enum IdValidatorError {
 pub struct IdValidator {
     id_allocator: IdAllocator,
     bucket_ids: HashMap<BucketId, usize>,
-    proof_ids: HashMap<ProofId, BucketId>,
+    proof_ids: HashMap<ProofId, Option<BucketId>>,
```
talekhinezh, line 18: Unclear what Option is for. Is this to distinguish between BucketProofs and VaultProofs?

talekhinezh: Maybe use enum instead of Option. Would make my brain hurt less :)

talekhinezh (approved):

radix-engine/src/engine/id_validator.rs
iamyulong: For proofs received at runtime (vault proofs) and virtual proofs, the bucket id is unknown.

Enum is good idea. Will implement it to make it explicit.

PR #178, opened 2022-03-12
Feature/refactor host errors

talekhinezh:


radix-engine/src/engine/component_objects.rs
```diff
@@ -87,13 +87,13 @@ impl ComponentObjectRefs {
         // Only allow vaults to be added, never removed
         for vault_id in &other.vault_ids {
             if !self.vault_ids.remove(&vault_id) {
-                return Err(RuntimeError::VaultRemoved(vault_id.clone()));
+                return Err(RuntimeError::VaultRemoved());
```
iamyulong, line 90: Why are we removing the arguments?

radix-engine/src/engine/process.rs
```diff
@@ -573,7 +573,13 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
         let result = module.invoke_export(invocation.export_name.as_str(), &[], self);
         re_debug!(self, "Invoke result: {:?}", result);
         let rtn = result
-            .map_err(RuntimeError::InvokeError)?
+            .map_err(|e| {
+                match e.into_host_error() {
+                    // Pass-through runtime errors
+                    Some(host_error) => *host_error.downcast::<RuntimeError>().unwrap(),
+                    None => RuntimeError::InvokeError(),
```
iamyulong, line 580: Does this handle recursion properly, i.e. call and call and trap?

radix-engine/src/errors.rs
```diff
@@ -61,10 +61,10 @@ pub enum RuntimeError {
     IdAllocatorError(IdAllocatorError),
 
     /// Error when invoking an export.
-    InvokeError(Error),
+    InvokeError(),
```
iamyulong, line 64: Can we remove the `()` if there is no arguments? This makes mapping of error much easier.

talekhinezh: Yup good point

radix-engine/src/engine/component_objects.rs
iamyulong: Ah, I see. I'd suggest we keep the arguments. We're only going to add more in the long run. They provide useful information for both debugging and testing.  The `matches!` macro can help with loose comparison (although I'd advocate full comparison for testing).

iamyulong (approved):
Thanks for improving our error reporting!

PR #180, opened 2022-03-14
Feature/generify resource container

iamyulong:
This PR adds the data structure of a generic resource container.

The container state isn't updated based on proof generation yet. Every proof accounts for one ref to the resource container, for the entire liquidity.

No new tests have been added. I will add tests to cover the full proof system, once vault proof and worktop proof are in place.

### DIFF
--- a/radix-engine/src/engine/process.rs
+++ b/radix-engine/src/engine/process.rs
@@ -1613,12 +1613,12 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
         Ok(CreateEmptyVaultOutput { vault_id })
     }
 
-    fn get_local_vault(&mut self, vault_id: VaultId) -> Result<&mut Vault, RuntimeError> {
+    fn get_local_vault(&mut self, vault_id: &VaultId) -> Result<&mut Vault, RuntimeError> {
         let wasm_process = self
             .wasm_process_state
             .as_mut()
             .ok_or(RuntimeError::IllegalSystemCall)?;
-        match wasm_process.process_owned_objects.get_vault_mut(&vault_id) {
+        match wasm_process.process_owned_objects.get_vault_mut(vault_id) {
             Some(vault) => Ok(vault),
             None => match &wasm_process.interpreter_state {
                 InterpreterState::Component {
@@ -1627,15 +1627,15 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
                     additional_object_refs,
                     ..
                 } => {
-                    if !initial_loaded_object_refs.vault_ids.contains(&vault_id)
-                        && !additional_object_refs.vault_ids.contains(&vault_id)
+                    if !initial_loaded_object_refs.vault_ids.contains(vault_id)
+                        && !additional_object_refs.vault_ids.contains(vault_id)
                     {
-                        return Err(RuntimeError::VaultNotFound(vault_id));
+                        return Err(RuntimeError::VaultNotFound(*vault_id));
                     }
-                    let vault = self.track.get_vault_mut(component_id, &vault_id);
+                    let vault = self.track.get_vault_mut(component_id, vault_id);
                     Ok(vault)
                 }
-                _ => Err(RuntimeError::VaultNotFound(vault_id)),
+                _ => Err(RuntimeError::VaultNotFound(*vault_id)),
             },
         }
     }
@@ -1651,7 +1651,7 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
             .remove(&input.bucket_id)
             .ok_or(RuntimeError::BucketNotFound(input.bucket_id))?;
 
-        self.get_local_vault(input.vault_id)?
+        self.get_local_vault(&input.vault_id)?
             .put(bucket)
             .map_err(RuntimeError::VaultError)?;
 
@@ -1660,7 +1660,7 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
 
     fn check_take_from_vault_auth(
         &mut self,
-        vault_id: VaultId,
+        vault_id: &VaultId,
         badge: Option<ResourceDefId>,
     ) -> Result<(), RuntimeError> {
         let resource_def_id = self.get_local_vault(vault_id)?.resource_def_id();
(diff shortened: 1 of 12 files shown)


radix-engine/src/model/resource.rs
```diff
@@ -0,0 +1,282 @@
...
+#[derive(Debug, Clone)]
+pub struct Proof {
+    /// The resource definition id
+    resource_def_id: ResourceDefId,
+    /// The resource type
+    resource_type: ResourceType,
+    /// Restricted proof can't be moved down along the call stack (growing down).
+    restricted: bool,
+    /// The total amount locked
+    total_amount: ResourceAmount,
```
talekhinezh, line 82: Is the `total_amount` here the sum of all the values in `amounts`? If so, should add a comment that this is optimization to get total_amount faster.

radix-engine/src/model/resource.rs
```diff
@@ -0,0 +1,282 @@
...
+    InvalidAmount(Decimal, u8),
+    /// The balance is not enough
+    InsufficientBalance,
+    /// The operation is not supported
+    UnsupportedOperation,
+}
+
+/// Represents the contained resource.
+#[derive(Debug, Clone, TypeId, Encode, Decode)]
+pub enum Resource {
```
talekhinezh, line 24: Perhaps description should be something like: "Represents the locked/liquid state of the contained resource". Perhaps also change the name from `Resource` to `ResourceContainerState`. `Resource` is a bit too generic.

radix-engine/src/model/resource.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        // add the other bucket into liquid pool
+        match (&mut self.resource, other.liquid_amount()) {
+            (Resource::Fungible { liquid_amount, .. }, ResourceAmount::Fungible { amount }) => {
+                *liquid_amount = *liquid_amount + amount;
+            }
+            (Resource::NonFungible { liquid_ids, .. }, ResourceAmount::NonFungible { ids }) => {
+                liquid_ids.extend(ids);
+            }
+            _ => panic!("Resource type should match!"),
```
talekhinezh, line 150: This is awesome matching logic :)

radix-engine/src/model/resource.rs
```diff
@@ -0,0 +1,282 @@
...
+        resource: Resource,
+    ) -> Self {
+        Self {
+            resource_def_id,
+            resource_type,
+            resource,
+        }
+    }
+
+    pub fn put(&mut self, other: Self) -> Result<(), ResourceError> {
```
talekhinezh, line 133: It feels like this is where rules like you can't put a vault in a bucket should be.

radix-engine/src/model/resource.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    pub fn non_fungible_ids(&self) -> Result<BTreeSet<NonFungibleId>, ResourceError> {
+        match self {
+            ResourceAmount::Fungible { .. } => Err(ResourceError::UnsupportedOperation),
+            ResourceAmount::NonFungible { ids } => Ok(ids.clone()),
+        }
+    }
+}
+
+impl ResourceContainer {
```
talekhinezh, line 120: Not sure if this is the right abstraction here as I think the way vaults and buckets implement `take` would be quite different I think

talekhinezh: Perhaps this would be better as a trait?

radix-engine/src/engine/process.rs
```diff
@@ -1756,8 +1694,11 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
     ) -> Result<GetNonFungibleIdsInVaultOutput, RuntimeError> {
         let vault = self.get_local_vault(input.vault_id)?;
         let non_fungible_ids = vault
-            .get_non_fungible_ids()
-            .map_err(RuntimeError::VaultError)?;
+            .liquid_amount()
+            .non_fungible_ids()
+            .map_err(|_| RuntimeError::UnsupportedOperation)?
```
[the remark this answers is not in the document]
iamyulong: This is indeed a good idea. Let me explore what options I have.

radix-engine/src/engine/process.rs
```diff
@@ -1908,8 +1835,11 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
 
         Ok(GetNonFungibleIdsInBucketOutput {
             non_fungible_ids: bucket
-                .get_non_fungible_ids()
-                .map_err(RuntimeError::BucketError)?,
+                .liquid_amount()
+                .non_fungible_ids()
+                .map_err(|_| RuntimeError::UnsupportedOperation)?
```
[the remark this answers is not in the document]
iamyulong: Yes

radix-engine/src/model/resource.rs
iamyulong: Yes, this is purely for optimization. It can be derived from the hashmap below. I will add comments.

radix-engine/src/model/resource.rs
iamyulong: Like the proposed name. Will change

sbor/src/type_id.rs
```diff
@@ -147,6 +148,13 @@ impl<T: TypeId> TypeId for Box<T> {
     }
 }
 
+impl<T: TypeId> TypeId for Rc<T> {
```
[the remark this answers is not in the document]
iamyulong: Yes, encoding and decoding of Rc and Arc may introduce multiple copies of the same data.  This is okay for our system, since only vaults are eventually persisted (and only one copy exists).

This is also consistent with other serialization framework, like Serde. 

A better approach would be to make it a feature, so external users of SBOR have the option to turn it off.

radix-engine/src/model/resource.rs
iamyulong: Haha, I've been struggling to find an alternative...

radix-engine/src/model/resource.rs
iamyulong: Such rules are currently implemented by the actual containers, buckets and vaults. Will take this into consideration.

radix-engine/src/model/resource.rs
iamyulong: The main difference I see is authorization and liquid amount representation. The abstraction here is to ensure the same accounting logic used by all containers. Make it a trait sounds a bit off to me.

A better solution is to have `ResourceContainer<T>` where `T` is any struct with trait for offering liquidity.

radix-engine/src/engine/process.rs
```diff
@@ -2011,9 +1927,11 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
 
         Ok(GetNonFungibleIdsInProofOutput {
             non_fungible_ids: proof
-                .bucket()
-                .get_non_fungible_ids()
-                .map_err(RuntimeError::BucketError)?,
+                .total_amount()
+                .non_fungible_ids()
+                .map_err(|_| RuntimeError::UnsupportedOperation)?
```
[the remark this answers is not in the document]
iamyulong: Yes

radix-engine/src/model/resource.rs
talekhinezh: ah makes sense

talekhinezh (approved):

PR #185, opened 2022-03-15
Cleanup Unit Tests

talekhinezh:


radix-engine/tests/test_runner.rs
```diff
@@ -0,0 +1,84 @@
...
+use radix_engine::model::Receipt;
+use radix_engine::model::Transaction;
+
+pub struct TestRunner<'l> {
+    executor: TransactionExecutor<'l, InMemorySubstateStore>,
+}
+
+impl<'l> TestRunner<'l> {
+    pub fn new(ledger: &'l mut InMemorySubstateStore) -> Self {
+        let executor = TransactionExecutor::new(ledger, true);
```
iamyulong, line 13: Minor: I'd suggest we make `TestRunner` own an `InMemorySubstateStore`, and allow the constructor to take a parameter `trace` (I personally found it useful when writing tests).

iamyulong (approved):
LGTM

iamyulong:
Looks like `rustfmt` is having problem with resolving modules under `tests`. Script `./format.sh`  stops working.

iamyulong:
A temporary workaround would be
```
#[rustfmt::skip]
mod test_runner;
```

radix-engine/tests/test_runner.rs
talekhinezh: Was looking to do this initially but then ran into awkwardness with certain borrowed structures. Will probably fix this in a later refactor.

PR #188, opened 2022-03-17
Replace pure RC with dedicated accounting

iamyulong:
This PR replaces RC-based proof system with accounting-based proof system.

This builds the foundation for generating partial/full/composite proofs on any combination of resource containers.


radix-engine/src/engine/process.rs
```diff
@@ -1239,31 +1261,37 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
-                Ok(ResourceContainerState::fungible(
-                    amount,
+                Ok(ResourceContainer::new_fungible(
+                    resource_def_id,
                     resource_def.resource_type().divisibility(),
+                    amount,
                 ))
             }
             MintParams::NonFungible { entries } => {
+                if !matches!(
```
talekhinezh, line 1286: Feels like this type checking should be done outside of `match mint_params` so that it's clear the check is made. Could we do this like:
```
impl MintParams {
    pub fn is_same_type(&self, resource_type: &ResourceType) -> bool {
        match self {
            Self::Fungible { .. } => matches!(resource_type, ResourceType::Fungible { .. }),
            Self::NonFungible { .. } => matches!(resource_type, ResourceType::NonFungible),
        }
    } 
 }
 ```

File: radix-engine/src/model/proof.rs [hunk 2]
```diff
@@ -0,0 +1,154 @@
...
+        Self {
+            resource_def_id: self.resource_def_id,
+            resource_type: self.resource_type,
+            restricted: self.restricted,
+            locked_in_total: self.locked_in_total.clone(),
+            locked_in_details: self.locked_in_details.clone(),
+        }
+    }
+
+    pub fn settle(self) {
```
talekhinezh, line 114: drop?

radix-engine/src/model/vault.rs
```diff
@@ -56,39 +49,44 @@ impl Vault {
...
-    pub fn borrow_container(&mut self) -> Result<&mut ResourceContainer, VaultError> {
-        Ok(Rc::get_mut(&mut self.container).ok_or(VaultError::VaultLocked)?)
+    pub fn borrow_container(&self) -> Ref<ResourceContainer> {
+        self.container.borrow()
     }
 
-    /// Takes the ownership of the container
-    pub fn take_container(self) -> Result<ResourceContainer, VaultError> {
-        Ok(Rc::try_unwrap(self.container).map_err(|_| VaultError::VaultLocked)?)
+    pub fn borrow_container_mut(&mut self) -> RefMut<ResourceContainer> {
```
talekhinezh, line 85: Shouldn't be `pub`

radix-engine/src/model/vault.rs
```diff
@@ -56,39 +49,44 @@ impl Vault {
...
+    }
+
+    pub fn is_locked(&self) -> bool {
+        self.borrow_container().is_locked()
     }
 
-    /// Creates a mutable reference to the container
-    pub fn borrow_container(&mut self) -> Result<&mut ResourceContainer, VaultError> {
-        Ok(Rc::get_mut(&mut self.container).ok_or(VaultError::VaultLocked)?)
+    pub fn borrow_container(&self) -> Ref<ResourceContainer> {
```
talekhinezh, line 81: Shouldn't be `pub`

radix-engine/src/model/bucket.rs
```diff
@@ -69,38 +60,51 @@ impl Bucket {
...
+
+    pub fn is_locked(&self) -> bool {
+        self.borrow_container().is_locked()
+    }
+
+    pub fn borrow_container(&self) -> Ref<ResourceContainer> {
+        self.container.borrow()
+    }
+
+    pub fn borrow_container_mut(&mut self) -> RefMut<ResourceContainer> {
```
talekhinezh, line 97: Shouldn't be `pub`

radix-engine/src/model/bucket.rs
```diff
@@ -69,38 +60,51 @@ impl Bucket {
...
+        self.borrow_container()
+            .total_ids()
+            .map_err(BucketError::ResourceContainerError)
+    }
+
+    pub fn is_locked(&self) -> bool {
+        self.borrow_container().is_locked()
+    }
+
+    pub fn borrow_container(&self) -> Ref<ResourceContainer> {
```
talekhinezh, line 93: Shouldn't be `pub`

radix-engine/src/model/vault.rs
```diff
@@ -56,39 +49,44 @@ impl Vault {
...
+        self.borrow_container().total_amount()
+    }
+
+    pub fn total_ids(&self) -> Result<BTreeSet<NonFungibleId>, VaultError> {
+        self.borrow_container()
+            .total_ids()
+            .map_err(VaultError::ResourceContainerError)
+    }
+
+    pub fn is_locked(&self) -> bool {
```
talekhinezh, line 77: Thoughts on making these methods implemented in a default Trait?

sbor/src/type_id.rs
```diff
@@ -155,6 +156,13 @@ impl<T: TypeId> TypeId for Rc<T> {
     }
 }
 
+impl<T: TypeId> TypeId for RefCell<T> {
```
talekhinezh, line 159: When is this needed for sbor?

radix-engine/src/engine/process.rs
iamyulong: Yes, will apply

radix-engine/src/model/proof.rs
iamyulong: Definitely

sbor/src/type_id.rs
iamyulong: This is for `Rc<RefCell<ResourceContainer>>` in the vault.

radix-engine/src/model/vault.rs
iamyulong: You mean a trait for these getters that forward request to the "resource container"?

I'm not sure how this is possible if I want to keep composition, i.e. bucket/vault/worktop contains one or more resource containers (implementation). This essentially is a good example of the "delegate call" idea.

An alternative approach is remove all these getters and redirect clients to use the `borrow_container` directly.

What do you think?

radix-engine/src/model/vault.rs
iamyulong: I was debating whether I should expose these "proxy" getter methods or a reference to the row resource container. 

As commented in another thread, if we decide to remove these getters, we will have to expose the borrow methods; if not, I'm totally happy to make them private.

radix-engine/src/engine/process.rs
```diff
@@ -208,36 +208,66 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
-        let bucket = match resource_spec {
-            ResourceSpecification::Fungible { amount, .. } => {
-                self.worktop.take(amount, resource_def_id)
-            }
-            ResourceSpecification::NonFungible { ids, .. } => {
-                self.worktop.take_non_fungibles(&ids, resource_def_id)
+        re_debug!(self, "(Transaction) Taking from worktop: {:?}", resource);
+        let new_bucket_id = self.new_bucket_id()?;
+        let bucket = match resource {
+            ResourceSpecifier::Amount(amount, resource_def_id) => self
```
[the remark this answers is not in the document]
iamyulong: Let me try it.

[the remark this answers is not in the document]
iamyulong: Actually, I will just remove this match logic and expose three methods. There is no point merging three transaction instructions (`take_from_worktop`, `take_non_fungibles_from_worktop` and `take_all_from_worktop`) into one and splitting it in the `Process`.

This will further reduce `ResourceSpecifier` to resim/tx_builder, to be aligned with our decision on amount types.

radix-engine/src/model/vault.rs
talekhinezh: More of the latter since a whole subset of functionality is essentially the same. But, I don't think it's too big of a deal for now.

talekhinezh (approved):

PR #189, opened 2022-03-18
Feature/vault proof

iamyulong:
This PR enables the creation of vault proof (full amount or IDs).


radix-engine/tests/proof.rs
```diff
@@ -0,0 +1,160 @@
+#[rustfmt::skip]
+pub mod test_runner;
+
+use crate::test_runner::TestRunner;
+use radix_engine::ledger::InMemorySubstateStore;
+use scrypto::prelude::*;
+
+fn instantiate_vault_proof_component(
```
talekhinezh, line 8: Just move this into TestRunner :)

radix-engine/tests/proof/src/bucket_proof.rs
```diff
@@ -0,0 +1,28 @@
...
+            assert_eq!(proof.amount(), amount);
+            assert_eq!(clone.amount(), amount);
+
+            clone.drop();
+            proof.drop();
+            bucket
+        }
+
+        pub fn use_bucket_proof_for_auth(bucket: Bucket, to_burn: Bucket) -> Bucket {
+            bucket.authorize(|| {
```
talekhinezh, line 21: We also have the resource system function `authorize()`. We should come to agreement on best interface for this.

radix-engine/tests/proof/src/vault_proof.rs
```diff
@@ -0,0 +1,34 @@
...
+            assert_eq!(self.vault.amount(), amount);
+            assert_eq!(proof.amount(), amount);
+            assert_eq!(clone.amount(), amount);
+
+            clone.drop();
+            proof.drop();
+        }
+
+        pub fn use_vault_proof_for_auth(&self, to_burn: Bucket) {
+            self.vault.authorize(|| {
```
talekhinezh, line 29: We also have the resource system function authorize(). We should come to agreement on best interface for this.

radix-engine/tests/proof.rs
iamyulong: Yes, I will try.

radix-engine/tests/proof/src/bucket_proof.rs
talekhinezh: the latter with overloading is possible as a macro, which I think is the right option. I prefer the latter as:
1. It continues the pattern that objects don't understand where they are located
2. The borrowing mechanism allows programmer to continue to refer to the bucket in their logic without doing anything too awkward

iamyulong: With the macro approach, how you imagine the macro would be and what overloading methods will be required? There are unbounded number of argument type combinations, so I think we will end up with additional recursive data structure.

For 1, I didn't understand what you mean. You're saying we want to educate developers that proofs are generated by resource system, provided references to containers. From a developer's perspective, it's clear to me that I'm using _this_ bucket/vault to authorise some actions.

For 2, `authorize(&self, xxx)` follows the same borrowing mechanism. In fact, I think it's identical to `func(bucket: &Bucket, xxx)` and the only difference is about namespacing.

talekhinezh: With macro approach I'd use macro overload where recursion is done in preprocessing step: https://doc.rust-lang.org/rust-by-example/macros/overload.html

Re:1 No, I mean more of the fact that we've taken the model that in general, container objects don't know about their location. For example we don't have the following interfaces:
```
bucket.send_to_component(c);
bucket.put_into_vault(v);
```

Re:2 This is true but imo it's more of an awkward interface to need to pass references as there's the indirection programmer would need to know:
```
bucket.authorize(|b| {
  b.total_amount() ...
});
```

vs.

```
authorize!(&bucket, || {
  bucket.total_amount() ...
});
```

iamyulong: I don't think `authorize(&self)` is making any assumption about the location though. To me, it's updating it's own state and producing a proof.

`send_to_component` is a different concept, as it requires `self`, not `&self`.  In `Bucket`, we have
```
pub fn burn(self);

// User can write burn in two syntax if they wish; both are passing ownership
some_bucket.burn();
Bucket::burn(some_bucket);
```

I feel like like there is some misunderstanding about `authorize(&self, F)`. The example should be
```
bucket.authorize(|| {
  bucket.total_amount() ...
});
```

iamyulong: The `authorize!` macro is making borrowing explicit by having that `&` symbol.

iamyulong: I still think the two approaches are identical in terms of borrowing. The only difference is about namespacing - whether it's a method as part of a resource container, or an independent function that operates on resource container references.

talekhinezh: ah yes, you're right in terms of borrowing and I see the point about location assumptin. Let's go with your way then :)

talekhinezh (approved):

PR #199, opened 2022-03-23
Scry 101 parallelize tests

r001:
Now you can build and test using `cargo` from the root folder, or use the updated `test.sh`. Tests are executed in a parallell fashion.
 To further speed-up tests the tracing should be disabled, so that decoding does not flood the screen with data.


iamyulong:
Let's close this PR for now. Will definitely revisit the CI time issue in not too distant future.

r001:
Sounds good to me. Yesterday I tried to do a bash solution: multiple
test processes starting, once one returns failure, kill all others. Does
not work right. :(
The 04/08/2022 03:21, Yulong Wu wrote:
> Let's close this PR for now. Will definitely revisit the CI time issue in not too distant future.
> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/radixdlt/radixdlt-scrypto/pull/199#issuecomment-1092710171
> You are receiving this because you authored the thread.
> 
> Message ID: ***@***.***>

-- 
Robert Horvath

PR #202, opened 2022-03-23
Feature/composite proof

iamyulong:
This PR allows developers to create composite from proofs in auth zone.


radix-engine/src/model/proof.rs
```diff
@@ -33,103 +45,314 @@ pub enum ProofError {
...
+            Self::Fungible {
+                resource_def_id,
+                restricted,
+                total_amount,
+                sources,
+            } => {
+                for (container, amount) in sources.values() {
+                    container
+                        .borrow_mut()
+                        .lock_amount(amount.clone())
```
talekhinezh, line 265: State update in `clone()` is something worth mentioning in comments.

radix-engine/src/model/proof.rs
```diff
@@ -33,103 +45,314 @@ pub enum ProofError {
...
+    }
+
+    pub fn compose_by_amount(
+        proofs: &[Proof],
+        total_amount: Option<Decimal>,
+        resource_def_id: ResourceDefId,
+        resource_type: ResourceType,
+    ) -> Result<Proof, ProofError> {
+        // check resource type
+        if matches!(resource_type, ResourceType::NonFungible) {
```
talekhinezh, line 115: Is this check temporary? I feel like we still want to be able to compose amount proofs for non fungibles where the user doesn't care about the set of NonFungibleIds but only needs a specific amount.

talekhinezh (approved):

radix-engine/src/engine/process.rs
```diff
@@ -461,18 +465,87 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
-            .create_proof_by_ids(ids)
+            .create_proof_by_ids(ids, ProofSourceId::Vault(vault_id))
             .map_err(RuntimeError::ProofError)?;
         self.proofs.insert(new_proof_id, new_proof);
 
         Ok(new_proof_id)
     }
 
+    // Creates a auth zone proof.
+    pub fn create_auth_zone_proof(
```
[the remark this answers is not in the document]
iamyulong: Yes, will add comment.

radix-engine/src/model/proof.rs
iamyulong: My original idea was not to support creating amount proofs from fungible resources, but realised that "amount" is just a way of expressing user intention, rather than fungibility. 

Will add support for this, the same way as in `take`  - check granularity and pick non-fungibles based on ID order.

radix-engine/src/model/proof.rs
iamyulong: Yes, will do

PR #206, opened 2022-03-25
Named fields in sbor path

talekhinezh:


radix-engine/src/model/component.rs
```diff
@@ -29,19 +36,177 @@ impl Component {
...
+                            CustomType::ResourceDefId => HardProofRuleResourceList::List(
+                                values
+                                    .iter()
+                                    .map(|v| {
+                                        if let Value::Custom(_, bytes) = v {
+                                            return ResourceDefId::try_from(bytes.as_slice())
+                                                .unwrap()
+                                                .into();
+                                        }
+                                        panic!("Unexpected type");
```
iamyulong, line 71: This is one of the assumptions that haven't been well documented. When decoding **ANY** value from a byte array, the decoder won't validate the "bytes" of custom types.

Here, we will have to handle the errors or to reuse `CustomValueValidator` with some refactoring.

radix-engine/src/model/method_authorization.rs
```diff
@@ -2,105 +2,164 @@ use crate::errors::RuntimeError;
...
+            HardProofRule::AnyOf(resource_list) => match resource_list {
+                HardProofRuleResourceList::SoftResourceListNotFound => Err(NotAuthorized),
+                HardProofRuleResourceList::List(resources) => {
+                    for resource in resources {
+                        if resource.check(proofs_vector) {
                             return Ok(());
                         }
                     }
+
+                    Err(NotAuthorized)
```
iamyulong, line 125: We might have discussed during team meeting, but it's a bit hard for me to accept `AllOf([]) == Ok` and `AnyOf([]) == NotAuthorized` without reading the implementation.

Do you think it's worth to add a "at least one" constraint on the list, for both static and dynamic ones?

scrypto/src/resource/schema_path.rs
```diff
@@ -0,0 +1,78 @@
...
+        self
+    }
+
+    pub fn to_sbor_path(&self, schema: &Type) -> Option<SborPath> {
+        let length = self.0.len();
+        let mut cur_type = schema;
+        let mut sbor_path: Vec<usize> = vec![];
+
+        for i in 0..length {
+            match self.0.get(i).unwrap() {
```
iamyulong, line 40: Minor: suggest `for element in self.0`

iamyulong (commented):
Looks very good to me, with some minor comments.

scrypto/src/resource/schema_path.rs
talekhinezh: good catch

radix-engine/src/model/method_authorization.rs
talekhinezh: Sure, I can replace `AnyOf` with `AtleastOneOf`. Is that what you mean?

radix-engine/src/model/component.rs
talekhinezh: I think I'm getting you but I'm assuming that we will validate the bytes of custom types in a future PR?

radix-engine/src/model/method_authorization.rs
iamyulong: Sorry, I was trying to avoid the empty set condition by issuing a `SoftResourceNotFound` if the resource list is empty. 

Now I'm thinking `AllOf` as logical `for all` and `AnyOf` as logical `there exists`. 
- The formula  `\forall {x}{\in}\emptyset P(x)}` is always true, regardless of the predicate P(x)
- The formula `\exists {x}{\in}\emptyset P(x)}` is always false, regardless of the predicate P(x)

The current implementation starts to make sense. Let's keep it as is.

radix-engine/src/model/component.rs
iamyulong: When do validate the bytes, when using `ValidateData` validator. 

This reminds me that the `dom: &Value` is the component state, which has already been validated by Radix Engine. We should be good unless we want to get rid of such assumption.

iamyulong (approved):

radix-engine/src/model/method_authorization.rs
talekhinezh: yup exactly

PR #211, opened 2022-03-30
Refactor read of return value

talekhinezh:
Cleaned up processing of return value:
* Removed call to `scrypto_free` which is unnecessary since process will get garbage collected.
* Removed memcopy of returned value and instead process the value directly from the buffer.

### DIFF
--- a/radix-engine/src/engine/process.rs
+++ b/radix-engine/src/engine/process.rs
@@ -1235,8 +1235,7 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
             return Err(RuntimeError::MemoryAccessError);
         }
 
-        ValidatedData::from_slice(&buffer[range])
-            .map_err(RuntimeError::DataValidationError)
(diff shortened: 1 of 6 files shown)


radix-engine/src/engine/process.rs
```diff
@@ -1215,36 +1214,24 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
-            .map_err(|_| RuntimeError::MemoryAccessError)?;
+        let start = ptr.checked_add(4).ok_or(RuntimeError::MemoryAccessError)?;
+        let end = start
+            .checked_add(len)
+            .ok_or(RuntimeError::MemoryAccessError)?;
+        let range = start as usize..end as usize;
+        let direct = wasm_process.vm.memory.direct_access();
 
-        Ok(data.to_vec())
+        ValidatedData::from_slice(&direct.as_ref()[range])
```
iamyulong, line 1233: Would this panic if range is invalid?

iamyulong (approved):

PR #215, opened 2022-03-31
Cleanup auth frontend

talekhinezh:
Some refactoring of the auth frontend:
* Add `globalize()` for explicit movement of a component into global space
* Add `LocalComponent` for simple auth setup
* Update of ProofRule macro names


iamyulong:
Unfortunately, I'm going to register some dislike about the changes to component instantiation process.

To me, `instantiate` and `globalize` are two steps (the second step may be done automatically when an instantiated component reaches transaction). The syntax gives developer the false impression that they're globalizing a strut. 

Additionally, the `LocalComponent` should be abstracted by radix engine and Scrypto owns a handle to it.

I would suggest something like:
- `struct { }.instantiate(NO_AUTH).globalize()`
- `struct { }.instantiate(AUTH_RULES).globalize()`

where `instantiate()` returns a `Component` and `globalize()` returns a `ComponentAddress`. 

Since we don't have `Component` YET, we can temporarily have `instantiate()` returns an address and `globalize()` is just another method of `ComponentAddress` that returns itself.

iamyulong:
Sounds good. Do you think if we should make component instantiation and auth setup atomic.

talekhinezh:
@iamyulong Not necessarily as we want users to be able to set their own auth. We do this by passing around these "LocalComponents" and then allowing one to set auth on it.

iamyulong (approved):
Looks good!

PR #224, opened 2022-04-05
Resource auth refactor

talekhinezh:


assets/system/src/lib.rs
```diff
@@ -16,17 +16,13 @@ blueprint! {
         pub fn new_resource(
             resource_type: ResourceType,
             metadata: HashMap<String, String>,
-            flags: u64,
-            mutable_flags: u64,
-            authorities: HashMap<ResourceAddress, u64>,
+            authorization: ComponentAuthorization,
```
iamyulong, line 19: Can we rename `ComponentAuthorization`? :)

assets/template/src/lib.rs
```diff
@@ -15,6 +15,7 @@ blueprint! {
             let my_bucket: Bucket = ResourceBuilder::new_fungible()
                 .metadata("name", "HelloToken")
                 .metadata("symbol", "HT")
+                .auth("take_from_vault", auth!(allow_all))
```
iamyulong, line 18: Feel like having an enum for all the resource methods would be nice, since they're finite and predefined.

radix-engine/src/ledger/traits.rs
```diff
@@ -170,9 +170,9 @@ pub trait SubstateStore {
             let mut xrd = ResourceManager::new(
                 ResourceType::Fungible { divisibility: 18 },
                 metadata,
-                0,
-                0,
-                HashMap::new(),
+                component_authorization! {
```
iamyulong, line 173: Maybe more generic macro name?

radix-engine/src/model/auth_converter.rs
```diff
@@ -0,0 +1,231 @@
...
+
+fn soft_to_hard_decimal(schema: &Type, soft_decimal: &SoftDecimal, dom: &Value) -> HardDecimal {
+    match soft_decimal {
+        SoftDecimal::Static(amount) => HardDecimal::Amount(amount.clone()),
+        SoftDecimal::Dynamic(schema_path) => {
+            let sbor_path = schema_path.to_sbor_path(schema);
+            if let None = sbor_path {
+                return HardDecimal::SoftDecimalNotFound;
+            }
+            match sbor_path.unwrap().get_from_value(dom) {
```
iamyulong, line 23: minor: `if let Some(dom)` can avoid a unwrap.

radix-engine/src/model/auth_converter.rs
```diff
@@ -0,0 +1,231 @@
...
+    match soft_count {
+        SoftCount::Static(count) => HardCount::Count(count.clone()),
+        SoftCount::Dynamic(schema_path) => {
+            let sbor_path = schema_path.to_sbor_path(schema);
+            if let None = sbor_path {
+                return HardCount::SoftCountNotFound;
+            }
+            match sbor_path.unwrap().get_from_value(dom) {
+                Some(Value::U8(count)) => HardCount::Count(count.clone()),
+                _ => HardCount::SoftCountNotFound,
```
iamyulong, line 46: Minor: wonder if it's worth to have a dedicated enum for type mismatch.

radix-engine/src/model/resource_manager.rs
```diff
@@ -1,209 +1,102 @@
...
-            "update_non_fungible_mutable_data".to_string(),
-            MethodState::new(IsSet(INDIVIDUAL_METADATA_MUTABLE), AlwaysTrue),
-        );
+        if let Some(take_auth) = auth.get("take_from_vault") {
+            authorization.insert(
+                "take_from_vault".to_string(),
+                convert(&Type::Unit, &Value::Unit, take_auth),
+            );
+        } else {
+            return Err(ResourceManagerError::TakeFromVaultNotDefined);
```
iamyulong, line 63: Do we really need this check? This looks strange as we're taking a generic approach but paternalistic about this specific method. Any thoughts?

iamyulong (commented):
Left some minor comments but overall looks very good.

radix-engine/tests/resource/src/lib.rs
```diff
@@ -22,59 +24,60 @@ blueprint! {
...
+                .auth("burn", auth!(require(badge.resource_address())))
                 .no_initial_supply();
             (badge, token_address)
         }
 
-        pub fn query() -> (Bucket, HashMap<String, String>, u64, u64, Decimal) {
-            let (badge, resource_address) = Self::create_fungible();
-            let resource_manager = resource_manager!(resource_address);
+        pub fn query() -> (Bucket, HashMap<String, String>, Decimal) {
+            let (badge, resource_def_id) = Self::create_fungible();
```
iamyulong, line 76: Can we rename these `resource_def_id`? :)

radix-engine/src/model/auth_converter.rs
talekhinezh: Is rust smart enough to know that the negation of that has already been handled though?

assets/template/src/lib.rs
talekhinezh: I'm actually thinking of taking this a step further and implement something like:
```
.take_auth(auth!(allow_all))
.mint_auth(auth!(require()))
```
Thoughts?

radix-engine/src/model/resource_manager.rs
talekhinezh: Agreed it's weird, maybe we move the paternalism to the builder side while keeping system open?

radix-engine/tests/resource/src/lib.rs
talekhinezh: oops

assets/template/src/lib.rs
iamyulong: I personally like `auth()` more. We will likely reuse the enum in other places, like `update_auth()`. Don't want to have too many `update_xxx_auth()` methods in the library.

radix-engine/src/model/auth_converter.rs
iamyulong: It's not

radix-engine/src/model/resource_manager.rs
iamyulong: Yeah, would prefer that.

radix-engine/src/model/auth_converter.rs
iamyulong: Sounds good

iamyulong (approved):

PR #225, opened 2022-04-06
Feature/add some signature algo

iamyulong:
This PR adds an actual signature algorithm (`secp256r1` for now) into Radix Engine.

Also, added some refactoring to the transaction builder.


radix-engine/src/transaction/validator.rs
```diff
@@ -8,15 +8,23 @@ use crate::errors::*;
 use crate::model::*;
 
 pub fn validate_transaction(
-    transaction: &Transaction,
+    transaction: &SignedTransaction,
```
talekhinezh, line 11: Minor: Seems like this can just be moved into SignedTransaction?

scrypto/src/crypto/ecdsa.rs
```diff
@@ -161,21 +176,23 @@ impl TryFrom<&[u8]> for EcdsaPrivateKey {
...
 
 impl EcdsaPrivateKey {
     pub fn to_vec(&self) -> Vec<u8> {
-        self.0.to_vec()
+        self.0.to_be_bytes().as_slice().to_vec()
     }
 }
 
-// Private key is not a custom type, as we don't expect it to be passed around
+// Temporarily for simulator
```
talekhinezh, line 194: Why needed? Should have larger alarm bells in comment such as `FIXME`

scrypto/src/types.rs
```diff
@@ -68,13 +69,14 @@ pub enum ScryptoType {
...
 // Need to update `scrypto-derive/src/import.rs` after changing the table below
-const MAPPING: [(ScryptoType, u8, &str); 13] = [
+const MAPPING: [(ScryptoType, u8, &str); 14] = [
     (ScryptoType::PackageAddress, 0x80, "PackageAddress"),
     (ScryptoType::ComponentAddress, 0x81, "ComponentAddress"),
     (ScryptoType::LazyMap, 0x82, "LazyMap"),
     (ScryptoType::Hash, 0x90, "Hash"),
     (ScryptoType::EcdsaPublicKey, 0x91, "EcdsaPublicKey"),
-    (ScryptoType::EcdsaSignature, 0x92, "EcdsaSignature"),
+    (ScryptoType::EcdsaPrivateKey, 0x92, "EcdsaPrivateKey"),
```
talekhinezh, line 78: Can we not include private key here? If not possible, let's add a `FIXME` comment that should be removed at some point.

talekhinezh:
Looks good overall, just want clarification that `EcdsaPrivateKey` into scrypto/sbor is only a temporary measure.

radix-engine/src/transaction/validator.rs
iamyulong: Yeah, let me move it there

scrypto/src/crypto/ecdsa.rs
iamyulong: The simulator needs a way to store the private key and it's using sbor.

I agree that it doesn't have to be in the custom type list, but I don't think this should be FIXME. It would be nice to have our library not encouraging clients to store private key. It's ultimately their responsibility to keep private key safely (they will find an alternative way to encode if sbor doesn't support it).

I will see if I can remove it.

scrypto/src/types.rs
iamyulong: Removed

scrypto/src/crypto/ecdsa.rs
talekhinezh: I think private key serialization/deserialization would flag any auditor immediately (we were flagged for this before). And from security perspective, we want to not have any private key handling in radix-engine\scrypto since we want to ensure no logic ever requires storing or signing with private keys.

iamyulong: Feel like the auditor is being lazy - don't want to check how they're being used. 

All crypto library I've seen supports `SEC1 ASN.1 DER` encoding and decoding.

scrypto/src/crypto/ecdsa.rs
```diff
@@ -237,7 +251,7 @@ impl FromStr for EcdsaPrivateKey {
 
 impl fmt::Display for EcdsaPrivateKey {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        write!(f, "{}", hex::encode(self.0))
+        write!(f, "{}", hex::encode(self.to_vec()))
```
talekhinezh, line 254: Could we also remove this? :)

scrypto/src/crypto/ecdsa.rs
```diff
@@ -1,60 +1,63 @@
...
 
-pub const ECDSA_PRIVATE_KEY_LENGTH: usize = 32;
-pub const ECDSA_PUBLIC_KEY_LENGTH: usize = 33;
-pub const ECDSA_SIGNATURE_LENGTH: usize = 65;
-
 /// Represents an ECDSA private key.
-#[derive(Clone, Copy, PartialEq, Eq, TypeId, Encode, Decode)]
-pub struct EcdsaPrivateKey(pub [u8; ECDSA_PRIVATE_KEY_LENGTH]);
+#[derive(Clone, PartialEq, Eq)]
+pub struct EcdsaPrivateKey(SecretKey);
```
talekhinezh, line 17: Do we really need `Clone`? I'm just being extra paranoid about not having secret keys everywhere.

scrypto/src/crypto/ecdsa.rs
iamyulong: What's the problem with this?

scrypto/src/crypto/ecdsa.rs
iamyulong: What's the issue? Even the `SecretKey` from 3rd-party supports Clone.

scrypto/src/crypto/ecdsa.rs
talekhinezh: Printing private keys?

scrypto/src/crypto/ecdsa.rs
talekhinezh: Possibly harder to track where private key has been cloned/moved. But this is minor.

scrypto/src/crypto/ecdsa.rs
iamyulong: I feel like were are being paternalistic.

If they want to print private keys, they know what they're doing (we dump private key to screen in simulator; Metamask displays private key on screen). 

On the other hand, I understand your standpoint. Disallowing all insecure operations.

talekhinezh: even if only by accident in this case, e.g. if EcdsaPrivateKey is some struct which wants to be printed. I think need to make interface such that it's "hard" to display but possible.

talekhinezh (approved):

PR #226, opened 2022-04-07
Add a workflow that triggers PTE environment upon merge to the main branch

collins-w:
This PR adds a workflow that triggers PTE environment upon merge to the main branch

### DIFF
--- a/.github/workflows/pte.yml
+++ b/.github/workflows/pte.yml
@@ -1,0 +1,22 @@
+name: PTE Deployment Trigger
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  trigger-pte-deployment:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger PTE deployment
+        env:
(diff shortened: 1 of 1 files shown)


iamyulong:
Hi @collins-w, we only update main branch right before a release. Does this match your expectation? Do we want to make continuous builds from the `develop`?

collins-w:
> Hi @collins-w, we only update main branch right before a release. Does this match your expectation? Do we want to make continuous builds from the `develop`?

@iamyulong yes

> Hi @collins-w, we only update main branch right before a release. Does this match your expectation? Do we want to make continuous builds from the `develop`?

@iamyulong yes, if that's the case, then I think it makes more sense to make builds from the develop branch. I'm going to update the workflow. Thanks

iamyulong:
That should eventually points to `v0.4` once it's released.

After more thinking, I don't think the PTE release should be triggered by upstream repo. It should be based on the release branch of the `resim-server` (or `main` if we don't distinguish staging and release).

PR #227, opened 2022-04-07
Decimal overflow detection added for sub() and add() methods

r001:
We did not check for overflow in Decimal::add() and sub() methods.
Now we do! ;)


iamyulong:
Rust does overflow check by default. :)

iamyulong (approved):

PR #229, opened 2022-04-08
Fix empty bucket creation error

iamyulong:
To allow creation of empty buckets in manifest

### DIFF
--- a/radix-engine/src/model/worktop.rs
+++ b/radix-engine/src/model/worktop.rs
@@ -40,6 +40,8 @@ impl Worktop {
                 .take_by_amount(amount)
                 .map(Bucket::new)
                 .map(Option::Some)
+        } else if !amount.is_zero() {
+            Err(ResourceContainerError::InsufficientBalance)
         } else {
             Ok(None)
         }
@@ -55,6 +57,8 @@ impl Worktop {
                 .take_by_ids(ids)
                 .map(Bucket::new)
                 .map(Option::Some)
+        } else if !ids.is_empty() {
+            Err(ResourceContainerError::InsufficientBalance)
         } else {
             Ok(None)
         }
--- a/radix-engine/src/model/worktop.rs
+++ b/radix-engine/src/model/worktop.rs
@@ -34,31 +34,29 @@ impl Worktop {
         &mut self,
         amount: Decimal,
         resource_address: ResourceAddress,
-    ) -> Result<Bucket, ResourceContainerError> {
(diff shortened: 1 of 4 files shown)


radix-engine/src/model/worktop.rs
```diff
@@ -34,31 +34,29 @@ impl Worktop {
         &mut self,
         amount: Decimal,
         resource_address: ResourceAddress,
-    ) -> Result<Bucket, ResourceContainerError> {
+    ) -> Result<Option<Bucket>, ResourceContainerError> {
```
talekhinezh, line 37: If returning `Option`, should just remove the `Result` part.

radix-engine/src/model/worktop.rs
```diff
@@ -34,31 +34,29 @@ impl Worktop {
...
-    ) -> Result<Bucket, ResourceContainerError> {
-        self.take_non_fungibles(&BTreeSet::from([id.clone()]), resource_address)
-    }
-
     pub fn take_non_fungibles(
         &mut self,
         ids: &BTreeSet<NonFungibleId>,
         resource_address: ResourceAddress,
-    ) -> Result<Bucket, ResourceContainerError> {
+    ) -> Result<Option<Bucket>, ResourceContainerError> {
```
talekhinezh, line 52: If returning Option, should just remove the Result part.

radix-engine/src/model/worktop.rs
iamyulong: I respectfully disagree though. `Option` means an optional return, while `Result` represents a fallible response. Not having the resource on worktop doesn't mean an error imo.

talekhinezh: But there is no possible error in this method?

talekhinezh: ie this method should be "infallible"

iamyulong: There are many errors, e.g. invalid amount/granularity and insufficient balance.

iamyulong: I think I missed one point though

iamyulong: Should still check if the amount is zero before issuing a `None`.

talekhinezh: Ahhh I see, didn't notice that `take_by_amount` is just mapped into the return value

talekhinezh (approved):

PR #233, opened 2022-04-12
Remove deprecated auth

talekhinezh:


iamyulong (approved):
I thought we removed this. Nice catch!

PR #235, opened 2022-04-12
Resource manager cleanup

talekhinezh:


radix-engine/src/engine/track.rs
```diff
@@ -456,31 +461,51 @@ impl<'s, S: SubstateStore> Track<'s, S> {
...
                 SubstateUpdate {
-                    prev_id: substate_update.prev_id,
+                    prev_id,
                     value: resource_manager,
                 },
             );
-            Ok(result)
         } else {
-            panic!("Resource Manager not found");
+            panic!("Component was never borrowed");
```
iamyulong, line 503: Message is misleading :)

radix-engine/src/engine/track.rs
```diff
@@ -456,31 +461,51 @@ impl<'s, S: SubstateStore> Track<'s, S> {
...
+        let maybe_resource = self.resource_managers.remove(&resource_address);
+        if let Some(SubstateUpdate { value, prev_id }) = maybe_resource {
+            self.borrowed_resource_managers
+                .insert(resource_address, prev_id);
+            Ok(value)
+        } else if self
+            .borrowed_resource_managers
+            .contains_key(&resource_address)
+        {
+            panic!("Invalid resource manager reentrancy");
```
iamyulong, line 477: Should this be an error instead of panic? Would also consider moving it up for readability.

iamyulong, line 477: Should this be an error instead of panic?

talekhinezh: No, panic for now since this logic is all enforced by system at the moment

iamyulong (approved):

PR #236, opened 2022-04-13
Output manifest for resim publish

iamyulong:
Note: some extra diffs have been introduced because of formatting with `./format.sh`


talekhinezh (approved):
Looks good, just suggest changing use of base64 to hex

iamyulong:
Agree on that base64 often causes confusion. Performance can be achieved by compression.

Replaced base64 with hex in the other PR.

PR #238, opened 2022-04-13
Resource Manager syscalls refactor

talekhinezh:


iamyulong (approved):
Looks good. May consider refactoring ResourceStatic to the same way as Blueprint/Component.

PR #241, opened 2022-04-13
Bucket refactor

talekhinezh:


radix-engine/src/engine/process.rs
```diff
@@ -865,27 +831,18 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
 
-        Ok(output)
+        Ok((output, moving_buckets, moving_proofs))
     }
 
     /// Calls a function/method.
     pub fn call(
         &mut self,
-        snode_ref: SNodeRef,
+        mut snode_ref: SNodeRef,
```
iamyulong, line 840: Why do we need `mut SNodeRef`?

radix-engine/src/engine/process.rs
```diff
@@ -962,21 +919,37 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
                     .get_auth(&function);
                 Ok((SNodeState::Bucket(bucket), method_auth.clone()))
             }
-            SNodeRef::Vault(vault_id) => {
+            SNodeRef::BucketRef(bucket_id) => {
```
iamyulong, line 922: Don't think it makes sense to have both `SNode::Bucket` and `SNode::BucketRef`.  

When we call bucket methods it's always `BucketRef`, except `XXX::func(self)` (which should be calling XXXStatic/XXXBlueprint, where bucket is just an argument).

radix-engine/src/engine/process.rs
```diff
@@ -1005,32 +988,47 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
+                    SNodeState::ResourceRef(resource_address, resource_manager) => {
+                        self.track.return_borrowed_global_resource_manager(
+                            resource_address,
+                            resource_manager,
+                        );
+                    }
+                    SNodeState::BucketRef(bucket_id, bucket) => {
+                        self.buckets.insert(bucket_id, bucket);
+                    }
+                    _ => {}
```
iamyulong, line 1030: A good practice is to explicitly write all the remaining matches, so we know what we're doing and not break when a new joiner adds another enum variant.

radix-engine/src/engine/process.rs
```diff
@@ -1052,23 +1050,37 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
             blueprint_name.to_string(),
         ));
-        let result = self.call(snode_ref, function.to_string(), args);
+        let result = self.call(snode_ref, function.to_string(), args)?;
+
+        // Auto move into auth_zone
+        for proof_id in &result.proof_ids {
+            let proof = self.proofs.remove(proof_id).unwrap();
+            self.auth_zone.push(proof);
+        }
```
iamyulong, line 1059: This reminds me that we need special handling of buckets at transaction layer as well. Is that properly covered?

radix-engine/src/engine/process.rs
talekhinezh: good point

radix-engine/src/engine/process.rs
talekhinezh: Yes transaction layer handling of buckets is still special handled. Unifying that comes in a future PR.

radix-engine/src/engine/process.rs
talekhinezh: good point

talekhinezh: Actually, I think we'll want to keep this for now as using BucketStatic won't work given that we'd need to expose arguments to retrieve the method auth. And this may be a useful pattern in the future.

talekhinezh: But will keep this in mind with further refactors.

iamyulong (approved):

PR #251, opened 2022-04-18
Deposit auth

talekhinezh:


radix-engine/src/model/vault.rs
```diff
@@ -153,8 +154,13 @@ impl Vault {
         &mut self,
         function: &str,
         args: Vec<ScryptoValue>,
+        bucket: Option<Bucket>, // FIXME: Hack in order to get deposit auth to work
     ) -> Result<Option<Bucket>, VaultError> {
         match function {
+            "put_into_vault" => {
+                self.put(bucket.unwrap()).map_err(VaultError::ResourceContainerError)?;
```
iamyulong, line 161: Would this crash RE if no bucket is passed via `InvokeSnode` instruction?

talekhinezh: Yes, will crash but so will the other methods as well. Was waiting to fix this until a refactor of methods/args into enums which will happen before v0.4.

iamyulong (approved):
Approved. We will refactor all SNode invocation as the next priority.

PR #254, opened 2022-04-19
Rename borrow macros

iamyulong:
New names:
- `borrow_package`, 
- `borrow_component`, 
- `borrow_resource_manager`


iamyulong:
Merging this rename PR so not to block Scrypto examples update prior to v0.4 release.

Should we have additional concern, can be easily reverted.

PR #262, opened 2022-04-22
Transaction as snode

talekhinezh:


radix-engine/src/transaction/executor.rs
```diff
@@ -171,7 +172,18 @@ impl<'l, L: SubstateStore> TransactionExecutor<'l, L> {
...
         );
         let mut proc = track.start_process(self.trace);
-        let (outputs, error) = validated.main(&mut proc);
+
+        let txn_process = TransactionProcess::new(validated.clone());
+        let mut txn_snode = SNodeState::Transaction(txn_process);
+        let error = match proc.run(&mut txn_snode, "execute".to_string(), vec![]) {
+            Ok(_) => None,
+            Err(e) => Some(e),
+        };
```
iamyulong, line 181: Passing the validated transaction to the transaction process would be cool lol

iamyulong (approved):
Looks good

radix-engine/src/transaction/executor.rs
talekhinezh: I was actually thinking per Instruction at some point :)

PR #267, opened 2022-04-23
Change resim output to use **writeln!** instead of **println!** 

cseidman:
Currently, resim functions which emit output to the console use println! to do so. This means they can only send output to the console. Other applications might wish to redirect the output to other IO objects such as files, network streams, or anything else that implements the std::io:Write trait.

Changes made here replaced the uses of println! with writeln! to use an output object such that what was previously: 
println!("This is some text") ; 
becomes
writeln!(out, "This is some text") ; 

The "out" in this case can be a stdout, a  file; or a TcpStream

Doing this allows the TCP server to use the same functions used by local resim without resorting to copying files for this purpose and then having them be out of sync.


iamyulong:
Closing this PR as it was wrongly branched. @cseidman will create a new one.

PR #268, opened 2022-04-24
Change resim output to use **writeln!** instead of **println!**

cseidman:
Currently, resim functions which emit output to the console use println! to do so. This means they can only send output to the console. Other applications might wish to redirect the output to other IO objects such as files, network streams, or anything else that implements the std::io:Write trait.

Changes made here replaced the uses of println! with writeln! to use an output object such that what was previously:
println!("This is some text") ;
becomes
writeln!(out, "This is some text") ;

The "out" in this case can be a stdout, a file; or a TcpStream

Doing this allows the TCP server to use the same functions used by local resim without resorting to copying files for this purpose and then having them be out of sync.

### DIFF
--- a/simulator/src/ledger/dumper.rs
+++ b/simulator/src/ledger/dumper.rs
@@ -1,3 +1,4 @@
+#![allow(unused_must_use)]
 use colored::*;
 use radix_engine::ledger::*;
 use radix_engine::model::*;
@@ -8,6 +9,7 @@ use std::collections::VecDeque;
 
 use crate::utils::*;
 
+
 /// Represents an error when displaying an entity.
 #[derive(Debug, Clone)]
 pub enum DisplayError {
@@ -17,21 +19,22 @@ pub enum DisplayError {
 }
 
 /// Dump a package into console.
-pub fn dump_package<T: SubstateStore>(
+pub fn dump_package<T: SubstateStore, O: std::io::Write>(
     package_address: PackageAddress,
     substate_store: &T,
+    output: &mut O
 ) -> Result<(), DisplayError> {
     let package: Option<Package> = substate_store
         .get_decoded_substate(&package_address)
         .map(|(package, _)| package);
     match package {
         Some(b) => {
-            println!(
+            writeln!(output,
                 "{}: {}",
                 "Package".green().bold(),
                 package_address.to_string()
             );
-            println!("{}: {} bytes", "Code size".green().bold(), b.code().len());
+            writeln!(output,"{}: {} bytes", "Code size".green().bold(), b.code().len());
             Ok(())
         }
         None => Err(DisplayError::PackageNotFound),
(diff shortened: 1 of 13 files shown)


simulator/src/resim/mod.rs
```diff
@@ -86,25 +86,27 @@ pub enum Command {
...
-        Command::GenerateKeyPair(cmd) => cmd.run(),
+        Command::ExportAbi(cmd) => cmd.run(&mut out),
+        Command::GenerateKeyPair(cmd) => cmd.run(&mut out),
         Command::Mint(cmd) => cmd.run(),
-        Command::NewAccount(cmd) => cmd.run(),
+        Command::NewAccount(cmd) => cmd.run(&mut out),
         Command::NewBadgeFixed(cmd) => cmd.run(),
         Command::NewBadgeMutable(cmd) => cmd.run(),
         Command::NewTokenFixed(cmd) => cmd.run(),
         Command::NewTokenMutable(cmd) => cmd.run(),
```
iamyulong, line 101: Is there a particular reason we're using `writeln` for some but not all?

simulator/src/ledger/dumper.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(unused_must_use)]
```
iamyulong, line 1: Minor: can we remove this warning suppression? I didn't see variables that is unused.

iamyulong (commented):
Looks good to me. Just left some minor comments.

simulator/src/resim/mod.rs
cseidman: Yes. The other ones don't produce any output - so we'd just be passing the IO parameter for nothing

iamyulong: That's perfect! Thanks for clarifying.

simulator/src/ledger/dumper.rs
cseidman: Yes we can - the warning actually is caused no by unused variables, but by the fact that writeln! has a return value and so the compiler warns that you need to read that variable (we really don't need to). The alternative is to use `let _ = writelin!(..` . I can goi ahead and do that if you'd like

iamyulong: I think this is because the return is a `Result` which may indicate a failure, especially when you're dealing with TCP stream.

The suggested use from the rust lang doc is to add `?` after `writeln!`. https://doc.rust-lang.org/std/macro.writeln.html

iamyulong: In our case `.map_err(Error::IOError)?`

cseidman: Yeah, but the TCP failure would be from a closed or broken client connection which is handled upstream. But yeah - I'll go ahead and add that in

iamyulong: Thanks!

cseidman: Done

iamyulong (approved):

PR #273, opened 2022-04-25
powi(i32) function for Decimal implemented

r001:
Decimal::powi(&self, exp:i32) -> Decimal function is now implemented with approximation. This solution will also work once we make Decimal::SCALE mutable.

### DIFF
--- a/scrypto/src/math/decimal.rs
+++ b/scrypto/src/math/decimal.rs
@@ -186,36 +186,21 @@ impl Decimal {
 
     /// Calculates power using exponentiation by squaring.
     pub fn powi(&self, exp: i32) -> Self {
-        let prod = BigInt::from(self.0);
         let one = BigInt::from(Self::ONE.0);
-        let exponent = exp.abs() as u32;
-        let get = |a: BigInt, ex: u32| if ex == 0 { one.clone() } else { a };
-        let fix_rounding = one.clone() / 2i8;
-        let fix_product = |product: BigInt| {
-            let fixed = (product + &fix_rounding) / &one;
-            if fixed.to_signed_bytes_be().len() > i128::BITS as usize {
-                panic!("overflow");
-            } else {
-                fixed
-            }
-        };
-        let nth_bit = |a: u32, n: u32| (a >> n) & 1;
-        let (_, res) = (1..i32::BITS - exponent.leading_zeros())
-            .map(|n| nth_bit(exponent, n))
-            .fold(
-                (fix_product(&prod * &prod), get(prod, nth_bit(exponent, 0))),
-                |acc, bit| {
-                    (
-                        fix_product(&acc.0 * &acc.0),
-                        fix_product(get(acc.0, bit) * acc.1),
-                    )
-                },
-            );
-
-        if exp.is_negative() {
-            Self((&one * &one / res).to_i128().expect("overflow"))
+        let base = BigInt::from(self.0);
+        let to_dec = |x: BigInt| Decimal(x.to_i128().expect("overflow"));
+        if exp < 0 {
+            return to_dec(&one * &one / base).powi(-exp);
+        }
+        if exp == 0 {
+            return Self::ONE;
+        }
+        if exp % 2 == 0 {
+            return to_dec(&base * &base / &one).powi(exp / 2);
         } else {
-            Self(res.to_i128().expect("overflow"))
+            return to_dec(
+                &base * BigInt::from(to_dec(&base * &base / &one).powi((exp - 1) / 2).0) / &one,
+            );
         }
     }
 }
@@ -320,7 +305,7 @@ impl<T: Into<Decimal>> Mul<T> for Decimal {
     fn mul(self, other: T) -> Self::Output {
         let a = BigInt::from(self.0);
         let b = BigInt::from(other.into().0);
-        let c = (a * b + Self::ONE.0 / 2) / Self::ONE.0;
+        let c = a * b / Self::ONE.0;
         Self(c.to_i128().expect("overflow"))
     }
 }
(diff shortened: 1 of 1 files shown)


scrypto/src/math/decimal.rs
```diff
@@ -170,6 +173,65 @@ impl Decimal {
                 }
             }
         }
+
+    }
+
+    pub fn powi(&self, exp: i32) -> Self {
+        let mut x: BigInt = self.0.clone().into();
+        if x.is_negative() {
+            panic!("powi is not supported for negative numbers");
```
iamyulong, line 182: Why not? :)

scrypto/src/math/decimal.rs
```diff
@@ -170,6 +173,65 @@ impl Decimal {
...
+        }
+        let s = Self::SCALE;
+        let bytes2 = Self::BITS / 4;
+        let b: BigInt = 10i128.pow(s).into();
```
iamyulong, line 186: `10i128.pow(s)` may overflow

scrypto/src/math/decimal.rs
```diff
@@ -170,6 +173,65 @@ impl Decimal {
...
+                }
+                n /= 2i8;
+            }
+        }   
+
+        if exp.is_negative() {
+            Self(big_int_to_i128(b.clone() * b / z))
+        } else {
+            Self(big_int_to_i128(z))
+        }
```
iamyulong, line 234: Personally, I feel like there are too many intermediary steps and rounding, which makes it super hard to describe the behaviour of the `powi` function and hard for me to verify the correctness.

````
(a * 10 ^ -18) ^ exp = a ^ exp * 10 ^ (-18 * exp) 
                     = a ^ exp * 10 ^ (-18 * exp + 18) * 10 ^ -18
                     = a ^ exp / 10 ^ (18 * exp + 18) * 10 ^ -18
````

It seems to we just need to raise `a` to the power of `exp`, raise `10` to the power of `18 * exp + 18`, and do the division of the these two numbers, which requires one rounding only.

devmannic:
Why not just implement traits from `num-traits` or reusing the macros that implement lots of these types of algorithms or reuse existing implementations altogether by converting to a different type do the operation then converting back?

  I don't think having a custom decimal implementation really makes sense apart from the existing lowest level functionality.

r001:
> Why not just implement traits from `num-traits` or reusing the macros that implement lots of these types of algorithms or reuse existing implementations altogether by converting to a different type do the operation then converting back?
> 
> I don't think having a custom decimal implementation really makes sense apart from the existing lowest level functionality.

@devmannic There are two issues with floating point types:
1. The very big and very small problem: If you want to add to a very big floating point number a very small one, you have the problem that addition is not going to take place - SILENTLY - as the fractional part will not be able to represent the small addition. This leads to unexpected behaviour, and thus an exploit. Decimal addresses this, by only enabling as big numbers as the precision of the number can actually hold.
2. Binary representation problem: Floating point numbers represent numbers with a binary fractional part, and a binary exponent. Because of this, they can't even express the number of `1.7` correctly, because of having to convert them to binary, and thus the closest `1.69999...` will actually be stored. And this will also lead to: exploits.
See [this link](https://www.sciencedirect.com/topics/engineering/floating-point-number) for details.

devmannic:
> > Why not just implement traits from `num-traits` or reusing the macros that implement lots of these types of algorithms or reuse existing implementations altogether by converting to a different type do the operation then converting back?
> > 
> > I don't think having a custom decimal implementation really makes sense apart from the existing lowest level functionality.
> 
> @devmannic There are two issues with floating point types:
> 1. The very big and very small problem: If you want to add to a very big floating point number a very small one, you have the problem that addition is not going to take place - SILENTLY - as the fractional part will not be able to represent the small addition. This leads to unexpected behaviour, and thus an exploit. Decimal addresses this, by only enabling as big numbers as the precision of the number can actually hold.
> 2. Binary representation problem: Floating point numbers represent numbers with a binary fractional part, and a binary exponent. Because of this, they can't even express the number of `1.7` correctly, because of having to convert them to binary, and thus the closest `1.69999...` will actually be stored. And this will also lead to: exploits.
> See [this link](https://www.sciencedirect.com/topics/engineering/floating-point-number) for details. 
> 

Yes I'm aware of these issues and agree we want solutions.  But my concern is in reimplementation of lots of operations and maintaining it let alone verifying it.  As an alternative but maybe slightly less performance approach see my implementation in defi/amm/bonding in the community repo for examples of how I solved these issues using num-rational and converting to/from Decimal.  It gets us a lot more for a lot less work in the short and long term.

https://github.com/radixdlt/community-scrypto-examples/blob/main/defi/amms/bonding/src/number.rs

beemdvp:
Just wanted to say, thank you for looking into this as this was needed during foohack! And others have also been asking for this operation. Great job 👍

r001:
This `powi()` works the following way: 
- It originates from the idea that `a ^ b = a * a * a ...` as b is integral, and a is also, we can do this. 
- Question: how to minimize multiplications? We use the fact that b = bit(0) * 2 ^ 0 + bit(1) * 2 ^ 1 + bit(2) * 2 ^ 2 + bit(3) * 2 ^ 3 + ... + bit(n) * 2 ^ n
 eg: if `bit(3)` of `b` is `1` then it means that we calculate `prod(n) = prod(n-1) * prod(n-1), prod(0) = a` and thus  `prod(3) = a ^ 8` and we multiply final result by this `res = res * prod(3)`. This way we only have to do around `log2(exponent)` rounds instead of `exponent` rounds.
- I have also renamed variables, and simplified code.
- Also fixed bug with negative power calculation.

I have also replaced a custom made BigInt -> Decimal function using trait `ToPrimitive` for `BigInt`.
@iamyulong @devmannic pls accept this one!

r001:
I have rewritten the whole thing again to have a more concise syntax. The algo is based on:
https://en.wikipedia.org/wiki/Exponentiation_by_squaring

iamyulong:
Awesome! We also need to use the checked `i32` arithmetic so overflow is captured for `release` build.

iamyulong (approved):
LGTM

PR #274, opened 2022-04-26
Refactored publish and the return of receipts to allow for them to write to an IO target

cseidman:
Changes make it so that the TCP server can best stay in sync with changes to to RESIM


simulator/src/resim/cmd_publish.rs
```diff
@@ -47,27 +48,50 @@ impl Publish {
...
+    }
+
+    pub fn publish_wasm<O: std::io::Write>(&self, out: &mut O) -> Result<(), Error> {
+        // Load wasm code
+        println!("Publishing ..");
+        let wasm_path = "wasm_dir/cur.wasm" ;
+        let code = fs::read(wasm_path).map_err(Error::IOError)?;
+        println!("Read code to variable");
+        self.store_package(out, &code)
+    }
```
iamyulong, line 61: Do we need `publish_wasm`? Can we move it to PTE server? If not, can we make the path a parameter?

iamyulong (approved):
LGTM with one minor comment.

simulator/src/resim/cmd_publish.rs
cseidman: Fixed that. Thanks!

PR #279, opened 2022-04-27
Implement Derived SubstateIds

talekhinezh:


simulator/src/ledger/radix_engine_db.rs
```diff
@@ -94,51 +94,54 @@ impl QueryableSubstateStore for RadixEngineDB {
...
+impl WriteableSubstateStore for RadixEngineDB {
+    fn put_substate(&mut self, address: &[u8], substate: Substate) {
+        self.write(address, &scrypto_encode(&substate));
+    }
+
+    fn put_child_substate(&mut self, address: &[u8], key: &[u8], substate: Substate) {
+        let mut id = address.to_vec();
+        id.extend(key.to_vec());
+        self.write(&id, &scrypto_encode(&substate));
+    }
```
iamyulong, line 134: Feel like we should get rid of the `put_child_substate` interface. It's causing very inconsistent physical keys in the database.

radix-engine/src/ledger/traits.rs
```diff
@@ -124,6 +71,37 @@ pub trait SubstateStore {
...
+    fn get_nonce(&self) -> u64;
+}
+
+pub trait WriteableSubstateStore {
+    fn put_substate(&mut self, address: &[u8], substate: Substate);
+    fn put_child_substate(&mut self, address: &[u8], key: &[u8], substate: Substate);
+
+    fn put_keyed_substate(&mut self, address: &[u8], value: Vec<u8>, phys_id: (Hash, u32)) {
+        self.put_substate(address, Substate { value, phys_id });
+    }
```
iamyulong, line 89: Didn't see the difference between `put_keyed_substate` and `put_substate`. 

I'm supposing that there will be some semantic difference in the next PR:
-  One uses conflict free substate id
-  The other uses derived substate id where existence check is required (if so, the interface should return a result).

iamyulong (commented):
The clean-up is nice

simulator/src/ledger/radix_engine_db.rs
talekhinezh: Yup agreed, refining this interface will be the subject of the next few PRs :)

radix-engine/src/ledger/traits.rs
talekhinezh: Yup, this is mostly temporary interface in fact.

iamyulong (approved):
Let's go!

PR #280, opened 2022-04-28
Migrate to enum methods

talekhinezh:


radix-engine/src/model/resource_manager.rs
```diff
@@ -63,30 +68,21 @@ impl MethodEntry {
...
 
-    pub fn get_update_auth(&self, args: &[ScryptoValue]) -> &MethodAuthorization {
-        let method: String = match scrypto_decode(&args[0].raw) {
-            Ok(m) => m,
-            _ => return &MethodAuthorization::Unsupported,
-        };
-        match method.as_str() {
-            "lock" | "update" => &self.update_auth,
-            _ => &MethodAuthorization::Unsupported,
+    pub fn get_update_auth(&self, method: MethodAccessRuleMethod) -> &MethodAuthorization {
```
iamyulong, line 71: `MethodAccessRuleMethod`  doesn't seem to be needed.

radix-engine/src/model/resource_manager.rs
```diff
@@ -117,53 +114,52 @@ impl ResourceManager {
         metadata: HashMap<String, String>,
         mut auth: HashMap<ResourceMethod, (AccessRule, Mutability)>,
     ) -> Result<Self, ResourceManagerError> {
+        let mut vault_method_table: HashMap<String, Option<ResourceMethod>> = HashMap::new();
+        vault_method_table.insert("take".to_string(), Some(Withdraw));
+        vault_method_table.insert("put".to_string(), Some(Deposit));
+        if let ResourceType::NonFungible = resource_type {
+            vault_method_table.insert("take_non_fungibles".to_string(), Some(Withdraw));
+        }
```
iamyulong, line 122: I think we should remove this `if` check. Calling this method on fungible resource should be a runtime error rather than authorisation error.

radix-engine/src/model/resource_manager.rs
```diff
@@ -173,39 +169,77 @@ impl ResourceManager {
...
+        let method: VaultMethod = match scrypto_decode(&arg.raw) {
+            Ok(m) => m,
+            Err(_) => return &MethodAuthorization::Unsupported,
+        };
+
+        match self.vault_method_table.get(method.name()) {
+            None => &MethodAuthorization::Unsupported,
+            Some(None) => &MethodAuthorization::AllowAll,
+            Some(Some(method)) => self.authorization.get(method).unwrap().get_method_auth(),
+        }
```
iamyulong, line 197: Why don't we store `MethodAuthorization` to the vault method? Looks like unnecessary indirection.

scrypto/src/resource/proof.rs
```diff
@@ -14,23 +14,46 @@ use crate::rust::vec::Vec;
...
+        }
     }
 }
 
 impl Proof {
+    invocations! {
+        SNodeRef::ProofRef(self.0) => {
+            pub fn amount(&self) -> Decimal { ProofMethod::Amount() }
+            pub fn resource_address(&self) -> ResourceAddress { ProofMethod::ResourceAddress() }
+            pub fn non_fungible_ids(&self) -> BTreeSet<NonFungibleId> { ProofMethod::NonFungibleIds() }
```
iamyulong, line 47: How are we going to deal with documentation on the individual methods?

scrypto/src/resource/resource_manager.rs
```diff
@@ -32,6 +31,51 @@ pub enum Mutability {
...
+            ResourceManagerMethod::GetResourceType() => "get_resource_type",
+            ResourceManagerMethod::GetMetadata() => "get_metadata",
+            ResourceManagerMethod::GetTotalSupply() => "get_total_supply",
+            ResourceManagerMethod::GetNonFungible(_) => "get_non_fungible",
+            ResourceManagerMethod::NonFungibleExists(_) => "non_fungible_exists",
+            ResourceManagerMethod::UpdateNonFungibleData(_, _) => "update_non_fungible_data",
+            ResourceManagerMethod::UpdateMetadata(_) => "update_metadata",
+            ResourceManagerMethod::CreateVault() => "create_vault",
+            ResourceManagerMethod::CreateBucket() => "create_bucket",
+        }
```
iamyulong, line 75: Didn't get why we still need these string literals.

radix-engine/src/engine/process.rs
```diff
@@ -360,24 +361,39 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
                 transaction_process.main(self)
             }
             SNodeState::PackageStatic => {
-                Package::static_main(&function, args, self).map_err(RuntimeError::PackageError)
+                let arg = if args.len() > 1 {
+                    Err(RuntimeError::InvalidInvocation)
+                } else {
+                    args.into_iter().nth(0).ok_or(RuntimeError::InvalidInvocation)
+                }?;
+                Package::static_main(arg, self).map_err(RuntimeError::PackageError)
```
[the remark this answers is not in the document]
talekhinezh: Reason this was kept in is that it will all be refactored out in next PR when all methods have the same interface

radix-engine/src/model/auth_zone.rs
```diff
@@ -80,61 +79,54 @@ impl AuthZone {
 
     pub fn main<S: SystemApi>(
         &mut self,
-        function: &str,
-        args: Vec<ScryptoValue>,
+        arg: ScryptoValue,
```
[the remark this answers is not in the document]
talekhinezh: Both method and args I suppose. I was taking the approach that from radix engine's point of view this is just an arg to the `main()` method.

radix-engine/src/engine/process.rs
```diff
@@ -603,24 +669,15 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
...
-                let method_auth = self
-                    .track
-                    .get_resource_manager(&resource_address)
-                    .unwrap()
-                    .get_auth(&function, &args);
-                Ok((
-                    Borrowed(BorrowedSNodeState::Bucket(bucket_id.clone(), bucket)),
-                    vec![method_auth.clone()],
-                ))
+                Ok((Borrowed(BorrowedSNodeState::Bucket(bucket_id.clone(), bucket)), vec![]))
```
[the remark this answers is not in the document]
talekhinezh: No, as you can see in all of the `BorrowedSNodeState` invariants, these are all owned objects. It's "borrowed" because it is exposed as borrowed through the `SNodeState` interface as something borrowed during execution. The borrowed states are then returned at the end of the call.

radix-engine/src/model/resource_manager.rs
talekhinezh: This is there to maintain what the interface would like once we do separate out MethodAccessRule, ie a rule per method

radix-engine/src/model/resource_manager.rs
talekhinezh: I don't view this as an authorization error but as a method does not exist error. I don't believe such methods should be shown in an abi for example for fungible tokens.

radix-engine/src/model/resource_manager.rs
talekhinezh: As all vaults from a resource get their authorization from a single auth source (required so that resource manager can update this easily). Having said that, a bunch of refactors still needed to be done to get this in a better state (e.g. moving the auth stuff outside of resource manager).

scrypto/src/resource/proof.rs
talekhinezh: We would be able to have better documentation once we have a code generator for these which I believe is the right approach.

scrypto/src/resource/resource_manager.rs
talekhinezh: This is required to be able to add the range of methods to maps. But agreed this needs more reworking.

radix-engine/src/engine/process.rs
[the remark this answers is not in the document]
iamyulong: Maybe we need a better strategy w.r.t. intermediate changes, especially when we're certain they will be thrown away right after? I'm trying to avoid introducing a lot of "unreviewed" code in the code base, which is not a good practice.

radix-engine/src/engine/process.rs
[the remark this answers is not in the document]
iamyulong: Yep, understood the implementation. I was just noting a direction that we may consider. The limitation of manual borrow semantics is hard to maintain and prone to bugs.

radix-engine/src/model/auth_zone.rs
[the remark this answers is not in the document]
iamyulong: Maybe we should then rename to `method_and_args` or `call_data` or `invocation` to ease the confusion?

radix-engine/src/model/resource_manager.rs
iamyulong: Got it!

radix-engine/src/model/resource_manager.rs
iamyulong: Realised that this is the second time we discuss about this. The issue with this design is introduction of polymorphism which we're not sure if it's the right design. We have a single `ResourceManager` blueprint, but different ABIs for `ResourceManager` components. 

If we want distinct method set, we should have `FungibleResourceAddress + FungibleResourceManager` and `NonFungibleResourceAddress + NonFungibleResourceManager`. Although, I didn't see much benefits from this since we're supporting handling non-fungibles in a fungible way.

radix-engine/src/model/resource_manager.rs
iamyulong: Not sure if I get the point. 

Saw a lot of redundant matching arms like this.
```rust
None => &MethodAuthorization::Unsupported,
Some(None) => &MethodAuthorization::AllowAll,
Some(Some(method)) => self.authorization.get(method).unwrap().get_method_auth(),
```

Originally, I think I didn't understand why we need an `Option` wrapping `MethodAuthorization`, as it seems we're always assigning a value to all methods and this is inconsistent with component auth. 

But digging into this a bit more, looks like we're only using it as a shortcut for "if the method auth is none, there is no update/lock_auth method", which related to my other point about polymorphism.  This worsens the situation as we're essentially introducing an ABI per resource configuration.

scrypto/src/resource/proof.rs
iamyulong: Before that, can we use the "internal" approach you've been using for the other types?

scrypto/src/resource/resource_manager.rs
iamyulong: Sorry, what do you mean by "the range of methods"? Wildcard matching?

radix-engine/src/engine/process.rs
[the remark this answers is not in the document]
talekhinezh: This is true, I can fix this then.

radix-engine/src/engine/process.rs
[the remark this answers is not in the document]
talekhinezh: Ah, it's just difficult/impossible to do because we ARE implementing `borrow_mut` essentially. The problem comes in the Rust constraint that you should never more than one mutable borrow so this makes it very difficult to both pass a borrowed object and a borrowed track to a called process.

scrypto/src/resource/proof.rs
talekhinezh: what's the internal approach?

iamyulong (approved):

PR #300, opened 2022-05-10
Added the ability to check an access rule from Scrypto

0xOmarA:
This pull request adds the ability to check whether a given vector of `Proof`s satisfies an `AccessRule` or not, inside of Scrypto. Example usage looks as follows:

```rust
pub fn is_allowed(proofs: Vec<Proof>)  {
    let rule: AccessRule = rule!(
        require_amount(dec!("20"), RADIX_TOKEN)
    );
    info!("Proofs satisfy AuthRule: {:?}", rule.check(&proofs[..])); 
}
```

### DIFF
--- a/radix-engine/src/engine/process.rs
+++ b/radix-engine/src/engine/process.rs
@@ -1254,11 +1254,14 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
             .iter()
             .map(|proof_id| self.proofs.get(&proof_id).ok_or(RuntimeError::ProofNotFound(proof_id.clone())).unwrap().clone())
             .collect::<Vec<Proof>>();
-        let simulated_auth_zone = AuthZone::new_with_proofs(proofs);
+        let mut simulated_auth_zone = AuthZone::new_with_proofs(proofs);
+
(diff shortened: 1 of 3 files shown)


iamyulong:
As discussed with @0xOmarA, it's intended to consume all the input proofs. Will see how developers are going to use this method.

iamyulong (approved):
LGTM

radix-engine/src/engine/process.rs
```diff
@@ -1248,6 +1248,19 @@ impl<'r, 'l, L: SubstateStore> Process<'r, 'l, L> {
             actor: wasm_process.vm.actor.clone(),
         });
     }
+    
+    fn handle_check_access_rule(&mut self, input: CheckAccessRuleInput) -> Result<CheckAccessRuleOutput, RuntimeError> {
+        let proofs = input.proof_ids
+            .iter()
+            .map(|proof_id| self.proofs.remove(&proof_id).ok_or(RuntimeError::ProofNotFound(proof_id.clone())).unwrap())
```
iamyulong, line 1255: Should be `self.proofs.get` since we're borrowing the proofs now.

PR #302, opened 2022-05-10
Update dependency versions

iamyulong:
I wanted to lock all our dependencies to avoid non-deterministic compilation issues, but realised that it's very against the `Cargo.lock` design.

From https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries
> For libraries the situation is somewhat different. A library is not only used by the library developers, but also any downstream consumers of the library. Users dependent on the library will not inspect the library’s Cargo.lock (even if it exists). This is precisely because a library should not be deterministically recompiled for all users of the library.

Most of our crates are libraries. The `Cargo.lock` file will be ignored by Scrypto packages. It's better to identify any potential build issues rather than masking them by locking dependencies.

**This PR is still up for review for dependency updates.**


iamyulong:
Merge this now as no meaningful review can be done, rather than CI check.

PR #303, opened 2022-05-10
Add naive WASM metering

iamyulong:
This PR introduces a naive implementation of WASM instrumentation using `wasm-instrument` library.

This is more of integration test, rather than a proper metering implementation.

Benchmark after the PR:
```
test bench_transfer ... bench:  14,797,367 ns/iter (+/- 2,313,080)
```
And before
```
test bench_transfer ... bench:  10,834,012 ns/iter (+/- 1,494,257)
```


radix-engine/src/wasm/wasmi.rs
```diff
@@ -202,6 +226,27 @@ impl ScryptoWasmValidator for WasmiEngine {
     }
 }
 
+impl ScryptoWasmInstrumenter for WasmiEngine {
+    fn instrument(&mut self, code: &[u8]) -> Result<Vec<u8>, InstrumentError> {
+        let mut module =
+            parity_wasm::deserialize_buffer(code).expect("Unable to parse wasm module");
```
talekhinezh, line 232: Will need to add validation that code itself doesn't include calls to `TBD_FUNCTION_INDEX`. Maybe add as a TODO comment?

radix-engine/tests/metering.rs
```diff
@@ -0,0 +1,40 @@
...
+    let package = executor
+        .publish_package(&compile_package!(format!("./tests/{}", "metering")))
+        .unwrap();
+
+    let transaction = TransactionBuilder::new()
+        .call_function(package, "Metering", call_data!(iterations(10_000u32)))
+        .build(executor.get_nonce([]))
+        .sign([]);
+    let receipt = executor.validate_and_execute(&transaction).unwrap();
+    assert!(receipt.result.is_ok());
```
talekhinezh, line 23: Can we use `.expect` rather than `assert` as `.expect` displays the actual error on failure

radix-engine/tests/metering.rs
```diff
@@ -0,0 +1,40 @@
+#[rustfmt::skip]
+pub mod test_runner;
+
+use radix_engine::ledger::*;
+use radix_engine::transaction::*;
+use radix_engine::wasm::InvokeError;
+use scrypto::call_data;
+use scrypto::prelude::*;
+
```
talekhinezh, line 9: Can we also add max_stack_depth test and memory_grow test?

radix-engine/src/wasm/wasmi.rs
iamyulong: Not sure if I agree. This function index is from the `env` module, separated from the Scrypto module. Are we concerned about Scryto burning fees for nothing?

radix-engine/tests/metering.rs
iamyulong: Yes yes

radix-engine/tests/metering.rs
iamyulong: Yes, will add

radix-engine/src/wasm/wasmi.rs
talekhinezh: Yes, exactly. Or rather, it is an assumption we make I believe and if assumption is broken, it may potentially break things (e.g. if we may need to change fees down the line and recreate a new wasm file.

talekhinezh (approved):

PR #304, opened 2022-05-10
Cleanup vault in track

talekhinezh:


radix-engine/src/engine/track.rs
```diff
@@ -343,6 +366,24 @@ impl<'s, S: ReadableSubstateStore> Track<'s, S> {
         address
     }
 
+    pub fn create_uuid_value_2<A: Into<Address>, V: Into<SubstateValue>>(
+        &mut self,
+        addr: A,
+        value: V,
+    ) {
+        self.up_substates.insert(addr.into().encode(), value.into());
+    }
```
iamyulong, line 375: Minor:  May need more thoughts on these secondary UUID-based value. `ComponentAddress + Transaction + Index` is too much for an address. Also, do we want to put LazyMap into this category as well, for map-level attributes?

iamyulong (approved):

radix-engine/src/engine/track.rs
talekhinezh: Why do you think it would be too much for an address? It seems to me to be an extension of the snode addressing scheme we talked about. And yes, I think LazyMap will at some point be added as well.

PR #308, opened 2022-05-12
Global Scrypto loader

iamyulong:


radix-engine/src/model/bucket.rs
```diff
@@ -152,7 +153,7 @@ impl Bucket {
         self.container.borrow_mut()
     }
 
-    pub fn main<S: SystemApi>(
+    pub fn main<S: SystemApi<W, I>, W: WasmEngine<I>, I: WasmInstance>(
```
talekhinezh, line 156: Bit of a code smell here in that `bucket.rs` needs to know about wasm

radix-engine/src/transaction/executor.rs
```diff
@@ -45,21 +61,34 @@ impl<'l, L: ReadableSubstateStore + WriteableSubstateStore> AbiProvider
...
+    S: ReadableSubstateStore + WriteableSubstateStore,
+    W: WasmEngine<I>,
+    I: WasmInstance,
+{
+    pub fn new(
+        substate_store: &'s mut S,
+        wasm_engine: &'w mut W,
+        trace: bool,
+    ) -> TransactionExecutor<'s, 'w, S, W, I> {
+        bootstrap(substate_store, wasm_engine);
```
talekhinezh, line 75: A bit weird to have bootstrap here

radix-engine/src/model/bucket.rs
iamyulong: Yep, was struggling a bit with Rust's generic type. The `Bucket` impl depends on `SystemApi` which is parameterized with `WasmEngine`.

radix-engine/src/transaction/executor.rs
iamyulong: Yep, I was trying to avoid bootstraping everywhere (which requires the wasm engine as in this PR). 

After more thoughts, I think it will make more sense if we:
- Remove `WasmEngine` in `Package` constructor, giving up validation cache since it's always billed. 
- Restore the `with_bootstrap` substate store constructor

Both changes will be done in https://github.com/radixdlt/radixdlt-scrypto/pull/316

radix-engine/src/model/bucket.rs
iamyulong: In the long term, we want to move away from compile-time-decided engine to runtime-decided engine, so multiple engines can coexist and there is not need for generic types (replaced with trait object).

talekhinezh (approved):

PR #310, opened 2022-05-19
Add external_blueprint and external_component macros

dhedey:
As discussed in a call a few weeks ago, I took an action to prototype these macros.

See the docs in `macro.rs` for how they're intended to be used.

## Context
They're not intended to _replace_ the ABI - I imagine we still want to store the SBOR of the ABI being stored in the engine, for validation.

But they're intended for two use cases:
* Replace the import! macro with something which looks much nicer, and is slightly clearer what it does.
* Allow users to create dynamic component bridges without importing an explicit interface.

## Suggested next steps

If you like the look of them, my proposal would be to take this PR and:
* Add some more tests 👍 
* --Change the ExportAbi resim method to output the bridge, as well as the JSON:
   * The external_blueprint! call will be created from JSON, and include the enums and structs as separate definitions; similar to the import! macro.-- Will discuss this and follow up later.
* Rebase on develop instead of main 👍 

Note - under this plan, the change to have ABI stored in the engine (and not generated from the WASM at runtime) is orthogonal to this PR, and these macros are only relevant to other people writing scrypto, not to the build of the original component.


radix-engine/tests/external_bridge.rs
```diff
@@ -0,0 +1,106 @@
...
+    let code_file_contents =
+        template_file_contents.replace("%%PACKAGE_ADDRESS%%", &package_address_hex);
+
+    // Recreate the file and dump the processed contents to it
+    let mut code_file = File::create(&code_file_path)?;
+    code_file.write(code_file_contents.as_bytes())?;
+    drop(code_file);
+
+    Result::Ok(())
+}
```
iamyulong, line 106: Just a small note: cargo run tests in parallel by default. We will need to avoid sharing crates in future if we continue this pattern of testing.

scrypto/src/macros.rs
```diff
@@ -201,3 +201,314 @@ macro_rules! include_package {
...
+}
+
+// This is a TT-Muncher, a useful guide for this type of use case is here: https://adventures.michaelfbryan.com/posts/non-trivial-macros/
+#[macro_export]
+macro_rules! bridge_blueprint_interface_members {
+    (
+        $blueprint_context:tt,
+        fn $method_name:ident(&self$(, $method_args:ident: $method_types:ty)*) -> $method_output:ty;
+        $($rest:tt)*
+    ) => {
```
iamyulong, line 371: How about function/method with no output? Will I have to use `()` ?

scrypto/src/macros.rs
```diff
@@ -201,3 +201,314 @@ macro_rules! include_package {
...
+    (
+        {
+            package: $package_address:literal,
+            blueprint: $blueprint_logical_name:literal $(,)?
+        }
+    ) => {
+        $blueprint_logical_name
+    };
+    () => {
+        compile_error!("Cannot call package functions (ie without &self or &mut self) on a external_component - use a external_blueprint instead.");
```
iamyulong, line 507: I got confused by this error message. How do we know it's used by an `external_component`?

iamyulong (commented):
Looks good to me. Just a few minor comments.

radix-engine/tests/external_bridge.rs
dhedey: Good point - I'll add some comments inside the crate to flag this issue.

scrypto/src/macros.rs
dhedey: Add a comment explaining :)

iamyulong (approved):

PR #311, opened 2022-05-24
Overflow

r001:
As Dave has pointed out, we can rust do checked math with setting `overflow-checks = true` in `Cargo.toml`.
In this PR this was done, along with all tests to make sure that we always have overflow-checks on Scrypto code.

### DIFF
--- a/scrypto/Cargo.toml
+++ b/scrypto/Cargo.toml
@@ -28,3 +28,8 @@ trace = ["scrypto-derive/trace"]
 
 # Enable serde derives
 serde = ["sbor/serde", "scrypto-abi/serde"]
+
+[profile.release]
+# Enable checking of overflow in Scrypto
+# Do not remove this as all scrypto code depends on overflow safe operations.
+overflow-checks = true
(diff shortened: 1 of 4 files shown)


r001:
@iamyulong All checks check out.

scrypto/Cargo.toml
```diff
@@ -26,4 +27,9 @@ alloc = ["hashbrown", "hex/alloc", "sbor/alloc", "scrypto-abi/alloc", "scrypto-d
...
 
 # Enable serde derives
-serde = ["sbor/serde", "scrypto-abi/serde"]
\ No newline at end of file
+serde = ["sbor/serde", "scrypto-abi/serde"]
+
+[profile.release]
+# Enable checking of overflow in Scrypto
+# Do not remove this as all scrypto code depends on overflow safe operations.
+overflow-checks = true
```
dhedey, line 35: Because the profile is chosen based on the root Crate that's actually compiled, we need to (also) add this to the template for Scrypto packages, as that's the root source of compilation for DApps

iamyulong:
Overflow check has been added to the template through https://github.com/radixdlt/radixdlt-scrypto/pull/428

PR #313, opened 2022-05-25
Refactor package

talekhinezh:
Package creation is done via struct rather than just a WASM array of bytes. This is in preparation for static ABI implementation.


iamyulong (approved):
LGTM

PR #317, opened 2022-05-29
Refactor vault abi

talekhinezh:


radix-engine/src/engine/call_frame.rs
```diff
@@ -404,12 +411,13 @@ impl<'r, 'l, L: ReadableSubstateStore> CallFrame<'r, 'l, L> {
...
         snode_ref: SNodeRef,
+        method_name: &str,
         call_data: ScryptoValue,
     ) -> Result<ScryptoValue, RuntimeError> {
         let function = if let Value::Enum { name, .. } = &call_data.dom {
             name.clone()
         } else {
-            return Err(RuntimeError::InvalidInvocation);
+            method_name.to_string()
         };
```
iamyulong, line 421: Why do we need this logic? I thought you were using `invoke_snode2` for the `name + input` style.

I guess I'm also asking what the strategy for this massive refactoring is.

scrypto/src/resource/vault.rs
```diff
@@ -14,36 +14,44 @@ use crate::engine::{api::*, call_engine, types::VaultId};
...
+
+#[derive(Debug, TypeId, Encode, Decode)]
+pub struct VaultCreateProofByAmountInput {
+    pub amount: Decimal,
+}
+
+#[derive(Debug, TypeId, Encode, Decode)]
+pub struct VaultCreateProofByIdsInput {
+    pub ids: BTreeSet<NonFungibleId>,
 }
```
iamyulong, line 56: I was wondering if it would be better to add some namespace for these input/output types, like encapsulate with a `mod` or `impl`. Not a big fan of prefix. :)

PR #322, opened 2022-06-01
Refactor proof abi

talekhinezh:


build.sh
```diff
@@ -15,7 +15,6 @@ cd "$(dirname "$0")"
 (cd transaction-manifest; cargo build)
 
 echo "Building assets and examples..."
-(cd assets/account; cargo build --target wasm32-unknown-unknown --release)
-(cd assets/system; cargo build --target wasm32-unknown-unknown --release)
+(cd assets; ./update-assets.sh)
```
iamyulong, line 18: Can we not use `update-assets.sh` before we have a better solution for deterministic compilation? I'm trying to avoid every commit including a change to the wasm files, as our machine environments are different.

iamyulong (approved):
LGTM

build.sh
talekhinezh: sure i'll update

PR #326, opened 2022-06-03
New Transaction Format

iamyulong:
Changes:
- Added models for transaction header, manifest, intent and notary signature
- Moved signing from Scrypto library to transaction module
- Split `TransactionBuilder` into `ManifestBuilder` and `TransactionBuilder`
- Refactored all Radix Engine unit tests to use `TestRunner`

Given the size of this PR, I will mark this PR as ready for review and will do the following enhancements in another PR:
- Complete transaction header validation
- Add tests for various "invalid" transactions
- Make `TestRunner` visible to Scrypto to simplify testing
- Use compressed public key and ED25519 crypto


transaction/src/model/transaction.rs
```diff
@@ -0,0 +1,145 @@
...
+use sbor::*;
+use scrypto::buffer::{scrypto_decode, scrypto_encode};
+use scrypto::crypto::{hash, EcdsaPublicKey, EcdsaSignature, Hash};
+
+use crate::manifest::{compile, CompileError};
+use crate::model::Instruction;
+
+#[derive(Debug, Clone, TypeId, Encode, Decode, PartialEq, Eq)]
+pub enum Network {
+    InternalTestnet,
```
talekhinezh, line 12: Not sure if we need this as a separate field as we could use the V byte in the signature to save a lot of space?

talekhinezh (approved):

PR #334, opened 2022-06-07
Refactor scrypto abi

talekhinezh:


radix-engine/src/engine/call_frame.rs
```diff
@@ -526,7 +526,7 @@ where
                 .map_err(RuntimeError::WorktopError),
             SNodeState::Scrypto(actor, package, export_name, component_state) => {
                 self.component_state = component_state;
-                package.invoke(actor, export_name, call_data, self)
+                package.invoke(actor, export_name, method_name, call_data, self)
```
iamyulong, line 529: Hate to talk about naming, but I think we should be more consistent about when to use "method" and "function", since there is a distinction at Scrypto language level. It's quite messy in the code base now.

Sometimes, the variable can be either a function or method, while in other places, only method is applicable.

Looking at the [official Rust Reference](https://doc.rust-lang.org/reference/items/associated-items.html#:~:text=Associated%20functions%20and%20methods,the%20name%20of%20the%20function.),
> Associated functions are functions associated with a type.

> Associated functions whose first parameter is named self are called methods and may be invoked using the method call operator

There are three terms "function", "associated function" and "method".

Moving forward, I think should pick one of the following approch:
1. Continue Scrypto's definition and use `function_or_method`, `function` (without self) and `method` (with self)
2. Use `function` everywhere and potentially merge `call function` and `call method` into `call` in Scrypto

Similarly, we have `call_data`, `arg`, `args` and `input` in the code base, which also needs a clean-up.

Happy to talk about alternatives as long as they are consistent. :)

What do you think?

iamyulong (approved):
Overall looks good to me.

radix-engine/src/engine/call_frame.rs
talekhinezh: Good point! I'm leaning towards using the same naming scheme as used in the rust `syn` library so maybe something like `fn_ident` and `input`?

PR #341, opened 2022-06-08
Refactor value references

talekhinezh:
* Refactored value references so that they are only available if it's parent has been loaded
* Remove the previous tree-search reference handling


radix-engine/src/engine/call_frame.rs
```diff
@@ -52,20 +52,37 @@ pub struct CallFrame<
...
     caller_auth_zone: Option<&'p AuthZone>,
 
     phantom: PhantomData<I>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ValueType {
+    Owned,
+    Ref(ValueRefType),
+}
```
iamyulong, line 73: It's strange that we have two hashmaps of values after introducing this enum.

radix-engine/src/engine/call_frame.rs
```diff
@@ -52,20 +52,37 @@ pub struct CallFrame<
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ValueRefType {
+    Uncommitted {
+        root: KeyValueStoreId,
+        ancestors: Vec<KeyValueStoreId>,
```
iamyulong, line 79: Minor: can be merged into `path: Vec<KeyValueStoreId>`?

iamyulong, line 79: Minor: can be merged into `path: Vec<KeyValueStoreId>`?

radix-engine/src/engine/call_frame.rs
```diff
@@ -568,6 +592,95 @@ where
...
+    ) -> Result<Vec<StoredValue>, RuntimeError> {
+        let values = self.take_set(value_ids)?;
+        for value in &values {
+            if let StoredValue::KeyValueStore(_, store) = value {
+                for id in store.all_descendants() {
+                    self.refed_values.remove(&id);
+                }
+            }
+        }
+        Ok(values)
```
iamyulong, line 608: Time to remove flat map of local vaults and kv stores, in favor of a tree of values?

```
Frame 
|- bucket_id => bucket
|- proof_id => proof
|- vault_id => vault
|- kv_store_id => kv_store 
             |- key => value
             |- vault_id => vault
             |- kv_store_id => kv_store
                     |- key => value
                     |- vault_id => vault
                     |- kv_store_id => kv_store
```

All these vault id and kv store id should be `u32` instead of `(Hash, u32)`. To access a "local" vault/kv_store, one must specify the full path; accessing local vault and globalized vault should be different APIs.

Another interesting thing we can then enforce is parent can not be globalized before all children are globalised (expect entries).

iamyulong, line 608: Time to remove flat map of local vaults and kv stores, in favor of a tree of values?

```
Frame 
|- bucket_id => bucket
|- proof_id => proof
|- kv_store_id => kv_store 
             |- vault_id => vault
             |- kv_store_id => kv_store
                     |- vault_id => vault
                     |- kv_store_id => kv_store
```

All these vault id and kv store id should be `u32` instead of `Hash, u32`. To access a "local" vault, one must specify the full path; accessing local vault and globalized vault should be different APIs.

Another interesting thing we can potentially enforce is a lazy_map can only be globalized when the "local" children is empty.

radix-engine/src/engine/call_frame.rs
talekhinezh: I had that originally but this way encodes the invariant that there must always be a root value at the compile time level (rather than requiring use of `unwrap()` later)

radix-engine/src/engine/call_frame.rs
talekhinezh: Which two hashmaps?

radix-engine/src/engine/call_frame.rs
iamyulong: Cool

radix-engine/src/engine/call_frame.rs
iamyulong: I see. Let's leave it for now then

radix-engine/src/engine/call_frame.rs
iamyulong: Ah, looks like I misread  `owned_values` and `refed_values`. Never mind.

iamyulong (approved):

PR #342, opened 2022-06-09
Feature/cost unit counter

iamyulong:
This PR adds a reusable `CostUnitCounter`.


radix-engine/src/engine/cost_unit_counter.rs
```diff
@@ -0,0 +1,87 @@
+pub struct CostUnitCounter {
```
talekhinezh, line 1: `CostUnit` still sounds very strange to me...

radix-engine/src/engine/wasm_runtime.rs
```diff
@@ -15,41 +15,41 @@ use crate::engine::SystemApi;
...
 where
     S: SystemApi<W, I>,
     W: WasmEngine<I>,
     I: WasmInstance,
 {
-    pub fn new(this: ScryptoActorInfo, system_api: &'s mut S, tbd_limit: u32) -> Self {
+    pub fn new(
+        this: ScryptoActorInfo,
+        system_api: &'s mut S,
+        cost_unit_counter: &'c mut CostUnitCounter,
```
talekhinezh, line 42: Does it make more sense for the runtime to own the counter?

radix-engine/src/wasm/constants.rs
```diff
@@ -1,7 +1,7 @@
 pub const RADIX_ENGINE_FUNCTION_INDEX: usize = 0;
 pub const RADIX_ENGINE_FUNCTION_NAME: &str = "radix_engine";
-pub const USE_TBD_FUNCTION_INDEX: usize = 1;
-pub const USE_TBD_FUNCTION_NAME: &str = "gas";
+pub const CONSUME_COST_UNIT_FUNCTION_INDEX: usize = 1;
+pub const CONSUME_COST_UNIT_FUNCTION_NAME: &str = "gas";
```
talekhinezh, line 4: "cost unit"

talekhinezh (approved):

radix-engine/src/engine/wasm_runtime.rs
iamyulong: It does because both WASM and RE need to update the counter. Next PR :)

radix-engine/src/wasm/constants.rs
iamyulong: No way to change this without forking the library.

PR #343, opened 2022-06-09
Cleanup abi

talekhinezh:
Various cleanups to abi names and unification of abi


iamyulong:
Merging so I can leverage the changes to the wasm tests :)

PR #344, opened 2022-06-10
Add engine call metering

iamyulong:
A simple engine call metering - static, fixed fee.


radix-engine/src/engine/call_frame.rs
```diff
@@ -565,6 +587,30 @@ where
 
         Ok((output, moving_buckets, moving_proofs, moving_vaults))
     }
+
+    pub fn take_cost_unit_counter(&mut self) -> CostUnitCounter {
```
talekhinezh, line 591: Do these need to be `pub`?

radix-engine/src/engine/call_frame.rs
```diff
@@ -565,6 +587,30 @@ where
...
+        self.cost_unit_counter
+            .take()
+            .expect("Frame doesn't own a cost unit counter")
+    }
+
+    pub fn cost_unit_counter(&mut self) -> &mut CostUnitCounter {
```
talekhinezh, line 597: I'm finding this type of method signature becoming problematic as these don't capture partial borrows of `self`.

radix-engine/tests/metering.rs
```diff
@@ -146,5 +146,23 @@ fn test_grow_memory_out_of_cost_unit() {
...
+    let (pk1, _, account1) = test_runner.new_account();
+    let (_, _, account2) = test_runner.new_account();
+    let manifest = ManifestBuilder::new()
+        .withdraw_from_account(RADIX_TOKEN, account1)
+        .call_method_with_all_resources(account2, "deposit_batch")
+        .build();
+    let receipt = test_runner.execute_manifest(manifest, vec![pk1]);
+
+    // Assert
+    assert_eq!(237268, receipt.cost_units_consumed);
```
talekhinezh, line 167: This seems to be a fragile test given any change in the fee table?

talekhinezh (commented):
Can we add another test which just makes engine calls and causes metering failure because of engine calls?

radix-engine/src/engine/call_frame.rs
talekhinezh: https://users.rust-lang.org/t/how-to-partially-borrow-from-struct/32221/2

radix-engine/src/engine/call_frame.rs
iamyulong: Looks like it doesn't need to be public. Will remove

radix-engine/src/engine/call_frame.rs
iamyulong: Wow, there is a workaround for partial borrow across functions. I will give it try.

radix-engine/tests/metering.rs
iamyulong: That's true. The idea was to capture the whole output to avoid any regression issue.  A better solution is to use an expression instead of fixed number to make it less fragile, bug given the size of the wasm, it's hard to do so. 

Thinking I will use the a manually crafted wasm for this test.

talekhinezh (approved):

PR #345, opened 2022-06-10
Access Rules abi check

talekhinezh:
Add fn_ident checks on access rules


iamyulong (approved):
LGTM. I'm assuming that we will have another PR that validates the input. :)

PR #351, opened 2022-06-14
Transaction validation cost and engine run cost

iamyulong:


radix-engine/src/engine/call_frame.rs
```diff
@@ -548,6 +548,19 @@ where
         ),
         RuntimeError,
     > {
+        let remaining_cost_units = self.cost_unit_counter().remaining();
+        self.sys_log(
```
talekhinezh, line 552: Can we use macros for these sys_logs? or only borrow specific values required? borrow checker is a pain with all of these self borrows

talekhinezh (approved):

radix-engine/src/engine/call_frame.rs
iamyulong: Lol, I will do that!

PR #352, opened 2022-06-15
Feature/enhance sbor

iamyulong:


sbor/src/encode.rs
```diff
@@ -379,9 +381,10 @@ impl<K: Encode + TypeId, V: Encode + TypeId> Encode for HashMap<K, V> {
         encoder.write_type(K::type_id());
         encoder.write_type(V::type_id());
         encoder.write_len(self.len());
-        for (k, v) in self {
-            k.encode_value(encoder);
-            v.encode_value(encoder);
+        let keys: BTreeSet<&K> = self.keys().collect();
```
dhedey, line 384: Is it worth explaining why we do this with a short comment about determinism?

dhedey (approved):
Thanks, looks good!

sbor/src/encode.rs
iamyulong: Yep, I will add that

PR #353, opened 2022-06-15
Ref cells for tracked values

talekhinezh:


radix-engine/src/engine/call_frame.rs
```diff
@@ -261,29 +266,12 @@ impl<'a> LoadedSNodeState<'a> {
...
-                        let store = owned_values
-                            .get_mut(&StoredValueId::KeyValueStoreId(root))
-                            .map(|v| match v {
-                                StoredValue::KeyValueStore(_, store) => store,
-                                _ => panic!("Expected KV store"),
-                            })
-                            .unwrap();
-                        store.put_child_vault(&ancestors, vault_id, vault);
-                    }
+                BorrowedSNodeState::TrackedVault(vault_id, vault, value_type) => match value_type {
```
iamyulong, line 269: `GlobalizedVault`?

radix-engine/src/engine/call_frame.rs
```diff
@@ -261,29 +266,12 @@ impl<'a> LoadedSNodeState<'a> {
...
                     ValueType::Ref(ValueRefType::Committed { component_address }) => {
                         track
                             .return_borrowed_global_mut_value((component_address, vault_id), vault);
                     }
+                    _ => panic!("Should not get here."),
```
iamyulong, line 274: Minor: in the long run, we should be careful with documenting our assumption. Instead of saying it's not reachable, we should say why it's not reachable and enforced by what. For this specific case, I would document "Tracked vaults are owned by Track and only references are passed to call frames"

iamyulong, line 274: Minor: in the long run, we should be careful with documenting our assumption. Instead of saying it's not reachable, we should say why it's not reachable and enforced by what. For this specific case, I would document "Tracked vaults are owned by Track and only references are passed to call frames" although it's quite obvious.

iamyulong (approved):
Looks good to me with minor comments

radix-engine/src/engine/call_frame.rs
talekhinezh: We reserve "globalize" for things that are accessible to all. Will keep it track for now

PR #354, opened 2022-06-15
Separate extern functions

talekhinezh:


radix-engine/src/wasm/prepare.rs
```diff
@@ -161,6 +163,32 @@ impl WasmModule {
...
+                    x.field().eq(func_name) && matches!(x.internal(), Internal::Function(_))
+                }) {
+                    return Err(PrepareError::MissingExport {
+                        export_name: func_name.to_string(),
+                    });
+                }
+            }
+        }
+
+        Ok(self)
```
iamyulong, line 189: Should we reject exports that's not specified in the ABI? :)

radix-engine/tests/abi/src/lib.rs
```diff
@@ -25,12 +25,12 @@ blueprint! {
 }
 
 #[no_mangle]
-pub extern "C" fn AbiComponent2_main(_input: *mut u8, _input2: *mut u8) -> *mut u8 {
+pub extern "C" fn AbiComponent2_main(_input: *mut u8) -> *mut u8 {
```
iamyulong, line 28: Still `Blueprint_main`? Are we supporting a hybrid mode?

iamyulong, line 28: Still `Blueprint_main`?

scrypto-derive/src/blueprint.rs
```diff
@@ -278,23 +243,28 @@ fn generate_dispatcher(bp_ident: &Ident, items: &[ImplItem]) -> Result<(Vec<Expr
...
-                    attrs: vec![],
-                    label: None,
-                    block: Block {
-                        brace_token: Brace {
-                            span: Span::call_site(),
-                        },
-                        stmts,
-                    },
-                }));
+                let function_ident = format_ident!("{}_{}_main", bp_ident, ident);
```
iamyulong, line 248: Should we get rid of the `_main` suffix?

iamyulong, line 248: Should we get rid of the `_main` suffix?

radix-engine/tests/abi/src/lib.rs
talekhinezh: ah, it's just another level of indirection, ie every `fn_ident` must map to some extern function (in this case it is to the same function).

scrypto-derive/src/blueprint.rs
talekhinezh: That's a good idea

radix-engine/src/wasm/prepare.rs
talekhinezh: hrm do you think that gives us more security? There may be interesting usecases if we allowed other exports?

iamyulong: Not really, just more control over the structure of the wasm. Let's keep it as-in for now.

iamyulong (approved):

PR #357, opened 2022-06-17
Enhance wasm validation

iamyulong:
This PR enhances wasm validation by introducing more constraints on the wasm module.


talekhinezh (commented):
Looks good but can we add tests for all of the wasm validation errors. For example, there isn't a test for `EmptyMemorySection`

PR #368, opened 2022-06-24
Merge Buckets/Proofs into owned_values

talekhinezh:


radix-engine/src/engine/call_frame.rs
```diff
@@ -143,17 +203,61 @@ fn verify_stored_key(value: &ScryptoValue) -> Result<(), RuntimeError> {
     Ok(())
 }
 
-pub enum ConsumedSNodeState {
-    Bucket(Bucket),
-    Proof(Proof),
+pub enum REValueRef<'a> {
+    Owned(RefMut<'a, REValue>),
```
iamyulong, line 207: What is an "owned" ref?

iamyulong (approved):
LGTM

radix-engine/src/engine/call_frame.rs
talekhinezh: Lol different levels of abstraction here:
* `Owned` part is the underlying value is "owned" by some frame
* `REValueRef` part is the fact that it's a reference to some value (which may be owned by a frame or in track)

PR #387, opened 2022-07-08
Typed components

talekhinezh:


radix-engine/tests/local_component/src/lib.rs
```diff
@@ -14,7 +14,7 @@ blueprint! {
             self.secret = next;
         }
 
-        pub fn new(secret: u32) -> Component {
+        pub fn new(secret: u32) -> crate::LocalComponent {
```
iamyulong, line 17: Do developers have to specify the full path after this PR?

iamyulong:
Looks good to me. 

The only point I'd like to have a discussion on is whether we should use different names for types:
- Component state struct
- Local component
- Global component

I would propose to use different names for these types for explicitness, e.g.
- `GumballMachine`
- `GumballMachineLocal` / `LocalComponent<GumballMachine>`
- `GumballMachineGlobal` / `GlobalComponent<GumballMachine>`

iamyulong (approved):

PR #389, opened 2022-07-11
Feature/bech32 support

0xOmarA:


transaction/src/manifest/decompiler.rs
```diff
@@ -28,7 +32,10 @@ pub fn decompile(manifest: &TransactionManifest) -> Result<String, DecompileErro
                 buckets.insert(bucket_id, name.clone());
                 buf.push_str(&format!(
                     "TAKE_FROM_WORKTOP ResourceAddress(\"{}\") Bucket(\"{}\");\n",
-                    resource_address, name
+                    bech32_encoder
+                        .encode_resource_address(&resource_address)
+                        .unwrap(),
```
iamyulong, line 37: Should return a DecompileError and similarly for other unwraps in this file.

0xOmarA: Yup, changed that. Will push the changed with the `address` macro once I'm done :)

radix-engine/tests/metering.rs
```diff
@@ -157,7 +157,7 @@ fn test_total_cost_units_consumed() {
     Cost analysis:
     1. Transaction validation cost = TX_VALIDATION_COST_PER_BYTE * 1
     2. Engine run cost
-       * invoke_function: 4515
+       * invoke_function: 4520
```
0xOmarA, line 160: @iamyulong Since you are working on the fees stuff, I wanted to tag you in this to make sure that everything here is correct. 

The fees seem to have increased since our addresses are not 27 bytes long instead of 26. I have updated the number here to reflect that. Does this look good to you based on your calculations?

scrypto-tests/tests/address.rs
```diff
@@ -0,0 +1,436 @@
...
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+        0,
+    ]
+}
```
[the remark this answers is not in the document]
0xOmarA: I've added the `address` macro but kept this method here since the macro creates an address with a correct entity byte and here I'm testing for addresses with mismatch entity bytes :)

radix-engine/tests/metering.rs
iamyulong: That's understandable. Part of invoke_function cost comes from payload size. The delta `10` seems to match the address size change.

iamyulong (approved):

PR #390, opened 2022-07-12
System as value

talekhinezh:
- Implement epoch as part of system substate
- Remove get/set epoch from substate store


scrypto/src/constants.rs
```diff
@@ -25,3 +25,8 @@ pub const RADIX_TOKEN: ResourceAddress = ResourceAddress([
 pub const ECDSA_TOKEN: ResourceAddress = ResourceAddress([
     0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5,
 ]);
+
+/// The system token which only validators have access to
```
dhedey, line 29: It's not just validators - it's more the system executor, or I think we called them "system transactions", I believe?

radix-engine/src/engine/track.rs
```diff
@@ -251,6 +256,22 @@ impl SubstateValue {
...
+        } else {
+            panic!("Not a system value");
+        }
+    }
+
+    pub fn system_mut(&mut self) -> &mut System {
+        if let SubstateValue::System(system) = self {
+            system
+        } else {
+            panic!("Not a vault");
```
dhedey, line 271: Not a system value*?

talekhinezh: good catch

iamyulong (approved):
LGTM

PR #395, opened 2022-07-18
Bugfix: transaction network fees

0xOmarA:
The `transaction_netwotk()` method in the system API was implemented in the middle of the implementation of fees; therefore, it accidentally did not consume any fees. This PR makes it consume fees equal to the amount consumed by the `transaction_hash()` since both are cashed by the track.


iamyulong (approved):
LGTM

PR #398, opened 2022-07-19
Recursive track

talekhinezh:


radix-engine/src/state_manager/staging.rs
```diff
@@ -0,0 +1,216 @@
+use sbor::rust::collections::*;
+use sbor::rust::vec::Vec;
+use scrypto::buffer::{scrypto_decode, scrypto_encode};
+
+use crate::ledger::*;
+
+/// Nodes form an acyclic graph towards the parent
+struct StagedExecutionStoreNode {
+    parent_id: u64,
```
iamyulong, line 9: Out of curiosity, why do we need `parent_id`? Looks like the id is maintained by the owner.

radix-engine/src/state_manager/staging.rs
```diff
@@ -0,0 +1,216 @@
...
+            spaces: BTreeMap::new(),
+            outputs: BTreeMap::new(),
+        }
+    }
+}
+
+/// Structure which manages the acyclic graph
+pub struct StagedExecutionStores<'s, S: ReadableSubstateStore + WriteableSubstateStore> {
+    parent: &'s mut S,
+    nodes: HashMap<u64, StagedExecutionStoreNode>,
```
iamyulong, line 29: Can we have a proper tree data structure instead of `HashMap`?

radix-engine/src/state_manager/staging.rs
```diff
@@ -0,0 +1,216 @@
...
+        if !remove_children {
+            self.set_root_parent(id);
+        }
+    }
+}
+
+pub struct ExecutionStore<'t, 's, S: ReadableSubstateStore + WriteableSubstateStore> {
+    stores: &'t mut StagedExecutionStores<'s, S>,
+    id: u64,
+}
```
iamyulong, line 117: Looks like `ExecutionStore` can be the node.

radix-engine/src/state_manager/staging.rs
```diff
@@ -0,0 +1,216 @@
...
+    fn put_substate(&mut self, address: &[u8], output: Output) {
+        if self.id == 0 {
+            self.stores.parent.put_substate(address, output);
+        } else {
+            let node = self.stores.nodes.get_mut(&self.id).unwrap();
+            node.outputs.insert(address.to_vec(), output);
+        }
+    }
+}
+
```
iamyulong, line 182: I'd also like to propose the following terms/names:

| Name                                             | Definition                                                                                                                                                                                                                |
|--------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `SubstateStore`                                  | (Not changed) the storage that physically stores substates, either in memory or in a database                                                                                                                             |
| `StagedSubstateStore`                            | A data structure that caches substate changes and points to a parent `SubstateStore` or `StagedSubstateStore` recursive substate reads/writes; it also implements the `ReadableSubstateStore` and `WriteableSubstateStore` traits. |
| `StagedSubstateStoreManager/Engine` | Manages the lifecyle of `StagedSubstateStore` and may also persisting them.

radix-engine/src/state_manager/track.rs
```diff
@@ -0,0 +1,350 @@
+use sbor::rust::collections::*;
```
iamyulong, line 1: As discussed, let's move this file back to the engine module for now.

radix-engine/tests/abi.rs
```diff
@@ -13,7 +14,8 @@ use transaction::builder::ManifestBuilder;
 #[test]
 fn test_invalid_access_rule_methods() {
     // Arrange
-    let mut test_runner = TestRunner::new(true);
+    let mut store = InMemorySubstateStore::with_bootstrap();
```
iamyulong, line 17: Why can't we have `TestRunner` owns the store? Looks easier to have a getter method to expose it if needed.

radix-engine/src/engine/call_frame.rs
```diff
@@ -118,6 +121,33 @@ fn verify_stored_key(value: &ScryptoValue) -> Result<(), RuntimeError> {
     Ok(())
 }
 
+pub fn insert_non_root_nodes<'s, S: ReadableSubstateStore>(
+    track: &mut Track<'s, S>,
+    values: HashMap<ValueId, RENode>,
+) {
+    for (id, node) in values {
+        match node {
+            RENode::Vault(vault) => {
```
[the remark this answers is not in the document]
talekhinezh: `RENode` contains one or more `Substates`. See that confluence page we went over a few weeks back.

radix-engine/src/state_manager/staging.rs
talekhinezh: Which owner?

radix-engine/src/state_manager/staging.rs
talekhinezh: We need an index into every node as well as a way to remove nodes in the middle of the tree. Is there a Rust data structure for that?

radix-engine/src/state_manager/staging.rs
talekhinezh: Can't because we need a reference to the entire store.

radix-engine/src/state_manager/staging.rs
talekhinezh: These sound good.

radix-engine/src/state_manager/track.rs
talekhinezh: Yup still need to make this change.

radix-engine/tests/abi.rs
talekhinezh: I needed to be able to hold a reference in the struct of `TestRunner` to the `store` for the purpose of staged execution. This was the simplest idea I could think of.

radix-engine/src/state_manager/staging.rs
iamyulong: That's exactly one of the reasons we need tree structures. 

It can be as simple as:
```
Node {
   node_state: BlaBla,
   children: Vec<Node>
}
```

radix-engine/tests/abi.rs
iamyulong: Is it possible to get a reference through the `TestRunner`? If you can point me to the code, I can also help take a look.

radix-engine/src/state_manager/staging.rs
iamyulong: Sounds like resolvable by a bi-directional tree structure, but we're in Rust not Java.  It's going to be ugly to fight with the ownership system, http://way-cooler.org/blog/2016/08/14/designing-a-bi-mutable-directional-tree-safely-in-rust.html

Kindly ignore my comment.

radix-engine/tests/abi.rs
iamyulong: In my other PR, I started to use `Rc<dyn ResableSubstateStore>`, so we'll never need a strong reference. I can also do that during merging time. :)

radix-engine/src/engine/call_frame.rs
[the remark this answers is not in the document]
iamyulong: Thanks for explaining to me again!

radix-engine/src/state_manager/staging.rs
talekhinezh: Yup like mentioned, this has repercussions in terms of how do we index to each node. Like git/PMT we need a map from id => state.

radix-engine/src/state_manager/staging.rs
talekhinezh: Yup, definitely different with Rust.

radix-engine/tests/abi.rs
talekhinezh: You can, but you can't self-reference to the same struct which is what we would need to do.

iamyulong: Cool. Let's leave it as-is. It's not a big deal.

iamyulong (approved):

PR #399, opened 2022-07-21
Transaction fee infrastructure

iamyulong:
This PR implements the following pre-requisites to transaction fee payment:
* A two-layer state track system to support partial commit or rollback
* A new mechanism for auto-generating substate operation receipt
* A `pay_fee` Scrypto interface
* Some Radix Engine clean-up and refactors, notably
   * Changed to use `Address` as function input for many engine interfaces
   * Added types for `NonFungible` and `KeyValueStoreEntry` wrappers
   * Removed `SubstateStore` completely from `CallFrame` and `Track`


radix-engine/src/engine/track.rs
```diff
@@ -87,54 +80,37 @@ pub enum SubstateParentId {
...
 pub enum Address {
-    Resource(ResourceAddress),
     GlobalComponent(ComponentAddress),
     Package(PackageAddress),
-    NonFungibleSet(ResourceAddress),
-    KeyValueStore(KeyValueStoreId),
+    ResourceManager(ResourceAddress),
+    NonFungibleSpace(ResourceAddress),
+    NonFungible(ResourceAddress, Vec<u8>),
+    KeyValueStoreSpace(KeyValueStoreId),
```
talekhinezh, line 97: Can we remove "Space"? I think we're moving away from the concept of a "space".

radix-engine/src/engine/call_frame.rs
```diff
@@ -2144,4 +2187,63 @@ where
     fn fee_table(&self) -> &FeeTable {
         self.fee_table
     }
+
+    fn pay_fee(&mut self, vault_id: VaultId, amount: Decimal) -> Result<(), RuntimeError> {
```
talekhinezh, line 2191: Is the reason this is here because you can't pay fees from a local vault? If so,
1. Why not?
2. There must be a better place to put this which checks whether the call frame should have access to the vault or not

radix-engine/src/engine/state_track.rs
```diff
@@ -0,0 +1,239 @@
...
+use crate::engine::Address;
+use crate::ledger::*;
+
+use super::{
+    track::VirtualSubstateId, SubstateOperation, SubstateOperationsReceipt, SubstateParentId,
+    SubstateValue,
+};
+
+/// Keeps track of state changes that that are non-reversible, such as fee payments
+pub struct BaseStateTrack {
```
talekhinezh, line 17: There must be a way to integrate this with the current recursive track. Seems like both structures essentially do the same thing?

radix-engine/tests/state_track.rs
```diff
@@ -0,0 +1,142 @@
...
+        .call_method_with_all_resources(other_account, "deposit_batch")
+        .build();
+    let receipt = test_runner.execute_manifest(manifest, vec![public_key]);
+
+    // Assert
+    receipt.expect_success();
+    let mut expected_downs = HashSet::new();
+    expected_downs.extend([
+        substate_id!(
+            "5feceb66ffc86f38d952786c6d696c79c2dbc239dd4e91b46729d73a27fb57e9",
```
talekhinezh, line 37: The current substate_ids aren't very stable yet (most likely to change soon). Is there another less fragile way to test without knowing "expected_down" ids?

radix-engine/src/engine/call_frame.rs
iamyulong: This is to prevent people from draining XRD borrowed through flash loan.

radix-engine/src/engine/state_track.rs
iamyulong: Not sure if we need recursiveness in transactional state track. Also, the difference is it doesn't care about substate ID.

radix-engine/tests/state_track.rs
iamyulong: Yes, this is ugly at the moment. I will try to enhance it once I have more fee logic implemented; then I can inspect the downed fee substates.

At the moment, it's more of a testing whether (not what) states have been rolled back/committed.

radix-engine/src/engine/call_frame.rs
```diff
@@ -1067,10 +1072,20 @@ where
...
+            "Invoking: {:?} {:?}",
+            snode_ref,
+            &fn_ident
+        );
+
         if self.depth == MAX_CALL_DEPTH {
             return Err(RuntimeError::MaxCallDepthLimitReached);
         }
 
+        let is_pay_fee = matches!(snode_ref, SNodeRef::VaultRef(..)) && &fn_ident == "pay_fee";
```
talekhinezh, line 1087: I think this should probably be within the `VaultRef` statement rather than outside of the `match` statement?

radix-engine/src/engine/call_frame.rs
```diff
@@ -1577,15 +1596,29 @@ where
                         .cloned()
                         .ok_or(RuntimeError::ValueNotFound(ValueId::Vault(*vault_id)))?
                 };
+                if is_pay_fee && !matches!(cur_location, REValuePointer::Track { .. }) {
```
talekhinezh, line 1599: Could just have the full `is_pay_fee` check here.

radix-engine/src/engine/call_frame.rs
```diff
@@ -1708,27 +1741,53 @@ where
...
+            // TODO: refactor after introducing `Lock` representation.
+            self.track
+                .release_lock(l.clone(), is_pay_fee && matches!(l, Address::Vault(..)));
         }
 
-        // move buckets and proofs to this process.
-        for (id, value) in received_values {
-            trace!(self, Level::Debug, "Received value: {:?}", value);
-            self.owned_values.insert(id, value);
+        if is_pay_fee {
```
talekhinezh, line 1754: Not sure why this part is here instead of in the method itself?

radix-engine/src/engine/call_frame.rs
iamyulong: Commented in another thread.

radix-engine/src/engine/call_frame.rs
iamyulong: Originally trying to avoid overloading the `SystemApi` interface to support fee accounting, but might be a good idea to have the logic within `Vault` model.

Will refactor.

talekhinezh (approved):

radix-engine/src/engine/values.rs
```diff
@@ -1,52 +1,32 @@
...
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+/// Represents a Radix Engine address. Each maps a unique substate key.
+///
+/// TODO: separate space addresses?
+///
+/// FIXME: RESIM listing is broken ATM.
+/// By using scrypto codec, we lose sorting capability of the address space.
+/// Can also be resolved by A) using prefix search instead of range search or B) use special codec as before
+#[derive(Debug, Clone, TypeId, Encode, Decode, PartialEq, Eq, Hash, PartialOrd, Ord)]
```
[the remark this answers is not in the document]
iamyulong: Yep, I have a FIXME there. Will sort this out then.

PR #401, opened 2022-07-25
Allow tx preview to run without paying fees

LukasGasior1:


radix-engine/src/engine/call_frame.rs
```diff
@@ -555,7 +555,7 @@ where
         track: &'g mut Track<'s, S>,
         wasm_engine: &'g mut W,
         wasm_instrumenter: &'g mut WasmInstrumenter,
-        cost_unit_counter: &'g mut CostUnitCounter,
+        cost_unit_counter: &'g mut dyn CostUnitCounter,
```
talekhinezh, line 558: Any reason to use dynamic dispatch here?

radix-engine/tests/test_runner.rs
```diff
@@ -9,15 +12,21 @@ use scrypto::core::Network;
...
+use transaction::validation::{
+    EpochManager, IntentHashManager, TestEpochManager, TestIntentHashManager,
+};
 
 pub struct TestRunner {
     substate_store: InMemorySubstateStore,
     wasm_engine: DefaultWasmEngine,
     wasm_instrumenter: WasmInstrumenter,
+    epoch_manager: Box<dyn EpochManager>,
+    intent_hash_manager: Box<dyn IntentHashManager>,
```
talekhinezh, line 29: Also here, why `dyn`?

radix-engine/tests/preview.rs
```diff
@@ -0,0 +1,97 @@
...
+use radix_engine::fee::{CostUnitCounterError, DEFAULT_MAX_TRANSACTION_COST};
+use scrypto::core::Network;
+use transaction::builder::ManifestBuilder;
+use transaction::builder::TransactionBuilder;
+use transaction::model::*;
+use transaction::signing::EcdsaPrivateKey;
+use transaction::validation::{TestEpochManager, TestIntentHashManager, TransactionValidator};
+
+#[test]
+fn test_transaction_preview_cost_estimate() {
```
talekhinezh, line 15: Can we split test to be more specific about what is being tested and use Arrange/Act/Assert pattern?

LukasGasior1: I've reorganized it a bit. Is: Arrange, (Act & Assert #1), (Act & Assert #2) acceptable? I think it's more readable that way.

LukasGasior1: on second thought, that first assert probably isn't needed anymore, since the preview config doesn't even specify the system loan amount

talekhinezh (approved):
Thanks for the changes!

PR #404, opened 2022-07-27
Cleanup output ids

talekhinezh:
Replaced the `OutputId` derived from a CommittedTransactionHash with a simple incrementing version. This is possible with the assumption that `Address` is itself a unique identifier.

This simple change also allows us to remove an unneeded `spaces` database which we needed previously to maintain the `OutputId` of the space.


iamyulong (approved):
Looks good to me.

Could you also do a global find & replace to change `output_id` to `version`? :)

talekhinezh:
As discussed, `output_id` includes `version` so no need to replace.

PR #405, opened 2022-07-28
Feature: Checking resource behavior in Scrypto

0xOmarA:
This PR adds the ability to check the resource behavior in Scrypto through the `ResourceManager` adding methods such as:

```rust
let resource_manager: &ResourceManager = borrow_resource_manager!(RADIX_TOKEN);
println!(
    "Is this token forever non-mintable?: {}", 
    resource_manager.mint_behavior().is_permanently_disabled()
);
```


talekhinezh:
I think it's a good feature but I think we may want to discuss the interface to this? Specifically I can see this being better implemented on the scrypto side rather than the radixEngine side by just reading the state of the ResourceManager.

0xOmarA:
@talekhinezh Thanks for the feedback! To clarify, by "reading the state of the ResourceManager" do you mean that you'd like the radix engine to return the `AccessRule` associated with the specified resource method to the Scrypto side which understands it and process it into nice boolean values?

This was the original design that I had in mind, but I found that I'd need to translate the `MethodAuthorization` to an `AccessRule` where in some of the cases there isn't a 1:1 mapping. If you think its worth it, I can implement the translation.

iamyulong:
What's the current state of this PR? Are we pending review?

0xOmarA:
@iamyulong I will follow up with Josh about it :)

0xOmarA:
Closing this for now, will rethink the approach.

PR #406, opened 2022-07-29
Feature: Validated Proofs

0xOmarA:
This PR adds `ValidatedProof`s as a language level concept to Scrypto. With this PR, a `Proof` would have two main methods:

* `validate_proof`: Takes in a `ResourceAddress` and validates that the proof has a matching resource address.
* `unsafe_skip_proof_validation`: Takes no arguments and simply skips the validation of the resource address.

All of the methods that were previously available on `Proof`s are now available on `ValidatedProof`s.


devmannic:
This concept could go say much farther to actually enforce secure coding.  As implemented here it doesn't seem to add much value and has many ways of failing.

If this concept is to be added at the scrypto level id really like to see a better more secure design.  My static types library goes much farther, but if something with less complexity is wanted, I think there is still room for improvement to this.

Are you open to a better design here, or is this already decided?

0xOmarA:
@devmannic Wow that was quick! :D 

> has many ways of failing.

Can you elaborate a little bit more on what you mean by that? 

> Are you open to a better design here, or is this already decided?

Yeah we are very open to better design here. We wanted `ValidatedProof`s to be a temporary solution to the proof validation problem until we figured out a more long term solution for the issue. 

That being said, I'm open to better design here and would love to get any feedback that you have :)

devmannic:
> @devmannic Wow that was quick! :D

My security spidy sense must have gone off. ;)

> > has many ways of failing.
> 
> Can you elaborate a little bit more on what you mean by that?

Yeah, and this is really my main critique of the design.  My starting assumption is that the entire point of providing a `ValidatedProof` is for security.  So, any way that a ValidatedProof can be created and *not* actually validate the proof resource address would be a design or implementation failure.  There's absolutely no good reason to have any trapdoors, implicit or explicit to create a ValidatedProof that is not actually validated.

So as a result we have a few problems:

1. the `try_into` implementation seems like an unintentional way to create a ValidatedProof without validating the address.  This should be fixed and isn't too hard.
2. The `unsafe_skip_proof_validation` is an intentional trapdoor, which should be removed

> > Are you open to a better design here, or is this already decided?
> 
> Yeah we are very open to better design here. We wanted `ValidatedProof`s to be a temporary solution to the proof validation problem until we figured out a more long term solution for the issue.

It's fine to think of things as temporary, and so maybe not wanting to over complicate, and knowingly miss some use cases, but for APIs specifically designed for security, every "legacy" API  which doesn't do something it should do from a security standpoint is just a future vulnerability, yet it's rarely considered that way by designers when first created.  Keep that in mind.

> That being said, I'm open to better design here and would love to get any feedback that you have :)

So in addition to the changes mentioned above to ensure that any instantiated `ValidatedProof` is actually validated, I would add the following:

While I think `sbor::Encode` and `sbor::Decode` etc implementations are important for a good API, I can see not wanted to add all of that at this stage.  But I would highly suggest then to separate concerns a bit add a new `UnValidatedProof` (you can think of a better name).  This would be something that *does have*  `Decode` and uses the same techniques I have created in scrypto_statictypes to decode a regular `Proof` into a `UnValidatedProof` transparently.  This means it would sit in the function parameters in place of a raw `Proof`.  This is one way a developer indicates they plan to use `ValidatedProof`s and most importantly creates some guarantees.  Then, this `UnValidatedProof` simply wraps the real Proof (but not pub) and instead of providing all of the methods (as my library does for typed proofs) it would *only have the method* `validate_proof` which acts like the current method, returning a `ValidatedProof`.  This provides devs and auditors a way to ensure the check has in fact been made because there's no other way to use the proofs content.  (Though it could still pass it along to another function, see next paragraph).

sbor Encode could also be defined for `UnValidatedProof` but it's less code to maintain to not do that when simply implementing From/Into so that `UnValidatedProof` can be changed `.into()` `Proof`. for returning from functions.  So implementing a transparent `Encode` is optional.

Also, I will add that if you're going to do all this and have these "wrappers" for Proof anyway, please just implement an "auto drop" already as I've done in static types so we don't need to manually do `proof.drop()` everywhere.  It would be a nice feature of `UnValidatedProof` and `ValidatedProof` without changing how `Proof` behaves wrt. `Drop` so it's a good opportunity to do this in a backwards compatible way (even though I'd much rather see auto drop on `Proof`).

That's all I can think of for now, but might have some more ideas.  What do you think?  Trying to keep this self contained while still providing safety and value.  I think these changes do that.

devmannic:
The more I think about the more I think implementing Encode is actually worth it for compatability with other components, and it removes the hole created by `into()` `Proof` which would no longer be necessary

0xOmarA:
>My starting assumption is that the entire point of providing a  `ValidatedProof`  is for security. 

You are indeed correct, the entire point of `ValidatedProof`s is security and to (as much as possible) ensure that developers are checking the resource address of the passed proof and not blindly trusting it.

> So, any way that a ValidatedProof can be created and  _not_  actually validate the proof resource address would be a design or implementation failure. There's absolutely no good reason to have any trapdoors, implicit or explicit to create a ValidatedProof that is not actually validated.
> So as a result we have a few problems:
> 1.  the  `try_into`  implementation seems like an unintentional way to create a ValidatedProof without validating the address. This should be fixed and isn't too hard.
> 2.  The  `unsafe_skip_proof_validation`  is an intentional trapdoor, which should be removed

For point 1, if you are talking about the about the `Into<Proof>` trait that is implemented for the `ValidatedProof` type. This is intentional and allows people to go from a validated proof to an unvalidated proof.

If you are talking about the `TryFrom<&[u8]>` for `ValidatedProof`s, then this is indeed unintentional and should indeed be removed as it indeed allows for the creation of ValidatedProofs without any validation happening.

For point 2, the reason that the `unsafe_skip_proof_validation` method is there is because there will be certain cases where the developer wishes to perform their own validation on the `Proof`. An example of this is as follows: a dApp could have a list of `ResourceAddress`es that it allows `Proof`s from. So, instead of trying to validate the proof against the entire list of resource addresses, the developer might choose to skip the build-in validation process in favor of their own validation process.

My main worry with removing the `unsafe_skip_proof_validation` method would be that it could make certain DeFi apps harder to model in Scrypto as the data in the `Proof` now becomes to difficult to obtain even when the developer knows what they're doing. 

> While I think `sbor::Encode` and `sbor::Decode` etc implementations are important for a good API, I can see not wanted to add all of that at this stage.

Regarding the `sbor::Encode` and `sbor::Decode` traits, they are not implemented on `ValidatedProof`s as each individual components are expected to perform the `Proof` validation on their own. So, a `ValidatedProof` is never meant to be used as an argument or return of a public function. 

> This would be something that _does have_  `Decode` and uses the same techniques I have created in scrypto_statictypes to decode a regular `Proof` into a `UnValidatedProof` transparently. This means it would sit in the function parameters in place of a raw `Proof`. This is one way a developer indicates they plan to use `ValidatedProof`s and most importantly creates some guarantees.

This might be the way that we choose to go if `ValidatedProof`s will remain as a concept in the long run.

talekhinezh (approved):

devmannic:
> Thank you for the super detailed feedback! I really appreciate that.

You're welcome!

> If you are talking about the `TryFrom<&[u8]>` for `ValidatedProof`s, then this is indeed unintentional and should indeed be removed as it indeed allows for the creation of ValidatedProofs without any validation happening.

Yup, this one, and it looks like you already removed it, which I think is the right thing.

 
> For point 2, the reason that the `unsafe_skip_proof_validation` method is there is because there will be certain cases where the developer wishes to perform their own validation on the `Proof`. An example of this is as follows: a dApp could have a list of `ResourceAddress`es that it allows `Proof`s from. So, instead of trying to validate the proof against the entire list of resource addresses, the developer might choose to skip the build-in validation process in favor of their own validation process.

If this is the only reason, I think there are much better ways for a dev to test more than one `ResourceAddress`.  A typical approach in Rust (which I think is a good one here) would be to provide a method which takes a `Iterator<ResourceAddress>` as a parameter in addition to the existing `validate_proof`.  Something like `validate_proof_with_any(iter: Iterator<ResourceAddress>)`.  Then devs can easily create an iterator using typical functional programming style using any technique they want to specify the list of valid addresses.  Or they can simply call `validate_proof` in a loop, and do so as many types as they want with whatever logic.  That is slightly more verbose but has the advantage of only supporting a single simple method call.  None of that is a compelling reason to have a "unsafe" trapdoor function IMHO.

> 
> My main worry with removing the `unsafe_skip_proof_validation` method would be that it could make certain DeFi apps harder to model in Scrypto as the data in the `Proof` now becomes to difficult to obtain even when the developer knows what they're doing.

I'm not really seeing the issue here.  Since `ValidatedProof` is not intended to pass (pub) function/method boundaries anyway I dont really see how not providing a "skip" option is in any way a problem.  If the dev wants to do some more complex checking then they could just use `Proof` anyway with no less security.  why make a `ValidatedProof` handle a use case `Proof` already handles perfectly and risk undermining the actual functionality and value of `ValidatedProof`  I really do not see any compelling reason for the `unsafe_skip_proof_validation` function to exist, it's all liability.

> > This would be something that _does have_  `Decode` and uses the same techniques I have created in scrypto_statictypes to decode a regular `Proof` into a `UnValidatedProof` transparently. This means it would sit in the function parameters in place of a raw `Proof`. This is one way a developer indicates they plan to use `ValidatedProof`s and most importantly creates some guarantees.
> 
> This might be the way that we choose to go if `ValidatedProof`s will remain as a concept in the long run.

I still don't get why it would not remain a concept.  Having a way to consistently,  in the "standard library" of scrypto, validate Proofs (and actually Buckets too) would be immensely beneficial for security and usability.  FWIW, I'd much rather see more machinery to actually make it easy to test multiple things in a Proof, not just the ResourceAddress in a convenient repeatable way.  That would go tons farther to getting security right on argument validation, incentivizes devs to use a new API instead of rolling their own comparisons, etc.  Something the the spirit of the `BucketRequirement` I added in hareswap I think is massively more useful than this version of `ValidatedProof` as a way to check a Proof for some criteria.  Granted that was for Buckets not Proofs, it could be adapted, and in fact I'd love to see a single "ResourceRequirement" type which does what that does for both Buckets and Proofs as a scrypto builtin.  You could even take the concept further by using the same approach as in AccessRules with the path strings which map to blueprint fields so you can make "soft" checks and add higher order logic if you wanna go nuts.  All this to say that if the goal is having devs correctly check incoming proofs, then a lot more actually needs to be done than what is currently proposed here.  Just putting a band-aid on the resource address check I think actually really misses an opportunity for more secure coding conventions and built-in implementations.

So for `UnValidatedProof` I actually I don't really see a point to "adding it later".  Without it I see no reason to even use `ValidatedProof` as it exists here.  The `ValidatedProof` as it exists in this implementation now I think is of really really limited value.  It can be replaced with a single line `if statement` and most types are elided in Rust anyway so asking new developers to understand this new type and API for what?  All devs already understand an if statement as a common language construct, this is more complexity, and cognitive load for almost no value. Of course the hope is that devs learn to use this for security, but you could already say that for the if statement.  I think the only way any of this is useful is if there are guarantees that the checks are made.  So there needs to be some "gated logic" to ensure checks are made at all (as in the `UnValidatedProof` concept) as well as no trapdoors that immediately undermine the entire approach.  Without the gated logic "guarantees" this is nothing more than an if statement by another name at best, and potentially a lot more baggage and room for error at worst.


So in summary, implement `UnValidatedProof` (or something like it) and remove `unsafe_skip_proof_validation` otherwise I think these changes as implemented are either net zero, or even net negative for overall security and usability, especially when considering support tail and developer impact for an API that adds very little. 

I think it's easy to be tricked into the "less is more" mentality with a change like this, but I really think doing nothing is better than doing what is currently in this PR.  That's not a slight against the @0xOmarA to be clear, and it's not even a problem with the overall idea, just the specific implementation (gaps) that area here.  I know RDX likes to iterate, but getting close to the Scrypto version that will be live at Babylon and I really hate the idea of an "unsafe" API which is little more than a rename of "if" like this being added and then being somewhat stuck with it, and actually more importantly I wouldn't want to lose out on the opportunity to train a cadre of new devs to do checks in a better way from the start.

I hope this didn't come across too harshly.  Just some strong feeling on how insecure software is born and I really don't want to see that happen for Radix since security a main pilliar of Scrypto and the more done right as early as possible the better, not just at the Radix Engine level but in terms of what real developers are going to produce it is really important for the ecosystem to get things like this right.

0xOmarA:
@devmannic I really appreciate you taking the time to continue this discussion :)

> If this is the only reason, I think there are much better ways for a dev to test more than one ResourceAddress. A typical approach in Rust (which I think is a good one here) would be to provide a method which takes a Iterator<ResourceAddress> as a parameter in addition to the existing validate_proof. Something like validate_proof_with_any(iter: Iterator<ResourceAddress>). Then devs can easily create an iterator using typical functional programming style using any technique they want to specify the list of valid addresses. Or they can simply call validate_proof in a loop, and do so as many types as they want with whatever logic. That is slightly more verbose but has the advantage of only supporting a single simple method call. None of that is a compelling reason to have a "unsafe" trapdoor function IMHO.

I quite like the idea with the iterator for people who would like to validate the resource address against a set of `ResourceAddress`es. This might be something that we add either in this PR or a future PR, but it makes sense especially since we're introducing fees with v0.5.0 so a fee-efficient method to validate the proof against a list of resource addresses might be a good thing to add here. 

I agree with you that there should not be a way to create `ValidatedProof`s without actually performing the validation. Long term, if `ValidatedProof`s ever become scrypto types that can actually be passed to methods and returned from methods then I agree that this would become an issue as methods now have a way to "lie" to other methods about having validated a `Proof`. This only becomes an issue once `ValidatedProof`s becomes a valid argument and return type. However, as you have said, it is better to deal with such an issue before it even happens. 

I am trying to understand your proposal a little better and how the `UnvalidatedProof` type tied into all of it. If I understood your proposal correctly, then you are proposing that we add an `UnvalidatedProof` type, such that all proofs begin their life as a `Proof` and then either get validated into `ValidatedProof` or have validation skipped and become `UnvalidatedProof`. Are you thinking that the `ValidatedProof` and `UnvalidatedProof` types should be identical? Both wrapping a `Proof` and exposing the same methods? 

Let me know what you are thinking here and we are certainly very open to improving this concept :) 

For the time being, since the `ValidatedProof` (and potentially `UnvalidatedProof`) are only meant to be used inside methods (and not as arguments or return types) then I would be hesitant to add `sbor::Encode` or `sbor::Decode` now. However, I understand and see the usefulness of having them, but we need to evaluate the idea of validated proofs and determine how proof validation will progress. 

> I still don't get why it would not remain a concept. Having a way to consistently, in the "standard library" of scrypto, validate Proofs (and actually Buckets too) would be immensely beneficial for security and usability.

I do not believe that the idea of validating proofs is going to go away, the `ValidatedProof` type is our first attempt at adding some kind of proof validation to scrypto. If the `ValidatedProof` type ever goes away at some point of time, then it would probably be replaced by a better way of doing proof validation and not something in the opposite direction. 

> If the dev wants to do some more complex checking then they could just use `Proof` anyway with no less security.

Can you please elaborate on this more and about how it ties into your proposal? Do you propose that the `Proof` type should have methods to get the resource address, non fungible id, data, etc...?

> I still don't get why it would not remain a concept. Having a way to consistently, in the "standard library" of scrypto, validate Proofs (and actually Buckets too) would be immensely beneficial for security and usability. FWIW, I'd much rather see more machinery to actually make it easy to test multiple things in a Proof, not just the ResourceAddress in a convenient repeatable way. That would go tons farther to getting security right on argument validation, incentivizes devs to use a new API instead of rolling their own comparisons, etc. Something the the spirit of the `BucketRequirement` I added in hareswap I think is massively more useful than this version of `ValidatedProof` as a way to check a Proof for some criteria. Granted that was for Buckets not Proofs, it could be adapted, and in fact I'd love to see a single "ResourceRequirement" type which does what that does for both Buckets and Proofs as a scrypto builtin. You could even take the concept further by using the same approach as in AccessRules with the path strings which map to blueprint fields so you can make "soft" checks and add higher order logic if you wanna go nuts. All this to say that if the goal is having devs correctly check incoming proofs, then a lot more actually needs to be done than what is currently proposed here. Just putting a band-aid on the resource address check I think actually really misses an opportunity for more secure coding conventions and built-in implementations.

We are not opposed to this at all, if you have something specific in mind, please do let us know and we'll look into it.

0xOmarA:
@devmannic I will go ahead and merge this PR into the develop branch. How about we continue our discussion in a Github Issue? I'm hoping to understand more about your proposed solution(s) to see if there is any chance any of them can make it to v0.5.0.

devmannic:
Sounds good.  I will be able to reply at length a little later tonight

Issue #367, opened 2022-06-24
CALL_METHOD_WITH_ALL_RESOURCES in transaction manifest is not authorized through Authzone


unghuuduc:
Recently I tried the `restrict_deposit` access rule on my resources and found this problem:
I have made sure that the Authzone contain the Proof to allow the deposit, but I cannot use `CALL_METHOD_WITH_ALL_RESOURCES` on the transaction manifest but manually `TAKE_FROM_WORKTOP` the resource and later `CALL_METHOD` to deposit the bucket contain that resource.
Anyway to fix this issue? I think the problem lie in the `CALL_METHOD_WITH_ALL_RESOURCES` of the [TransactionBuilder](https://radixdlt.github.io/radixdlt-scrypto/src/radix_engine/transaction/builder.rs.html#396-406)

cbisaillon:
Can you show us your transaction manifest please?

unghuuduc:
@cbisaillon [here](https://github.com/unghuuduc/community-scrypto-examples/blob/main/defi/real-estate-manager/test/transaction_manifest/fill_order#:~:text=CALL_METHOD_WITH_ALL_RESOURCES)

emit those line
`TAKE_FROM_WORKTOP
    ResourceAddress("${LAND}")
    Bucket("landreturn");

CALL_METHOD 
    ComponentAddress("${R_ACC}") "deposit" Bucket("landreturn");

TAKE_FROM_WORKTOP
    ResourceAddress("${BUILDING}")
    Bucket("buildingreturn");

CALL_METHOD 
    ComponentAddress("${R_ACC}") "deposit" Bucket("buildingreturn");`
    
and it shows the error `AuthorizationError`, normally if I don't use `AccessRule` `restrict_deposit` on the `Resource`, it would just work fine.

0xOmarA:
Hey Peter! The `CALL_METHOD_WITH_ALL_RESOURCES` instruction drops all of the proofs first before making the method call. When a bucket has a proof created of it, the bucket becomes restricted and can not be moved around. So, the `CALL_METHOD_WITH_ALL_RESOURCES` instruction first drops all proofs (to un-restrict the buckets) and then performs the method call. 

We have discussed this and we have agreed to stop the `CALL_METHOD_WITH_ALL_RESOURCES` instruction from dropping all of the proofs in the AuthZone when it's called. Instead, we will leave it up to the developer to call the `DROP_ALL_PROOFS` instruction when they need to use it. 

[...]

PR #412, opened 2022-08-02
`CALL_METHOD_WITH_ALL_RESOURCES` instruction refactor

0xOmarA:
This PR refactors the logic behind the `CALL_METHOD_WITH_ALL_RESOURCES` instruction so that the instruction no longer drops all proofs when called. Instead, if the caller wishes to drop all proofs, they can do that with the newly added `DROP_ALL_PROOFS` instruction. 

In summary, currently, the `CALL_METHOD_WITH_ALL_RESOURCES` instruction does the following:
1. Clears the AuthZone
2. Drops all proofs in the transaction
3. Collects all buckets to use for a method call.

This PR changes it so that 1 and 2 are done by a `DROP_ALL_PROOFS` instruction and 3 is done by the `CALL_METHOD_WITH_ALL_RESOURCES` instruction,

This is in relation to issue #367.


iamyulong:
Looks good to me. The only question I have is if we should rename `CALL_METHOD_WITH_ALL_RESOURCES ` to `CALL_METHOD_WITH_ALL_WORKTOP_RESOURCES` to reflect on the fact that buckets taken from worktop are not included?

iamyulong (approved):

0xOmarA:
@iamyulong Both names are okay with me, should we add this as an item to the to discuss list for tomorrow's call?

iamyulong:
Let's add it to the to discuss list. You can merge this PR if want. No need to wait for that.

0xOmarA:
@iamyulong Sounds good, will resolve the merge conflicts and go ahead and merge :)

PR #413, opened 2022-08-02
Draft - Semi-major SBOR overhaul

dhedey:
As discussed, there are a few changes that I think could neaten up the SBOR abstractions. At a high level, they are:

* Having the type ids be more generic / less Rust focused:
  * This allows for cleaner cross-language use of SBOR
  * The explicit rust types would be represented as type hints in the "describe" schema (not yet implemented)
* Being more explicit about what is (static) "type data" and what isn't. This allows the payload size to be smaller, particularly when you have collection types -- as the static type data can be included in the collection parent type data.

I spent a little time today pulling together an example of how this might look - it's by no means anywhere near complete (and the code doesn't compile) - but I wanted to get input to check the direction was sound, before investing another day on it.

Suggestion - check out the `type_id.rs` file first, which has a comment explaining the key concepts behind this new SBOR paradigm, and then `encode.rs` and `decode.rs` to see how this looks. Don't pay too much attention to TODOs / bugs / breakages, I'm more keen to get input on high level design and approach.

Still to come (if the direction is sound) is:
* Some fixes to the existing things, such as byte array encoding/decoding (due to removal of TypeId - maybe this was an error)
* Fixing the any decoder and encoder (this will be a little painful)
* Fixing the derive macros for Structs + Enums
* Updating describe and code generation to work with type hints

### DIFF
--- a/radix-engine/src/model/resource.rs
+++ b/radix-engine/src/model/resource.rs
@@ -41,7 +41,7 @@ pub enum ResourceContainer {
         locked_ids: HashMap<NonFungibleId, usize>,
         /// The liquid non-fungible ids.
         liquid_ids: BTreeSet<NonFungibleId>,
-    },
+    }
 }
 
 /// The locked amount or non-fungible IDs.
--- a/sbor/src/any.rs
+++ b/sbor/src/any.rs
@@ -138,19 +138,19 @@ fn encode_any_internal(ty_ctx: Option<u8>, value: &Value, enc: &mut Encoder) {
         // struct & enum
         Value::Struct { fields } => {
             if ty_ctx.is_none() {
-                enc.write_type(TYPE_STRUCT);
+                enc.write_type_id(TYPE_STRUCT);
             }
-            enc.write_len(fields.len());
+            enc.write_variable_size(fields.len());
             for field in fields {
                 encode_any_internal(None, field, enc);
             }
         }
         Value::Enum { name, fields } => {
             if ty_ctx.is_none() {
-                enc.write_type(TYPE_ENUM);
+                enc.write_type_id(TYPE_ENUM);
             }
             name.encode_value(enc);
-            enc.write_len(fields.len());
+            enc.write_variable_size(fields.len());
             for field in fields {
                 encode_any_internal(None, field, enc);
             }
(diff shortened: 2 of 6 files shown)


iamyulong (commented):
**EDIT: renamed value to data**

As a long-term objective, we may use SBOR as the data model for an independent programming language. I'm striving for a balance between generic data model and specific language requirements.

In general, I want the type encoding to be recursive and decodable (which is aligned with your goal), and type section to be completely separate from the data section and may get stripped.

Example:
````
// primitive
type(()) = 1;
type(bool) = 2;
type(i8) = 3;
...
type(i128) = 4;
type(u8) = 5;
...
type(u128) = 6;
type(string) = 7;

// array and tuple
type([T;N]) = 8 + type(T) + N
type((A, B, C)) = 9 + len(fields) + type(A) + type(B) + type(C)

// collection
type(List<T>) = 14 + type(T)
type(Set<T>) = 15 + type(T)
type(Map<K, V>) = 16 + type(K) + type(V)

// struct 
type(struct) = 15 + len(fields) + type(field_1) + .. + type(field_n)

// enum variant
type(enum_variant) = 16 + len(fields) + type(field_1) + .. + type(field_n)

// options and results are enum variants but given special status
type(None) = 17
type(Some<T>) = 18 + type(T)
type(Ok(T)) = 19 + type(T)
type(Err(E)) = 20 + type(E)
````

````
// primitive
data(()) = empty;
data(bool) = 0 or 1;
data(i8) = 1 byte;
...
data(i128) = 16 bytes;
data(u8) = 1 byte;
...
data(u128) = 16 bytes;
data(string) = len(content) + content;

// array and tuple
data([T;N]) = data(a[1]) + data(a[2]) + ... + data(a[n])
data((A, B, C)) = data(a.1) + data(a.2) + ... + data(t.n)


// collection
data(List<T>) = len(elements) + data([1]) + .. + data([n])
data(Set<T>) = len(elements) + data([1]) + .. + data([n])
data(Map<K, V>) = len(elements) + data([1][0])+ data([1][1]) + .. + data([n][0])+ data([n][1])

// struct 
data(struct) = data(field_1) + .. + data(field_n) 

// enum variant
data(enum_variant) = index/name + data(field_1) + .. + data(field_n)

// option and result
data(Some<T>) = 0 + data(T)
data(None) = 1
data(Ok<T>) = 0 + data(T)
data(Err<E>) = 1 + data(E)
````
Both enum index/name and size of value are in the data section.

iamyulong:
Changes have evolved into 
- https://github.com/radixdlt/radixdlt-scrypto/pull/419
- https://github.com/radixdlt/radixdlt-scrypto/pull/423

As discussed, we will go through REP process for 1) encode other type data, besides `type_id`,  in the `encode_static_type_data` method, 2) generic product and sum types, and 3) dynamic and static collection types

PR #419, opened 2022-08-04
Reduce number of collection types in value model

iamyulong:
- Changed 
  ```
  TYPE_VEC: u8 = 0x30;
  TYPE_TREE_SET: u8 = 0x31;
  TYPE_TREE_MAP: u8 = 0x32;
  TYPE_HASH_SET: u8 = 0x33;
  TYPE_HASH_MAP: u8 = 0x34;
  ```
  to
  ```
  TYPE_LIST: u8 = 0x30;
  TYPE_SET: u8 = 0x31;
  TYPE_MAP: u8 = 0x32;
  ```
- Updated type id of `Option` and `Result` (to be more consistent with other enum representation)


transaction/src/manifest/lexer.rs
```diff
@@ -372,15 +373,12 @@ impl Lexer {
...
             "Tuple" => Ok(TokenKind::Tuple),
             "Result" => Ok(TokenKind::Result),
-            "Vec" => Ok(TokenKind::Vec),
-            "TreeSet" => Ok(TokenKind::TreeSet),
-            "TreeMap" => Ok(TokenKind::TreeMap),
-            "HashSet" => Ok(TokenKind::HashSet),
-            "HashMap" => Ok(TokenKind::HashMap),
+            "List" => Ok(TokenKind::List),
+            "Set" => Ok(TokenKind::Set),
+            "Map" => Ok(TokenKind::Map),
```
0xOmarA, line 381: I really like this change. Just wondering, would we like to keep aliases in the transaction manifest Lexer to make it more intuitive for people? Something like:

```rust
"Vec" => Ok(TokenKind::List), 
```

That way when people see a method that takes a vec, they provide it with a Vec and we do the magic of translating that for them :)

iamyulong: That might be a good idea. Will apply.

dhedey: In which case, perhaps we also do the same for `HashMap`, `BTreeMap` etc... Although I'm not so sure about these.

I do like the alias idea, but I'm also torn by keeping the grammar as simple as possible; as this code needs to (eg) be copied onto a Ledger where space is limited.

I'm also happy with supporting Vec but not the others

dhedey (approved):
Really nice, thanks Yulong! 👍

PR #420, opened 2022-08-04
Improved Proof Validation

0xOmarA:
This PR adds a number of different proof validation modes that can be used with the `Proof` type. In addition to adding more complex ways of performing proof validation, this PR still keeps the 90% case (validating against a resource address) fairly easy and frictionless.

### DIFF
--- a/scrypto/src/resource/proof.rs
+++ b/scrypto/src/resource/proof.rs
@@ -43,16 +43,15 @@ pub enum ProofValidationMode {
     /// the `Proof`'s resource address belongs to the set, then its valid.
     ValidateResourceAddressBelongsTo(BTreeSet<ResourceAddress>),
 
-    /// Specifies that the `Proof` should be validated against a single `ResourceAddress` and `NonFungibleId` pair.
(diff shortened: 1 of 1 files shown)


scrypto/src/resource/proof.rs
```diff
@@ -96,8 +130,103 @@ impl Proof {
...
+    }
+
+    fn validate_contains_non_fungible_id(
+        &self,
+        non_fungible_id: NonFungibleId,
+    ) -> Result<(), ProofValidationError> {
+        if self.non_fungible_ids().get(&non_fungible_id).is_some() {
+            Ok(())
+        } else {
+            Err(ProofValidationError::InvalidNonFungibleId(non_fungible_id))
```
iamyulong, line 195: Minor: suggest renaming to `NonFungibleIdNotFound`

iamyulong (approved):

scrypto/src/resource/proof.rs
0xOmarA: @iamyulong Thanks Yulong, will make this change and merge :)

PR #422, opened 2022-08-04
Add an optional `proofs` argument to resim's transaction builder

0xOmarA:
This PR adds an additional optional argument to the resim transaction builder subcommands which allow the user to specify the proofs that they wish to have in the Auth Zone in this transaction. Example usage:

```sh
resim call-method $component add_admin --proofs 1,$admin_badge 1,$owner_badge
```

This has been added to the following subcommands:

- `resim call-function`
- `resim call-method`
- `resim mint`
- `resim transfer`

While the `resim call-function` subcommand does not quite need this additional argument right now, it will be useful once we allow for functions to have system-level auth tied to them.


iamyulong:
Can we add some tests to cover this new feature? :)

0xOmarA:
@iamyulong Indeed, will do that :)

iamyulong (approved):
Thanks for spending time adding extra tests!

PR #423, opened 2022-08-05
Various enhancements to SBOR protocol

iamyulong:
Changes:
- (breaking) Updated the type IDs
- (breaking) Renamed flag `with_type` to `with_static_info`
  * Static info includes type IDs and static sizes
  * Although, in the future, we may introduce dynamic type ID to support `Any` type
- (breaking) Split `read/write_len` into
  * `read/write_dynamic_size` 
  * `read/write_static_size`
- Renamed `read/write/check_type` to `read/write/check_type_id`
- Split `read/write_u8` into 
   *  `read/write_variant_index`
   * `read/write_byte`
- Added `read/write_variant_label` 
  * In the future, we may use less bytes to represent the label size
  * May also replace label with index
- Made `Encode::encode_type_id` a static function


cseidman (approved):

sbor/src/decode.rs
```diff
@@ -87,16 +87,32 @@ impl<'de> Decoder<'de> {
...
+        self.read_byte()
+    }
+
+    pub fn read_variant_label(&mut self) -> Result<String, DecodeError> {
+        let n = self.read_dynamic_size()?;
+        let slice = self.read_bytes(n)?;
+        String::from_utf8(slice.to_vec()).map_err(|_| DecodeError::InvalidUtf8)
+    }
+
+    pub fn read_static_size(&mut self) -> Result<usize, DecodeError> {
```
dhedey, line 103: Just a thought - this probably shouldn't be a pub - because it should be protected behind a with_static_info check -- realistically, for now, people would just use `check_static_size` I imagine?

dhedey (commented):
Looks good!

You claimed:
> (breaking) Updated the type IDs

I don't see any type id changes? Just a slight change to schemaless SBOR for Arrays/Tuples?
I guess maybe you meant the type id encoding/decoding being static?

sbor/src/decode.rs
iamyulong: This method allows developers to implement their custom codec. It's not part of the `Decode` trait.

iamyulong:
Thanks @dhedey! You're right. There is actually no type id change.

I wanted to change String type id to allow wider integer types, but reverted that change. Will correct PR description.

sbor/src/decode.rs
dhedey: Ahh okay, well then I think it should include a check of whether the `with_static_info` boolean is active before reading? (and conversely for writing?)

dhedey: I'd kinda advocate for a separate encoder/decoder for with/without type information to remove these runtime checks, but that would be a fun refactor adding in generics everywhere :')

PR #425, opened 2022-08-08
Cleanup invocation

talekhinezh:
Refactors to get closer to a fully unified `invoke_method`.
* Moved Proof/Authzone logic into separate AuthModule
* Combined all native ref receivers into a single `NativeRENodeRef`
* Refactored node pointers to be absolute rather than relative to their stack frame.


radix-engine/src/engine/call_frame.rs
```diff
@@ -162,14 +158,15 @@ impl RENodePointer {
 
     fn borrow_native_ref<'p, 's>(
         &self, // TODO: Consider changing this to self
+        self_frame_id: usize,
```
iamyulong, line 161: Minor: Isn't `self_frame_id` part of self?

iamyulong (approved):
LGTM

radix-engine/src/engine/call_frame.rs
talekhinezh: Ah no, probably named badly but `self_frame_id` is the current `frame_id` of the call frame requiring access to this node pointer. This will get cleaned up when heap is implemented.

Issue #358, opened 2022-06-17
Geting component/vault from Substore


a137x:
Good evening.
In version 0.4.1 I can't find a good replacement for thngs that existed in version 0.3.0

in ```radix_engine::ledger::memory::InMemorySubstateStore```
there were two functions:
```rust
   fn get_component(&self, address: Address) -> Option<Component> {
        self.components.get(&address).map(Clone::clone)
    }

    fn get_vault(&self, component_address: &Address, vid: &Vid) -> Option<Vault> {
        self.vaults
            .get(&(component_address.clone(), vid.clone()))
            .map(Clone::clone)
    }
```
Which existed prior version 0.4.0, is gone now.

This function were useful, in conjunction with other stuff, for getting of vaults for particular component.

The thing looked like this:
```rust
    let mut ledger = InMemorySubstateStore::with_bootstrap();
    
    pub fn get_account_vaults(
        &mut self,
        component_address: ComponentAddress,
    ) -> HashMap<ResourceAddress, Contents> {
        let ledger = self.executor.substate_store();
        let component = ledger.get_component(component_address).unwrap();

        let state = component.state();
        let validated_data = validate_data(state).unwrap();
        validated_data
            .lazy_maps
            .iter()
            .flat_map(|mid| TestEnv::get_lazymap_info(ledger, &component_address, &mid))
            .collect()
    }

    fn get_lazymap_info(
        ledger: &L,
        component_address: &ComponentAddress,
        id: &Mid,
    ) -> Vec<(ComponentAddress, Contents)> {
        let lazy_map = ledger.get_lazy_map(component_address, id).unwrap();
        lazy_map
            .map()
            .iter()
            .flat_map(|(_, data)| {
                let validated_data = validate_data(data).unwrap();
                validated_data
                    .vaults
                    .iter()
                    .map(|vid| TestEnv::get_vault_info(ledger, component_address, vid))
                    .collect::<Vec<(ResourceAddress, Contents)>>()
            })
            .collect()
    }

    pub enum Contents {
        Amount(Decimal),
        NonFungibleIds(Vec<NonFungibleId>),
    }

```
How to achieve the desirable in 0.4.1 ?

0xOmarA:
Hey @a137x, I've looked into this a little bit and it indeed seems to not be possible due to the `InMemorySubstateStore` having no methods to obtain the keys in the substate store. We have discussed this internally and this is something we're looking to add with v0.5.0.

[...]

PR #426, opened 2022-08-08
Implemented the `QueryableSubstateStore` trait for the `InMemorySubstateStore`

0xOmarA:
This is a small PR implementing the `QueryableSubstateStore` trait for the `InMemorySubstateStore` type. This is done in response to #358.


a137x:
Hey @0xOmarA , thanks for implementing this feature.
I briefly looked into the code of the commit but still perplexed how to use it in scenario like I mentioned in #358
My idea now is to grab your code from commit and use it with 0.4.1 since 0.5.0 is still in dev stage and I need this functionality.
Could you please elaborate on usage?
Particularly:
How can I get vaults of known component with this in 0.5?

Thanks

PR #428, opened 2022-08-08
Refactor `TestRunner` into its own crate

0xOmarA:
The PR refactors the `TestRunner` struct into its own create allowing it to be used in new Scrypto blueprints to import this crate and the `TestRunner` struct and make use of it.

Recent changes to Scrypto made it harder to test Scrypto code and lengthier to test Scrypto packages. Since more convenient testing tools are already available (the `TestRunner`), it is worth adding a way for developers to use it when testing their packages.


scrypto-unit/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "scrypto-unit"
+version = "0.1.0"
```
iamyulong, line 3: Can we use the same version as other crates to avoid confusion?

iamyulong (approved):
Have been waiting for this for too long!

scrypto-unit/Cargo.toml
0xOmarA: Indeed, will go ahead and correct the version :)

PR #429, opened 2022-08-09
Architecture: Unify invoke_method

talekhinezh:
Another PR unifying `invoke_method` in call_frame and various other cleanups.

* Removed `Receiver::Scrypto`, replaced with `Function::Scrypto`
* Unified `Receiver::Consumed` and `Receiver::Ref`


scrypto/src/core/invocation.rs
```diff
@@ -16,11 +15,24 @@ pub enum TypeName {
...
-    Component(ComponentAddress),
+    Ref(RENodeId),
     AuthZoneRef,
 }
 
+#[derive(Debug, Clone, TypeId, Encode, Decode)]
+pub enum Function {
+    Scrypto(String),
+    Native(String),
+}
```
iamyulong, line 26: Feels like wrong abstraction to me. What's defined here is more of function identifier rather than function. I would consider "function = function identifier + inputs + body definition".

Whether it's Scrypto or native should NOT be determined from the identifier. Instead,
- For function call, it's the type that contains the function definition
- For method call, it's the type of the receiver/self

iamyulong: Here is how `syn` models function and method calls (which I mentioned briefly earlier).
- https://docs.rs/syn/latest/syn/struct.ExprCall.html
- https://docs.rs/syn/latest/syn/struct.ExprMethodCall.html

talekhinezh: Ah wait, reading these comments a little closer it sounds a little different from what we talked about earlier.

I don't necessarily agree with the "For method call, it's the type of the receiver" which identifies the function. If following Rust conventions, methods may also resolve to trait implementations: https://doc.rust-lang.org/reference/expressions/method-call-expr.html. We don't have traits but we have something a little similar in the form of substates.

I think we need another call tomorrow :)

iamyulong: This is consistent with what we've discussed, although a bit outdate as it's before our discussion.  

The approach described above is more of the separate Function/Method invocation path. 

Your approach (which I also agreed) is to have a unified invocation path:
```
enum Function {
     Native(RENodeType, String)
     Scrypto(PackageAddress, String, String)
}

// And, RENode/RENodeRef can be just part of the argument list
```

iamyulong (approved):
Like all these function enums!

PR #438, opened 2022-08-12
Fix `NonFungibleId` and `NonFungibleAddress` in manifest.

0xOmarA:
Currently, if a `NonFungibleId` is created of a `Decimal` or any Scrypto type, it can not be passed through the transaction manifest. 

This means that `NonFungibleId`s of certain types can be created, and can have NFTs created of them, but the `NonFungibleId` can not be passed through the manifest. 

This is because the implementation of the `TryFrom<&[u8]>` trait used `ScryptoValue::from_slice_no_custom_values` to create a `ScryptoValue` out of the slice. Since `NonFungibleId`s can now be any vector of SBOR bytes, we should use `ScryptoValue::from_slice`.

### DIFF
--- a/scrypto/src/resource/non_fungible_id.rs
+++ b/scrypto/src/resource/non_fungible_id.rs
@@ -66,8 +66,8 @@ impl TryFrom<&[u8]> for NonFungibleId {
     type Error = ParseNonFungibleIdError;
 
     fn try_from(slice: &[u8]) -> Result<Self, Self::Error> {
-        let value = ScryptoValue::from_slice(slice)
-            .map_err(|_| ParseNonFungibleIdError::InvalidValue)?;
+        let value =
+            ScryptoValue::from_slice(slice).map_err(|_| ParseNonFungibleIdError::InvalidValue)?;
         Ok(Self(value.raw))
     }
 }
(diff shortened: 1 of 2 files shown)


0xOmarA:
Wondering if it would be better to make `NonFungibleId` have a constructor which ensures that it does not have a `Bucket`, `Proof`, `Vault`, etc...?

scrypto/src/resource/non_fungible_id.rs
```diff
@@ -66,8 +66,8 @@ impl TryFrom<&[u8]> for NonFungibleId {
     type Error = ParseNonFungibleIdError;
 
     fn try_from(slice: &[u8]) -> Result<Self, Self::Error> {
-        let value = ScryptoValue::from_slice_no_custom_values(slice)
-            .map_err(|_| ParseNonFungibleIdError::InvalidValue)?;
+        let value =
+            ScryptoValue::from_slice(slice).map_err(|_| ParseNonFungibleIdError::InvalidValue)?;
```
talekhinezh, line 70: Instead of using `from_slice` can we modify `from_slice_no_custom_values` to include `Decimal`. This is mainly so that we don't accept buckets/vaults/etc.

talekhinezh (changes requested):

scrypto/src/resource/non_fungible_id.rs
0xOmarA: Thought about it, this is non-urgent and might be better to wait until that becomes an Enum to add that in with that refactor.

PR #441, opened 2022-08-14
Minor Improvements

0xOmarA:
This pull request makes a number of minor improvements.

* Adds `Ord` and `PartialOrd` to a number of the types in the Scrypto repo to allow them to be used as keys in HashMaps. 
* Makes the ABI exported through `resim export-abi` match that required by the `import!` macro.
* Adds a way to get and set the current epoch through the `TestRunner`. 
* Increases the cost unit limits to allow for the publishing of larger packages without issues.


scrypto-unit/src/test_runner.rs
```diff
@@ -438,6 +441,71 @@ impl<'s, S: ReadableSubstateStore + WriteableSubstateStore> TestRunner<'s, S> {
...
+        });
+    }
+
+    pub fn get_current_epoch(&mut self) -> u64 {
+        let current_epoch: ScryptoValue = self.system_kernel_call(|kernel| {
+            kernel
+                .invoke_method(
+                    Receiver::Ref(RENodeId::System),
+                    FnIdentifier::Native(NativeFnIdentifier::System(SystemFnIdentifier::SetEpoch)),
+                    ScryptoValue::from_typed(&SystemGetCurrentEpochInput {}),
```
talekhinezh, line 463: Should be `GetEpoch`

scrypto-unit/src/test_runner.rs
```diff
@@ -438,6 +441,71 @@ impl<'s, S: ReadableSubstateStore + WriteableSubstateStore> TestRunner<'s, S> {
...
+    fn system_kernel_call<F>(&mut self, fun: F) -> ScryptoValue
+    where
+        F: FnOnce(
+            &mut Kernel<DefaultWasmEngine, DefaultWasmInstance, SystemLoanFeeReserve>,
+        ) -> ScryptoValue,
+    {
+        let tx_hash = hash(self.next_transaction_nonce.to_string());
+        let mut substate_store = self.execution_stores.get_output_store(0);
+        let mut wasm_engine = DefaultWasmEngine::new();
+        let mut wasm_instrumenter = WasmInstrumenter::new();
```
talekhinezh, line 480: Minor: Wasm Engine and Wasm Instrumenter should exist in `self` already

scrypto-unit/src/test_runner.rs
talekhinezh: Not to mention, should have tests which test `set_current_epoch` and `get_current_epoch`

0xOmarA: Sounds good, I will add that. The tests are likely to fail through until the issue with the `resim set-current-epoch` is fixed since they both use the same logic :)

scrypto-unit/src/test_runner.rs
0xOmarA: Yes you are right, changing that :)

scrypto-unit/src/test_runner.rs
0xOmarA: Fixed and added tests for that.

talekhinezh (approved):

PR #442, opened 2022-08-15
Add execution trace module

LukasGasior1:
This PR introduces the execution trace module, which is used to trace some specific events that happen inside the engine and extract the information out into the receipt.

Initially, it's used to track fungible resource changes in vaults belonging to components.

There are a few missing bits to be addressed in subsequent PRs:
* lock_fee isn't included
* non-fungible resources aren't traced
* neither are withdrawals/deposits on vaults that do not have a component parent (for example, created before the component is globalized)
* resource changes from all call frames are combined, there's no per-frame breakdown


radix-engine/src/engine/execution_trace_module.rs
```diff
@@ -0,0 +1,143 @@
...
+use scrypto::engine::types::*;
+use scrypto::prelude::{VaultPutInput, VaultTakeInput};
+use scrypto::values::*;
+
+use crate::engine::*;
+use crate::model::*;
+
+pub struct ExecutionTraceModule;
+
+impl ExecutionTraceModule {
```
talekhinezh, line 13: Seems that this should just be `ExecutionTrace`

radix-engine/tests/execution_trace.rs
```diff
@@ -0,0 +1,54 @@
...
+            package_address,
+            "ExecutionTraceTest",
+            "transfer_resource_between_two_components",
+            to_struct!(transfer_amount),
+        )
+        .build();
+    let receipt = test_runner.execute_manifest(manifest, vec![public_key]);
+
+    // Assert
+    if let TransactionStatus::Succeeded(output) = receipt.status {
```
talekhinezh, line 31: This can be replaced with `let output = receipt.expect_success();`

radix-engine/tests/execution_trace.rs
```diff
@@ -0,0 +1,54 @@
+use radix_engine::ledger::TypedInMemorySubstateStore;
+use radix_engine::transaction::TransactionStatus;
+use scrypto::core::Network;
+use scrypto::prelude::*;
+use scrypto::to_struct;
+use scrypto_unit::*;
+use transaction::builder::ManifestBuilder;
+
+#[test]
+fn test_trace_resource_transfers() {
```
talekhinezh, line 10: Can we also add checks in the `account.rs` test with the new receipt model?

radix-engine/src/engine/execution_trace_module.rs
LukasGasior1: yeah, makes sense

talekhinezh (approved):

PR #454, opened 2022-08-20
feat: Support external network definitions

dhedey:
Theo reported that we didn't hook the Network up correctly for Enkinet.

There was a question of where these should live - Node or Engine - I think I'm tempted to stick with the source of truth being in the Node for now. 

In any case, I realised first we need support for non-const Network Definitions in the Engine.

The following changes are in this PR:
* Separation of Network (a simple enum to select a Network) and NetworkDefinition (the actual Network Configuration). Alternative names welcome.
* The TransactionHeader includes a network_id byte, instead of the Network Enum (which was essentially the Network Name).

I'd welcome input on naming etc. 

Still to do:
* Transaction rejects for mismatching Network - instead of failing an assert as per currently - I'll fix this in a follow-up PR where I'll rearrange the Transaction Receipt structure to enable this more smoothly.


iamyulong:
Looks like both transaction decoding and verification cost have decreased.

You see the costing analysis [here](https://github.com/radixdlt/radixdlt-scrypto/runs/7932101608?check_suite_focus=true) so to be able update the numbers.

_(OR run `cargo run --test metering` under radix-engine)_

scrypto/src/core/network.rs
```diff
@@ -1,42 +1,62 @@
...
-                }
-            }
+// TODO: we may be able to squeeze network identifier into the other fields, like the `v` byte in signature.
+/// Network is intended to be an "easy" method to get common Network configuration for (eg) tests
+#[derive(Debug, Clone, TypeId, Encode, Decode)]
+pub enum Network {
+    LocalSimulator,
+    Mainnet,
+    Custom(NetworkDefinition),
+}
```
iamyulong, line 20: I'm not sure if this enum is necessary. And the abstraction doesn't seem right (there is no logical separation between the three variants: `Custom` can be `Mainnet`).

For convenience, I would consider using helper functions: `NetworkDefinition::mainnet()`  returns the definition of mainnet.

iamyulong (commented):
Other than the above comment, the changes look good to me.

scrypto/src/core/network.rs
dhedey: Yeah, that's a good shout - I'll make this change :)

iamyulong: Also would consider having network ID as single source of truth and deriving name and HRP from id.

```rust
struct Network(u8);

impl Network {
     pub fn id() -> u8;
     pub fn name() -> String;  // computed based on id
     pub fn hrp() -> String;  // computed based on id
}
```

dhedey: Well that would depend on where the source of truth lives for the Network Definition.

Chatting to Josh, we seemed to think the Node could/should own it for now? 👍

iamyulong: Cool

iamyulong (approved):

PR #456, opened 2022-08-22
Transaction receipt neatening

dhedey:
refactor: Improved Transaction Receipt

* TransactionReceipt is now separated into sub-parts.
* The result can be reject or commit. Commit includes Success and Failure outcomes.

The goal of these changes is two-fold:
* Support early returning from transaction executor on network mismatch (the assert I left in on my last PR, #454 )
* Better abstract the execution receipt to make more things optional, and to align with the commit intention

### DIFF
--- a/radix-engine/src/transaction/transaction_executor.rs
+++ b/radix-engine/src/transaction/transaction_executor.rs
@@ -1,4 +1,3 @@
-use scrypto::core::NetworkDefinition;
 use transaction::model::*;
 
 use crate::constants::{DEFAULT_COST_UNIT_PRICE, DEFAULT_MAX_CALL_DEPTH, DEFAULT_SYSTEM_LOAN};
@@ -12,7 +11,6 @@ use crate::types::*;
 use crate::wasm::*;
 
 pub struct ExecutionConfig {
-    pub network_definition: NetworkDefinition,
     pub cost_unit_price: Decimal,
     pub max_call_depth: usize,
     pub system_loan: u32,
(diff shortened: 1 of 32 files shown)


radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -116,28 +118,40 @@ where
...
+                transaction_contents,
+                None,
+                TransactionResult::Reject(RejectResult {
+                    error: RejectionError::MismatchedNetwork {
+                        transaction_network_id: transaction.transaction_network_id(),
+                        expected_network_id: params.network_definition.id,
+                    },
+                }),
+            );
+        }
```
iamyulong, line 153: Don't think we need to validate the transaction network id here. It's part of the static header validation process, same as other header fields. TransactionExecutor only issue stateful execution-based rejections.

Given network ID is no longer used or exposed to Scrypto, network definition is no longer required by TransactionExecutor.

iamyulong (commented):
Good refactoring to the TransactionReceipt models. Only one comment left.

radix-engine/src/transaction/transaction_executor.rs
dhedey: Hmmm fair, but the receipt needs to somehow know the network configuration in order to format itself, and the header only contains the network id at present.

So the question is:
1. Where does that configuration come from?
2. How do we sanity-check we're using the right one, to limit possibility for confusion and provide defence-in-depth?

Where does it come from = either the (a) Executable Transaction or the (b) ExecutionParams.

If (a), we'd need to clone it inside ValidatedTransaction and TestTransaction.

If (b), the only real way to achieve (2) is to do as above -- otherwise someone could run the transaction through and get a receipt for a different network to the header; which would be weird.

I went for (b) because the network configuration felt like it belonged as a parameter, configuring the engine. But you're saying possibly that's not the direction of vision/travel? Would you prefer option (a) in that case?

iamyulong: > The receipt needs to somehow know the network configuration in order to format itself

It can take a network id parameter. Formatting of addresses are broken ATM (hardcoded network id in `ResourceAddress`/`PackageAddress`).

> How do we sanity-check we're using the right one, to limit possibility for confusion and provide defence-in-depth?

It's all part of the `TransactionValidator` (and `ValidationConfig`).

dhedey: For anyone else tracking this, Yulong and I caught up:
* Formatting is currently broken (see eg) format! of PackageAddress which assumes LocalSim
* For now, let's break formatting of the receipt further, and have a separate task

Therefore:
* Remove Network Definition from the execution params and executor for this PR

We'll have future conversations about:
* How to format stuff correctly
* Validation - particularly of Network Id and Epoch; and whether that should be in the Executor or Validator (in particular, so dependency tracking works right for a "clever" mempool)

PR #460, opened 2022-08-24
fix: Preview doesn't assume local sim

dhedey:
Preview tests were breaking when I integrated with the engine - this is why.

Have removed the hardcoding of network in preview too.


iamyulong (approved):
LGTM. Only need to fix CI failures before merge.

PR #466, opened 2022-08-25
Fix scrypt fmt to work with external_blueprint

iamyulong:


0xOmarA:
Should we in line 156 reverse the replace changes back? :)

dhedey (approved):
Looks good. I'd be tempted to add a comment to the newly added line explaining the hack, as it's perhaps a little unexpected :)

iamyulong:
@0xOmarA It's not required. Added comment to explain.

PR #469, opened 2022-08-25
Feature/expression

iamyulong:
Changes:
- Removed `CALL_METHOD_WITH_ALL_WORKTOP_RESOURCES` instruction
- Added `Expression` Scrypto value which is evaluated by `TransactionProcessor` at runtime
    * Currently supports `ENTIRE_WORKTOP` and `ENTIRE_AUTH_ZONE`

Note that `ENTIRE_WORKTOP` does **NOT** include named buckets


0xOmarA:
Maybe I'm looking too early into this, but I feel like the expressions are better suited as enums `Expression::EntireWorktop` or `Expression::AllAuthZoneProofs`. Thoughts?

iamyulong:
@0xOmarA This is indeed a good point to discuss. 

My initial thinking is we should keep `Expression` open so we can easily add support for new expressions without breaking the full stack (VSCode extension, manifest compiler, transaction and engine). 

The drawback of this approach is losing Rust compiler check, for which I'm not extremely concerned given the main usage is within manifest.

talekhinezh:
Been meaning to get to this but let's chat about this tomorrow :)

talekhinezh (approved):
After discussion with Yulong. Agreed to merge this and re-look into how this may be implemented better in the future. For example, by adding an `enum` wrapper around all enums passed into the `TransactionProcessor`

PR #470, opened 2022-08-28
Helpful functions for TransactionReceipt on RadixEngine

unghuuduc:
I have reorganized the functions on TransactionReceipt a bit and included 4 more helpful functions:

+ `expect_commit_success()` now returned the exact encrypted transaction output (Last time, the function return a Vector with index [0] return only `[0]`? and index [1] return the transaction output).

+ `expect_commit_failure()` now returned the `RuntimeError` of transaction failure.

+ Change the last `expect_commit_failure()` into `expect_specific_failure()` .

+ Add `fn output()` return the decoded type (user choice) of transaction output (if success).

+ Add 3 functions return the new package, component, resource addresses on the transaction.

### DIFF
--- a/radix-engine/src/transaction/transaction_receipt.rs
+++ b/radix-engine/src/transaction/transaction_receipt.rs
@@ -82,10 +82,10 @@ impl TransactionReceipt {
         }
     }
 
-    pub fn expect_commit_success(&self) -> &[u8] {
+    pub fn expect_commit_success(&self) -> &Vec<Vec<u8>> {
         match &self.result {
             TransactionResult::Commit(c) => match &c.outcome {
-                TransactionOutcome::Success(x) => &x[1][..],
+                TransactionOutcome::Success(x) => x,
                 TransactionOutcome::Failure(err) => {
                     panic!("Expected success but was failed:\n{:?}", err)
                 }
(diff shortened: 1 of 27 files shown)


radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -84,10 +75,17 @@ impl TransactionReceipt {
...
+            TransactionResult::Commit(..) => panic!("Expected rejection but was commit"),
+            TransactionResult::Reject(ref r) => &r.error,
+        }
+    }
+
+    pub fn expect_commit_success(&self) -> &[u8] {
         match &self.result {
             TransactionResult::Commit(c) => match &c.outcome {
-                TransactionOutcome::Success(x) => x,
+                TransactionOutcome::Success(x) => &x[1][..],
```
iamyulong, line 88: I don't think it's proper to assume the output of the second instruction is the expected output. With Radix transaction model, a transaction may include "any" number of instructions and developers need to be able to inspect output of any instruction.

Suggest keeping the original `expect_commit_success` and change `output()` to `output(nth: usize)`.

iamyulong (commented):
Left one comment. The other convenience methods look good to me.

unghuuduc:
Resolved, please check.

iamyulong:
Thanks for contributing to Scrypto! Merging...

iamyulong (approved):

PR #476, opened 2022-08-29
Feature: Encodeable Transaction Receipts

talekhinezh:
This PR makes a few refactors in order to allow transaction receipts to be SBOR encoded/decoded. Most notably an `InvokeError` enum is implemented to allow us to catch downstream `RuntimeErrors` rather than using Box.


iamyulong (approved):
Looks good. Nice and clean!

dhedey (approved):
Nice! :)

PR #481, opened 2022-09-01
Fixed the display of NonFungibleIds in resim

0xOmarA:


simulator/src/ledger/dumper.rs
```diff
@@ -187,7 +187,7 @@ fn dump_resources<T: ReadableSubstateStore, O: std::io::Write>(
                     .map(|s| s.into())
                     .unwrap();
                 if let Some(non_fungible) = non_fungible.0 {
-                    let id = ScryptoValue::from_slice(&id.to_vec()).unwrap();
+                    let id = ScryptoValue::from_typed(id);
```
iamyulong, line 190: I can't remember why we need to covert `NonFungibleId` to `ScryptoValue`. Can't we just dump it directly?

0xOmarA: We do that so we can make use of the `ScryptoValueFormatter` (used in the `ScryptoValue.to_string` method). This allows `NonFungibleId`s (and everything else that we are printing) to be printed as `NonFungibleId("some hex")` instead of `"some hex"`. I like that we are using the  `ScryptoValueFormatter` in this case since the way that it represents most values matches that required by the transaction manifest.

iamyulong (approved):

simulator/src/ledger/dumper.rs
iamyulong: Don't have a strong opinion anyway. Was wondering why we previous attempt to decode the id.

PR #484, opened 2022-09-05
Make into great again

beemdvp:
Hope we can bring this convenience back again :D

This allows us to do:

```rust
let test_component: TestComponent = self.test_component_address.into();
```


iamyulong (approved):
LGTM

PR #485, opened 2022-09-06
Fix `NonFungibleAddress` from and to string

0xOmarA:


scrypto/src/resource/non_fungible_address.rs
```diff
@@ -100,18 +114,18 @@ impl FromStr for NonFungibleAddress {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         let bytes =
             hex::decode(s).map_err(|_| ParseNonFungibleAddressError::InvalidHex(s.to_owned()))?;
-        if bytes.get(0) != Some(&3u8) {
+        if bytes.get(0) != Some(&EntityType::Resource.id()) {
             return Err(ParseNonFungibleAddressError::InvalidPrefix);
         }
```
iamyulong, line 119: Legacy issue. I think we should move the above check to `try_from`.

iamyulong (approved):
LGTM. Wonder if we should bech32 non-fungible addresses.

iamyulong:
Sounds good to me. Thanks @0xOmarA!

scrypto/src/resource/non_fungible_address.rs
0xOmarA: I have removed this check in my latest commit. The reason why I removed it is because the `ResourceAddress::try_from` method will perform this check either way for us, so there was no need to do the check twice. 

I will give you a chance to take a look at my latest commit. If you are satisfied with it, then feel free to merge this PR.

PR #495, opened 2022-09-14
Vault and VaultSubstate

iamyulong:
Changes:
- Split model `ResourceContainer` to `Resource` and `LockableResource`
- Added `VaultSubstate` to distinguish from `Vault` RENode, and the conversion back and forth
- Refactored `Track` to NOT flush updates during transaction unless it's in `write_through` mode
   - This allows us to store transient state, like `lockable_resource` in a `Vault`


radix-engine/src/engine/errors.rs
```diff
@@ -67,7 +70,8 @@ pub enum KernelError {
     RENodeNotInTrack,
 
     // Substate
-    Reentrancy(SubstateId),
+    NodeToSubstateFailure(NodeToSubstateFailure),
```
talekhinezh, line 73: Since Node to Substate mapping is static/deterministic feels like this error should be an unbroken invariant and cause a panic if it ever occurs?

radix-engine/src/engine/kernel.rs
```diff
@@ -717,6 +720,7 @@ where
                         }
                     }
                 };
+                drop(current_frame);
```
talekhinezh, line 723: I think better to use rust scoping to implement this drop. e.g.:
```
let node_pointer = {
    let current_frame = Self::current_frame(&self.call_frames);
    if current_frame.owned_heap_nodes.contains_key(&node_id) {
        RENodePointer::Heap {
            frame_id: current_frame.depth,
            root: node_id.clone(),
            id: None,
         }
    } else if let Some(pointer) = current_frame.node_refs.get(&node_id) {
          pointer.clone()
     } else {
         match node_id {
             // Let these be globally accessible for now
             // TODO: Remove when references cleaned up
             RENodeId::ResourceManager(..) | RENodeId::System => {
                 RENodePointer::Store(*node_id)
             }
             _ => {
                 return Err(RuntimeError::KernelError(
                     KernelError::InvokeMethodInvalidReceiver(*node_id),
                 ))
             }
          }
      };
}
```

talekhinezh (approved):

radix-engine/src/engine/kernel.rs
iamyulong: Good idea. Will apply

radix-engine/src/engine/errors.rs
iamyulong: I'm happy to either A) make this an invariant - every node must be persistable at the end of a transaction OR B) make `Track` more dynamic and allow nodes to define whether it's ready for persistence.  

Imagine, we have:
```
struct Node {
   substate: A,
   transient: B,
}
```
Then we can define interesting rules, like the transaction must abort if `transient > 3`.

If it's okay for you, I can leave a TODO to explore both options as a follow-up.

talekhinezh: Mmm, I think there's a better third option where persisted vs. transient is just a property of the substate (or whatever name we want to put to it) but yeah let's just leave this as a TODO

PR #497, opened 2022-09-14
Replace blob path with blob hash

iamyulong:


simulator/src/resim/cmd_run.rs
```diff
@@ -13,6 +12,14 @@ pub struct Run {
     /// The path to a transaction manifest file
     path: PathBuf,
 
+    /// The network to use when outputting manifest, [simulator | adapanet | nebunet | mainnet]
+    #[clap(short, long)]
+    network: Option<String>,
+
+    /// Comma separated paths to blob files
+    #[clap(short, long)]
+    blobs: Option<String>,
```
0xOmarA, line 21: Minor: should we do `#[clap(multiple = true)]` or keep this as comme separated paths?

0xOmarA (commented):
Really happy with this PR. Thanks a lot for looking into this Yulong :) I have left a small comment on a small item. It is not essential so if you would like, feel free to merge without it :)

0xOmarA (approved):

simulator/src/resim/cmd_run.rs
iamyulong: Didn't know `multiple`. Seems to be better. Will update.

PR #504, opened 2022-09-16
Misc/misc alphanet improvement

0xOmarA:
This PR introduces a number of miscellaneous alphanet improvements mostly in relation to the [transaction library](https://github.com/radixdlt/transaction-library).

1. Made some of the validation functions in the `TransactionValidator` structs public to be able to use them in the transaction library.
2. The `ValidationConfig` does not need to have a reference to a `NetworkDefinition` when all it needs is a network id. 
3. Derive the `Hash` trait on some of the cryptographic types as it would make it possible to have a `HashSet` of those types. 
4. Added a `serde` feature to the `transaction` crate and made the `TransactrionHeader` serializable and deserializable (this is part of a larger effort to make almost everything we have serde serializable and deserializable, but this is just the part we need for AlphaNet)
5. Made the crypto types serde serializable and deserializable and made some small changes to the structure of the `SignatureWithPublicKey` type.


iamyulong (approved):
LGTM

PR #507, opened 2022-09-19
Feature: Native Methods in Manifest

talekhinezh:
* Add NativeMethod call in Manifest
* Remove SysUtils
* Add BURN_BUCKET and MINT_FUNGIBLE manifest instructions


radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -242,13 +242,15 @@ impl fmt::Debug for TransactionReceipt {
                         ScryptoValue::from_slice(&args).expect("Failed parse call data")
                     ),
                     ExecutableInstruction::CallMethod {
-                        component_address,
-                        method_name,
+                        method_identifier: MethodIdentifier::Scrypto {
+                            component_address,
+                            ident
+                        },
```
iamyulong, line 248: Would be nice if we format the native call args as well.

transaction/src/manifest/decompiler.rs
```diff
@@ -295,36 +298,78 @@ pub fn decompile(
...
+                                buf.push_str(&format!(
+                                    "MINT_FUNGIBLE ResourceAddress(\"{}\") Decimal(\"{}\") ;\n",
+                                    bech32_encoder.encode_resource_address(&resource_address),
+                                    amount,
+                                ));
+                            }
+                            _ => return Err(DecompileError::Unsupported),
+                        }
+                    }
+                    _ => return Err(DecompileError::Unsupported),
```
iamyulong, line 370: Suggest more explicit name: `UnrecognizedNativeFunction`

iamyulong (approved):
LGTM with minor comments.

PR #508, opened 2022-09-22
Feature: Refactor Transaction Keys

talekhinezh:
* Add ability for for transaction to be executed in Supervisor mode


transaction/src/validation/transaction_validator.rs
```diff
@@ -307,39 +303,64 @@ impl TransactionValidator {
...
-        // verify notary signature
-        let signed_intent_payload = transaction.signed_intent.to_bytes();
-        if !verify(
-            &signed_intent_payload,
-            &transaction.signed_intent.intent.header.notary_public_key,
-            &transaction.notary_signature,
-        ) {
-            return Err(SignatureValidationError::InvalidNotarySignature);
-        }
+            IntentActorProof::Supervisor => {
```
dhedey, line 349: What is the purpose of the supervisor? What does Supervisor mean?

To me, Supervisor suggests a higher-tier of user, and  suggests there's a third, even higher tier. I think I'd rather the term System :)

But yes - just to flag, I'm very concerned with the current impl that the TransactionValidator allows Supervisor signatures through unchallenged - because this means users could submit Supervisor transactions through the current node and have these work :)

IMO either:
* Supervisor transactions shouldn't use the TransactionValidator - in fact, they should go via an entirely different route, as they shouldn't take a NotarizedTransaction as input imo :)
* As a temporary workaround, TransactionValidator should have an option whether to allow/require Supervisor or User transactions - and default to only allowing User transactions.

transaction/src/model/validated_transaction.rs
```diff
@@ -35,8 +43,25 @@ impl ExecutableTransaction for ValidatedTransaction {
...
+                    PublicKey::EddsaEd25519(pk) => NonFungibleAddress::new(
+                        ED25519_TOKEN,
+                        NonFungibleId::from_bytes(pk.to_vec()),
+                    ),
+                    PublicKey::EcdsaSecp256k1(pk) => {
+                        NonFungibleAddress::new(ECDSA_TOKEN, NonFungibleId::from_bytes(pk.to_vec()))
+                    }
+                })
+                .collect(),
+            ValidatedTransactionActor::Supervisor => vec![NonFungibleAddress::new(
```
dhedey, line 60: It would be nice if there was a function to create this System Badge, as we're duplicating this code in a few places?

dhedey (changes requested):
Mostly looks good - except a bit of a concern about how "Supervisor" mode works in the TransactionValidator

transaction/src/validation/transaction_validator.rs
dhedey: In fact, to double down, we should keep: `transaction.signed_intent.intent_signatures` and not change it to `intent_actor_proof` imo.

A NotarizedTransaction should always be a user transaction.

talekhinezh: `Supervisor` refers to the updates which the validatorSet can perform (e.g. roundUpdates). A higher tier would be `System` which is something like genesis where almost arbitrary updates can be performed (This would be the same mechanism used for forking updates).

> because this means users could submit Supervisor transactions through the current node and have these work... which is scary to me
Yes, this was going to be resolved at the node side checking for these Supervisor transactions but I think you have a good point that `TransactionValidator` may be the right place to put this. This is essentially like a `sudo` check.

I don't think these should go down an entirely different route. We've got a good auth model in the engine and we should just leverage that. There is definitely some difference between the two though and I think we can slowly move the enum split between `User` and `Supervisor` higher as gain more info on where the right place is. I'd prefer not to split from `NotarizedTransaction` right now though.

transaction/src/validation/transaction_validator.rs
```diff
@@ -85,190 +161,110 @@ impl TransactionValidator {
...
         // semantic analysis
         let mut id_validator = IdValidator::new();
         for inst in &intent.manifest.instructions {
             match inst.clone() {
-                Instruction::TakeFromWorktop { resource_address } => {
+                Instruction::TakeFromWorktop { .. } => {
                     id_validator
                         .new_bucket()
                         .map_err(TransactionValidationError::IdValidationError)?;
-                    instructions.push(ExecutableInstruction::TakeFromWorktop { resource_address });
```
dhedey, line 111: I'm not sure I agree with this convergence of (Manifest)Instruction and (Executable)Instruction.

It removes the ability to have some instructions which can't be put into manifests / exposed to users => for someone writing a  Rust manifest, validation would be in code, not enforced at compile time.

Moreover, we will need (Manifest)Instructions to be versioned to a given Transaction version, whereas versioning of ExecutableInstructions will be tied to the Engine's execution format versioning. A given Engine version would have one type for ExecutableTransaction, but might need to support multiple Transaction formats. I really imagine we'd want to be able to version them separately.

Also, by changing this to be a clone at end, and not pushing instructions, we remove the ability to map Transaction Instructions to Executable Instructions.

transaction/src/validation/transaction_validator.rs
```diff
@@ -1,81 +1,157 @@
...
 pub struct ValidationConfig {
     pub network_id: u8,
     pub current_epoch: u64,
     pub max_cost_unit_limit: u32,
     pub min_tip_percentage: u32,
 }
 
-pub struct TransactionValidator;
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub struct NetworkTransactionValidator {
```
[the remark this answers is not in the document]
dhedey: I don't feel strong enough for this to block the PR on these fronts

iamyulong (approved):
LGTM

PR #512, opened 2022-09-27
Architecture: Authzone as RENode

talekhinezh:


radix-engine/src/engine/modules/auth_module.rs
```diff
@@ -19,16 +19,47 @@ impl AuthModule {
...
+            .find(|e| {
+                matches!(
+                    e,
+                    HeapRootRENode {
+                        root: HeapRENode::AuthZone(..),
+                        ..
+                    }
+                )
+            })
+            .expect("Could not find auth zone")
```
iamyulong, line 38: I think we should create an `AuthZone` even if the receiver is `&AuthZone` to have a simplified invariant - every frame has a call frame. :)

radix-engine/src/engine/node.rs
```diff
@@ -291,6 +293,20 @@ impl HeapRENode {
...
+        match self {
+            HeapRENode::AuthZone(auth_zone, ..) => auth_zone,
+            _ => panic!("Expected to be a resource manager"),
+        }
+    }
+
+    pub fn auth_zone_mut(&mut self) -> &mut AuthZone {
+        match self {
+            HeapRENode::AuthZone(auth_zone, ..) => auth_zone,
+            _ => panic!("Expected to be a resource manager"),
```
iamyulong, line 306: Needs to update panic messages.

radix-engine/src/fee/fee_table.rs
```diff
@@ -49,11 +70,15 @@ pub enum SystemApiCostingEntry<'a> {
...
+    ReadBlob {
+        size: u32,
+    },
     /// Generates a UUID.
     GenerateUuid,
     /// Emits a log.
-    EmitLog { size: u32 },
+    EmitLog {
+        size: u32,
+    },
```
iamyulong, line 81: Minor: I guess we have a different formatter :)

iamyulong (approved):
Looks better than what's presented during our last discussion. 👍

radix-engine/src/engine/modules/auth_module.rs
talekhinezh: sure :)

PR #513, opened 2022-09-28
Separate Node and Substate Types

iamyulong:
This PR introduces explicit types for RE nodes and substates, which helps us _NOT_ misuse substate for node or vice visa,  with compiler's assistance.  It further enforces the `node = substate*` paradigm and allows new capabilities to be built atop.

Given the scope of the changes introduced, some engine features will be temporarily disabled, like node rootness tracking, and many implementations are very rough which require further clean-up.


radix-engine/src/engine/kernel.rs
```diff
@@ -177,9 +176,12 @@ where
                 ))
             })?;
 
-        if matches!(substate_id, SubstateId::ComponentInfo(..)) {
+        if let SubstateId::ComponentInfo(address) = substate_id {
             node_pointer
-                .acquire_lock(substate_id.clone(), false, false, track)
+                .acquire_lock(SubstateId::ComponentState(*address), false, false, track)
```
talekhinezh, line 181: Not sure if ComponentState needs to be locked here?

talekhinezh (approved):

radix-engine/src/engine/kernel.rs
iamyulong: This is because dynamic loading of component state is not supported yet. In order to create a node representation, we need both the state and info, thus requiring both substates to be locked upfront.

Will be fixed in a follow-up PR. :)

PR #515, opened 2022-09-28
Add auth zone execution config that allows to virtualize proofs and a…

LukasGasior1:
… preview flag that uses it.

The **Situation** (describes where we are now)
When running a transaction preview all required signature badges (that the caller wishes to be available for execution) have to be explicitly specified (via the `signer_public_keys` parameter). Any attempt to create a proof for a key that's not listed will fail. 

The **Target** (describes what would be the desired outcome)
The caller of a tx preview should be able to set a single preview flag that would make a signature proof for any public key always implicitly available (and, as a result, key-based auth rules to always match).

The **Proposal** (describes a good way to get from Situation to Target)
Internal mechanism: I've added a new execution config parameter: `virtualizable_proofs_resource_addresses`, which specifies resource addresses for which virtual proofs can be created. A virtual proof is created ad-hoc in the `CreateProofByIds` call (only if it's allowed for the given resource addr and if a real proof is missing).
Preview flag:
Added a new preview flag `assume_all_signature_proofs`, which uses the `virtualizable_proofs_resource_addresses` config (it adds two virtualizable resource addresses: `ECDSA_TOKEN` and `ED25519_TOKEN`).

This PR covers two scenarios:
1. method authorization (access rules): in this case a virtual proof is not really created, but the code (in method_authorization.rs) inspects `virtualizable_proofs_resource_addresses` (by calling the `is_proof_virtualizable` method) and authorizes the call if a proof can be virtualized.
2. creating a proof (via a manifest instruction): a virtual proof is created in the auth zone (the `CreateProofByIds` fn)


radix-engine/src/model/method_authorization.rs
```diff
@@ -62,6 +62,16 @@ impl HardResourceOrNonFungible {
     }
 
     pub fn check(&self, auth_zones: &[&AuthZone]) -> bool {
+        // Check if a proof can be virtualized
+        for auth_zone in auth_zones {
+            if let HardResourceOrNonFungible::NonFungible(non_fungible_address) = self {
+                if auth_zone.is_proof_virtualizable(&non_fungible_address.resource_address()) {
```
talekhinezh, line 68: Seems to be wrong abstraction to need to check if proof is virtualized but fine for now. Can we just add a TODO comment that we should move this logic into `AuthZone` at some point

radix-engine/tests/preview.rs
```diff
@@ -37,9 +45,85 @@ fn test_transaction_preview_cost_estimate() {
...
+        &preview_flags,
+    );
+
+    test_runner
+        .execute_preview(preview_intent_proof, &network)
+        .unwrap()
+        .receipt
+        .expect_commit_success();
+
+    // Act & Assert #2: Check method authorization (withdrawal) without a proof in the auth zone
```
talekhinezh, line 97: Can we create two separate tests

radix-engine/src/engine/kernel.rs
```diff
@@ -70,6 +70,7 @@ where
     pub fn new(
         transaction_hash: Hash,
         initial_proofs: Vec<NonFungibleAddress>,
```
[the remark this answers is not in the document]
LukasGasior1: Cleaner for sure :+1:

radix-engine/src/engine/call_frame.rs
```diff
@@ -47,15 +47,23 @@ impl CallFrame {
...
         Y: SystemApi<'s, W, I, R>,
         W: WasmEngine<I>,
         I: WasmInstance,
         R: FeeReserve,
     {
-        let auth_zone = AuthZone::new();
+        let mut auth_zone = AuthZone::new();
+
+        // Copy the parent auth zone's virtualizable_proofs_resource_addresses over
+        let prev_frame_auth_zone = &system_api.auth_zone(depth - 1);
```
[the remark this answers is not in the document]
LukasGasior1: It's needed for the `CREATE_PROOF_FROM_AUTH_ZONE_BY_IDS` instruction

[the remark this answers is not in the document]
talekhinezh: Ah gotcha, can we add a TODO to clean this up in the future. I reckon this logic would exist in the AuthModule itself.

radix-engine/src/engine/call_frame.rs
```diff
@@ -104,4 +105,21 @@ impl CallFrame {
 
         Ok((taken, missing))
     }
+
+    pub fn auth_zone(&self) -> &AuthZone {
```
talekhinezh, line 109: Can we remove this from `CallFrame`? CallFrame should not there is such a thing as AuthZone. Better to have this function in AuthModule

LukasGasior1: ahh, gotcha, sure

PR #517, opened 2022-09-29
Architecture: Cleanup Method Invocation Interface

talekhinezh:
* Split Native Functions and Methods
* Scrypto Methods to only require the method string


radix-engine/src/engine/actor.rs
```diff
@@ -1,51 +1,73 @@
...
+        blueprint_name: String,
+        ident: String,
+    },
+    Native(NativeMethodFnIdent),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, TypeId, Encode, Decode)]
+pub struct FullyQualifiedMethod {
+    pub receiver: Receiver,
+    pub fn_ident: FullyQualifiedMethodFn,
```
iamyulong, line 16: I think we only need the function ident here and the fully qualified path can be resolved from receiver and function identifier, no?

transaction/src/model/instruction.rs
```diff
@@ -95,15 +82,15 @@ pub enum Instruction {
     ///
     /// Buckets and proofs in arguments moves from transaction context to the callee.
     CallFunction {
-        fn_identifier: FnIdentifier,
+        fn_identifier: FunctionIdent,
```
iamyulong, line 85: Minor: Should have consistent name like `function_identifier`

iamyulong (commented):
I know we've previously agreed to use `fn` to represent `function` OR `method`, but I've found our naming scheme quite overwhelming now.

Especially, I'd suggest we rename `FnIdent` to just `Name`, so `AuthZoneMethodFnIdent` becomes `AuthZoneMethodName`.
 
`FnIdent` is only for the top-level enum
```
#[derive(Debug, Clone, Eq, PartialEq, TypeId, Encode, Decode)]
pub enum FnIdent {
    Function(FunctionIdent),
    Method(MethodIdent),
}
```

Also, suggest we use either `identifier` or `ident` consistently.

radix-engine/src/engine/actor.rs
talekhinezh: Yup but the idea is that by the time the actor is created the fully qualified path has been resolved.

transaction/src/model/instruction.rs
talekhinezh: good point

talekhinezh:
@iamyulong I've updated some of the naming, see if this makes better sense

radix-engine/src/engine/actor.rs
iamyulong: That's fine. The reason I brought it up again is because of code like this:
``` 
(Receiver::Ref(RENodeId::Proof(proof_id)), NativeMethod::Proof(method)) => {
   Proof::main(proof_id, method, input, system_api).map_err(|e| e.into())
}
```
which can be simplified into 
```
Receiver::Ref(RENodeId::Proof(proof_id)) => {
   Proof::main(proof_id, method, input, system_api).map_err(|e| e.into())
}
```

The structure definition can prevent mismatching `Receiver` and `Method` statically.

And it's more object oriented which you've been advocating. :)

iamyulong (approved):
Much much easier to read now. Thanks for making the changes!

I left a comment on the `FullyQualifiedReceiverMethod`, but not a blocker.

PR #523, opened 2022-10-03
Architecture/key value map

iamyulong:
- Refactored `read_substate` and `write_substate` to use the substate Rust types
- Allowed component state to be dynamically loaded


radix-engine/src/engine/kernel.rs
```diff
@@ -178,32 +180,14 @@ where
...
-        }
-
         // Read current value
-        let current_value = {
+        let substate = {
             let mut node_ref = node_pointer.to_ref_mut(call_frames, track);
-            node_ref.read_scrypto_value(&substate_id)?
+            node_ref.read_substate(&substate_id)?
         };
+        let contained_value = extract_value_from_substate(substate_id, &substate);
```
talekhinezh, line 188: Seems like `extract_value_from_substate` should be managed by the caller of this method rather than in `read_substate_internal`

radix-engine/src/engine/kernel.rs
```diff
@@ -1158,7 +1121,7 @@ where
         // TODO: Authorization
 
         // Take any required child nodes
-        let children = re_node.get_child_nodes()?;
+        let children = re_node.get_loaded_child_nodes()?;
```
talekhinezh, line 1124: Why `loaded_child_nodes`? Feels wrong as this would mean that only loaded ones are taken?

radix-engine-stores/src/memory_db.rs
```diff
@@ -56,14 +56,17 @@ impl WriteableSubstateStore for SerializedInMemorySubstateStore {
 }
 
 impl QueryableSubstateStore for SerializedInMemorySubstateStore {
-    fn get_kv_store_entries(&self, kv_store_id: &KeyValueStoreId) -> HashMap<Vec<u8>, Substate> {
+    fn get_key_value_store_entries(
```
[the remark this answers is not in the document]
iamyulong: I don't mind either way, as long as they're consistent. :)

radix-engine/src/engine/kernel.rs
```diff
@@ -1588,3 +1536,28 @@ where
...
     }
 }
+
+// Parse contained value (for reference management and children detection)
+// TODO: introduce trait/enum for readable substates.
+fn extract_value_from_substate(
+    substate_id: &SubstateId,
+    substate: &ScryptoValue,
+) -> Option<ScryptoValue> {
+    match substate_id {
```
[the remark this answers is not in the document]
iamyulong: As discussed, this is because we don't support `any` value in `ScryptoValue` and this function help extract the value from the `Vec<u8>`

radix-engine/src/engine/kernel.rs
```diff
@@ -1588,3 +1536,28 @@ where
         Ok(())
     }
 }
+
+// Parse contained value (for reference management and children detection)
+// TODO: introduce trait/enum for readable substates.
```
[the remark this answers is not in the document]
iamyulong: Answered in the other comment.

radix-engine-stores/src/memory_db.rs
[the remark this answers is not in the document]
iamyulong: Renamed all to `kv_store`.

radix-engine/src/engine/kernel.rs
iamyulong: Updated.

radix-engine/src/engine/kernel.rs
```diff
@@ -632,7 +616,7 @@ where
                 node_pointer
                     .acquire_lock(substate_id.clone(), false, false, &mut self.track)
                     .map_err(RuntimeError::KernelError)?;
-                let node_ref = node_pointer.to_ref(&self.call_frames, &mut self.track);
+                let mut node_ref = node_pointer.to_ref(&self.call_frames, &mut self.track);
```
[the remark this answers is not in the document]
iamyulong: As discussed, this mainly because we need to update the cache of the track. If we move the node initialisation to when a reference is created, we will be able to get rid of the `mut` requirement.

radix-engine/src/engine/kernel.rs
iamyulong: Renamed to the original version and added note that it should NOT be used for tracked nodes.

talekhinezh: Still looks like the new version? :)

iamyulong: Don't know what happened to my change. :(

Will add again.

talekhinezh (approved):

PR #527, opened 2022-10-04
Architecture/non fungible store

iamyulong:
This PR introduces `NonFungibleStore` - child node of `ResourceManager`.


radix-engine/src/model/nodes/bucket.rs
```diff
@@ -293,37 +296,30 @@ impl Bucket {
...
+                let bucket = node_ref.bucket();
+                let resource_address = bucket.resource_address();
 
-                Ok(ScryptoValue::from_typed(&()))
+                let bucket_id = match node_id {
+                    RENodeId::Bucket(bucket_id) => bucket_id,
+                    _ => panic!("Unexpected"),
+                };
+
+                system_api
```
talekhinezh, line 310: Can we keep the original TODO comment? I think there may be better ways of signaling updates to total supply over a method call.

radix-engine/src/model/nodes/bucket.rs
```diff
@@ -293,37 +296,30 @@ impl Bucket {
...
-                    .node_drop(&node_id)
-                    .map_err(InvokeError::Downstream)?
-                    .into();
-
-                // Notify resource manager, TODO: Should not need to notify manually
-                let resource_address = bucket.resource_address();
-                let mut value = system_api
-                    .borrow_node_mut(&RENodeId::ResourceManager(resource_address))
+                let node_ref = system_api
+                    .borrow_node(&node_id)
```
talekhinezh, line 300: Why just borrow here? Is this a change of abstraction?

radix-engine/src/model/nodes/bucket.rs
iamyulong: Yes, sure!

radix-engine/src/model/nodes/bucket.rs
iamyulong: This is because when invoking `Bucket::burn`, the call frame only needs to **read** the resource address, such that it can call `ResourceManager::burn` by passing the `self`.

talekhinezh (approved):

PR #528, opened 2022-10-04
Various alphanet fixes - AKA the big displayable refactor

dhedey:
I could be argued into this being merged into `develop` instead of `alphanet`.

The original goal of this PR was simply to fix two issues:
* The fact that the node error messages display simulator addresses, which is confusing - much better to just display some name and then some hex to avoid confusing people.
* Some hacky implementations of Clone/Eq in babylon-node for the tests.

The first point resulted in my wanting to remove the from string implementation on the addresses, but this had many knock-on issues:
* (1) The import! macro -- this was actually partly broken - some tests assumed the package address was hex, others assumed it was a simulator address. I've gone back to assuming this is just hex for now... Although I think we might want to redesign how it works (as per (3)) or even potentially consider dropping the `import!` macro in future.
* (2) FromStr was needed for the resim command line. My workaround was to introduce wrapper types, and implement FromStr from simulator addresses there.
* (3) This was needed for the `external_blueprint!` macro... After some humming and harring, I decided to change the `external_blueprint!` macro to only be for blueprint/package calls, and to take a package address at runtime. I think this is much nicer to be honest! Although the docs will need an update.

I ended up removing the `fmt::Display` implementations for Addresses, ScryptoValues and Receipts, replacing it with a common `.displayable(...)` abstraction. Debug impls still exist, but these should be restricted to debugging, not printing for user messages. I quite like this abstraction, but would be interested to hear your thoughts on it... and the other workarounds / fixes.

NB: All the engine changes are actually compatible with a developer using the `main` branch, so I suggest we merge this into alphanet, but thoughts appreciated.


scrypto/src/values.rs
```diff
@@ -539,119 +598,120 @@ impl ScryptoValueFormatter {
...
                 buf.push_str(", ");
             }
-            buf.push_str(Self::format_value(x, bucket_ids, proof_ids).as_str());
+            buf.push_str(Self::format_value(x, context)?.as_str());
         }
-        buf
+        Ok(buf)
     }
+
     pub fn from_custom_value(
```
iamyulong, line 620: I realised this should be `format_custom_value`.

scrypto/src/values.rs
```diff
@@ -413,13 +461,30 @@ impl CustomValueVisitor for ScryptoCustomValueChecker {
...
+    bech32_encoder: Option<&'a Bech32Encoder>,
+    bucket_ids: &'a HashMap<BucketId, String>,
+    proof_ids: &'a HashMap<ProofId, String>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ScryptoValueFormatterError {
+    InvalidTypeId(u8),
+    InvalidCustomData(ScryptoCustomValueCheckError),
+}
```
iamyulong, line 474: The original intention of `ScryptoValueFormatter` is to format `ScryptoValue` (validated) so we should panic if unable to format, so to help identify implementation bug. 

With many of the methods in the formatter being public and can be used for other use cases, I'm also okay with this error code approach.

iamyulong (approved):
Looks good to me overall. 

Only concern I had is about the additional display struct types, but not a blocker.

PR #533, opened 2022-10-05
Architecture: Substate Offsets

talekhinezh:


radix-engine/src/engine/node_properties.rs
```diff
@@ -66,62 +62,30 @@ impl RENodeProperties {
...
-            SubstateId::Worktop => false, // TODO: Fix
+    pub fn can_own_nodes(offset: &SubstateOffset) -> bool {
+        match offset {
+            SubstateOffset::Global(..) => true,
+            SubstateOffset::AuthZone(..) => false,
+            SubstateOffset::Component(ComponentOffset::State) => true,
+            SubstateOffset::Component(ComponentOffset::Info) => false,
+            SubstateOffset::Resource(ResourceManagerOffset::NonFungible(..)) => false,
+            SubstateOffset::Resource(ResourceManagerOffset::NonFungibleSpace) => false,
+            SubstateOffset::Resource(ResourceManagerOffset::ResourceManager) => false,
```
iamyulong, line 80: Can we rename to `SubstateOffset::ResourceManager`?

radix-engine/src/engine/node_ref.rs
```diff
@@ -16,30 +16,56 @@ pub enum RENodePointer {
 }
 
 impl RENodePointer {
+    pub fn node_id(&self) -> RENodeId {
+        match self {
+            RENodePointer::Heap { root, id, .. } => id.unwrap_or(*root),
```
iamyulong, line 21: Why do we return the root node id? can the pointer points to a child node?

iamyulong: Never mind. Misread the code. It falls back to root.

iamyulong (approved):

PR #538, opened 2022-10-09
Architecture: Implement Lock Handles

talekhinezh:


radix-engine/src/engine/kernel.rs
```diff
@@ -501,8 +452,28 @@ where
...
+        for (_, lock) in call_frame.drain_locks() {
+            let SubstateLock {
+                pointer: (node_pointer, offset),
+                ..
+            } = lock;
+            if !(matches!(offset, SubstateOffset::KeyValueStore(..))
+                || matches!(
+                    offset,
+                    SubstateOffset::ResourceManager(ResourceManagerOffset::NonFungible(..))
+                ))
```
iamyulong, line 467: This logic is duplicated a few times. Suggest we add a helper method  `is_key(&self) -> bool` to `SubstateOffset`

iamyulong (approved):

PR #539, opened 2022-10-10
Add fee payments to ExecutionTrace resource changes receipt

LukasGasior1:
Current situation:
Fee payments are not being recorded in execution trace resource changes.

Desired situation:
Resource changes should include fee payments.

Proposed solution:
Nearly all needed information to keep track of the changes is already in FeeSummary. What's missing is vault's component addresses. Execution trace stores those addresses and later combines them with fee summary to create corresponding ResourceChange entries.

### DIFF
--- a/radix-engine/src/engine/modules/execution_trace.rs
+++ b/radix-engine/src/engine/modules/execution_trace.rs
@@ -19,13 +19,20 @@ pub struct ExecutionTraceReceipt {
 
 #[derive(Debug)]
 pub struct ExecutionTrace {
+    /// Stores resource changes that resulted from vault's put/take operations.
     pub resource_changes: HashMap<ComponentId, HashMap<VaultId, (ResourceAddress, Decimal)>>,
+
+    /// Stores component IDs associated with vaults that have been used to lock a fee.
+    /// This, together with a FeeSummary, is later used to create ResourceChange entries
+    /// for fee payments (incl. any refunds back to the vault).
+    pub fee_vaults_components: HashMap<VaultId, ComponentId>,
 }
 
 impl ExecutionTrace {
     pub fn new() -> ExecutionTrace {
         Self {
             resource_changes: HashMap::new(),
+            fee_vaults_components: HashMap::new(),
         }
     }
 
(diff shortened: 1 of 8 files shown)


radix-engine/src/engine/track.rs
```diff
@@ -518,7 +519,7 @@ impl<'s, R: FeeReserve> Track<'s, R> {
     pub fn finalize(
         mut self,
         invoke_result: Result<Vec<Vec<u8>>, RuntimeError>,
-        resource_changes: Vec<ResourceChange>, // TODO: wrong abstraction, resource change should be derived from track instead of kernel
+        execution_trace: ExecutionTrace, // TODO: wrong abstraction, resource change should be derived from track instead of kernel
```
LukasGasior1, line 522: I'm not sure if/why the abstraction is wrong here?

iamyulong: Ah, I left the comment. I was thinking that resource changes (not resource movements) should be generated through the `Track` rather that execution trace, so no need to pass the argument to `finalize`.

iamyulong: If we look at all the node/substate changes made during the transaction, we should be able to tell which component has gained/lost XRD by how much.

iamyulong: On the other hand, if we want to trace resource movements from X to Y, we will need to extract from execution trace.

LukasGasior1: Hmh, maybe I'll need to refine my mental model :) I view track as being quite low-level (i.e. working with substates etc; inseparable component of the engine), while the resource changes, or the execution trace module in general, is a higher-level abstraction (ideally pluggable at some point - i.e. only enabled for tx preview), which provides some insights of the execution to the user.

iamyulong: The track is doing a bit more than substates - it collects all transaction wide side effects. 

And yes some insights must be from the execution trace, especially resource flow and call stack. Track is more of a flattened view of the state/resource changes, which make it easier to be used for generating diffs.

Since this PR is out, I will review it using your model for now.

radix-engine/src/engine/modules/execution_trace.rs
```diff
@@ -164,11 +181,22 @@ impl ExecutionTrace {
...
 
         vault_change.1 += amount;
-
-        Ok(())
     }
 
-    pub fn to_receipt(self) -> ExecutionTraceReceipt {
+    pub fn to_receipt(
+        mut self,
+        fee_payments: HashMap<VaultId, (ResourceAddress, Decimal)>,
```
iamyulong, line 188: Only need a reference to the `HashMap` right?

iamyulong (commented):
LGTM with minor comment.

radix-engine/src/engine/modules/execution_trace.rs
iamyulong: Looks like this is a map constructed for receipt generation only, so no big deal.

iamyulong (approved):

PR #540, opened 2022-10-11
Refactor/move execution trace to module

iamyulong:


radix-engine/src/engine/modules/execution_trace.rs
```diff
@@ -18,41 +18,89 @@ pub struct ExecutionTraceReceipt {
...
+        _call_frames: &mut Vec<CallFrame>,
+        _vault_id: VaultId,
+        fee: Resource,
+        _contingent: bool,
+    ) -> Result<Resource, ModuleError> {
+        Ok(fee)
+    }
 }
 
 impl ExecutionTrace {
```
LukasGasior1, line 85: now that it's a full-fledged module, should we rename it to ExecutionTraceModule?

radix-engine/src/engine/modules/execution_trace.rs
```diff
@@ -62,157 +110,194 @@ impl ExecutionTrace {
...
+
+    pub fn new(
+        ops: Vec<(REActor, VaultId, VaultOp)>,
+        actual_fee_payments: HashMap<VaultId, Decimal>,
+        state_track: &mut AppStateTrack,
+        is_commit_success: bool,
+    ) -> Self {
+        let mut vault_changes = HashMap::<ComponentId, HashMap<VaultId, Decimal>>::new();
+        let mut vault_locked_by = HashMap::<VaultId, ComponentId>::new();
+        for op in ops {
```
LukasGasior1, line 221: any reason not to name the tuple components here? (`for (actor, vault_id, vault_op) in ops`)
not a big fan of using indices

radix-engine/src/engine/modules/execution_trace.rs
```diff
@@ -62,157 +110,194 @@ impl ExecutionTrace {
...
+
+impl ExecutionTraceReceipt {
+    // TODO: is it better to derive resource changes from substate diff, instead of execution trace?
+    // The current approach relies on various runtime invariants.
+
+    pub fn new(
+        ops: Vec<(REActor, VaultId, VaultOp)>,
+        actual_fee_payments: HashMap<VaultId, Decimal>,
+        state_track: &mut AppStateTrack,
+        is_commit_success: bool,
```
LukasGasior1, line 217: Does this correspond to `TransactionResult Commit/Reject` or `TransactionOutcome Success/Failure`? Should we use one of those names here for consistency?

LukasGasior1 (approved):
Just a few minor comments; Nice refactoring :+1: LGTM

radix-engine/src/engine/modules/execution_trace.rs
iamyulong: Yes, will do.

radix-engine/src/engine/modules/execution_trace.rs
iamyulong: Good idea. That's definitely better!

radix-engine/src/engine/modules/execution_trace.rs
iamyulong: Yes, but I think it's consistent regards to naming:
- `Commit`
   * `Success`
   * `Failure`
- `Reject`

LukasGasior1: Ahh, right `is_Commit->Success`, makes sense

talekhinezh (approved):

PR #541, opened 2022-10-11
Architecture: Cleanup InvokeError::Downstream

talekhinezh:
Implemented `From<RuntimeError>` so that error cruft is reduced

### DIFF
--- a/radix-engine/src/model/nodes/resource_manager.rs
+++ b/radix-engine/src/model/nodes/resource_manager.rs
@@ -314,8 +314,7 @@ impl ResourceManager {
 
                 let resource_node_id = if matches!(input.resource_type, ResourceType::NonFungible) {
                     let non_fungible_store_node_id = system_api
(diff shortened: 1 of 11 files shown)


radix-engine/src/model/nodes/resource_manager.rs
```diff
@@ -399,17 +392,14 @@ impl ResourceManager {
                         ),
                     };
                     let bucket_id = system_api
-                        .node_create(HeapRENode::Bucket(Bucket::new(container)))
-                        .map_err(InvokeError::Downstream)?
+                        .node_create(HeapRENode::Bucket(Bucket::new(container)))?
                         .into();
```
iamyulong, line 396: Can remove the `into()` after question mark. There are a few other occurrences too.

talekhinezh: `.into()` is required to change `RENodeId` to `BucketId`

PR #542, opened 2022-10-11
Architecture: Replace Read/Write Substate interface with SubstateRef and SubstateRefMut

talekhinezh:
* Replace substate read/write semantics with SubstateRef and SubstateRefMut
* Moved substate read/write semantics into WASM Runtime
* Use SubstateRef interface in some native blueprints (`resource_manager` and `component`)


radix-engine/src/model/substates/substate.rs
```diff
@@ -266,3 +261,337 @@ impl Into<GlobalAddressSubstate> for Substate {
...
+                Substate::KeyValueStoreEntry(next),
+            ) => *current = next,
+            (RawSubstateRefMut::NonFungible(current), Substate::NonFungible(next)) => {
+                *current = next
+            }
+            (RawSubstateRefMut::System(current), Substate::System(next)) => *current = next,
+            _ => return Err(RuntimeError::KernelError(KernelError::InvalidOverwrite)),
+        }
+
+        self.flush()
```
iamyulong, line 500: Better not to flush automatically?

radix-engine/src/model/substates/substate.rs
```diff
@@ -266,3 +261,337 @@ impl Into<GlobalAddressSubstate> for Substate {
...
+    }
+
+    pub fn get_raw_mut(&mut self) -> RawSubstateRefMut {
```
iamyulong, line 503: Why do we need two write interfaces, `overwrite` and `get_raw_mut`? With a raw mutable reference, we can already overwrite the value.

radix-engine/src/model/substates/substate.rs
```diff
@@ -266,3 +261,337 @@ impl Into<GlobalAddressSubstate> for Substate {
...
+            track,
+        };
+        Ok(substate_ref_mut)
+    }
+
+    pub fn offset(&self) -> &SubstateOffset {
+        &self.offset
+    }
+
+    pub fn flush(&mut self) -> Result<(), RuntimeError> {
```
iamyulong, line 443: Should we consume the `RefMut` upon `flush`? It seems flush can only be done once.

radix-engine/src/engine/modules/module.rs
```diff
@@ -31,9 +31,8 @@ pub enum SysCallInput<'a> {
     ReadSubstate {
         lock_handle: &'a LockHandle,
     },
-    WriteSubstate {
+    GetMut {
```
iamyulong, line 34: Minor: Can we rename to `GetRefMut`/`get_ref_mut` for consistence?

radix-engine/src/model/substates/substate.rs
talekhinezh: This was really a convenience method to "flash" a substate. Will see if it makes sense to remove this.

radix-engine/src/model/substates/substate.rs
talekhinezh: `overwrite` was really a convenience method to "flash" a substate. Will see if it makes sense to remove this.

radix-engine/src/model/substates/substate.rs
talekhinezh: hrm, good idea! flush could be called multiple times currently but probably simpler to only allow it once.

iamyulong (approved):

PR #544, opened 2022-10-12
Fix/contextual display neatenings

dhedey:
A follow-up PR to https://github.com/radixdlt/radixdlt-scrypto/pull/528

Includes:
* Allowing ScryptoValue / Value / TransactionReceipt to be formatted using ContextualDisplay
* ScryptoValue is formatted without large string allocations
* Added rust docs to ContextualDisplay trait to aid with its usage


iamyulong (approved):
LGTM

PR #547, opened 2022-10-12
Feature: Allow pre-reads on fee vault

talekhinezh:
This is a feature required for better substate management. Track layer load state should not get exposed to application layer. 

This break of abstraction prevented allowing the application layer from obtaining locks on it's own since the AuthModule acquires these locks first.


radix-engine/src/engine/track.rs
```diff
@@ -233,10 +247,13 @@ impl<'s, R: FeeReserve> Track<'s, R> {
...
             LockState::Read(n) => loaded_substate.lock_state = LockState::Read(n - 1),
-            LockState::Write => loaded_substate.lock_state = LockState::no_lock(),
+            LockState::Write => {
+                loaded_substate.lock_state = LockState::no_lock();
+                loaded_substate.metastate = SubstateMetaState::Updated;
+            }
         }
 
-        if write_through {
+        if unmodified_base {
```
iamyulong, line 256: I think `unmodified_base` only capture one aspect - the substate must pre-exist and unchanged. It does not indicate that any change must be immediately committed.

How about two flags `unmodified | write_through`?

iamyulong (approved):
Nice improvement!

radix-engine/src/engine/track.rs
talekhinezh: hrm good point

talekhinezh: what if we called it `force_write` instead

iamyulong: Sounds good.

PR #548, opened 2022-10-12
Architecture: Unify to use of SubstateRef/RefMut for all native execution

talekhinezh:
* Updated all native execution to use substateRef/RefMut interface. All substate locking is now done dynamically controlled by the application layer
* Removed track node interface
* Separated `Substate` into `PersistedSubstate` and `RuntimeSubstate`


radix-engine/src/model/substates/substate.rs
```diff
@@ -20,23 +18,116 @@ pub enum Substate {
...
-impl Substate {
+impl Into<VaultSubstate> for PersistedSubstate {
+    fn into(self) -> VaultSubstate {
+        if let PersistedSubstate::Vault(vault) = self {
+            vault
+        } else {
+            panic!("Not a vault");
+        }
+    }
+}
```
iamyulong, line 29: Why do we need to add conversion for `VaultSubstate` only?

radix-engine/src/model/substates/substate.rs
```diff
@@ -263,12 +356,16 @@ impl Into<GlobalAddressSubstate> for Substate {
 }
 
 pub enum SubstateRef<'a> {
+    AuthZone(&'a AuthZone),
+    Worktop(&'a Worktop),
+    Proof(&'a Proof),
+    Bucket(&'a Bucket),
```
iamyulong, line 362: Can we have `AuthZone/Worktop/Proof/BucketRuntimeSubstate`?

radix-engine/src/engine/node.rs
```diff
@@ -8,7 +8,7 @@ pub enum HeapRENode {
     Bucket(Bucket),
     Proof(Proof),
     AuthZone(AuthZone),
-    Vault(Vault),
+    Vault(VaultRuntimeSubstate),
```
iamyulong, line 11: Can we have a structure for Vault?
```
struct Vault {
    substate: VaultRuntimeSubstate
}
```
I'm imagining we will have multiple states for `Vault` node.

radix-engine/src/model/substates/substate.rs
talekhinezh: Currently used for `test_runner` to inspect the vault without seeing the runtime representation. All other substates just use `to_runtime().into()` since they have the same runtime/persisted representations.

radix-engine/src/model/substates/substate.rs
talekhinezh: Yup

radix-engine/src/engine/node.rs
iamyulong: Cool

iamyulong (approved):

PR #551, opened 2022-10-17
Feature/native invoke

iamyulong:
- Added syntax for native invoke
- Refactored manifest decompiler to be more lenient
- Refactored `FnIdent` types to separate Scrypto and Native calls (with decoupling of `ResolvedFunction/Method` from `FnIdent`)
- Limited Scrypto function call to package and method call to component only


scrypto/src/core/invocation.rs
```diff
@@ -1,102 +1,151 @@
...
-pub enum FunctionIdent {
-    Scrypto {
-        package_address: PackageAddress,
-        blueprint_name: String,
-        ident: String,
-    },
-    Native(NativeFunction),
+pub struct NativeFunctionIdent {
+    pub blueprint_name: String,
+    pub function_name: String,
```
talekhinezh, line 20: What's the reasoning to keep these as strings? Seems like we lose the nice type checking and possibly integrating with args in the future.

scrypto/src/engine/api.rs
```diff
@@ -31,7 +31,10 @@ macro_rules! native_functions {
 
 #[derive(Debug, TypeId, Encode, Decode)]
 pub enum RadixEngineInput {
-    Invoke(FnIdent, Vec<u8>),
+    InvokeScryptoFunction(ScryptoFunctionIdent, Vec<u8>),
+    InvokeScryptoMethod(ScryptoMethodIdent, Vec<u8>),
+    InvokeNativeFunction(NativeFunctionIdent, Vec<u8>),
+    InvokeNativeMethod(NativeMethodIdent, Vec<u8>),
```
talekhinezh, line 37: Would rather see this as a single entry. Want to keep abstraction about which interpreter to call outside of this interface.

transaction/examples/call_method.rtm
```diff
@@ -0,0 +1,38 @@
+# Inovke scrypto method (both global & and)
+CALL_METHOD ComponentAddress("component_sim1qgvyxt5rrjhwctw7krgmgkrhv82zuamcqkq75tkkrwgs00m736") "free_xrd";
+CALL_METHOD Component("000000000000000000000000000000000000000000000000000000000000000005000000") "free_xrd";
+
+# Invoke native method (ref)
+TAKE_FROM_WORKTOP ResourceAddress("resource_sim1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzqu57yag") Bucket("xrd");
+CREATE_PROOF_FROM_AUTH_ZONE ResourceAddress("resource_sim1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqzqu57yag") Proof("proof");
+CALL_NATIVE_METHOD &Bucket("xrd") "get_resource_address";
```
talekhinezh, line 8: I think seems wrong to include `&` in these. Compiler should be smart enough to know.

iamyulong: The idea was to give some visual representation of passing by reference, in distinction from passing by ownership. Since we're removing support of passing by ownership, I don't mind removing the `&` by that time.

Also, I didn't want to introduce too much compiler inference about nagive funciton/method ABI, i.e. the compiler/decompiler should continue to work even if we've added/deleted native methods.

scrypto/src/engine/api.rs
iamyulong: I've left a TODO to combine `fn_ident` and `args` into something like `ScryptoMethodInvocation`. I will do it after this PR.

scrypto/src/core/invocation.rs
iamyulong: Two reasons:
- Decouple the interface from engine implementation
- Be consistent with Scrypto invocation

We're not losing compiler check since we immediately convert it into an enum upon `invoke_native` sys calls. 

The only downside of using string as discriminator is performance loss. We will need to address this separately if it proves to be a problem.

radix-engine/src/engine/actor.rs
```diff
@@ -32,17 +45,14 @@ pub enum ResolvedFunction {
 #[derive(Debug, Clone, Eq, PartialEq, TypeId, Encode, Decode)]
 pub enum REActor {
     Function(ResolvedFunction),
-    Method(ResolvedReceiverMethod),
+    Method(ResolvedMethod, Receiver),
```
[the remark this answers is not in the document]
iamyulong: Let's chat about this, but the motivation for moving the method first was to emphasise the Function/Method theme. Most of the engine logic only distinguishs between funciton and method, regardless the state of the receiver.

Anothe refactor I'm going to propose is to move down the generic `Recevier` and replace it with concrete types, like
```
pub enum ResolvedMethod {
    BucketMethod(String, BucketId),
    VaultMethod(String, VaultId)
}
```
This is to eliminate the state of mismatching method and receiver type, so no runtime assumption is needed.

scrypto/src/engine/api.rs
iamyulong: Actually, let me do that within this PR. Doesn't seem to be too big a change.

iamyulong: Looks like I misunderstood the question. Are you proposing we have an enum that captures all four variants OR we combine the ident with the args?

radix-engine/src/engine/actor.rs
```diff
@@ -44,44 +52,89 @@ impl ResolvedReceiver {
...
 
+#[derive(Clone, Eq, PartialEq, TypeId, Encode, Decode)]
+pub enum ResolvedMethod {
+    Scrypto {
+        package_address: PackageAddress,
+        blueprint_name: String,
+        ident: String,
+        export_name: String,
+        return_type: Type,
+        code: Vec<u8>,
```
talekhinezh, line 56: perfect!

radix-engine/src/engine/kernel.rs
```diff
@@ -469,139 +446,346 @@ where
...
-                            InvokeError::Error(error) => RuntimeError::InterpreterError(
-                                InterpreterError::InvalidScryptoFunction(
-                                    function_ident.clone(),
-                                    error,
-                                ),
-                            ),
-                        })
-                    })?;
+                // Deref if global
+                let resolved_receiver =
```
talekhinezh, line 553: Resolving deref shouldn't be a scrypto interpreter decision but decision made in the kernel side. Can we move this back?

radix-engine/src/engine/kernel.rs
```diff
@@ -679,20 +863,20 @@ where
         &Self::current_frame(&self.call_frames).actor
     }
 
-    fn invoke(
+    fn invoke_scrypto(
```
talekhinezh, line 803: Not sure why we're splitting up this logic again after much work went into unifying? We should only be splitting at the points where interpreter should be making the decision.

iamyulong: Good point! Will do.

radix-engine/src/engine/kernel.rs
iamyulong: As discussed, main reason for moving deref into fn resolver is because there are already derefing happening there, e.g. package address deref. 

Easier to have all fn resoliving logic in one place, so can be moved to the interepreter as a whole later on.

talekhinezh (approved):

PR #554, opened 2022-10-20
lint: fix `clippy::redundant_clone`

mozi-rdx:
This commit is the result of running
```
cargo clippy --fix -- --allow clippy::all --warn clippy::redundant-clone
```

> What it does
> Checks for a redundant clone() (and its relatives) which clones an owned value that is going to be dropped without further use.
>
> Why is this bad?
> It is not always possible for the compiler to eliminate useless allocations and deallocations generated by redundant clone()s.
>
> Known problems
> False-negatives: analysis performed by this lint is conservative and limited.


iamyulong (approved):
Nice!

PR #557, opened 2022-10-21
Architecture: Flatten Heap Substates

talekhinezh:
* Flatten Heap to substates so that it's structure is closer to the Store
* Refactored DropNode logic
* Add DropNodeVisibility property


radix-engine/src/engine/heap.rs
```diff
@@ -112,3 +149,63 @@ impl Heap {
...
+}
+
+impl Into<ProofSubstate> for HeapRENode {
+    fn into(mut self) -> ProofSubstate {
+        self.substates
+            .remove(&SubstateOffset::Proof(ProofOffset::Proof))
+            .unwrap()
+            .into()
+    }
+}
```
iamyulong, line 211: This is a general comment. I think we should use `TryInto` trait for conversions that may fail.

radix-engine/src/engine/kernel.rs
```diff
@@ -339,6 +339,23 @@ where
...
+        for node_id in owned_nodes {
+            if let RENodeId::Worktop = node_id {
+                worktops.push(node_id);
+            } else {
+                self.drop_node(node_id)?;
+            }
+        }
+        for worktop_id in worktops {
+            self.drop_node(worktop_id)?;
+        }
```
iamyulong, line 354: This is interesting. Why do we need to drop other nodes first? For the proofs?

radix-engine/src/engine/kernel.rs
```diff
@@ -949,7 +970,10 @@ where
 
         // TODO: Move this into higher layer, e.g. transaction processor
         if self.current_frame.depth == 0 {
-            self.current_frame.drop_frame(&mut self.heap)?;
+            self.current_frame.drop_all_locks(&mut self.track)?;
+            self.execution_mode = ExecutionMode::Application;
+            self.drop_nodes_in_frame()?;
+            self.execution_mode = ExecutionMode::Kernel;
```
iamyulong, line 976: I think the current of execution mode management is very fragile and error prone. Needs a better machanism for execution mode switching. :)

radix-engine/src/model/substates/package.rs
```diff
@@ -1,14 +1,21 @@
 use sbor::rust::fmt::Debug;
+use std::fmt::Formatter;
```
iamyulong, line 2: Minor: can we use `sbor::rust::fmt::Formatter;`?

iamyulong (commented):
LGTM with minor comments.

radix-engine/src/engine/kernel.rs
talekhinezh: Yup, I didn't totally finish this part, just wanted to get the PR out as soon as possible :) will cleanup :)

iamyulong (approved):

PR #558, opened 2022-10-21
Feature/expose genesis tx and receipt

LukasGasior1:


scrypto/src/component/component.rs
```diff
@@ -208,8 +208,6 @@ pub enum ComponentAddress {
     System([u8; 26]),
 }
 
-impl ComponentAddress {}
```
LukasGasior1, line 211: those empty impl blocks were shadowing the impl below (with `to_vec`), not sure if that's expected in rust, or a bug (or I didn't know how to use it)

iamyulong: It should no shadow other `impl` for the struct, but we should remove it.

iamyulong (approved):

scrypto/src/component/component.rs
iamyulong: It should not shadow other `impl`s, but it's nice to remove them.

PR #560, opened 2022-10-25
Feature/system address

iamyulong:
This PR introduces `SystemAddress` as a top-level global address, along with some clean-up and renaming of epoch manager and faucet.

Previously, we were using system component to represent this category of entities, but it's not accurate. Such entities don't have any of the traits implemented for a component.

Now, a `SystemAddress` points to an RENode owned and exposed by the system, such as epoch manager. `invoke_native` should be used for interacting with entities behind a system address.

```rust
pub enum SystemAddress {
     EpochManager,  // Entity ID: 0x04
     // More variants and entity IDs may be introduced in future, like staking
}
```

**Breaking changes:**
- Added `SystemAddress` type and `GlobalAddress::System` variant
- Removed `ComponentAddress::System` variant
- Renamed `System` RENode to `EpochManager`
- Moved `GetTransactionHash` to a top-level Radix Engine API
- Removed `sys-` prefix for `Faucet`, as they're normal Scrypto components 
   * Note that faucet component Bech32 address starts with `component_`
- Added `SystemAddress` support for manifest


radix-engine-stores/src/rocks_db.rs
```diff
@@ -80,10 +80,6 @@ impl RadixEngineDB {
 
     pub fn list_components(&self) -> Vec<ComponentAddress> {
         let mut addresses = Vec::new();
-        addresses.extend(self.list_components_helper(
-            ComponentAddress::System([0u8; 26]),
-            ComponentAddress::System([255u8; 26]),
```
talekhinezh, line 85: Let's talk to Russ/Matt about this change as like we talked about I am under the impression that we only want a few types of top level things: Component/Resource/Package.

radix-engine/src/engine/interpreters/wasm_runtime.rs
```diff
@@ -186,6 +186,10 @@ where
         return Ok(self.actor.clone());
     }
 
+    fn handle_get_transaction_hash(&mut self) -> Result<Hash, RuntimeError> {
```
talekhinezh, line 189: Can we move this into another RENode rather than a separate system call? It would be a RENode which contains other info like time/validator proposing/etc.

radix-engine-stores/src/rocks_db.rs
iamyulong: Yes, I will add it to our to discuss list.

PR #561, opened 2022-10-26
Feature: Simple Virtual Account

talekhinezh:
* Implement virtual accounts through lazy loading
* Use hashed public key instead of public key for virtual proofs


iamyulong (approved):
Very elegant implementation.

PR #563, opened 2022-10-26
Feature/update manifest model

iamyulong:
Manifest "any" value model is bit outdated, which may fail the the decompiler when certain inputs are provided. This PR updates it to be consistent with the latest `ScryptoValue` model.

Notably, I also re-arranged the type IDs for Scrypto types, which may impact downsteam tools. cc @0xOmarA


transaction/src/manifest/parser.rs
```diff
@@ -282,19 +282,30 @@ impl Parser {
...
-            | TokenKind::ResourceAddress
-            | TokenKind::Hash
-            | TokenKind::Bucket
-            | TokenKind::Proof
-            | TokenKind::NonFungibleId
-            | TokenKind::NonFungibleAddress
-            | TokenKind::Expression
-            | TokenKind::Blob => self.parse_scrypto_types(),
+            /* Global address */
+            TokenKind:: PackageAddress |
```
dhedey, line 286: There's inconsistent spacing after the :: here, I assume not intentional? :)

dhedey (approved):
Nice changes!! 👍 

I guess at some point we might want to extract a Scrypto Value AST possibly distinct from the Manifest AST -- which Omar can re-use etc

transaction/src/manifest/parser.rs
iamyulong: For some reason, the formatter doesn't work for this blob of code. Will make them consistent.

PR #564, opened 2022-10-26
Use self-hosted runners for the linux jobs

muzuke:


.github/workflows/ci.yml
```diff
@@ -1,19 +1,38 @@
...
 
 jobs:
+  cancel_running_workflows:
+    name: Cancel running workflows
+    runs-on: ubuntu-20.04
+    steps:
+      - name: cancel running workflows
+        uses: styfle/cancel-workflow-action@0.8.0
+        with:
+          access_token: ${{ github.token }}
```
iamyulong, line 22: Can someone explain to me how running cancellation works? Is it per PR or global?

.github/workflows/ci.yml
```diff
@@ -1,19 +1,38 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
iamyulong, line 9: Can we not limit to `main` branch only? We have multiple active branches (main, develop, release/*, docs/*, alphanet, etc.) and personally I also use CI for feature branch.

iamyulong (commented):
Also, would be nice to see the time comparison between self-hosted linux and github equiv. 

The current CI bottleneck is for windows, but if this self-hosted linux proves to be faster, we can also benefit from this PR by seeing results earlier for one platform.

iamyulong:
> The bottleneck on the workflow is still the windows cli tests which take ~30 minutes. The linux jobs finish in less that 20 mins.
> Looking over at https://github.com/radixdlt/radixdlt-scrypto/actions?query=is%3Asuccess we can see that this PR run at 33mins while the previous jobs were mostly around the 40min-1h mark.

Nice! 50% reduction!

.github/workflows/ci.yml
iamyulong: Thanks for explaining. Per PR/branch is good.

.github/workflows/ci.yml
iamyulong: Cool. Let's include all the listed branches except feature branch.  I can create a draft PR for my feature branch to utilise the CI, if needed.

iamyulong: Still think it's not right.  Both push and PR to any of the above branch should trigger a build.

iamyulong (approved):
LGTM.

PR #565, opened 2022-10-26
Architecture: Remove consuming receiver

talekhinezh:


iamyulong (approved):
Nice move by replacing consumed receiver call with function call!

PR #566, opened 2022-10-27
Bump rust-cache version

iamyulong:
Switch back to the official `rust-cache` Github Action, since it now supports additional cache directories.


.github/workflows/ci.yml
```diff
@@ -159,14 +145,12 @@ jobs:
...
         os: [ubuntu-latest]
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-    - uses: bmwill/rust-cache@v1
+    - uses: actions/checkout@v3
+    - run: rustup toolchain install stable --profile minimal
+    - uses: Swatinem/rust-cache@master
```
dhedey, line 150: It's good practice to hardcode the version of github actions, to decrease the surface for supply chain attacks - would it be possible to do that in each copy of this here?

dhedey (approved):

.github/workflows/ci.yml
iamyulong: I'm actually waiting for the author to make a release. Never though we can hardcode a version. Will try.

PR #567, opened 2022-10-27
Trace sys call resource-like (buckets, proofs) inputs/outputs and match them against manifest instructions

LukasGasior1:
The goal is to be able to tell which proofs/resources have been consumed/produced by each manifest instruction.

The execution trace module uses a stack-based approach to match sys call outputs to their corresponding inputs.

The PR includes:
* support for kernel application events
* support for track output events


radix-engine/src/model/resource.rs
```diff
@@ -571,6 +571,31 @@ impl LockableResource {
     }
 }
 
+impl Into<Resource> for &LockableResource {
```
LukasGasior1, line 574: is there a way to reuse this (`&LockableResource`) impl for non-ref `LockableResource` (with an extra `is_locked` check), without unnecessary cloning?

dhedey:
I'm slightly concerned about the performance overhead of this. I was imagining the ExecutionTraceModule would just listen to the ScryptoValues passed in and returned, for buckets and proofs?

As for the knowing which instruction we're on, I imagine this would be some kind of more generic event, which the Module can listen to?

radix-engine/src/engine/heap.rs
```diff
@@ -123,6 +123,10 @@ impl Heap {
             .remove(&node_id)
             .ok_or(CallFrameError::RENodeNotOwned(node_id))
     }
+
+    pub fn nodes(&self) -> &HashMap<RENodeId, HeapRENode> {
+        &self.nodes
+    }
```
iamyulong, line 129: Instead of asking heap for global node list, we should check the owned the nodes by the current frame.

LukasGasior1: _(for the record: the tracing architecture has been changed, as discussed; this method no longer exists)_

radix-engine/src/engine/kernel.rs
```diff
@@ -1526,4 +1526,30 @@ where
 
         Ok(())
     }
+
+    fn pre_execute_instruction(&mut self, instruction: &Instruction) -> Result<(), RuntimeError> {
```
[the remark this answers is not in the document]
LukasGasior1: _(for the record: the tracing architecture has been changed, as discussed; this method no longer exists)_

radix-engine/src/engine/interpreters/native_interpreter.rs
```diff
@@ -188,7 +189,7 @@ pub trait NativeInvocation: NativeExecutable + Encode + Debug {
 }
 
 pub trait NativeExecutable: Invocation {
-    type Output: Debug;
+    type Output: Traceable + 'static;
```
LukasGasior1, line 192: not sure if adding 'static is okay here?

radix-engine/src/engine/modules/execution_trace.rs
```diff
@@ -25,30 +27,225 @@ pub enum VaultOp {
...
+    }
+
+    fn proofs(&self) -> Vec<ProofId> {
+        vec![]
+    }
+}
+
+impl Traceable for CallFrameUpdate {
+    fn buckets(&self) -> Vec<BucketId> {
+        self.nodes_to_move
```
LukasGasior1, line 168: not sure if we also want to trace the refs (`node_refs_to_copy`) here? :thinking:

radix-engine/src/engine/kernel.rs
```diff
@@ -717,6 +717,14 @@ where
 
         Ok(output)
     }
+
+    pub fn finalize(&mut self) -> Result<(), RuntimeError> {
```
talekhinezh, line 721: Can we integrate this with `track` finalize and also make this function be a consuming `self` one?

radix-engine/src/engine/modules/module.rs
```diff
@@ -4,9 +4,15 @@ use crate::fee::FeeReserve;
 use crate::model::Resource;
 use crate::types::*;
 
+#[derive(Debug)]
+pub enum InvocationInfo<'a> {
```
talekhinezh, line 8: Would rather not use enums here. Can we use something part of the trait?

radix-engine/src/engine/modules/module.rs
```diff
@@ -49,8 +55,10 @@ pub enum SysCallInput<'a> {
     },
 }
 
+#[derive(Debug)]
 pub enum SysCallOutput<'a> {
-    Invoke { rtn: String },
+    Invoke { rtn: &'a dyn Traceable },
+    ReadOwnedNodes,
```
talekhinezh, line 61: Is this being used anywhere? If not, would rather not add this, as most likely the `read_visible_nodes` syscall will get removed in the near future.

LukasGasior1: I just added it to match the `pre_sys_call` call, happy to remove both in this PR; neither is used (other than for logger)

radix-engine/src/engine/kernel.rs
LukasGasior1: Done. I've moved track ownership to kernel and `Kernel.finalize` now outputs a TrackReceipt. I could figure out how to make the modules `on_finished_processing` consume `self` as well, not sure if that's even possible because of them being `Box<dyn >>` with statically unknown size.

radix-engine/src/model/resource.rs
LukasGasior1: _(removed this `Into<Resource>` impl)_

radix-engine/src/engine/kernel.rs
```diff
@@ -717,6 +717,28 @@ where
...
+    pub fn finalize(mut self, result: InvokeResult) -> TrackReceipt {
+        let final_result = match result {
+            Ok(res) => self.finalize_modules().map(|_| res),
+            Err(err) => {
+                // If there was an error, we still try to finalize the modules,
+                // but forward the original error (even if module finalizer also errors).
+                let _silently_ignored = self.finalize_modules();
+                Err(err)
+            }
+        };
```
iamyulong, line 730: This section of code is a bit hard to read. Suggest:
```rust
let final_result = self.finalize_modules().and_then(|_| result);
self.track.finalize(final_result)
```

simulator/src/resim/cmd_set_current_epoch.rs
```diff
@@ -46,21 +46,22 @@ impl SetCurrentEpoch {
...
         // Invoke the system
-        kernel
+        let output = kernel
             .invoke(EpochManagerSetEpochInvocation {
                 epoch: self.epoch,
                 receiver: EPOCH_MANAGER,
             })
             .map_err(Error::TransactionExecutionError)?;
 
+        let receipt = kernel.finalize(Ok(Vec::new()));
```
iamyulong, line 62: Shoudn't we pass the output to `finalize`?

scrypto-unit/src/test_runner.rs
```diff
@@ -772,17 +773,18 @@ impl<'s, S: ReadableSubstateStore + WriteableSubstateStore> TestRunner<'s, S> {
...
-            &mut track,
+            track,
             &mut self.scrypto_interpreter,
             Vec::new(),
         );
 
         // Invoke the system
         let output = fun(&mut kernel);
 
+        let receipt = kernel.finalize(Ok(Vec::new()));
```
iamyulong, line 784: Same

simulator/src/resim/cmd_set_current_epoch.rs
LukasGasior1: not sure :sweat_smile: I thought it was done like that on purpose

scrypto-unit/src/test_runner.rs
LukasGasior1: _(same answer)_

dhedey (commented):
Looks good! Some small thoughts, but I like this much more than the first version.

radix-engine/src/engine/event.rs
```diff
@@ -0,0 +1,13 @@
...
+use sbor::{Decode, Encode, TypeId};
+use transaction::model::Instruction;
+
+pub enum ApplicationEvent<'a> {
+    PreExecuteInstruction { instruction: &'a Instruction },
+    PostExecuteInstruction { instruction: &'a Instruction },
+}
+
+#[derive(Debug, Clone, TypeId, Encode, Decode)]
+pub enum OutputEvent {
```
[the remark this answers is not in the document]
dhedey: > Can we name it Event? Whether an event should be outputed or not is a node decision.

I don't think that's necessarily true. EG certain modules might wish to output events directly to track which weren't application events.

Having a distinction between RuntimeEvent and OutputEvents I think is reasonable?

I also approve of `TrackEvent` or `TrackedEvent`

radix-engine/src/model/nodes/transaction_processor.rs
```diff
@@ -250,6 +250,12 @@ impl TransactionProcessor {
         let auth_zone_id: AuthZoneId = auth_zone_ref.into();
 
         for inst in input.instructions.as_ref() {
+            system_api
+                .emit_application_event(ApplicationEvent::PreExecuteInstruction {
+                    instruction: &inst,
```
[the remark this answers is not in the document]
LukasGasior1: Good idea :+1:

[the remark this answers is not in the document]
LukasGasior1: (Also a note to self: there's a corner case when handling any sys calls that happen after all instructions have been processed. Their associated index should be None, instead of the latest idx)

radix-engine/tests/fuzz_transactions.rs
```diff
@@ -43,6 +43,7 @@ fn execute_single_transaction(transaction: NotarizedTransaction) {
     let execution_config = ExecutionConfig {
```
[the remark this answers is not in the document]
LukasGasior1: :+1:

simulator/src/resim/cmd_set_current_epoch.rs
LukasGasior1: ah, that's because the invoke output is now typed and `finalize` specifically requires the output of `TransactionProcessorRunInvocation`

radix-engine/src/engine/kernel.rs
```diff
@@ -1205,24 +1235,23 @@ where
             .map_err(RuntimeError::ModuleError)?;
         }
 
-        let substate_ref =
-            self.current_frame
-                .get_ref(lock_handle, &mut self.heap, &mut self.track)?;
-
+        // A little hacky: this post sys call is called before the sys call happens.
```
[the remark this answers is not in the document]
LukasGasior1: ah, that's the TODO below; it should actually be after the sys call, but can't be due to a mutable borrow for substate_ref.
I'll explain that better in the comment

radix-engine/src/engine/modules/execution_trace.rs
```diff
@@ -25,30 +27,225 @@ pub enum VaultOp {
...
+    ScryptoMethod(ScryptoMethodIdent),
+    NativeFunction(NativeFunction),
+    NativeMethod(NativeMethod),
+    CreateNode,
+    DropNode,
+    /// Anything else that isn't traced on its own, but the trace exists for its children
+    Opaque,
+}
+
+pub trait Traceable: Debug {
```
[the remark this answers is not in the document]
LukasGasior1: The word `Resource` kept tempting me for various naming choices in this PR (eg SysCallResourcesTrace) :smile:  But it doesn't really fit as technically Proof is not a Resource. Also possibly we'll want to trace more information at some point.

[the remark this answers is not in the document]
dhedey: Yeah, `Traceable` is growing on me :D

simulator/src/resim/cmd_set_current_epoch.rs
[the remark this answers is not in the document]
LukasGasior1: The reason for that is that `finalize` specifically requires the (typed) output of `TransactionProcessorRunInvocation` (that is `Vec<Vec<u8>>`), which isn't the case here (`EpochManagerSetEpochInvocation` outputs unit). Similarly in test_runner.rs.
I don't think it really matters in those cases, but might be worth adding a comment.

radix-engine/src/model/nodes/transaction_processor.rs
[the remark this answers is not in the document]
LukasGasior1: _(addressed the corner case too)_

simulator/src/resim/cmd_set_current_epoch.rs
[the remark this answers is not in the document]
dhedey: Makes sense :D.

Tbh this `Vec<Vec<u8>>` has never really sat that well with me - it seems a little arbitrary that we're keeping the instruction return values in this way, and that it's not just a part of (eg) tracing or something...

Not something to change now, more of a reflection.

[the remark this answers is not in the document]
LukasGasior1: Added a comment

scrypto-unit/src/test_runner.rs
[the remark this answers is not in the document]
LukasGasior1: Added a comment

radix-engine/src/engine/kernel.rs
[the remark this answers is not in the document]
LukasGasior1: Added a comment

radix-engine/src/engine/kernel.rs
LukasGasior1: I've reverted the change you suggested. It doesn't really work in the same way:
if both `result` and `finalize_modules` are `Err`, we want to return the earlier one, the one from `result` (and ignore the latter). While your snippet returns the latter Err (the result of `finalize_modules`) and ignores the one from `result`.

iamyulong: Sounds good!

radix-engine/src/engine/kernel.rs
```diff
@@ -1374,6 +1407,14 @@ where
 
         Ok(())
     }
+
+    fn emit_application_event(&mut self, event: ApplicationEvent) -> Result<(), RuntimeError> {
+        for m in &mut self.modules {
+            m.on_application_event(&self.current_frame, &mut self.heap, &mut self.track, &event)
+                .map_err(RuntimeError::ModuleError)?;
+        }
+        Ok(())
```
iamyulong, line 1416: As discussed over slack, the should be refactored to:
1. call back `pre_syscall` on modules
2. insert application events to `track.tracked_events`
3. call back `post_syscall` on modules.

Would also consider renaming to `emit_event` to be consistent with `emit_log`, but no blocker.

radix-engine/src/engine/modules/execution_trace.rs
```diff
@@ -94,11 +291,266 @@ impl<R: FeeReserve> Module<R> for ExecutionTraceModule {
...
+    }
+
+    fn read_proof(heap: &mut Heap, proof_id: &ProofId) -> Result<ProofSnapshot, ModuleError> {
+        let node_id = RENodeId::Proof(proof_id.clone());
+        let substate_ref = heap
+            .get_substate(node_id, &SubstateOffset::Proof(ProofOffset::Proof))
+            .map_err(|e| {
+                ModuleError::ExecutionTraceError(ExecutionTraceError::CallFrameError(e))
+            })?;
+        Ok(substate_ref.proof().into())
```
iamyulong, line 540: Ah, missed this `into`. Suggeset using associated method `.snapshot` instead of `into` trait.

Into is for
> A value-to-value conversion that consumes the input value. 

`.into()` gives false impression that the proof is consumed (but in fact it's the reference that gets consumed).

iamyulong (commented):
Overall looks good to me.

radix-engine/src/engine/modules/execution_trace.rs
LukasGasior1: Ahh, right. Fixed

dhedey (commented):
I'll leave approvals with Yulong/Josh. But looks good! The implementation feels really neat now

iamyulong (approved):
LGTM

PR #568, opened 2022-10-27
Add cargo workspace and lock file

dhedey:
The key changes in this PR are as follows:
* Add a Cargo.toml workspace - this makes IDEs behave a lot more nicely - and I get better discovery of dependent code in VS Code with Rust Analyzer; AND it makes CLion actually work for me
* Source control a Cargo.lock file - I don't know how we didn't have this source controlled already! This is important for mitigating broken/malicious dependency updates etc.
* I also cleared up the build and format scripts to pick up new packages automatically

Outstanding work:
* Consider what to do with the scrypto packages. I avoided putting these as part of the main workspace because Cargo didn't like each scrypto package defining its own profile... I suggest we possibly create a separate sub-workspace for each of the sets of scrypto packages we build - but thoughts welcome 👍 
* Similarly, we still need to source control Cargo.locks for the various scrypto packages... This is painful because it would mean un-git-ignoring files that everyone has on their computers, which will create developer pain. I thought I'd wait until the previous question has an answer before going through this pain.


iamyulong (approved):
LGTM. Thanks for making the changes!

PR #569, opened 2022-10-28
SBOR generics, refs and Cow support

dhedey:
This PR can give us increased performance with SBOR encoding 👍

By using the "Cow" (copy-on-write) smart pointer: https://doc.rust-lang.org/std/borrow/enum.Cow.html we can now have a typed struct which can _encode_ a reference or owned, and output an owned value ✨. I've then shown an example of how it can be used to reduce the cloning in the transaction processor.

To complete this magic, I had to extend the Encode/Decode/TypeId macros to work with generics.

I have also added a nicer error message for Describe when used with a type with generics. The SBOR schema doesn't currently support generics (for eg recreation of the code), and I didn't fancy adding that! But at least we get a clearer error message that generics aren't supported.


iamyulong (approved):
LGTM

PR #570, opened 2022-10-28
Add lock_fee_and_withdraw methods to account

talekhinezh:


transaction/src/builder/manifest_builder.rs
```diff
@@ -715,6 +715,56 @@ impl ManifestBuilder {
         .0
     }
 
+    pub fn lock_fee_and_withdraw(
+        &mut self,
+        amount: Decimal,
+        resource_address: ResourceAddress,
+        account: ComponentAddress,
```
dhedey, line 722: Your change is consistent with the rest, but it feels a little confusing to me that the account is last because it doesn't match how you'd write the manifest line...

What are your thoughts on refactoring these methods on the manifest builder to put the account first? (I imagine CLion might be able to make that easy?)

dhedey (approved):
LGTM! 👍

transaction/src/builder/manifest_builder.rs
iamyulong: Looks like some existing methods are already taking `account` parameter as the last argument.

But, I like the idea of making them consistent with manifest, so up to change them all?

iamyulong (approved):

PR #572, opened 2022-10-29
Feature/immutable transaction executor

dhedey:
_Base branch will change after its parent branch is merged_

**Breaking changes:**
* **Transaction Executor** - Is now methods instead of an Executor. Takes immutable references to substate store and to scrypto interpretor
* **Preview Executor** - Is now methods instead of an Executor. Takes immutable references to substate store and to scrypto interpretor
* **Scrypto Interpreter** - Now uses a size-limited, threadsafe cache. New constructors for WasmiEngine and WasmInterpreter allow for adjusting the size of the cache.


iamyulong (commented):
Looks like `radix_engine` with `no_std + no_wasmer` build has been broken and this PR breaks it further. Needs CI enforcement! 

1. `Arc` - There is a alloc version of it, we should include it into `sbor::rust` facade, so to support both `std` and `no_std`.
2. Moka - Doesn't seem to support `no_std`

iamyulong (approved):

dhedey:
Merging the PR. After chatting to Yulong, we agreed that changes / fixes to fix/allow the Engine to run with no_std can be separate...

The likely back-fix for this PR would be to:
* Make use of alloc's Arc; exported in sbor::rust
* Introduce a wrapper Cache trait for the Cache, which:
  * Compiles using moka behind the std feature
  * Compiles using an `Arc<Rwlock<HashMap>>` or something instead for no-std (or maybe even an `Rc`?)

PR #579, opened 2022-11-03
refactor: WasmEngine trait changes, ScryptoInterpreter is Sync

dhedey:
Trying to merge the changes into the node, I realised that ScryptoInterpreter still wasn't `Sync`... And it transpires that's because of the `PhantomData<I>` which captures the WasmEngine's `WasmModule` type. I realised that this could become an Associated Type - which is much nicer for all our type signatures, and also fixes the Sync problem.

[PhantomData<T>](https://doc.rust-lang.org/std/marker/struct.PhantomData.html) is intended to be used to pretend to the compiler that the type contains a type that behaves like `T` - and is intended for use on FFI, to eg capture lifetimes, or the fact something isn't Sync/Send. In contrast, I think our use of PhantomData is due to us not previously finding the associated type solution. And so does not relate to any want for ScryptoInterpreter to act as if it's carrying round a non-`Sync` WasmModule all the time... but please shout if I've missed something! :)

This PR also adds a test that `ScryptoInterpreter<DefaultWasmEngine>` is indeed `Sync` (and `Send`, although I don't care about that one particularly)

**Breaking Changes:**
* **Transaction Executor** - Anyone making their own can drop the WasmInstance from the generic type signature.

### DIFF
--- a/radix-engine/benches/radix_engine.rs
+++ b/radix-engine/benches/radix_engine.rs
@@ -21,7 +21,6 @@ fn bench_transfer(c: &mut Criterion) {
             InstructionCostRules::tiered(1, 5, 10, 5000),
             512,
         ),
-        phantom: PhantomData,
     };
 
     // Create a key pair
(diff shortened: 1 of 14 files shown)


talekhinezh (approved):

talekhinezh:
Nice!

PR #589, opened 2022-11-15
Refactor/avoid any codec

iamyulong:
- Remove any codec dependency in Scrypto std


scrypto/src/resource/non_fungible_id.rs
```diff
@@ -65,18 +63,38 @@ impl fmt::Display for ParseNonFungibleIdError {
 // binary
 //========
 
+fn validate_id(slice: &[u8]) -> Result<(), DecodeError> {
+    let mut decoder = ScryptoDecoder::new(slice);
```
dhedey, line 67: I think it'd be worth adding a comment explaining why we're manually doing this instead of using an any encoder, as someone naive might just add it back in.

dhedey (approved):
Nice stuff! This should hopefully cut our compile size for anything which is using NFId - 40KB less for the account component is pretty good going!

But hmm I wonder if we should ditch validation of the NonFungibleId entirely inside scrypto, and leave that to the Transaction/Engine layer?

PR #590, opened 2022-11-17
Refactor/various enhancements

iamyulong:
**No model change but interfaces and implementations:**
- Added `&self` to `encode_type_id` in prep for transparent `SborValue`
- Renamed `encode_value` to `encode_body`
- Refactored any codec (removing optional type id)
- Removed `check_type_id` from `Decode` traits
- Moved rust type codec to a module


sbor/src/codec/array.rs
```diff
@@ -0,0 +1,72 @@
...
+    #[inline]
+    fn encode_type_id(&self, encoder: &mut Encoder<X>) {
+        encoder.write_type_id(Self::type_id());
+    }
+    #[inline]
+    fn encode_body(&self, encoder: &mut Encoder<X>) {
+        encoder.write_type_id(T::type_id());
+        encoder.write_size(self.len());
+        if T::type_id() == SborTypeId::U8 || T::type_id() == SborTypeId::I8 {
+            let mut buf = Vec::<u8>::with_capacity(self.len());
```
mstrug-rdx, line 17: Maybe this could be done without additional allocation and copy using something like this:
let ptr = self.as_ptr().cast::<u8>();
unsafe {
    std::slice::from_raw_parts(ptr, self.len())
}

sbor/src/codec/collection.rs
```diff
@@ -0,0 +1,161 @@
...
+        if T::type_id() == SborTypeId::U8 || T::type_id() == SborTypeId::I8 {
+            let slice = decoder.read_slice(len)?; // length is checked here
+            let mut result = Vec::<T>::with_capacity(len);
+            unsafe {
+                copy(slice.as_ptr(), result.as_mut_ptr() as *mut u8, slice.len());
+                result.set_len(slice.len());
+            }
+            Ok(result)
+        } else {
+            let mut result = Vec::<T>::with_capacity(if len <= 1024 { len } else { 1024 });
```
mstrug-rdx, line 106: Wouldn't be better: `if let >= 1024 { len } else { 1024 }` ?

iamyulong: This is to avoid allocating too much memory than what's actually encoded, to mitigate fake size attacks like:
```
[EXTREMTLY_LARGE_SIZE] [NO_ENTRIES_AFTERWARDS]
```

sbor/src/codec/array.rs
iamyulong: I think I ran into some memory issue before with this approach, but let me try it again. Thanks!

sbor/src/decode.rs
```diff
@@ -40,14 +29,17 @@ pub enum DecodeError {
...
-        Self::decode_value(decoder)
+        let type_id = decoder.read_type_id()?;
+        Self::decode_with_type_id(decoder, type_id)
     }
 
-    fn check_type_id(decoder: &mut Decoder<X>) -> Result<(), DecodeError>;
-
-    fn decode_value(decoder: &mut Decoder<X>) -> Result<Self, DecodeError>;
+    /// Decodes from the byte array encapsulated by the given decoder, with a preloaded type id.
+    fn decode_with_type_id(
```
dhedey, line 39: For symmetry, should this be called `decode_body_with_type_id`?

sbor/src/decode.rs
```diff
@@ -113,13 +105,24 @@ impl<'de, X: CustomTypeId> Decoder<'de, X> {
...
+        } else {
+            Err(DecodeError::UnexpectedTypeId {
+                actual: type_id.as_u8(),
+                expected: expected.as_u8(),
+            })
         }
+    }
 
-        Ok(())
+    pub fn check_type_id(&mut self, expected: SborTypeId<X>) -> Result<SborTypeId<X>, DecodeError> {
```
dhedey, line 123: I find this method as written slightly confusingly named - as it does a read as well as a check - unlike the method above?

Perhaps `read_and_check_type_id`? and the method below `read_and_check_size`?

dhedey (approved):
Nice! I'm liking the neatenings we can do in `value` for this. And I like separating the codecs by class 👍

sbor/src/decode.rs
iamyulong: Yes!

sbor/src/decode.rs
iamyulong: Yes!

sbor/src/codec/array.rs
iamyulong: Applied. Seems to be much better :)

PR #591, opened 2022-11-17
Use LEB128 for sizes

iamyulong:
Use LEB128 (max 4 bytes) for sizes to reduce payload size.


sbor/src/decode.rs
```diff
@@ -86,9 +88,21 @@ impl<'de, X: CustomTypeId> Decoder<'de, X> {
     }
 
     pub fn read_size(&mut self) -> Result<usize, DecodeError> {
-        let mut bytes = [0u8; 4];
-        bytes.copy_from_slice(self.read_slice(4)?);
-        Ok(u32::from_le_bytes(bytes) as usize)
+        // LEB128 and 4 bytes max
+        let mut size = 0usize;
```
dhedey, line 92: We possibly don't care, but this would silently overflow on 16-bit machines...

Is it worth some const assert that usize is at least u32 size here?

sbor/src/encode.rs
```diff
@@ -38,8 +38,19 @@ impl<'a, X: CustomTypeId> Encoder<'a, X> {
         self.write_slice(discriminator.as_bytes());
     }
 
-    pub fn write_size(&mut self, len: usize) {
-        self.buf.extend(&(len as u32).to_le_bytes());
+    pub fn write_size(&mut self, mut size: usize) {
+        // LEB128 and 4 bytes max
+        assert!(size <= 0x0FFFFFFF); // 268,435,455
```
dhedey, line 43: So if we're encoding bytes, this sets a max size of about 250MB. This feels... eminently reasonable? 👍 

That said, this should be an EncodeError, not an assert - I guess we can fix this in a future PR when we add the EncodeError?

radix-engine/tests/metering.rs
```diff
@@ -151,21 +151,21 @@ fn test_basic_transfer() {
         10000 /* base_fee */
         + 0 /* blobs */
         + 2000 /* create_node */
-        + 1212 /* decode_manifest */
+        + 978 /* decode_manifest */
```
dhedey, line 154: Yay! Already a 25% smaller transaction payload 👍

dhedey (approved):
Looks great, thanks

sbor/src/decode.rs
iamyulong: Come on, 16-bit machines.

sbor/src/encode.rs
iamyulong: Yes, was planning to do that in one batch.

PR #592, opened 2022-11-17
Architecture: Remove Traceable

talekhinezh:


iamyulong:
Quite like the direct exposion of the `CallFrameUpdate`, but will defer to @LukasGasior1 for a review.

radix-engine/src/engine/modules/execution_trace.rs
```diff
@@ -249,15 +137,127 @@ impl<R: FeeReserve> Module<R> for ExecutionTraceModule {
...
+                    Self::handle_vault_lock_fee(track, caller, &vault_id)
+                }
+                _ => {}
+            }
+        }
+
+        Ok(())
+    }
+
+    fn post_execute_invocation(
```
LukasGasior1, line 210: This method duplicates a lot of code from `handle_post_sys_call`. Wdyt about cleaning it up a bit and extracting common code into methods?

talekhinezh: Yup sounds good

talekhinezh: Seems like this may be harder than it looks without making code even more convoluted. I'd suggest we keep code as is and then add comments that we still need to find correct abstraction. How does that sound?

LukasGasior1: That sounds like a veiled way to say "let's push this problem to the next person changing this file" :D

LukasGasior1: I removed some duplication in my commit (please review), but it still isn't ideal.
It's quite tricky to follow the internal state (specifically `current_sys_call_depth` and the inputs stack) - it's now being updated in two distinct places. I understand this is not a priority though so I'm okay to keep it as is for now, and perhaps add some explanatory comments to make that easier to grasp for our future selves.

talekhinezh: Yes :) My point is that extracting common code based on just "pattern matching" (rather than correct abstraction) may actually lead to the wrong abstractions and make later refactors more difficult than necessary.

So imo better to have "messy, duplicated code with comments" rather than "de-duped but incorrectly abstracted code".

LukasGasior1 (approved):

PR #593, opened 2022-11-18
Architecture: Separate access rules substate

talekhinezh:


radix-engine-interface/src/engine/wasm_input.rs
```diff
@@ -57,7 +57,7 @@ pub enum NativeFunctionInvocation {
 #[derive(Debug)]
 #[scrypto(TypeId, Encode, Decode)]
 pub enum ComponentMethodInvocation {
-    AddAccessCheck(ComponentAddAccessCheckInvocation),
+    AddAccessCheck(AccessRulesAddAccessCheckInvocation),
```
iamyulong: Why do we still need `ComponentMethodInvocation`, given `add_access_check` is universal?

iamyulong (commented):
LGTM

radix-engine-interface/src/engine/wasm_input.rs
talekhinezh: Good catch

PR #595, opened 2022-11-18
Feature/engine no std

iamyulong:
- Restore engine `no_std` build

### DIFF
--- a/radix-engine/src/engine/interpreters/scrypto_interpreter.rs
+++ b/radix-engine/src/engine/interpreters/scrypto_interpreter.rs
@@ -116,7 +116,8 @@ mod tests {
         // RFC 2056
         fn assert_all() {
             assert_send::<ScryptoInterpreter<DefaultWasmEngine>>();
-            assert_sync::<ScryptoInterpreter<DefaultWasmEngine>>();
+            // TODO: make sure engine is multi-thread safe!
+            // assert_sync::<ScryptoInterpreter<DefaultWasmEngine>>();
         }
     };
 }
--- a/radix-engine/src/wasm/wasm_instrumenter.rs
+++ b/radix-engine/src/wasm/wasm_instrumenter.rs
@@ -1,16 +1,18 @@
-use moka::sync::Cache;
 use radix_engine_interface::crypto::hash;
-use std::sync::Arc;
+use sbor::rust::cell::RefCell;
+use sbor::rust::collections::HashMap;
+use sbor::rust::sync::Arc;
 
 use crate::types::*;
 use crate::wasm::{WasmMeteringConfig, WasmModule};
 
 pub struct WasmInstrumenter {
-    cache: Cache<(Hash, Hash), Arc<Vec<u8>>>,
+    cache: RefCell<HashMap<(Hash, Hash), Arc<Vec<u8>>>>,
 }
 
 #[derive(Debug, Clone)]
 pub struct InstrumenterOptions {
+    #[allow(dead_code)]
     max_cache_size_bytes: u64,
 }
 
@@ -28,17 +30,9 @@ pub struct InstrumentedCode {
 }
 
 impl WasmInstrumenter {
-    pub fn new(options: InstrumenterOptions) -> Self {
-        let cache = Cache::builder()
-            .weigher(|_key: &(Hash, Hash), value: &Arc<Vec<u8>>| -> u32 {
-                value
-                    .len()
-                    .checked_add(Hash::LENGTH * 2)
-                    .and_then(|total| total.try_into().ok())
-                    .unwrap_or(u32::MAX)
-            })
-            .max_capacity(options.max_cache_size_bytes)
-            .build();
+    pub fn new(_options: InstrumenterOptions) -> Self {
+        // TODO: limit size
+        let cache = RefCell::new(HashMap::new());
 
         Self { cache }
     }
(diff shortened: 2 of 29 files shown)


radix-engine/src/wasm/wasm_instrumenter.rs
```diff
@@ -28,17 +30,9 @@ pub struct InstrumentedCode {
...
-                    .len()
-                    .checked_add(Hash::LENGTH * 2)
-                    .and_then(|total| total.try_into().ok())
-                    .unwrap_or(u32::MAX)
-            })
-            .max_capacity(options.max_cache_size_bytes)
-            .build();
+    pub fn new(_options: InstrumenterOptions) -> Self {
+        // TODO: limit size
+        let cache = RefCell::new(HashMap::new());
```
dhedey, line 35: This feels like a definite reversion of behaviour. We want a safe and performant cache implementation here, for use in the node.

Can we instead just fall back to a RefCell / HashMap in the no-std case?

radix-engine/src/wasm/wasm_instrumenter.rs
```diff
@@ -1,16 +1,18 @@
...
+use sbor::rust::cell::RefCell;
+use sbor::rust::collections::HashMap;
+use sbor::rust::sync::Arc;
 
 use crate::types::*;
 use crate::wasm::{WasmMeteringConfig, WasmModule};
 
 pub struct WasmInstrumenter {
-    cache: Cache<(Hash, Hash), Arc<Vec<u8>>>,
+    cache: RefCell<HashMap<(Hash, Hash), Arc<Vec<u8>>>>,
```
dhedey, line 10: I believe this will panic if multiple threads try to lock it at the same time - as can happen in the Node when we are accepting multiple mempool submissions at once.

Can we wrap it in a Mutex instead of a RefCell?

radix-engine/src/wasm/wasm_instrumenter.rs
dhedey: Better yet, what about creating our own `ThreadsafeCache` type, and swap out the implementation for either wrapping Moka or a Mutex<HashMap<K, Arc<X>>> depending on if we use `alloc` or not?

dhedey (changes requested):
I'm not sure if this is ready for review yet, but it's not a draft PR, so as I noticed it, I've taken a look.

I approve of the want to restore a no-std build for the engine, but I worry about a few of the details here. Happy to chat if you'd like :)

radix-engine/src/engine/interpreters/scrypto_interpreter.rs
```diff
@@ -116,7 +116,8 @@ mod tests {
         // RFC 2056
         fn assert_all() {
             assert_send::<ScryptoInterpreter<DefaultWasmEngine>>();
-            assert_sync::<ScryptoInterpreter<DefaultWasmEngine>>();
+            // TODO: make sure engine is multi-thread safe!
```
dhedey, line 119: We create a new Radix engine for each transaction and single thread it - it's the ScryptoInterpreter caches we're sharing.

#FearlessConcurrency should mean this is fine? 👍

iamyulong: As the runtime shares pointer to the cache, they may run into illegal state due to memory volatility. Like I said, we need dedicated story to make sure multi-thread work rather pretend to.

radix-engine/src/wasm/wasm_instrumenter.rs
iamyulong: `Mutex` is not part of the `no_std`.

iamyulong:
It's now ready for another review. @dhedey

dhedey (approved):
Nice, thanks! 👍 

Happy for you to merge as-is or to extract out a separate Cache type.

EDIT: Looks like you still need to fix some failing tests: https://github.com/radixdlt/radixdlt-scrypto/actions/runs/3515933512/jobs/5891823809

radix-engine/src/engine/interpreters/scrypto_interpreter.rs
dhedey: I mean I've already checked this:

* The moka cache is thread safe
* The multiple copies of the engine share the WasmModule which is just data, and from which they instantiate separate / independent WasmInstances

Separately, the only other interaction between engine threads is in the database layer - where RocksDB is threadsafe. (and we ensure there are no commits during executing)

Is there anything I've missed?

radix-engine/src/engine/interpreters/scrypto_interpreter.rs
```diff
@@ -116,6 +116,8 @@ mod tests {
         // RFC 2056
         fn assert_all() {
             assert_send::<ScryptoInterpreter<DefaultWasmEngine>>();
+            // TODO: make sure engine is indeed multi-thread safe!
```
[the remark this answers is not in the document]
dhedey: See my [previous comment](https://github.com/radixdlt/radixdlt-scrypto/pull/595#discussion_r1028570742).

Do you have specific worries or things you want to check? Else I think this code comment should probably be removed.

radix-engine/src/engine/interpreters/scrypto_interpreter.rs
iamyulong: Mainly the `wasmer` implementation which stores a pointer which could be volatible.

dhedey: But that's the `WasmerInstance` which contains the pointer, not the `WasmerModule`. The `WasmerInstance` is not shared - it's created for each execution from the (shared) `WasmerModule` template.

dhedey: Indeed - Wasmer is written in Rust, and they have [good understanding of Send/Sync (It appears even Instance implements send!)](https://github.com/wasmerio/wasmer/pull/807), so we can trust that if Module is Sync then it is Sync. Indeed - I think it's essentially a data object :).

talekhinezh: You don't know what Rust is doing with the data under that pointer though since Rust doesn't know that it's a mutable object potentially being written to across multiple threads.

dhedey: I believe you're referring to the pointer that we create ourselves. This pointer isn't in the object which is being shared. This pointer is in the `WasmerInstance` (not shared) not the `WasmerModule` (shared).

This pointer in `WasmerInstance` has no affect on whether `WasmerModule` is `Sync`.

On a related note, I'm quite surprised that we use `    runtime_ptr: Arc<Mutex<usize>>,` and use `usize`, not a raw pointer type here, but anyway... I agree WasmerInstance possibly has an unsound impl of `Sync` (without looking into it further)... But feeling slightly like a broken record, we're not sharing `WasmerInstance`. And I've double-checked - the `WasmerInstance` pointer does not point into `WasmerModule` (which would be unsound for sure if it did). :)

PR #598, opened 2022-11-20
Architecture: Cleanup invocation modules

talekhinezh:


radix-engine/src/engine/modules/module.rs
```diff
@@ -83,58 +83,72 @@ pub enum SysCallOutput<'a> {
 pub trait Module<R: FeeReserve> {
     fn pre_sys_call(
         &mut self,
-        call_frame: &CallFrame,
-        heap: &mut Heap,
-        track: &mut Track<R>,
-        input: SysCallInput,
-    ) -> Result<(), ModuleError>;
+        _call_frame: &CallFrame,
```
iamyulong, line 86: This is very unconventional. Why do we need to add `_` to parameters in traits?

radix-engine/src/engine/kernel.rs
```diff
@@ -529,72 +428,78 @@ where
...
-        for node_id in &call_frame_update.nodes_to_move {
-            self.prepare_move_downstream(*node_id, &actor)?;
+        // New Call Frame pre-processing
+        {
+            // TODO: Abstract these away
+            self.execute_in_mode(ExecutionMode::AuthModule, |system_api| {
+                AuthModule::on_call_frame_enter(&mut call_frame_update, &actor, system_api)
+            })?;
+            self.execute_in_mode(ExecutionMode::NodeMoveModule, |system_api| {
+                NodeMoveModule::on_call_frame_enter(&mut call_frame_update, &actor, system_api)
```
iamyulong, line 443: Good time to merge all these `on_xxx` events into `Module` trait?

radix-engine/src/engine/modules/module.rs
talekhinezh: Added default implementations, so `_` used to show that these parameters aren't used in the default implementation.

radix-engine/src/engine/kernel.rs
talekhinezh: It's not quite that simple yet...

There are two issues:
1. `dyn Module` trait object cannot have generics on method calls
2. Modules list is being used for both callee and system api so we need to come up with the right abstraction here

But yes, in the long term will be looking to have these merged :)

radix-engine/src/engine/modules/module.rs
iamyulong: You mean IDE generated default implementations? Should we just change the implementation if it's not using the arguments, rather than the traits? 

My concer is that it will disable unused varilable checks, and makes documentation "ugly".

radix-engine/src/engine/kernel.rs
iamyulong: Why not generics? Module is using `R: FeeReserve` generics already.

> Modules list is being used for both callee and system api

Sorry, could you elaborate on this?

radix-engine/src/engine/modules/module.rs
talekhinezh: Not IDE generated, but the Rust trait default implementation (you can see all of these returning `Ok(())`).

radix-engine/src/engine/kernel.rs
talekhinezh: The issue is when we have a `dyn` type object accept another generic in a method call. e.g. if we add the following to the `Module` trait:
```
pub trait Module<R: FeeReserve> {
...
fn on_execute<Y: SystemApi>(&mut self, system_api: Y);
...
}
```

You get the following error: ```note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>```

talekhinezh: > Modules list is being used for both callee and system api

By this I mean that the `system_api` uses the module list. And we want to pass the `system_api` as we iterate through the module list, so there's a circular dependency that needs to be cleaned up.

radix-engine/src/engine/modules/module.rs
iamyulong: I see.

radix-engine/src/engine/kernel.rs
iamyulong: Thanks for the explaination. We definitely need to figure out how to do it properly.

iamyulong (approved):

PR #599, opened 2022-11-20
Feature: Recallable Vaults

talekhinezh:


radix-engine/src/engine/call_frame.rs
```diff
@@ -45,6 +45,12 @@ pub enum RENodeLocation {
     Store,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RENodeVisibilityOrigin {
+    Normal,
+    IgnoredOwner,
```
iamyulong, line 51: Suggest renaming `IgnoredOwner` to `DirectAccess`

radix-engine/src/engine/modules/auth_module.rs
```diff
@@ -194,7 +197,25 @@ impl AuthModule {
                             system_api.lock_substate(node_id, offset, LockFlags::read_only())?;
                         let substate_ref = system_api.get_ref(handle)?;
                         let resource_manager = substate_ref.resource_manager();
-                        let auth = vec![resource_manager.get_vault_auth(*vault_fn).clone()];
+
+                        let auth = match visibility {
```
iamyulong, line 201: Seems bad abstraction to me. Auth should be based on method rather than visibility origin, although I can see the motivation for resuing the same methods. :)

radix-engine/src/model/transaction_processor/executables.rs
```diff
@@ -80,14 +80,30 @@ impl<'a> NativeInvocation for TransactionProcessorRunInvocation<'a> {
...
+                Instruction::CallNativeMethod { args, method_ident } => {
+                    let scrypto_value =
+                        IndexedScryptoValue::from_slice(&args).expect("Invalid CALL arguments");
+                    for global_address in scrypto_value.global_references() {
+                        node_refs_to_copy.insert(RENodeId::Global(global_address));
+                    }
+
+                    // TODO: This needs to be cleaned up
+                    // TODO: How does this relate to newly created vaults in the transaction frame?
+                    // TODO: Will probably want different spacing for refed vs. owned nodes
```
iamyulong, line 99: I thought we agreed to limit receiver type to `&mut self` only. Owned objects need another layer of indirection for addressing, like buckets and proofs, subject to product agreement.

radix-engine/tests/authorization_resource.rs
```diff
@@ -92,6 +97,29 @@ fn test_resource_auth(action: Action, update_auth: bool, use_other_auth: bool, e
...
             ),
+        Action::Recall => {
+            let vaults = test_runner.get_component_vaults(account, token_address);
+            let vault_id = vaults[0];
+
+            builder
+                .add_instruction(Instruction::CallNativeMethod {
+                    method_ident: NativeMethodIdent {
+                        receiver: RENodeId::Vault(vault_id),
+                        method_name: "take".to_string(),
```
iamyulong, line 108: Why is action `recall` and method `take`?

radix-engine/src/engine/modules/auth_module.rs
talekhinezh: Agreed it's a bit weird but struggled to find a better abstraction. Suggest to keep it simple like this for now but add a comment to continue looking for better abstraction?

radix-engine/src/model/transaction_processor/executables.rs
talekhinezh: it is `&mut self` only. I'm talking about refed vs. owned in caller frame NOT whether the function is consuming or non-consuming.

radix-engine/tests/authorization_resource.rs
talekhinezh: `recall` action refers to "Direct Access" + "Take"

radix-engine/src/engine/modules/auth_module.rs
iamyulong: The current implementation is to make auth a function of **ref origin**, method, and access rules.

The alternative approach is to have dedicated methods for recall at the API. This introduces the question whether/how we want to limit "direct access" to a particular method, instead of any method. Yulong believed that we could rely on auth to do the right job, but ultimately it's a product question.

Will keep this topic in the background for future discussion.

iamyulong:
As discussed, we don't have enough evidence to decide what is the right abstraction for "trait methods". 

Yulong believed that all methods should be resolved to a particular, specific type and the mapping is beyond the engine's responsibility, like `impl AccessRules for T where T::RENode {}`, for simplicity.

Josh believed that these traits methods are special and are guaranteed to have single implementation, which is important for DeFi.

We both agreed that we should properly namespace these trait methods in distinction from normal methods.

And the reason for associating common data to RE nodes, rather than RE global nodes, is primarily due to product reasons.

iamyulong (approved):

PR #600, opened 2022-11-20
fix: Substate encode errors caught during execution

dhedey:
Builds on https://github.com/radixdlt/radixdlt-scrypto/pull/600

* Ensures that we can't return unencodable substates in the transaction result. Prior to this change, with an SBOR stack size of 32, I saw exceptions through at commit time(!) from some tests which happened to hit the sweet-spot where returning an ABI value from a component was okay, but when wrapped in a substate, it was too large to encode.
* Adds a number of tests to show the consequences of adding the sbor limit, and ensure that we don't panic in the engine (or outside of the engine!)

No breaking changes.

### DIFF
--- a/radix-engine/src/engine/track.rs
+++ b/radix-engine/src/engine/track.rs
@@ -792,3 +792,108 @@ impl<'s> FinalizingTrack<'s> {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use transaction::model::DEFAULT_COST_UNIT_LIMIT;
+
+    use crate::model::ComponentInfoSubstate;
+
+    use super::*;
+
+    #[test]
+    fn unencodable_substate_gives_failed_transaction() {
+        // Arrange
+        let store = TypedInMemorySubstateStore::default();
+
+        let successful_invoke_result = Ok(vec![]);
+
+        let finalizing_track_with_small_depth_substate = FinalizingTrack {
+            substate_store: &store,
+            new_global_addresses: vec![],
+            loaded_substates: create_loaded_substate_with_sbor_depth(10),
+            vault_ops: vec![],
+        };
+
+        let finalizing_track_with_large_depth_substate = FinalizingTrack {
+            substate_store: &store,
+            new_global_addresses: vec![],
+            loaded_substates: create_loaded_substate_with_sbor_depth(100),
+            vault_ops: vec![],
+        };
+
+        // Act
+        let small_substate_txn_result = finalizing_track_with_small_depth_substate
+            .calculate_commit_result(successful_invoke_result.clone(), &dummy_fee_summary());
+
+        small_substate_txn_result
+            .expect_commit()
+            .outcome
+            .expect_success();
+
+        let large_substate_txn_result = finalizing_track_with_large_depth_substate
+            .calculate_commit_result(successful_invoke_result.clone(), &dummy_fee_summary());
+
+        let error = large_substate_txn_result
+            .expect_commit()
+            .outcome
+            .expect_failure();
+        assert!(matches!(
+            error,
+            RuntimeError::KernelError(KernelError::InvalidSborValueOnEncode(
+                EncodeError::MaxDepthExceeded { .. }
+            ))
+        ))
+    }
+
+    fn create_loaded_substate_with_sbor_depth(
+        sbor_depth: u8,
+    ) -> BTreeMap<SubstateId, LoadedSubstate> {
+        let runtime_substate = RuntimeSubstate::ComponentInfo(ComponentInfoSubstate {
+            package_address: PackageAddress::Normal([0; 26]),
+            blueprint_name: "test".to_owned(),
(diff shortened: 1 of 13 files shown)


radix-engine/src/engine/track.rs
```diff
@@ -695,23 +673,86 @@ impl<'s> FinalizingTrack<'s> {
...
-                Err(error) => TransactionOutcome::Failure(error),
-            },
+            outcome,
             state_updates: Self::generate_diff(self.substate_store, to_persist),
             entity_changes: EntityChanges::new(new_global_addresses),
             resource_changes: execution_trace_receipt.resource_changes,
         })
     }
 
+    fn calculate_to_persist(
```
iamyulong, line 690: Not sure if this is the right abstraction. Feels like we should emit an error during execution, instead of finalization.

dhedey: I agree with you, and I considered this (although in the end went with the approach as written)...

An error during execution would look like:
* When a mut-lock is dropped, attempt to encode and emit an error if not 👍 
* This would result in mapping the loaded substates into persisted substates each time a mut lock is dropped, and would result in more work; but maybe we need to do this anyway?

What do you think?

There's a broader re-work of track which we need to do in any case

iamyulong: I'm thinking specifically at the problem - user input is too deep, which fails our runtime. 

Does it make more sense to add additional validation over the input? If the system is going to wrap user data into a struct and encode with `MAX_DEPTH = 36`.  We should probabaly limit the `max_depth` of untrusted any user input to something like `35`?

iamyulong: I don't think it relates to performance and agree we need to ensure it doesn't reach the gap.

What I've been striving for is a clearer definition of the contract between users and the system. This is more about system specification and implementation. I don't want our system to be "oh, we will try to encode your stuff after whatever transformation; and we don't know/care what your limit will be".

dhedey: Yes - that's a very good point which I support.

Perhaps:
* We use 64 depth limit for the runtime substate; checked at save time.
* We have a 70 depth limit for the persisted substates. We explicitly derive with an X = PersistedSubstateTypeId, so that it doesn't implement ScryptoEncode and so we have to use a PersistedSubstateEncode/Decode method.
* We have some tests/comments regarding the need to keep the depth differential between Runtime and Persisted to <= 6 --- and we assume this is true; so we don't need to test encode again at the end

iamyulong: > We use 64 depth limit for the runtime substate; checked at save time.

This assumes the runtime never persists values.

> We explicitly derive with an X = PersistedSubstateTypeId, so that it doesn't implement ScryptoEncode and so we have to use a PersistedSubstateEncode/Decode method.

I feel this is overkill. 

I'm leaning towards understanding what user inputs can be of variable depth. Off of the top of my head, they are
- Component state
- Call data
- NonFungibleData
- KeyValueEntry

Then, we enforce a lower limit like `48` for such inputs (or even all user inputs) and any `MaxDepthExceeded` at runtime should panic (illegal state).

iamyulong: Also, how do you feel about moving `MAX_DEPTH` type generic to an argument of the de-/encoder?

dhedey: Just to note here - following some good chats with Yulong, we've realised that for lots of different payload types there are limits we'd like to be able to clearly spell out to a user --- but also, even if they're embedded inside other payloads, we will still want to capture those limits.

There are quite a few things we need to do, to do this right:

* As well as whole payloads themselves, different (sub)payload types will have maximum depths - and potentially lengths (for their bodies)
* Ideally we can handle this at a Sub-payload level - that is, codecs can define their own limits, which form a kind of stack in the Encoder/Decoder

Some really rough potential depths:
```
            PayloadType::ENGINE_PERSISTED_SUBSTATE => 64,
            PayloadType::ENGINE_RUNTIME_SUBSTATE => 50,
            PayloadType::SCRYPTO_INTERFACE => 48,
            PayloadType::NON_FUNGIBLE_ID => 32,
            PayloadType::TRANSACTION_ARGUMENT => 48,
            PayloadType::TRANSACTION_ARGUMENTS_WRAPPER_STRUCT => 49, // 1 more than argument
            PayloadType::TRANSACTION_MANIFEST => 55,
            PayloadType::TRANSACTION_SIGNED => 60,
            PayloadType::TRANSACTION_NOTARIZED => 64,
            PayloadType::CUSTOM { max_depth } => *max_depth, // For use in other systems (eg the node)
```

This isn't something I plan to do as part of this PR, so I've added it to our SBOR tracking doc; and will reduce this PR to a small number of other tests, which we can actually hopefully merge for now.

dhedey:
Closing, as discussed in this comment: https://github.com/radixdlt/radixdlt-scrypto/pull/600#discussion_r1031889972 - to reform into separate PRs in due course

PR #601, opened 2022-11-20
feature: SBOR payloads start with a prefix byte

dhedey:
You need to know if you're encoding a payload or a payload-part. Typically this is obvious, the main caveat is where we pull together multiple sub-parts into a larger struct - notably Transaction Manifest args and Non Fungible Data.

In these cases, they're encoded as a complete struct, without a separate prefix byte for each argument/field - this felt like the only real correct approach.

I also had to do some hand-cranked WASM which was fun :).

The tests aren't running on this PR :( - but I've run them locally and they pass.

## Breaking changes
* **SBOR** - Basic payloads start with 0x5b (think [5b]or) and scrypto payloads start with 0x5c (think [5c]rypto).


iamyulong:
Unfortunatley, I have to register some dislike about this implementation.

Packing version byte into `CustomTypeId` doesn't seem correct from an abstraction perspective. Further, it prevents SBOR applicaion from not using version.

Suggest adding `scrypto_encode_with_version` and `scrypto_decode_with_version` variants and removing `encode/encode_payload` from SBOR encoder/decoder (logic can be relocated to the `scrypto_encode/decode*` functions).

dhedey:
> Unfortunatley, I have to register some dislike about this implementation.
> 
> Packing version byte into CustomTypeId doesn't seem correct from an abstraction perspective. Further, it prevents SBOR applicaion from not using version.
> 
> Suggest adding scrypto_encode_with_version and scrypto_decode_with_version variants.

@iamyulong

Understood. Just to understand:
* What's your use-case for encoding without a version? Shouldn't encoding with a version be the default / only option?
* Would you suggest moving the encoding of the byte into the scrypto_encode/scrypto_decode/basic_encode/basic_decode methods themselves?

iamyulong:
> What's your use-case for encoding without a version? Shouldn't encoding with a version be the default / only option?

One example is to be able to encode/decode part of a payload, like for nested values.

To me, the version byte is top-level concept that specifies the protocol, like `[SCRYPTO_SBOR_V1]`, `[SCRYPTO_SBOR_V2]`, `[SCRYPTO_JSON_V1]` (although less likely).

I'm imagining the engine is going to do something like the following to deal with a multi-version future.
```rust
fn scrypto_decode_any_version(buffer: &[u8]) -> Result<ValueEnum, ()> {
    let version = buffer[0];
    match version {
        0x5c => scrypto_decode(&bufer[1..]).into(),
        0x5d => scrypto_decode_v2(&bufer[1..]).into(),
        _ => Err(()),
    }
}
```

> Would you suggest moving the encoding of the byte into the scrypto_encode/scrypto_decode/basic_encode/basic_decode methods themselves?

Personally prefer composable, primitive operations, but won't go against a conveninence function that does all.

dhedey:
@iamyulong 

> One example is to be able to encode/decode part of a payload, like for nested values.

They can drop down and use the encoder directly then... But actually, looking in the code base, it's very rare to do this (it happens once, inside a macro) - I prefer the explicitness.

This way, any Vec<u8> not inside an encoder should be a full payload.

> To me, the version byte is top-level concept that specifies the protocol, like `[SCRYPTO_SBOR_V1]`, `[SCRYPTO_SBOR_V2]`, `[SCRYPTO_JSON_V1]` (although less likely).

Indeed - I was imagining different custom value versions would have different prefixes. But I'm more than happy to move the prefix up into the scrypto_encode(_v1) method instead - which could also capture changes to SBOR itself / the underlying VecEncoder etc :)

That said, I don't feel that we should have a `scrypto_encode_without_prefix` method - it kinda defeats the point of the prefix :)

iamyulong:
Happy with the proposed approach. Thanks for the inputs!

iamyulong (approved):

PR #602, opened 2022-11-21
SBOR Type Schema

dhedey:
This is now ready for a second review - hopefully can get this merged so I can start adding lots of smaller reviews instead.

Just to recap - the purpose of this task is to enable a more powerful SBOR schema which can:
* Work with recursive types
* Be smaller and more performant (and not arbitrarily deep)
* Be an attachment point for future data which needs to be attached to types (eg i18n; dynamic/contextual validation; ...)

For some examples of some types where Schemas can be constructed, see `sbor-tests/tests/schema.rs`

## Code Location
I tried out a few different locations for where schema lived, and realised how it was set up just naturally fitted the same way as the separation we have for Basic/Scrypto -- that means, putting the base implementation in SBOR. But of course, we don't want all of this code compiled into the actual build of packages -- only the "ABI creation" part - so this is where the "schema" feature flags come in.

## Naming
I've replaced all the naming of the types added in this PR with the names discussed with Yulong.

## Future work
* The Schema needs to be able to be validated to ensure it's self-consistent
* Engine validates against schema
* Package ABI uses schema
* JSON encoding with schema
* Type generation from schema (eg via `import!`?)
* Various attributes on the derive: EG Transparent (for wrapper structs)
* We swap out Describe<->Schema


sbor/src/codec/result.rs
```diff
@@ -51,3 +51,25 @@ impl<X: CustomTypeId, D: Decoder<X>, T: Decode<X, D>, E: Decode<X, D>> Decode<X,
...
+#[cfg(feature = "schema")]
+impl<C: CustomTypeKind<GlobalTypeId>, T: Describe<C>, E: Describe<C>> Describe<C> for Result<T, E> {
+    const TYPE_ID: GlobalTypeId = GlobalTypeId::novel("Result", &[T::TYPE_ID, E::TYPE_ID]);
+
+    fn type_data() -> Option<TypeData<C, GlobalTypeId>> {
+        #[allow(unused_imports)]
+        use crate::rust::borrow::ToOwned;
+        Some(TypeData::named_enum(
+            "Result",
+            crate::rust::collections::btree_map::btreemap![
```
iamyulong, line 64: We should probably use array to retain the original order.

sbor/src/schema/schema.rs
```diff
@@ -0,0 +1,41 @@
+use crate::rust::borrow::Cow;
+use crate::rust::vec::Vec;
+use crate::*;
+
+/// An array of custom type kinds, and associated extra information which can attach to the type kinds
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Schema<C: CustomTypeExtension> {
+    pub type_kinds:
+        Vec<TypeKind<C::CustomTypeId, C::CustomTypeKind<LocalTypeIndex>, LocalTypeIndex>>,
+    pub type_metadata: Vec<NovelTypeMetadata>,
```
iamyulong, line 10: Minor: use `Vec<(kind, metadata)>` to enforce matching length?

sbor/src/schema/type_link.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// This is a global identifier for a type.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, Encode, Decode, TypeId)]
+pub enum GlobalTypeId {
+    /// This takes a well_known type index.
+    ///
+    /// This wraps a `[u8; 1]` because it needs to be able to be turned into a `[u8]` in a const context.
+    WellKnown([u8; 1]),
+    /// The global type hash of a type - used for types which aren't well known.
+    Novel(TypeHash),
```
iamyulong, line 16: `Novel` is fun.

iamyulong (approved):
LGTM with minor comments.

sbor-derive/src/describe.rs
```diff
@@ -0,0 +1,387 @@
...
+                            // different types depending on the generic types providing. Adding in the generic types' TYPE_IDs solves that issue.
+                            //
+                            // It's still technically possible to get a collision (by abusing type namespacing to have two types with identical code
+                            // reference other types) but it's good enough - you're only shooting yourself in the food at that point.
+                            //
+                            // Note that it might seem possible to still hit issues with infinite recursion, if you pass a type as its own generic type parameter.
+                            // EG (via a type alias B = A<B>), but these types won't come up in practice because they require an infinite generic depth
+                            // which the compiler will throw out for other reasons.
+                            &[#(#generic_type_idents::TYPE_ID,)*],
+                            &#code_hash
```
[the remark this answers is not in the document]
dhedey: Yup, as mentioned on slack

sbor/src/codec/result.rs
dhedey: I originally used `vec` here... but then decided that at a typekind level, we don't actually care about order, and canonically forcing an ordering by variant key feels cleaner.

And I'm perfectly comfortable with losing original order at type reconstruction time. Type reconstruction for me is about creating DTOs rather than getting a perfect one-for-one mapping of code.

Are you happy with this call?

sbor/src/schema/type_link.rs
dhedey: Glad you like it. I originally had `Custom` but that was used to mean other things. So I did synonym lookup and liked the term `Novel` 👍

sbor/src/codec/result.rs
iamyulong: It's going to be tricky if we use `u8` as descriminator,  so we don't know how to encode the generated enum types.

dhedey: Discussed on call - we can keep using btreemap when we pivot to `u8`, as this allows for us to eg have a variant with discriminants 5 and 7, and not necessarily start from 0.

Issue #585, opened 2022-11-10
Decimal multiplication and division overflow for no reason 


arthurvinci:
The functions `Div` and `Mul` overflow for no reason. The following pieces of code return overflow errors:
```Rust
let test = Decimal::MAX*1;
```
```Rust
let test = Decimal::MAX/1;
```

For the Mul function, the error comes from the following piece of code in `decimal.rs`:
```Rust
let a = self.0;
let b_dec: Decimal = other.try_into().expect("Overflow");
let b: I256 = b_dec.0;
let c: I256 = a * b / Self::ONE.0;
```
If we call `1_b` the binary number representing `1` in `Decimal`
Then `a*b` can be in [2^256 + 1_b -1; 0]  but should still not throw an overflow. Therefore, the multiplication function will overflow for all `a`, `b` such that `a*b` is in [2^256 +1_b - 1; 2^256 -1].
Note that changing the line `c = a * b / Self::ONE.0;` to `c = a / Self::ONE.0 * b` creates an issue. If `a` is smaller than `1` then the multiplication will always return `0`. 

For the Division function, the error comes from the following piece of code in `decimal.rs`:
```Rust
let a = self.0;
let b_dec: Decimal = other.try_into().expect("Overflow");
let b: I256 = b_dec.0;
let c: I256 = a * Self::ONE.0 / b;
```
We have a similar issue.
`a*Self::ONE.0` can be in [2^256 + c -1; 0] but the final division `a/b` not overflow in theory.
Note: Changing the order of the operations in the last line also creates an issue. If `a<b` then `a/b`will always return 0.

arthurvinci:
Proposed solutions:
For multiplication:
```Rust
    fn mul(self, other: T) -> Self::Output {
        let a: I512 = I512::from(self.0);
        let b_dec: Decimal = other.try_into().expect("Overflow");
        let b: I512 = I512::from(b_dec.0);
        let c: I512 = a * b / I512::from(Self::ONE.0);
        let c_256 = I256::try_from(c).expect("Overflow");
        Decimal(c_256)
    }
```
For division:
```Rust
fn div(self, other: T) -> Self::Output {
        let a: I512 = I512::from(self.0);
        let b_dec: Decimal = other.try_into().expect("Overflow");
        let b: I512 = I512::from(b_dec.0);
        let c: I512 = a * I512::from(Self::ONE.0) / b;
        let c_256 = I256::try_from(c).expect("Overflow");
[...]

Issue #584, opened 2022-11-10
`try_from` function for `IXXX` does not work as expected


arthurvinci:
The `try_from` function for `IXXX` elements overflows without panicking.
The following code :
```Rust
let c_256 = I256::try_from(c).expect("Overflow");
``` 
overflows with no error :
```Rust
I512 number c : 57896044618658097769681537123002051638420484837164235946363784336776846839695
Max I256:       57896044618658097711785492504343953926634992332820282019728792003956564819967
c_256:         -57896044618658097653889447885685856214849499828476328093093799671136282800241
```

arthurvinci:
Solved by #604

PR #604, opened 2022-11-21
Fixes reported bugs on Decimal and PreciseDecimal

arthurvinci:
For issue: #584 the bug came from the following code from the macro `try_from_large_into_large` in `convert.rs` :
```Rust
   if (other.leading_zeros() as i32) < <$o>::BITS as i32 - <$t>::BITS as i32
```
Changing the strict inequality to a large inequality solves the issue.

For issue: #585 the bug was explained in the issue. To solve it, I implemented `I320`, `I728`, `U320` and `U728` integer types, closest multiples of 8 to (256 + log_2(10^18)) and (512 + log_2(10^64)).


radix-engine-interface/src/math/decimal.rs
```diff
@@ -307,11 +307,12 @@ where
...
 
     fn mul(self, other: T) -> Self::Output {
-        let a = self.0;
+        let a = I320::from(self.0);
         let b_dec: Decimal = other.try_into().expect("Overflow");
-        let b: I256 = b_dec.0;
-        let c: I256 = a * b / Self::ONE.0;
-        Decimal(c)
+        let b: I320 = I320::from(b_dec.0);
+        let c: I320 = a * b / I320::from(Self::ONE.0);
```
iamyulong, line 313: We should promote to `I512` instead of `I320`, otherwise `a * b` will overflow.

iamyulong: After more thoughts, `I320` should also be fine. If it were to overflow `a * b`, we won't be able to convert it back to `Decimal` anywa.

iamyulong (approved):

PR #606, opened 2022-11-23
Feature: Package owner badge + metadata

talekhinezh:
* Add Package owner badge which is returned when package is created
* Add Package metadata which may be updated by owner of owner badge
* Simplified native invocation architecture


radix-engine/src/engine/interpreters/native_interpreter.rs
```diff
@@ -76,77 +81,21 @@ impl Into<ApplicationError> for EpochManagerError {
...
-}
-
-pub trait NativeExecutable: Invocation {
-    type NativeOutput: Debug;
-
-    fn execute<Y>(
-        invocation: Self,
-        system_api: &mut Y,
-    ) -> Result<(<Self as Invocation>::Output, CallFrameUpdate), RuntimeError>
+pub trait NativeProgram {
```
iamyulong, line 84: Program sounds a bit heavy. Suggest:
```rust
NativeProcedure {
    fn run()
}
```

radix-engine/src/engine/modules/auth_module.rs
```diff
@@ -84,25 +85,76 @@ impl AuthModule {
...
                 match (method, resolved_receiver) {
+                    (ResolvedMethod::Native(NativeMethod::Metadata(MetadataMethod::Set)), ..) => {
+                        if let Some((
+                            RENodeId::Global(GlobalAddress::Package(package_address)),
+                            ..,
+                        )) = resolved_receiver.derefed_from
+                        {
+                            let bytes = scrypto_encode(&package_address).map_err(|_| {
+                                RuntimeError::ModuleError(ModuleError::AuthError(
+                                    AuthError::CouldNotEncodePackageAddress,
```
iamyulong, line 96: Seems like we should panic.

radix-engine/src/engine/native_wrapper.rs
```diff
@@ -47,13 +47,22 @@ where
...
             }
-            NativeFunction::Package(PackageFunction::Publish) => {
-                let invocation: PackagePublishInvocation = scrypto_decode(&args)
-                    .map_err(|e| RuntimeError::KernelError(KernelError::InvalidSborValue(e)))?;
-                system_api
-                    .sys_invoke(invocation)
-                    .map(|a| IndexedScryptoValue::from_typed(&a))
-            }
+            NativeFunction::Package(package_function) => match package_function {
+                PackageFunction::PublishNoOwner => {
```
iamyulong, line 51: Same

radix-engine-interface/src/wasm/input.rs
```diff
@@ -0,0 +1,394 @@
...
+pub enum NativeFnInvocation {
+    Method(NativeMethodInvocation),
+    Function(NativeFunctionInvocation),
+}
+
+#[derive(Debug)]
+#[scrypto(TypeId, Encode, Decode)]
+pub enum NativeMethodInvocation {
+    AccessRules(AccessRulesMethodInvocation),
+    Metadata(MetadataMethodInvocation),
```
iamyulong, line 42: I think, at some point, we will need to limit and isolate these "any" methods. My inclination is to make it:
```rust
pub enum NativeMethodInvocation {
    Global(GlobalMethodInvocation)
}
```

iamyulong (commented):
I started to worry that mixing generic substates (i.e. access rule and metadata) with type substatess will cause more pain in the long run. This is essentially unneeded flexibilty.

Would very much like to explore if we should limit them to just globals from a practical viewpoint.

radix-engine-interface/src/api/types/invocation/native.rs
```diff
@@ -322,7 +344,8 @@ pub enum WorktopMethod {
 #[scrypto(TypeId, Encode, Decode, Describe)]
 #[strum(serialize_all = "snake_case")]
 pub enum PackageFunction {
-    Publish,
+    PublishNoOwner,
+    PublishWithOwner,
```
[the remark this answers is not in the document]
talekhinezh: Will add a note here regarding NoOwner

radix-engine/src/engine/interpreters/native_interpreter.rs
talekhinezh: sure sounds good

radix-engine/src/engine/modules/auth_module.rs
talekhinezh: okay

iamyulong:
Discussed offline. 

Thoughts on trait methods have been documented on https://github.com/radixdlt/radixdlt-scrypto/pull/599.

Will keep the "NoOwner" variant for now, otherwise we will have to find another place to store the system entity owner badge.  Josh may challenge the decision of not having NoOwner again. :)

iamyulong (approved):

PR #607, opened 2022-11-24
Architecture: Cleanup NodeCreate syscalls

talekhinezh:


radix-engine-interface/src/api/types/re_node.rs
```diff
@@ -11,6 +11,28 @@ pub enum ScryptoRENode {
...
+pub enum RENodeType {
+    Bucket,
+    Proof,
+    AuthZoneStack,
+    Worktop,
+    GlobalAccount,
+    GlobalComponent,
+    GlobalResourceManager,
+    GlobalPackage,
+    GlobalEpochManager,
```
iamyulong, line 26: Feel like we should represent globals as `Global(EntityTypeId)`.

radix-engine/src/engine/system_api.rs
```diff
@@ -71,8 +71,13 @@ pub trait SystemApi {
...
     fn drop_node(&mut self, node_id: RENodeId) -> Result<HeapRENode, RuntimeError>;
 
+    /// Allocates a new node id useable for create_node
+    fn allocate_node_id(&mut self, node_type: RENodeType) -> Result<RENodeId, RuntimeError>;
+
     /// Creates a new RENode
-    fn create_node(&mut self, re_node: RENode) -> Result<RENodeId, RuntimeError>;
+    /// TODO: Remove, replace with lock_substate + get_ref_mut use
+    fn create_node(&mut self, node_id: RENodeId, re_node: RENode)
+        -> Result<RENodeId, RuntimeError>;
```
iamyulong, line 80: Understand you have a TODO there, but is there reason we return `RENodeId`?

talekhinezh: More to keep interface there once we implement local handles. Would you rather it just be removed?

radix-engine-interface/src/api/types/re_node.rs
talekhinezh: Not sure if that gives any benefit at the moment? I'd rather avoid complexity until we get to the point of refactoring this area.

iamyulong: More from representation accuracy perspective. None of the variants are really a `RENodeType`.

PR #610, opened 2022-11-25
Fee Reserve Update and Primitive Royalty Accuring

iamyulong:
Changes:
- Updated fee reserve to lazily convert XRD into cost units
- Added basic royalty accruing for packages and components


radix-engine/src/engine/interpreters/native_interpreter.rs
```diff
@@ -102,12 +102,13 @@ impl<N: NativeInvocation> Resolver<N> for NativeResolver {
                 // TODO: Move this logic into kernel
                 let resolved_receiver =
                     if let Some((derefed, derefed_lock)) = deref.deref(receiver)? {
+                        call_frame_update.node_refs_to_copy.insert(receiver);
+                        call_frame_update.node_refs_to_copy.insert(derefed);
```
talekhinezh, line 106: Why include this? I think this will possibly break abstractions and some invariants.

radix-engine-interface/src/api/api.rs
```diff
@@ -57,6 +57,8 @@ pub trait SysInvokableNative<E>:
     + SysNativeInvokable<AuthZoneClearInvocation, E>
     + SysNativeInvokable<AuthZoneDrainInvocation, E>
     + SysNativeInvokable<AccessRulesAddAccessCheckInvocation, E>
+    + SysNativeInvokable<ComponentSetRoyaltyConfigInvocation, E>
+    + SysNativeInvokable<PackageSetRoyaltyConfigInvocation, E>
```
talekhinezh, line 61: I think these two should be merged into a single invocation as these should be invocations related to royalties and not Component/Package.

radix-engine/src/engine/interpreters/native_interpreter.rs
iamyulong: Vaguely remember this was due to visibility loss, but will double check again.

radix-engine-interface/src/api/api.rs
talekhinezh: As discussed we can keep these two as separate invocations for now and reassess later as we gather more evidence/info on the right abstraction

radix-engine/src/engine/interpreters/native_interpreter.rs
talekhinezh: Yes, and if this is actually a requirement can you add a TODO comment to figure out how to remove that deref in the future

talekhinezh (approved):

PR #611, opened 2022-11-25
Feature: Added NonFungibleId type

mstrug-rdx:
### Summary

Added support for different ID types: String, Number, UUID and Bytes array to `NonFungibleId`.
Task ref ID: SCRY-286

### Detailed description

Added new enum type: `NonFungibleIdType` which represents different types of Non Fungible ID. `NonFungibleIdType` added as a new field to `NonFungibleId` struct. Internally `NonFungibleId` data has been left unchanged as `Vec<u8>`, but changed visibility to private and added appropriate functions to get/set values. Also optimised `from_bytes()` function. 
Implemented `Default` trait for `NonFungibleIdType` which returns `UUID` as default ID type (according to requirement), but this change requires some updates in tests which uses `NonFungibleId::from_u32()` function (`Number` type id) - this could be changed to `from_uuid()` to eliminate the need for setting id type on `ResourceBuilder` or change default id type to `Number`.

Changes in `NonFungibleResourceBuilder`: added new field `id_type` and function `set_id_type()`.
Changes in `ResourceType`: added new field `id_type` to `NonFungible` variant which is used to validate Non Fungible Ids types in Mint Params.
Changes in `Resource` and `LockableResource`: similarly added new field `id_type` to `NonFungible` variant which is used to validate id type in function `put()` and `lock_by_ids()`.

### Testing

Added new tests in `radix-engine-interface/src/model/resource/non_fungible_id.rs` file and in `radix-engine/tests/non_fungible.rs` (including `non_fungible` blueprint update).


radix-engine-interface/src/model/resource/non_fungible_id.rs
```diff
@@ -9,25 +9,88 @@ use sbor::*;
...
 use crate::scrypto_type;
+use crate::math::Decimal;
+use crate::Describe;
 
 /// Represents a key for a non-fungible resource
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct NonFungibleId(pub Vec<u8>);
+pub struct NonFungibleId{ 
+    id_type: NonFungibleIdType,
+    value: Vec<u8>
```
talekhinezh, line 19: Would it make more sense fo have `NonFungibleId` as an `enum` and the value being the correct type e.g.:
```
pub enum NonFungibleId {
  String(String),
  Bytes(Vec<u8>),
  ..
}
```

radix-engine-interface/src/model/resource/non_fungible_id.rs
```diff
@@ -8,26 +8,89 @@ use sbor::*;
...
 use crate::abi::*;
 use crate::data::*;
+use crate::math::Decimal;
 use crate::scrypto_type;
+use crate::Describe;
 
 /// Represents a key for a non-fungible resource
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct NonFungibleId(pub Vec<u8>);
+pub struct NonFungibleId {
```
iamyulong, line 17: As Josh mentioned, it's better to refactor `NonFungibleId` into an array, and do the byte conversion at encode/decode time.

radix-engine-interface/src/model/resource/non_fungible_id.rs
```diff
@@ -127,9 +212,26 @@ impl FromStr for NonFungibleId {
...
+impl fmt::Debug for NonFungibleIdType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self)
+    }
+}
+
 impl fmt::Display for NonFungibleId {
     fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
-        write!(f, "{}", hex::encode(&self.0))
+        write!(f, "{}", hex::encode(&self.value))
```
iamyulong, line 234: Given non-fungible IDs are strongly typed, I think it makes more sense to design a syntx, similar to manifest, instead of hex.

radix-engine-interface/src/model/resource/resource_type.rs
```diff
@@ -8,14 +9,21 @@ pub enum ResourceType {
...
 impl ResourceType {
     pub fn divisibility(&self) -> u8 {
         match self {
             ResourceType::Fungible { divisibility } => *divisibility,
-            ResourceType::NonFungible => 0,
+            ResourceType::NonFungible { .. } => 0,
+        }
+    }
+
+    pub fn id_type(&self) -> NonFungibleIdType {
```
iamyulong, line 23: Suggest full name `non_fungible_id_type`

radix-engine-interface/src/model/resource/non_fungible_id.rs
mstrug-rdx: Reimplemented to `enum`.

radix-engine-interface/src/model/resource/non_fungible_address.rs
```diff
@@ -137,11 +137,11 @@ impl FromPublicKey for NonFungibleAddress {
         match public_key {
             PublicKey::EcdsaSecp256k1(public_key) => NonFungibleAddress::new(
                 ECDSA_SECP256K1_TOKEN,
-                NonFungibleId::from_bytes(hash(public_key.to_vec()).lower_26_bytes().into()),
+                NonFungibleId::from_bytes(&hash(public_key.to_vec()).lower_26_bytes()),
```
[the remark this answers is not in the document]
mstrug-rdx: After reimplementation NonFungibleId as `enum` this is not relevant.

radix-engine-interface/src/model/resource/non_fungible_id.rs
iamyulong: 100% agree.

radix-engine-interface/src/model/resource/non_fungible_id.rs
```diff
@@ -8,26 +8,44 @@ use sbor::*;
...
-pub struct NonFungibleId(pub Vec<u8>);
-
-impl NonFungibleId {
-    /// Creates a non-fungible ID from an arbitrary byte array.
-    pub fn from_bytes(v: Vec<u8>) -> Self {
-        Self(scrypto_encode(&v).expect("Error encoding byte array"))
-    }
+pub enum NonFungibleId {
+    String(String),
+    NumberU32(u32),
```
iamyulong, line 19: Minor: I personally like `U32/U64`.

iamyulong (approved):

radix-engine-interface/src/model/resource/non_fungible_id.rs
mstrug-rdx: New task for this: SCRY-292

radix-engine-interface/src/model/resource/non_fungible_id.rs
mstrug-rdx: I've updated this.

PR #612, opened 2022-11-28
Implement Royalty Disbursements

iamyulong:


radix-engine-interface/src/api/types/re_node.rs
```diff
@@ -144,12 +144,14 @@ pub enum ComponentOffset {
...
     RoyaltyConfig,
+    RoyaltyAccumulator,
 }
 
 #[derive(Debug, Clone, TypeId, Encode, Decode, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub enum PackageOffset {
-    Package,
+    Info,
     RoyaltyConfig,
+    RoyaltyAccumulator,
```
talekhinezh, line 154: As per discussion, would like to see Royalty `Config` + `Accumulator` be a part of a unified interface/substate-offsets. But for the purpose of getting things integrated asap, this is good enough for now.

radix-engine/src/model/component/substates.rs
```diff
@@ -32,3 +33,9 @@ impl ComponentInfoSubstate {
 pub struct ComponentRoyaltyConfigSubstate {
     pub royalty_config: RoyaltyConfig,
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[scrypto(TypeId, Encode, Decode)]
+pub struct ComponentRoyaltyAccumulatorSubstate {
+    pub royalty: Resource, // TODO: wrap with a vault?
```
talekhinezh, line 40: Yes, definitely should wrap with vault

talekhinezh (approved):

radix-engine/src/model/component/substates.rs
iamyulong: Yeah, I can definitely see the point -  all the resource are stored in vaults. 

The only concern I had was - do you need the extra functionality provided by vaults? Do we accept the additional cost associated with an extra hop? Is there any problem allowing component accessing royalty the same way accessing vaults?

PR #615, opened 2022-11-28
Add system clock RENode

LukasGasior1:


radix-engine-interface/src/api/types/invocation/native.rs
```diff
@@ -304,6 +306,49 @@ pub enum WorktopMethod {
...
+    EnumVariantNames,
+    IntoStaticStr,
+    AsRefStr,
+    Display,
+)]
+#[scrypto(TypeId, Encode, Decode, Describe)]
+#[strum(serialize_all = "snake_case")]
+pub enum ClockMethod {
+    SetCurrentTime,
+    GetCurrentTimeToMinutePrecision,
```
LukasGasior1, line 349: might be better to rename it to `GetCurrentTimeInMinutes`, "to minute precision" doesn't necessarily mean that the unit is also minutes

radix-engine-interface/src/api/types/re_node.rs
```diff
@@ -194,6 +204,13 @@ pub enum WorktopOffset {
     Worktop,
 }
 
+#[derive(Debug, Clone, TypeId, Encode, Decode, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub enum ClockOffset {
+    CurrentTimeInMillis,
+    CurrentTimeInSeconds,
+    CurrentTimeInMinutes,
```
talekhinezh, line 211: I think we had decided in an earlier builders meeting to just do `Minutes` as I think having too many variations can make things confusing

LukasGasior1: Hmh, wasn't that about what we expose in scrypto, while still keeping more granular substates around (possibly for future use)? So far in this PR I've only exposed the time rounded to minutes

LukasGasior1:
Marking as "ready for review" as this piece is ready; Remaining stuff can go in a follow up PR (mostly adding more user friendly methods on the scrypto side).

radix-engine/src/model/global/substates.rs
```diff
@@ -8,8 +8,7 @@ pub enum GlobalAddressSubstate {
     Component(ComponentId),
     Resource(ResourceManagerId),
     Package(PackageId),
-    // We may consider using another enum here so we can map system addresses to nodes of different type.
-    System(EpochManagerId),
+    System(SystemId),
```
talekhinezh, line 11: Can we just separate `System` into `Clock` and `EpochManager`. I think may be simpler if these were more flattened out?

radix-engine-interface/src/api/types/re_node.rs
talekhinezh: I don't think that was the result of our convo back then. I'd definitely much rather have this simplified and just have a single substate for time, atleast as the first pass.

radix-engine/src/model/substates.rs
```diff
@@ -12,6 +12,9 @@ use radix_engine_interface::data::IndexedScryptoValue;
 pub enum PersistedSubstate {
     Global(GlobalAddressSubstate),
     EpochManager(EpochManagerSubstate),
+    CurrentTime(CurrentTimeSubstate),
+    CurrentTimeRoundedToSeconds(CurrentTimeRoundedToSecondsSubstate),
+    CurrentTimeRoundedToMinutes(CurrentTimeRoundedToMinutesSubstate),
```
talekhinezh, line 17: One option I think we may want to think about is whether we even want to persist time. It might be better to just be something that is inputted at the beginning of execution.

talekhinezh:
Looks good overall! Just want to the multiple substate thing fixed and I think good to go.

radix-engine/src/model/global/substates.rs
LukasGasior1: yup, sure

talekhinezh (approved):

PR #619, opened 2022-11-29
change: SBOR struct now uses Tuple type id

dhedey:
NB:
* We might not want to merge this before Betanet, I don't mind
* We could consider keeping `Struct` in manifests as an alias for `Tuple`, but possibly best to bite the bullet and just remove it?

Breaking changes:
* SBOR Struct type id (0x10 / 16) has been removed - use the Tuple type id (0x21 / 33)
* In the manifest, structs should be represented as `Tuple(..)`, `Struct(..)` no longer works


iamyulong:
Nice!

And yes, better not give alias. Short-term pain.

0xOmarA:
Changes look good to me. 

> We might not want to merge this before Betanet, I don't mind

I am okay with this being merged before or after Betanet. It does impact the RET but it looks like it will only be a few simple changes to the RET. So, from my side, feel free to merge before or after 👍.

iamyulong (approved):

PR #617, opened 2022-11-29
Architecture: Cleanup Auth

talekhinezh:

PR #624, opened 2022-11-30
Mutable component access rules

0xOmarA:
This PR builds upon PR #617 the adds the ability for component access rules to be mutated and updated at a later point of time if they mutable.

When instantiating a new component, developers now have the ability to set the `AccessRule`'s mutability on the `AccessRules` struct. This looks like the following:

```rust
let access_rules = AccessRules::new()
    .method("my_method", rule!(allow_all), MUTABLE(rule!(require(admin_badge))))
    .method("deposit", rule!(allow_all), LOCKED)
    .method("deposit_batch", rule!(allow_all), LOCKED)
    .default(withdraw_rule);
```

When the developer wishes to update the `AccessRule` on a specific method, they can borrow the component and update the access rules as following:

```rust
let component = borrow_component!(component_address);
component.access_rules()[0].set_method_auth("my_method", rule!(deny_all));
```

Alternatively, developers can update their component access rules through the manifest by using the `CallNativeMethod` instruction.


assets/blueprints/account/src/lib.rs
```diff
@@ -13,10 +13,10 @@ blueprint! {
...
 
             let access_rules = AccessRules::new()
-                .method("balance", rule!(allow_all))
-                .method("deposit", rule!(allow_all))
-                .method("deposit_batch", rule!(allow_all))
-                .default(withdraw_rule);
+                .method("balance", rule!(allow_all), LOCKED)
+                .method("deposit", rule!(allow_all), LOCKED)
+                .method("deposit_batch", rule!(allow_all), LOCKED)
+                .default(withdraw_rule, LOCKED);
```
talekhinezh, line 19: Not sure if we want the ability to mutate default rule, what was the idea behind this?

radix-engine-interface/src/model/auth/access_rules/invocations.rs
```diff
@@ -33,7 +33,7 @@ impl Into<NativeFnInvocation> for AccessRulesAddAccessCheckInvocation {
 pub struct AccessRulesSetAccessRuleInvocation {
     pub receiver: RENodeId,
     pub index: u32,
-    pub key: AccessRuleKey,
+    pub selector: AccessRuleSelector,
```
talekhinezh, line 36: Not sure if we gain anything by using a selector here over creating a new set of invocations for default?

scrypto/src/component/stateful_access_rules.rs
```diff
@@ -0,0 +1,129 @@
...
+            component: component.into(),
+            index,
+        }
+    }
+
+    pub fn component_address(&self) -> ComponentIdentifier {
+        self.component.clone()
+    }
+
+    pub fn access_rules(&self) -> AccessRules {
```
talekhinezh, line 37: The problem with this is that AccessRules object won't reflect correct values if `set_method_auth` is called after `access_rules` call. If going down this route I'd rather expose a `RefMut` of AccessRules instead so that AccessRules may be mutated in place.

assets/blueprints/account/src/lib.rs
0xOmarA: Josh and I discussed this during a call. We agreed that default case on access rules should not be mutable and that only methods and groups should be mutable. This has been addressed in commit [6d81f6a](https://github.com/radixdlt/radixdlt-scrypto/pull/624/commits/6d81f6aa6383f4443c18c87fa04b2a9bbe55cc7f).

scrypto/src/component/stateful_access_rules.rs
0xOmarA: Josh and I discussed this during a call. We agreed that in the long run it is desirable to use `RefMut` for cases like this. For the time being, we have agreed that Scrypto doesn't need to have the ability to read the full `AccessRules`. The only knowledge Scrypto needs is to know how many layers of `AccessRules` an entity has so that it can create `StatefulAccessRules` (name TBD) for each layer. 

For this, I implemented an `AccessRulesGetLengthInvocation` in [6d81f6a](https://github.com/radixdlt/radixdlt-scrypto/pull/624/commits/6d81f6aa6383f4443c18c87fa04b2a9bbe55cc7f)

radix-engine-interface/src/model/auth/access_rules/invocations.rs
0xOmarA: Josh and I discussed this during a call. We have agreed to remove the `AccessRuleSelector` and have a set of invocations for methods and another set for groups. This has been addressed in [23bdce8](https://github.com/radixdlt/radixdlt-scrypto/pull/624/commits/23bdce82262394d2f47a952b2d0906ef9fb1fb42)

talekhinezh (approved):

PR #626, opened 2022-11-30
Add access rules for royalty methods

iamyulong:


radix-engine-interface/src/api/types/invocation/native.rs
```diff
@@ -28,7 +28,7 @@ pub enum NativeFn {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
 #[scrypto(TypeId, Encode, Decode, Describe)]
 pub enum NativeMethod {
-    AccessRules(AccessRulesMethod),
+    AccessRulesChain(AccessRulesChainMethod),
```
talekhinezh, line 31: With "Chain" suffix, suggest this just be renamed to `AccessRuleChain`

radix-engine-interface/src/model/package/invocations.rs
```diff
@@ -12,6 +12,8 @@ use sbor::rust::string::String;
 pub struct PackagePublishNoOwnerInvocation {
     pub code: Blob,
     pub abi: Blob,
+    pub royalty_config: HashMap<String, RoyaltyConfig>,
```
talekhinezh, line 15: Can we add a TODO: comment that we need to verify with product to what extent do we want royalty configuration to be customizable?

radix-engine/src/model/component/executables.rs
```diff
@@ -78,18 +78,43 @@ impl NativeProcedure for ComponentGlobalizeWithOwnerInvocation {
         };
         let owner_badge_bucket: Bucket = api.sys_invoke(mint_invocation)?;
 
-        let mut access_rules = AccessRules::new().default(AccessRule::AllowAll);
-        access_rules.set_access_rule_and_mutability(
+        // TODO: move to sdk
```
talekhinezh, line 81: Need to verify with product :)

radix-engine-interface/src/model/package/invocations.rs
iamyulong: Just to clarify: do you mean the methods available for royalty customisation? or the mutability of royalty configuration, like fully mutable/partially mutable/immutable?

radix-engine/src/model/component/executables.rs
iamyulong: Not sure if this is a product question. The implementation is problematic as it may conflict with existing access rules that've already included configuration for metadata/royalty etc.

radix-engine-interface/src/api/types/invocation/native.rs
talekhinezh: Ah yes, you're right. For 2 I think we voted for `AccessCheck` (name for method is `add_access_check`) but don't quite like it (maybe something like `ComponentAuthGate`? to go along with `ComponentRoyaltyGate`). But okay with what you have now for now.

radix-engine-interface/src/model/package/invocations.rs
talekhinezh: I mean configurability of auth to claim/update royalties.

radix-engine/src/model/component/executables.rs
talekhinezh: Ah I misunderstood the comment then

radix-engine-interface/src/model/package/invocations.rs
iamyulong: Decided to remove `royalty_config` and `access_rules_chain` for package no owner route; and remove `access_rules_chain` for package with owner route.

radix-engine-interface/src/api/types/invocation/native.rs
iamyulong: Yeah, I like `AccessRule`, `AuthGate`, `AuthGateList` too. 

Will keep it as-is for now before we finalise on the names. :)

radix-engine-interface/src/model/package/invocations.rs
iamyulong: As we're removing owner badge as a whole, we need to change this parameter to a `NonFungibleAddress`. I will do it after the removal if you don't mind.

talekhinezh (approved):

PR #627, opened 2022-11-30
NonFungibleId Display trait change

mstrug-rdx:
### Summary
Changed `NonFungibleId` Display trait to show type and value instead of encoded plain hex value.
Task ref ID: SCRY-292

### Detailed description
Now format of NonFungibleId looks as follow:
- String: `"value"`
- U32 number: `123u32`
- U64 number: `456u64`
- Decimal: `Decimal("789")`
- Bytes: `Bytes("12ab34cd56ef")`
- UUID: `1234567890u128`

### Testing
Added new test which validates format.
Updated all tests that used the old format.


iamyulong (approved):
LGTM

PR #632, opened 2022-12-02
feature: Can serialize a scrypto value

dhedey:
Both simple and invertible modes are supported - I'll opt for "simple" for now in the Core API sbor responses and see how it goes down.

Eventually, I'd like to support a choice in the Core and Gateway APIs between what kind of SBOR response you want (raw_hex / simple / invertible / simple_from_schema).

I decided to implement this in scrypto instead of import the toolkit's values for a couple of reasons:
* I don't want to require updating the toolkit in order to pull an updated scrypto into the node - adding an extra layer into the update cycle would cause real pain
* I don't want to have to remap the scrypto value to encode it / I wanted an implementation consistent with the other code in the repo (ie ContextDisplayable)

**Breaking changes**: None

### DIFF
--- a/radix-engine-interface/src/data/value_serializer.rs
+++ b/radix-engine-interface/src/data/value_serializer.rs
@@ -48,7 +48,9 @@ impl<'a> ScryptoValueFormattingContext<'a> {
     }
 }
 
-pub trait SerializableScryptoValue: for<'a> ContextualSerialize<ScryptoValueFormattingContext<'a>> {
+pub trait SerializableScryptoValue:
+    for<'a> ContextualSerialize<ScryptoValueFormattingContext<'a>>
+{
     fn simple_serializable<'a, 'b, TContext: Into<ValueFormattingContext<'b>>>(
         &'a self,
         context: TContext,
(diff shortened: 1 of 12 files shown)


radix-engine-interface/src/data/value_serializer.rs
```diff
@@ -0,0 +1,667 @@
...
+                },
+                SborValue::Array {
+                    element_type_id: SborTypeId::U8,
+                    elements: vec![SborValue::U8 { value: 0x3a }, SborValue::U8 { value: 0x92 }],
+                },
+                SborValue::Array {
+                    element_type_id: SborTypeId::U32,
+                    elements: vec![SborValue::U32 { value: 153 }, SborValue::U32 { value: 62 }],
+                },
+                SborValue::Tuple {
```
iamyulong, line 548: Looks like we're missing enum test here. Would also like one instance for each custom type, so we're less likely to break it.

radix-engine-interface/src/data/value_serializer.rs
```diff
@@ -0,0 +1,667 @@
...
+        let proof_names = HashMap::new();
+
+        let context = ValueFormattingContext::with_manifest_context(
+            Some(&encoder),
+            &bucket_names,
+            &proof_names,
+        );
+
+        assert_simple_json_matches(&value, context, expected_simple);
+        assert_invertible_json_matches(&value, context, expected_invertible);
```
iamyulong, line 639: Can we add test to verify the JSON representation is indeed invertible?

radix-engine-interface/src/data/value_serializer.rs
```diff
@@ -0,0 +1,667 @@
...
+            // In the invertible case, we have to add the tag anyway
+            serialize_type_name_tagged_value(serializer, context, type_name, value)
+        }
+    }
+}
+
+pub fn serialize_type_name_tagged_value<S: Serializer, T: Serialize + ?Sized>(
+    serializer: S,
+    _context: &ScryptoValueFormattingContext,
+    type_name: &'static str,
```
iamyulong, line 443: Possibly to use `ScryptoTypeId`?

dhedey: I considered it, but it's possibly less performant - happy to do it if you'd rather though... will do it in the next PR :)

radix-engine-interface/src/data/value_serializer.rs
iamyulong: Yes, that's good.

radix-engine-interface/src/data/value_serializer.rs
iamyulong: I suppose Rust simple enum matching is performant. Would prefer enum for compiler check to avoid mistakes, considering we're constantly renaming things. :)

dhedey: I've changed this now 👍

iamyulong (approved):

PR #636, opened 2022-12-02
Change royalty config to be cost unit-based (optimised)

iamyulong:
From https://github.com/radixdlt/radixdlt-scrypto/pull/630, we noticed significant performance loss with Decimal-based accounting.

This PR attempts to address this problem so we can safely release it.


radix-engine-constants/src/lib.rs
```diff
@@ -1,18 +1,41 @@
...
 
-/// The default cost unit limit for a transaction.
-pub const DEFAULT_COST_UNIT_LIMIT: u32 = 100_000_000;
+/// The minimum value of tip percentage
+pub const DEFAULT_MIN_TIP_PERCENTAGE: u8 = 0;
 
-/// The default cost unit price.
-pub const DEFAULT_COST_UNIT_PRICE: &'static str = "0.0000001";
+/// The maximum value of tip percentage
+pub const DEFAULT_MAX_TIP_PERCENTAGE: u8 = u8::MAX;
```
mstrug-rdx, line 25: Max tip percentage shouldn't be 100?

iamyulong: Haha, we deliberately allowed values greater than 100%.

PR #643, opened 2022-12-05
NonFungibleAddress manifest specification change

mstrug-rdx:
### Summary
Changed `NonFungibleAddress` transaction manifest format.
Task ref ID: SCRY-303

### Detailed description
Now format of NonFungibleAddress looks as follow:
`NonFungibleAddress(<ResourceAddress>, <NonFungibleId>)`
where:
`<ResourceAddress>` - bech32 context formatted resource address `"resource_xxxx"`
`<NonFungibleId>` - string representation of NonFungibleId type, ex.: `Bytes("a1b2c3")`, `1u32`, `"value"`

Also updated `NonFungibleId` manifest representation tests.


transaction/src/manifest/ast.rs
```diff
@@ -251,7 +252,7 @@ pub enum Value {
     // Other interpreted types
     Expression(Box<Value>),
     Blob(Box<Value>),
-    NonFungibleAddress(Box<Value>),
+    NonFungibleAddress(Vec<Value>),
```
iamyulong, line 255: Better to be represented as `(Box<Value>, Box<Value>)`

transaction/src/manifest/compiler.rs
```diff
@@ -190,7 +190,7 @@ mod tests {
                 Instruction::PublishPackageWithOwner {
                     code: Blob(code_hash),
                     abi: Blob(abi_hash),
-                    owner_badge: NonFungibleAddress::from_str("00ed9100551d7fae91eaf413e50a3c5a59f8b96af9f1297890a8f45c200721031b84c5567b126440995d3ed5aaba0565d71e1834604819ff9c17f5e9d5dd078f").unwrap(),
+                    owner_badge: NonFungibleAddress::from_str("0000000000000000000000000000000000000000000000000000045c200721031b84c5567b126440995d3ed5aaba0565d71e1834604819ff9c17f5e9d5dd078f").unwrap(),
```
iamyulong, line 193: Better to construct directly, like `NonFungibleAddress("resource_xxx", Bytes("xxx"))`?

transaction/src/manifest/generator.rs
```diff
@@ -1124,6 +1146,10 @@ fn generate_value(
                 value: ScryptoCustomValue::NonFungibleId(v),
             })
         }
+        ast::Value::Bytes(value) =>  
+            Ok( SborValue::String {
+                value: generate_string(value)?
```
iamyulong: This is incorrect. Should be mapped to `SborValue::Array<u8>`

transaction/src/manifest/generator.rs
```diff
@@ -1203,6 +1229,7 @@ fn generate_type_id(ty: &ast::Type) -> ScryptoSborTypeId {
         ast::Type::Decimal => SborTypeId::Custom(ScryptoCustomTypeId::Decimal),
         ast::Type::PreciseDecimal => SborTypeId::Custom(ScryptoCustomTypeId::PreciseDecimal),
         ast::Type::NonFungibleId => SborTypeId::Custom(ScryptoCustomTypeId::NonFungibleId),
+        ast::Type::Bytes => SborTypeId::String,
```
iamyulong, line 1235: Same

transaction/src/manifest/parser.rs
```diff
@@ -300,7 +300,8 @@ impl Parser {
             TokenKind::EddsaEd25519Signature |
             TokenKind::Decimal |
             TokenKind::PreciseDecimal |
-            TokenKind::NonFungibleId  => self.parse_scrypto_types(),
+            TokenKind::NonFungibleId |
+            TokenKind::Bytes => self.parse_scrypto_types(),
```
iamyulong, line 304: Bytes are not "custom" scrypto type. Suggest independent method.

radix-engine-interface/src/model/resource/non_fungible_id.rs
```diff
@@ -43,10 +40,16 @@ impl NonFungibleId {
             NonFungibleId::String(..) => NonFungibleIdType::String,
             NonFungibleId::U32(..) => NonFungibleIdType::U32,
             NonFungibleId::U64(..) => NonFungibleIdType::U64,
-            NonFungibleId::Decimal(..) => NonFungibleIdType::Decimal,
             NonFungibleId::UUID(..) => NonFungibleIdType::UUID,
         }
     }
+
+    /// Converts ordinary bytes buffer encoded as hex string
+    pub fn try_bytes_from_hex(hex_str: &str) -> Result<Self, ParseNonFungibleIdError> {
```
dhedey, line 48: I got a bit confused by this method, assuming it did SBOR decoding or something.

On reflection, I don't think we should have this method :) - people can construct more simply with `NonFungibleId::Bytes(hex::decode("abcd"))` themselves.

transaction/src/manifest/compiler.rs
dhedey: Just to add to Yulong's point - please can we remove `from_str` from NonFungibleAddress completely?

Having these methods encourages people to think of NonFungibleAddress / NonFungibleId as sbor hex, which I'm very keen we move away from.

(If people want to SBOR encode them, they can scrypto_encode / scrypto_decode themselves)

transaction/src/manifest/decompiler.rs
```diff
@@ -150,7 +150,7 @@ pub fn decompile_instruction<F: fmt::Write>(
                 f,
                 "TAKE_FROM_WORKTOP_BY_IDS Array<NonFungibleId>({}) ResourceAddress(\"{}\") Bucket(\"{}\");",
                 ids.iter()
-                    .map(|k| format!("NonFungibleId(\"{}\")", hex::encode(&k.to_vec())))
+                    .map(|k| format!("NonFungibleId({})", k))
```
dhedey, line 153: I'd rather we didn't rely on the fact that `k`'s display method writes it in manifest-speak, and call an explicit method here (and ditto for other methods).

EG I could imagine we have separate methods for the two different display methodologies from the table:
* `to_manifest_string(&self) -> String` / `from_manifest_string(string: &str) -> Self`
* `to_simple_string(&self) -> String` / `from_simple_string(id_type: NonFungibleIdType, string: &str) -> Self`

And we call `to_manifest_string` here.

I would also rather that display outputs the simple string - because display is intended for users reading, and debug outputs the manifest string - but thoughts welcome.

Separately, we should have the following on NonFungibleAddress:
* `to_canonical_string(&self, bech32_encoder: &Bech32Encoder) -> String` / `from_canonical_string(bech32_decoder: &Bech32Decoder, id_type: NonFungibleIdType, string: &str) -> Self`

Which uses that which encodes the resource address, then `:`, then the simple string representation of the id.

Finally, we can use that to implement `impl<'a> ContextualDisplay<AddressDisplayContext<'a>> for NonFungibleAddress` (see eg `model/component/data.rs` for reference)

radix-engine-interface/src/data/value_formatter.rs
```diff
@@ -256,7 +256,12 @@ pub fn format_custom_value<F: fmt::Write>(
             write!(f, "Blob(\"{}\")", value)?;
         }
         ScryptoCustomValue::NonFungibleAddress(value) => {
-            write!(f, "NonFungibleAddress(\"{}\")", value)?;
+            f.write_str("NonFungibleAddress(\"")?;
+            value
+                .resource_address()
+                .format(f, context.bech32_encoder)
+                .expect("Failed to format address");
+            write!(f, "\", {})", value.non_fungible_id())?;
```
dhedey, line 264: I'm not sure I like the fact that we rely on `display` of non fungible id being the scrypto value / manifest representation - could we have an explicit method or a method here instead?

transaction/src/manifest/ast.rs
mstrug-rdx: Implemented.

radix-engine-interface/src/model/resource/non_fungible_id.rs
mstrug-rdx: Removed.

transaction/src/manifest/parser.rs
mstrug-rdx: Changed.

transaction/src/manifest/generator.rs
mstrug-rdx: Changed.

transaction/src/manifest/generator.rs
mstrug-rdx: Changed.

radix-engine-interface/src/data/value_formatter.rs
mstrug-rdx: Updated.

transaction/src/manifest/decompiler.rs
mstrug-rdx: Removed NonFungibleAddress::from_str() function, in Simulator/publish command switched to String and added code to parse from String to hex, not sure how else this could be done without changing input arguments.

radix-engine-interface/src/model/resource/non_fungible_address.rs
```diff
@@ -91,6 +93,41 @@ impl NonFungibleAddress {
...
+                    {
+                        Ok(NonFungibleAddress::new(raddr, nfid))
+                    } else {
+                        Err(ParseNonFungibleAddressError::InvalidNonFungibleId)
+                    }
+                } else {
+                    Err(ParseNonFungibleAddressError::InvalidResourceAddress)
+                }
+            } else {
+                Err(ParseNonFungibleAddressError::InvalidLength(s.len()))
```
iamyulong, line 125: Minor: Misleading error message. I would use `.split().collect()` and give `InvalidNumberOfParts` error.

radix-engine-interface/src/model/resource/non_fungible_address.rs
```diff
@@ -155,11 +210,11 @@ impl FromPublicKey for NonFungibleAddress {
 mod tests {
     use super::*;
     use crate::address::Bech32Decoder;
-    use sbor::rust::string::ToString;
 
     #[test]
     pub fn non_fungible_address_from_and_to_string_succeeds() {
-        // Arrange
+        let expected_address = "00a6b27633f925c38a123dcfa488fbed09f3c36e97d055e2d603075c20071630071000000071dba5dd36e30de857049805fd1553cd";
```
iamyulong, line 216: Minor: I would remove hex representation everywhere. :)

radix-engine-interface/src/model/resource/non_fungible_id.rs
```diff
@@ -43,10 +42,89 @@ impl NonFungibleId {
...
+            Ok(NonFungibleId::U32(
+                s[0..s.len() - 3]
+                    .parse::<u32>()
+                    .map_err(|_| ParseNonFungibleIdError::InvalidValue)?,
+            ))
+        } else if s.len() > 2 && s.starts_with("\"") && s.ends_with("\"") {
+            Ok(NonFungibleId::String(s[1..s.len() - 1].to_string()))
+        } else {
+            Err(ParseNonFungibleIdError::InvalidValue)
+        }
```
iamyulong, line 90: This is a very fragile implementation. Can be wrong for many edge cases, e.g. `Bytes( "deadbeef" )` and `"abc"d"`. We should probabaly use manifest parser to do the this job.

In the interest of time, I'd suggest we add a `TODO` note and merge it as-is.

simulator/src/resim/cmd_publish.rs
```diff
@@ -95,9 +95,14 @@ impl Publish {
             );
             writeln!(out, "Package updated!").map_err(Error::IOError)?;
         } else {
+            let owner_badge = hex::decode(self.owner_badge.clone())
```
iamyulong, line 98: Let's remove hex. We can use manifest string for this case.

- `resim publish <path>  resource_xxxx,123u32`
- `resim publish <path>  resource_xxxx,"abc"` (users will type `\"` to work around shell)
- `resim publish <path>  resource_xxxx,456u128`

iamyulong: Similarily, need to check if `resim new-owner-badge` outputs the same syntax.

radix-engine-interface/src/model/resource/non_fungible_address.rs
mstrug-rdx: done

radix-engine-interface/src/model/resource/non_fungible_id.rs
mstrug-rdx: done

radix-engine-interface/src/model/resource/non_fungible_address.rs
mstrug-rdx: done

simulator/src/resim/cmd_publish.rs
mstrug-rdx: @dhedey please go ahed with implementation

iamyulong (approved):
LGTM

PR #645, opened 2022-12-05
Feature/manifest aliases for enums

dhedey:
Adds aliases for well-known enums to the manifest.

Also adds in better tests that compile and recompile are inverses, and fixes a non-determinism during transaction compilation by using an IndexMap for the blobs.

**Breaking Changes:**
* None per-se - this adds back support for "None", "Some", "Ok" and "Err" in the manifest.


0xOmarA (approved):
Looks good to me. Happy with us moving to `IndexMap` for blobs as this was the source of an old bug in the RET.

PR #662, opened 2022-12-07
Dedicated instructions for native invocation (Part 2)

iamyulong:
Follows https://github.com/radixdlt/radixdlt-scrypto/pull/653

- Updated `MINT_RESOURCE`
- Added `RECALL_RESOURCE`
- Added `SET_METHOD_ACCESS_RULE`
- Removed `CALL_NATIVE_FUNCTION` and `CALL_NATIVE_METHOD`


radix-engine-interface/src/constants.rs
```diff
@@ -11,284 +14,303 @@ use crate::model::*;
...
+    94
 );
 
 pub const EPOCH_MANAGER_BLUEPRINT: &str = "EpochManager";
 pub const CLOCK_BLUEPRINT: &str = "Clock";
 pub const RESOURCE_MANAGER_BLUEPRINT: &str = "ResourceManager";
 pub const PACKAGE_BLUEPRINT: &str = "Package";
 pub const TRANSACTION_PROCESSOR_BLUEPRINT: &str = "TransactionProcessor";
+
+pub struct AuthAddresses;
```
talekhinezh, line 299: I would prefer having these constants closer to their module rather than type (ie should be under Auth rather than constants) so that we can keep things modular (e.g. compile a kernel with no auth enabled).

radix-engine-interface/src/constants.rs
```diff
@@ -11,284 +14,303 @@ use crate::model::*;
...
+
+impl AuthAddresses {
+    pub fn genesis_role() -> NonFungibleAddress {
```
talekhinezh, line 302: Can we rename this to `system_role`, the idea is that this role would also be capable of doing hardfork state updates not normally possible.

iamyulong: No problem with that.

radix-engine-interface/src/constants.rs
iamyulong: Not sure how practical this is, but will move it closer to auth module.

talekhinezh (approved):

PR #666, opened 2022-12-08
No more hashing in wasm cache

iamyulong:
- Use `(PackageAddress, WasmMeteringConfig)` as cache key, instead of `hash(code)`;
- Share `ScryptoInterpreter` between bootstrap and subsequent transactions;
- `TestRunner` now owns a substate store.

### DIFF
--- a/simulator/src/resim/cmd_show_ledger.rs
+++ b/simulator/src/resim/cmd_show_ledger.rs
@@ -13,12 +13,12 @@ pub struct ShowLedger {}
 
 impl ShowLedger {
     pub fn run<O: std::io::Write>(&self, out: &mut O) -> Result<(), Error> {
-        let ledger = RadixEngineDB::new(get_data_dir()?);
-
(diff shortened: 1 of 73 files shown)


simulator/src/resim/cmd_show_ledger.rs
```diff
@@ -13,7 +13,7 @@ pub struct ShowLedger {}
 
 impl ShowLedger {
     pub fn run<O: std::io::Write>(&self, out: &mut O) -> Result<(), Error> {
-        let ledger = RadixEngineDB::with_bootstrap(get_data_dir()?);
```
dhedey, line 16: If we no longer do (optional) bootstrapping in all these commands, how is a user supposed to get started / bootstrapped? Does this make using resim harder?

iamyulong: Thought empty ledger state dump should be okay, but maybe it's more user friendly if we always bootstrap for any resim commands that read/write the substate store. I will revert.

PR #669, opened 2022-12-08
Add Scrypto DateTime util

LukasGasior1:
Uses signed integers for the Unix timestamp (which is not uncommon), so that `Instant` can represent any date from year 1 up to u32::MAX.


radix-engine-interface/src/time/instant.rs
```diff
@@ -0,0 +1,63 @@
+use crate::time::constants::*;
+use sbor::*;
+
+#[derive(Encode, Decode, TypeId, Copy, Clone, Debug, Eq, PartialEq)]
+pub struct Instant {
+    pub seconds_since_unix_epoch: i64,
```
LukasGasior1, line 6: This file is mostly unchanged, but gh fails to show a proper diff (I've moved it) :(
the changes are:
* u64 -> i64 for seconds_since_unix_epoch
* and added `checked_mul` in add_xyz methods

radix-engine-interface/src/time/date_time.rs
```diff
@@ -0,0 +1,512 @@
...
+// to a base date that is a multiple of a 400-year leap cycle.
+// Used in Instant -> DateTime conversion.
+// The date we're using is 2000-03-01 00:00:00, for two reasons:
+// a) It's a multiple of 400, to make it easier to work with leap years
+// b) We're also shifting the month to 1st March, so that
+//    the extra day on leap years is added to the last month (Feb),
+//    not in the middle of a year (makes some calculations easier)
+const SHIFT_FROM_UNIX_TIME_TO_MARCH_Y2K: i64 = 946684800 + 86400 * (31 + 29);
+const MIN_SUPPORTED_TIMESTAMP: i64 = -62135596800;
+const MAX_SUPPORTED_TIMESTAMP: i64 = 135536014634284799;
```
iamyulong, line 24: How are these numbers calculated? Worth citing where all these constants are ported from if no formula is provided?

radix-engine-interface/src/time/date_time.rs
```diff
@@ -0,0 +1,512 @@
...
+
+impl DateTime {
+    pub fn new(
+        year: u32,
+        month: u8,
+        day_of_month: u8,
+        hour: u8,
+        minute: u8,
+        second: u8,
+    ) -> Result<Self, String> {
```
iamyulong, line 44: Worth adding a proper enum for all error codes and moving the user-friendly messages to `fmt::Display` impl?

radix-engine-interface/src/time/date_time.rs
```diff
@@ -0,0 +1,512 @@
...
+            return Err(
+                "Invalid minute. Expected a value between 0 (inclusive) and 59 (inclusive)"
+                    .to_string(),
+            );
+        }
+
+        if second > 59 {
+            return Err(
+                "Invalid second. Expected a value between 0 (inclusive) and 59 (inclusive)"
+                    .to_string(),
```
iamyulong, line 85: Regarding inclusiveness and exclusiveness, are we following a standard?

radix-engine-interface/src/time/date_time.rs
```diff
@@ -0,0 +1,512 @@
...
+            year,
+            month,
+            day_of_month,
+            hour,
+            minute,
+            second,
+        })
+    }
+
+    pub fn from_instant(instant: &Instant) -> Result<Self, String> {
```
iamyulong, line 99: Same Error representation.

radix-engine-interface/src/time/date_time.rs
```diff
@@ -0,0 +1,512 @@
...
+        if instant.seconds_since_unix_epoch < MIN_SUPPORTED_TIMESTAMP
+            || instant.seconds_since_unix_epoch > MAX_SUPPORTED_TIMESTAMP
+        {
+            return Err(format!(
+                "Instant out of supported range [{}, {}]",
+                MIN_SUPPORTED_TIMESTAMP, MAX_SUPPORTED_TIMESTAMP
+            ));
+        }
+
```
iamyulong, line 108: All maths below are guaranteed not to overflow because of the range check above right?

radix-engine-interface/src/time/date_time.rs
```diff
@@ -0,0 +1,512 @@
...
+        let mut year =
+            remaining_years +
+                4 * num_4_year_cycles +
+                100 * num_100_year_cycles +
+                400 * num_400_year_cycles
+                + 2000 /* Add the base year (after shifting) */;
+
+        let mut days_in_months_starting_on_march = LEAP_YEAR_DAYS_IN_MONTHS.clone();
+        days_in_months_starting_on_march.rotate_left(2);
```
iamyulong, line 156: We don't consider non-leap year here?

radix-engine-interface/src/time/date_time.rs
```diff
@@ -0,0 +1,512 @@
...
+        let minute = remaining_secs / SECONDS_IN_A_MINUTE as i64 % SECONDS_IN_A_MINUTE as i64;
+        let second = remaining_secs % SECONDS_IN_A_MINUTE as i64;
+
+        Ok(Self {
+            year: u32::try_from(year).map_err(|_| "year overflow")?,
+            month: u8::try_from(month).map_err(|_| "month overflow")?,
+            day_of_month: u8::try_from(day_of_month).map_err(|_| "day_of_month overflow")?,
+            hour: u8::try_from(hour).map_err(|_| "hour overflow")?,
+            minute: u8::try_from(minute).map_err(|_| "minute overflow")?,
+            second: u8::try_from(second).map_err(|_| "second overflow")?,
```
iamyulong, line 188: Shouldn't no-error be expected here?

iamyulong (commented):
I don't think I can verify the correctness of the algorithm, but left a few general comments in case it's helpful.

radix-engine-interface/src/time/date_time.rs
LukasGasior1: They correspond to min/max values that the DateTime can represent (`Jan 1, year 1, 00:00:00` up to `Dec 31, year u32::MAX, 23:59:59`). I'll add a comment :+1:

radix-engine-interface/src/time/date_time.rs
LukasGasior1: I believe 0-23 (0-59 respectively) is more or less a standard

radix-engine-interface/src/time/date_time.rs
LukasGasior1: correct

radix-engine-interface/src/time/date_time.rs
LukasGasior1: true, none of these should ever overflow

radix-engine/src/model/clock/executables.rs
```diff
@@ -203,7 +204,10 @@ impl NativeProcedure for ClockGetCurrentTimeExecutable {
                 let substate_ref = system_api.get_ref(handle)?;
                 let substate = substate_ref.current_time_rounded_to_minutes();
                 let instant = Instant::new(
-                    substate.current_time_rounded_to_minutes_ms / SECONDS_TO_MS_FACTOR,
+                    i64::try_from(
+                        substate.current_time_rounded_to_minutes_ms / SECONDS_TO_MS_FACTOR,
```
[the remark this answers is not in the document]
iamyulong: Cool

radix-engine-interface/src/time/date_time.rs
iamyulong: Maybe we can document the range of  `year`, `month`, `day_of_month`, `hour`, `minute`, `second`  on the constructor `new` if no standard has been strictly followed.

radix-engine-interface/src/time/date_time.rs
LukasGasior1: Yeah, this surely isn't obvious :sweat_smile: but `remaining_days` is at most 365 (29 Feb on leap years), and, thanks to shifting the base to March (see comment above `const SHIFT_FROM_UNIX_TIME_TO_MARCH_Y2K`) Feb comes last, so, after iterating through all other months, the while loop predicate for the last month [Feb]:  `days[..] <= remaining_days` never matches.

In other words, we don't care about the leap year for dates before Mar 1: if the month happens to be Feb we just end up with a correct number of remaining_days, and if the date is on or after Mar 1, then it's just another full year that's already been handled earlier.

iamyulong: Woooow, that's very convoluted and interesting!

radix-engine/src/model/clock/executables.rs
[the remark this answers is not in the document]
LukasGasior1: I've changed it. I don't think there was a strong reason to keep it as u64, and now our system can also run in 1969 :D as a bonus we don't need those type conversions anymore

radix-engine-interface/src/time/instant.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub fn add_days(&self, days_to_add: i64) -> Option<Instant> {
+        days_to_add
+            .checked_mul(SECONDS_IN_A_DAY as i64)
+            .and_then(|to_add| self.seconds_since_unix_epoch.checked_add(to_add))
+            .map(Instant::new)
+    }
+
+    pub fn add_hours(&self, hours_to_add: i64) -> Option<Instant> {
+        hours_to_add
+            .checked_mul(SECONDS_IN_AN_HOUR as i64)
```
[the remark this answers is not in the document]
LukasGasior1: yup, changed most of those constants to i64

radix-engine-interface/src/time/date_time.rs
LukasGasior1: (changed to `expect`)

radix-engine-interface/src/time/date_time.rs
LukasGasior1: Hmh, I'd prefer not to reference the standard in the doc. For example, apparently some versions of it allow hour=24 (https://en.wikipedia.org/wiki/ISO_8601#:~:text=As%20of%20ISO%208601%2D1%3A2019/Amd,permitted%20in%20earlier%20versions%20of%20the%20standard.) and `60` seems to be a valid value for `second` (for leap seconds, which we don't support).

radix-engine/src/model/clock/executables.rs
```diff
@@ -262,7 +263,10 @@ impl NativeProcedure for ClockCompareCurrentTimeExecutable {
                 let substate_ref = system_api.get_ref(handle)?;
                 let substate = substate_ref.current_time_rounded_to_minutes();
                 let current_time_instant = Instant::new(
-                    substate.current_time_rounded_to_minutes_ms / SECONDS_TO_MS_FACTOR,
+                    i64::try_from(
+                        substate.current_time_rounded_to_minutes_ms / SECONDS_TO_MS_FACTOR,
+                    )
+                    .expect("i64 overflow on current_time_rounded_to_minutes_ms"),
```
iamyulong, line 269: We can remove the `try_from` now :)

LukasGasior1: missed this one :+1:

dhedey (approved):

PR #674, opened 2022-12-09
BugFix: Fix ResourceManager Metadata Auth

talekhinezh:


radix-engine/src/model/resources/executables/resource_manager.rs
```diff
@@ -276,8 +281,25 @@ fn build_substates(
...
+        ))),
+        update_metadata_access_rule,
+        update_metadata_mutability,
+    );
+    access_rules.set_access_rule_and_mutability(
+        AccessRuleKey::Native(NativeFn::Method(NativeMethod::Metadata(
+            MetadataMethod::Get,
+        ))),
+        AllowAll,
+        DenyAll,
```
iamyulong, line 301: `AllowAll` + `DenyAll` here is by design, even for the `WithOwner` variant, right?

iamyulong (approved):

radix-engine/src/model/resources/executables/resource_manager.rs
talekhinezh: Yup, that's right.

PR #678, opened 2022-12-11
Architecture: Refactor Transaction Processor

talekhinezh:
* Cleanup TransactionProcessor logic
* Removed `NativeProcedure` and `NativeExecutor ` which are remnants of the old architecture and don't do anything at the moment except delegate.
* Change Transaction return value from `Vec<Vec<u8>>` to `Vec<InstructionOutput>`. This allows us to skip serialization of output unless when needed. Moreover this allows us to use type checked results and skip serialization/deserialization when not crossing process boundaries.


radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -230,8 +226,16 @@ impl TransactionReceipt {
...
-        scrypto_decode::<T>(&self.expect_commit_success()[nth][..])
-            .expect("Wrong instruction output type!")
+        match &self.expect_commit_success()[nth] {
+            InstructionOutput::Native(native) => {
+                // TODO: Use downcast
+                let value = IndexedScryptoValue::from_typed(&native.as_ref());
+                scrypto_decode::<T>(&value.raw).expect("Wrong instruction output type!")
+            }
+            InstructionOutput::Scrypto(value) => {
+                scrypto_decode::<T>(&value.raw).expect("Wrong instruction output type!")
```
lrubasze, line 236: maybe it would be useful to print different error message depending on Instruction output kind?
but that's really a minor thing

radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -230,8 +226,16 @@ impl TransactionReceipt {
     }
 
     pub fn output<T: ScryptoDecode>(&self, nth: usize) -> T {
-        scrypto_decode::<T>(&self.expect_commit_success()[nth][..])
-            .expect("Wrong instruction output type!")
+        match &self.expect_commit_success()[nth] {
```
lrubasze, line 229: is there no need to check if nth is not out of bounds?

radix-engine/src/transaction/transaction_receipt.rs
talekhinezh: Yup good point

radix-engine/src/transaction/transaction_receipt.rs
talekhinezh: Don't think it's too necessary here as the caller should always know how many instructions/outputs there are. It may be that there is a better abstraction that would avoid this invariant being upheld by the client but don't think it's too important now.

lrubasze: ok, sounds reasonable

0xOmarA (approved):
left one minor comment about making the match statement exhaustive, aside from that, all looks good to me.

radix-engine/src/model/transaction_processor/executables.rs
```diff
@@ -47,851 +47,840 @@ pub enum TransactionProcessorError {
...
+            }
+            | BasicInstruction::CreateProofFromAuthZoneByIds {
+                resource_address, ..
+            }
+            | BasicInstruction::MintResource {
+                resource_address, ..
+            } => {
+                update.add_ref(RENodeId::Global(GlobalAddress::Resource(*resource_address)));
+            }
+            _ => {}
```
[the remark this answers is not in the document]
talekhinezh: Yup, good point!

PR #679, opened 2022-12-12
Architecture: Refactor Modules

talekhinezh:


radix-engine/src/model/fee/module.rs
```diff
@@ -27,6 +35,12 @@ impl<R: FeeReserve> Module<R> for CostingModule {
                 value_count,
                 ..
             } => {
+                if depth == self.max_depth {
```
0xOmarA, line 38: Conceptually speaking, I am not sure I understand why is the costing module the one performing this check. It seems to me like this is somewhat out of its scope?

0xOmarA (approved):
Looks good to me. This change captures the intent that set of modules doesn't change during runtime quite well.

radix-engine/src/model/fee/module.rs
talekhinezh: It's in the costing module as the costing module's main responsibility is to prevent unbounded resource use, more specifically it does this by adding economic bounds to resource useage. With the case of `max_depth`, we decided on a simple cap rather than spending cost units, but definitely possible to change.

Also, the change is relatively small and related to module configuration so was added to this PR.

PR #680, opened 2022-12-13
Common Manifests

0xOmarA:
This PR adds a number of common manifests which may be needed by different teams.

### DIFF
--- a/radix-engine/tests/common_transactions.rs
+++ b/radix-engine/tests/common_transactions.rs
@@ -1,0 +1,488 @@
+use radix_engine::ledger::TypedInMemorySubstateStore;
+use radix_engine::types::{
+    hash, Bech32Encoder, Blob, ComponentAddress, Decimal, FromPublicKey, NonFungibleAddress,
+    NonFungibleId, ResourceAddress, ACCOUNT_PACKAGE, FAUCET_COMPONENT, RADIX_TOKEN,
+};
+use radix_engine_interface::core::NetworkDefinition;
+use scrypto_unit::TestRunner;
+use transaction::manifest::compile;
+use transaction::signing::EcdsaSecp256k1PrivateKey;
+
+/// An example manifest for freeing some funds from the faucet
+#[test]
+fn free_funds_from_faucet_succeeds() {
+    test_manifest(|account_component_address, bech32_encoder| {
+        let manifest = format!(
+            r#"
+        # Locking 10 XRD in fees from the account component. 
+        CALL_METHOD 
+            ComponentAddress("{account_component_address}") 
+            "lock_fee"
+            Decimal("10");
+    
+        # Calling the "free" method on the faucet component which is the method responsible for 
+        # dispensing funds from the faucet.
+        CALL_METHOD 
+            ComponentAddress("{faucet_component_address}") 
+            "free";
+    
+        # Depositing all of the XRD dispensed from the faucet into our account component.
+        CALL_METHOD
+            ComponentAddress("{account_component_address}") 
+            "deposit_batch"
+            Expression("ENTIRE_WORKTOP");
+        "#,
+            faucet_component_address =
+                bech32_encoder.encode_component_address_to_string(&FAUCET_COMPONENT),
+            account_component_address =
+                bech32_encoder.encode_component_address_to_string(&account_component_address)
+        );
+        (manifest, Vec::new())
+    });
+}
+
+/// An example manifest for the creation of non-virtual (physical?) accounts
+#[test]
+fn creating_a_non_virtual_account_succeeds() {
+    test_manifest(|_, bech32_encoder| {
+        let private_key = EcdsaSecp256k1PrivateKey::from_u64(12).unwrap();
+        let public_key = private_key.public_key();
+        let virtual_badge_non_fungible_address = NonFungibleAddress::from_public_key(&public_key);
+
+        let manifest = format!(
+            r#"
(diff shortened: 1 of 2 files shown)


radix-engine/tests/common_transactions.rs
```diff
@@ -0,0 +1,338 @@
...
+use transaction::manifest::compile;
+use transaction::signing::EcdsaSecp256k1PrivateKey;
+
+/// An example manifest for freeing some funds from the faucet
+#[test]
+fn free_funds_from_faucet_succeeds() {
+    test_manifest(|account_component_address, bech32_encoder| {
+        let manifest = format!(
+            include_str!("../../transaction/examples/faucet/free_funds.rtm"),
+            faucet_component_address =
```
dhedey, line 17: If you import the `ContextualDisplay` trait, you should be able to do `FAUCET_COMPONENT.to_string(bech32_encoder)` :)

radix-engine/tests/common_transactions.rs
```diff
@@ -0,0 +1,338 @@
...
+                owner_badge_non_fungible_address.non_fungible_id()
+            {
+                *non_fungible_id
+            } else {
+                panic!("expected a u32 non-fungible-id");
+            };
+
+            // TODO: Update the complex.abi and complex.code files that are used for testing.
+            // Using the WASM and ABI from the account blueprint here as they are up to date. The
+            // complex.code and complex.abi files from the transaction crate are not.
```
dhedey, line 174: 👍

Do you think it might be worth separating the current contents of `transaction/examples` to eg `transaction/test-cases` or something? So that `transaction/examples` is full of only easy to use/follow manifests? :D

transaction/examples/account/multi_account_resource_transfer.rtm
```diff
@@ -0,0 +1,46 @@
+# This transaction shows how funds can be transferred from one account into multiple accounts all
```
dhedey, line 1: Nice! Thanks for explaining these 👍

transaction/examples/resources/creation/non_fungible/with_initial_supply.rtm
```diff
@@ -0,0 +1,39 @@
...
+        "NonFungible", 
+        Enum("U32")
+    ) 
+    Array<Tuple>(
+        Tuple("name", "MyResource"), 
+        Tuple("symbol", "RSRC"),
+        Tuple("description", "A very innovative and important resource"), 
+    ) 
+    Array<Tuple>(
+        Tuple(Enum("Withdraw"), Tuple(Enum("AllowAll"), Enum("DenyAll"))),
```
dhedey, line 22: Do you think it's worth a quick comment explaining what the two different rules are? (ie the first is for performing the action; the second is for updating either of the auth rules associated with the action)

transaction/examples/account/account_creation.rtm
```diff
@@ -0,0 +1,43 @@
...
+        "Protected", 
+        Enum(
+            "ProofRule", 
+            Enum(
+                "Require", 
+                Enum(
+                    "StaticNonFungible", 
+                    NonFungibleAddress(
+                        "{virtual_badge_resource_address}", 
+                        Bytes("{virtual_badge_non_fungible_id}")
```
dhedey, line 37: Is it worth possibly explaining where these come from?

transaction/examples/resources/mint_fungible.rtm
```diff
@@ -0,0 +1,19 @@
+# This transaction manifest shows how a mintable fungible resource may be minted through the
```
dhedey, line 1: It would be nice to pair this with a resource which is mintable via an auth rule - and then this would show proof from the account to do the minting? :)

As I imagine minting with auth will be the main way of minting?

dhedey (approved):
Really nice work, thanks Omar! 

Happy for any/all of my comments to be addressed in a future PR / after betanet, so marking as approved 👍

radix-engine/tests/common_transactions.rs
0xOmarA: Agreed. I have just made the change to use `ContextualDisplay::display` for all addresses.

radix-engine/tests/common_transactions.rs
0xOmarA: Done. All of the manifests and blobs that used to be in the root of the examples directory have now been moved to `examples/test-cases` so that this is the one directory that people do not look at :D

transaction/examples/account/account_creation.rtm
0xOmarA: Yes and No. I anticipate that this manifest will be primarily used by core-apps for account creation. I think that it's easier for them to ask and based on who is asking I can say "use the `NonFungibleAddress.from_public_key` in the kotlin-engine-toolkit" or use the `DeriveNonFungibleAddressFromPublicKeyRequest` in the swift-engine-toolkit".

Happy to add an explanation if you think that it is worth it :)

transaction/examples/resources/mint_fungible.rtm
0xOmarA: Yeah you are right. I will add another example now of how somebody can mint a restricted minting fungible and non-fungible resources :)

transaction/examples/account/account_creation.rtm
dhedey: Sounds good to me :)

PR #685, opened 2022-12-16
Bugfix: Fix import macro

talekhinezh:


scrypto-derive/src/import.rs
```diff
@@ -343,13 +343,13 @@ fn get_native_type(ty: &SchemaType) -> Result<(Type, Vec<Item>)> {
...
+        SchemaType::Proof => parse_quote! { ::scrypto::model::Proof},
+        SchemaType::Vault => parse_quote! { ::scrypto::model::Vault},
         SchemaType::Expression => parse_quote! {::scrypto::runtime::Expression },
-        SchemaType::Blob => parse_quote! { ::scrypto::engine_lib::crypto::Blob},
+        SchemaType::Blob => parse_quote! { ::scrypto::model::Blob},
         SchemaType::NonFungibleAddress => {
-            parse_quote! { ::scrypto::engine_lib::resource::NonFungibleAddress}
+            parse_quote! { ::scrypto::model::NonFungibleAddress}
         }
         SchemaType::Hash => parse_quote! { ::scrypto::crypto::Hash},
```
dhedey, line 354: I don't believe `::scrypto::crypto::` hash exists either...

Perhaps we should create it as:
```
/// Scrypto crypto types.
pub mod crypto {
    pub use radix_engine_interface::crypto::*;
}
```

dhedey (changes requested):
Still a few things left to fix.

Ideally we'd have a test which tests construction of everything, but given that we'll be adapting how schema works at some point soon, maybe it's not worth it.

dhedey (approved):
Approving, pending your fix - to allow for an easy merge.

scrypto-derive/src/import.rs
talekhinezh: Nice catch!

PR #686, opened 2022-12-17
feat: Improved Encode/Decode derive to work with generics

dhedey:
Small, independent change that may as well be merged in separately from the type schema work.

Basically it enables you to do this now - with a generic type parameter like `T`:

```
#[derive(Debug, Clone, PartialEq, Eq, Default, TypeId, Encode, Decode)]
pub struct NumericValidation<T> {
    pub min: Option<T>,
    pub max: Option<T>,
}
```

And the macro now expands to the following - which includes the bound `T: ::sbor::Encode<X, E>`
```
impl<T: ::sbor::Encode<X, E>, E: ::sbor::encoder::Encoder<X>, X: ::sbor::type_id::CustomTypeId>
    ::sbor::Encode<X, E> for NumericValidation<T>
{
    #[inline]
    fn encode_type_id(&self, encoder: &mut E) -> Result<(), ::sbor::EncodeError> {
        encoder.write_type_id(::sbor::type_id::SborTypeId::Tuple)
    }
    #[inline]
    fn encode_body(&self, encoder: &mut E) -> Result<(), ::sbor::EncodeError> {
        use ::sbor::{self, Encode};
        encoder.write_size(2)?;
        encoder.encode(&self.min)?;
        encoder.encode(&self.max)?;
        Ok(())
    }
}
```

I've also neatened up the default generic parameter idents to be X/D/E instead of CTI/DEC/ENC... and implemented automatic clash-avoidance so that they automatically get numbers to avoid clashes with other generic parameters if required - see test cases :)

### DIFF
--- a/sbor-derive/src/decode.rs
+++ b/sbor-derive/src/decode.rs
@@ -41,7 +41,7 @@ pub fn handle_decode(input: TokenStream) -> Result<TokenStream> {
                         #[inline]
                         fn decode_body_with_type_id(decoder: &mut #decoder_generic, type_id: ::sbor::SborTypeId<#custom_type_id_generic>) -> Result<Self, ::sbor::DecodeError> {
                             use ::sbor::{self, Decode};
-                            decoder.check_preloaded_type_id(type_id, ::sbor::type_id::SborTypeId::Tuple)?;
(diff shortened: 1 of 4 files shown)


iamyulong (approved):
Looks good

PR #687, opened 2022-12-17
Dedicated instructions for Resource creation and minting

0xOmarA:


radix-engine-interface/src/model/package/access_rules.rs
```diff
@@ -0,0 +1,31 @@
+use crate::api::types::{AccessRuleKey, MetadataMethod, NativeFn, NativeMethod, PackageMethod};
+use crate::scrypto;
+use sbor::*;
+
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone)]
+#[scrypto(TypeId, Encode, Decode)]
+pub enum PackageMethodAuthKey {
```
talekhinezh, line 7: Though this does make this interface temporarily cleaner I think this will need to get refactored when we implement "Associated Nodes". So can we revert this for now?

radix-engine/src/model/transaction_processor/executables.rs
```diff
@@ -609,31 +609,20 @@ impl TransactionProcessor {
                     })
                     .map(|rtn| IndexedScryptoValue::from_typed(&rtn))
                     .map_err(InvokeError::Downstream),
-                Instruction::Basic(BasicInstruction::PublishPackageWithOwner {
```
talekhinezh, line 612: I think we still want to keep the manifest instruction with owner. This was the original reason we had to have the `*WithOwner` invocation in the first place.

radix-engine/src/model/transaction_processor/executables.rs
```diff
@@ -645,17 +634,19 @@ impl TransactionProcessor {
                         }
                         Ok(result)
                     }),
-                Instruction::Basic(BasicInstruction::CreateResourceWithOwner {
```
talekhinezh, line 648: Same thing here.

talekhinezh (changes requested):

radix-engine/src/model/transaction_processor/executables.rs
0xOmarA: Happy to provide some insight into this. The two main reasons as to why I made this change are:

1. I find it awkward that `*WithOwner` instructions take a `NonFungibleAddress` for the owner field. It feels like this is remnants of the old model where things could have an owner and that owner was some system-created `NonFungible`. In this new model, the choice to restrict an owner to non-fungibles seems rather strange to me. 
2. With the concept of "owner" now gone, it seems strange that the radix-engine still needs to know about it (in the tx processor)

We could add them back, however, I imagine that a better way would be to say that the owner is some `AccessRule` (not `AccessRules`) instead of being strictly a `NonFungible`.

radix-engine-interface/src/model/package/access_rules.rs
0xOmarA: Sounds good, will revert this change.

radix-engine/src/model/transaction_processor/executables.rs
talekhinezh: I don't disagree but we should re-visit this with Product as `*WithOwner` with `NonFungibleAddress` was the initial agreement we had.

0xOmarA: Sounds good, will add this back.

talekhinezh (approved):

PR #689, opened 2022-12-19
Feature: Implement From<ComponentAddress>

talekhinezh:


scrypto-derive/src/blueprint.rs
```diff
@@ -499,11 +501,55 @@ fn generate_stubs(
...
+            fn claim_royalty(&self) -> Bucket {
+                self.component.claim_royalty()
+            }
+            fn access_rules_chain(&self) -> Vec<ComponentAccessRules> {
+                self.component.access_rules_chain()
+            }
+        }
+
+        impl #component_ref_ident {
+            #(#methods)*
```
0xOmarA, line 551: Is `#(#functions)*` intentionally left out here?

0xOmarA (approved):

0xOmarA:
Looks good to me. Left one minor comment.

I think that at some point in the future, it might be worth unifying the component struct wrappers generated by the different macros. Right now what we have is:

* `${name}GlobalComponentRef` for the `blueprint` macro.
* `${name}` for the `import` macro.
* `${name}` for the `external_component` macro.

Additionally, the different component struct wrappers currently provide varying degrees of functionality (aside from the most common one - i.e. being a wrapper for calls)

* The wrapper generated by the `blueprint` macro implements `GlobalComponent` and `LocalComponent` on the appropriate structs and therefore allows for metadata to be updated, royalties to be set, etc...
* The wrappers generated by the `import` and `external_component` macros do not implement `GlobalComponent` (despite wrapping a `ComponentAddress`) which means that they do not have the methods from that trait.

scrypto-derive/src/blueprint.rs
talekhinezh: Yup. Didn't want to add all stubs to the global version.

talekhinezh:
> Looks good to me. Left one minor comment.
> 
> I think that at some point in the future, it might be worth unifying the component struct wrappers generated by the different macros. Right now what we have is:
> 
> * `${name}GlobalComponentRef` for the `blueprint` macro.
> * `${name}` for the `import` macro.
> * `${name}` for the `external_component` macro.
> 
> Additionally, the different component struct wrappers currently provide varying degrees of functionality (aside from the most common one - i.e. being a wrapper for calls)
> 
> * The wrapper generated by the `blueprint` macro implements `GlobalComponent` and `LocalComponent` on the appropriate structs and therefore allows for metadata to be updated, royalties to be set, etc...
> * The wrappers generated by the `import` and `external_component` macros do not implement `GlobalComponent` (despite wrapping a `ComponentAddress`) which means that they do not have the methods from that trait.
> 
> I do not imagine that we will get to the above by betanet, however, I think that we should keep it in mind.

Yup totally agree. I did try and clean this up in this PR actually using a `Own<MyComponent>` vs. `Ref<MyComponent>` structure but found that there is a bit more underlying plumbing that's required to get this to work.

PR #691, opened 2022-12-19
feat: staged substate store and execution cache

malexandru-rdx:
This PR implements a staged substate store using an immutable (persistent) store (im::hashmap) in order to facilitate implementation of a performant execution cache.


radix-engine/src/state_manager/staging.rs
```diff
@@ -1,189 +1,581 @@
...
 
-impl<'t, 's, S: ReadableSubstateStore> WriteableSubstateStore for StagedSubstateStore<'t, 's, S> {
-    fn put_substate(&mut self, substate_id: SubstateId, output: OutputValue) {
-        if self.id == 0 {
-            panic!("Cannot write to root store");
-        } else {
-            let node = self.stores.nodes.get_mut(&self.id).unwrap();
-            node.outputs.insert(substate_id, output);
+impl<'t, S: ReadableSubstateStore> ReadableSubstateStore for StagedSubstateStore<'t, S> {
+    fn get_substate(&self, substate_id: &SubstateId) -> Option<OutputValue> {
```
talekhinezh, line 307: Does this fail if a substate is deleted in the child but not in the root?

radix-engine/src/state_manager/staging.rs
```diff
@@ -1,189 +1,581 @@
-use crate::ledger::*;
+use crate::transaction::TransactionReceipt;
 use crate::types::*;
-use radix_engine_interface::api::types::SubstateId;
+use crate::{ledger::*, transaction::TransactionResult};
 
-/// Nodes form an acyclic graph towards the parent
-struct StagedSubstateStoreNode {
```
talekhinezh, line 6: Can we retain old SubstateManager in order to be able to compare/benchmark between the two implementations?

radix-engine/src/state_manager/staging.rs
```diff
@@ -1,189 +1,581 @@
...
-    fn merge_to_parent_recurse(&mut self, id: u64, remove_children: bool) {
-        if id == 0 {
-            if remove_children {
-                self.remove_children(0);
-            }
-            return;
+    /// Recursively deletes all nodes that are not in new_root_key subtree and returns the
+    /// sum of weights from current root to new_root_key. Updates to ImmutableStore on this
+    /// path will persist even after deleting the nodes.
+    fn delete_recursive<V: StagedSubstateStoreVisitor>(
```
talekhinezh, line 223: Probably better to use `V: FnMut(&StagedSubstateStoreNodeKey)`?

radix-engine/src/state_manager/staging.rs
malexandru-rdx: What do you mean by failing? Right now there is no implementation for deleting and the StagedStore reflects that. When we will have delete supported by the root store, this is easily fixed here by adding a Sentinel/Tombstone value.

radix-engine/src/state_manager/staging.rs
malexandru-rdx: Sure, should I keep it under same name and rename the new one? Also it's not really quite the same since this one is basically an immutable store that supports branching natively while the old implementation doesn't.

radix-engine/src/state_manager/staging.rs
talekhinezh: Ah cool kk. Can you add a comment regarding to implement Tombstoning here when required?

radix-engine/src/state_manager/staging.rs
talekhinezh: We could just rename the old one to something like "DeprecatedStagingStore"

radix-engine/src/state_manager/staging.rs
```diff
@@ -1,189 +1,581 @@
...
+        match self.key {
+            StagedSubstateStoreKey::RootStoreKey => self.manager.root.get_substate(substate_id),
+            StagedSubstateStoreKey::InternalNodeStoreKey(key) => {
+                match self
+                    .manager
+                    .nodes
+                    .get(key)
+                    .unwrap()
+                    .store
+                    .get_substate(substate_id)
```
iamyulong, line 317: It seems that we do not merge the updates from the receipt into the store in `new_child_node`. Are we sure  this will return the correct substate? Suggest unit testing.

radix-engine/src/state_manager/staging.rs
```diff
@@ -1,189 +1,581 @@
...
+impl ImmutableStore {
+    fn new() -> Self {
+        ImmutableStore {
+            outputs: ImmutableHashMap::new(),
+        }
+    }
+
+    fn from_parent(parent: &ImmutableStore) -> Self {
+        ImmutableStore {
+            outputs: parent.outputs.clone(),
```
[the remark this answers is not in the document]
malexandru-rdx: It's not really independent. I will add a comment to make this more clear. `outputs` is an ImmutableStore which is implemented with an im::HashMap which is immutable (when `clone`-ing, really under the hood, only the root node is cloned), this is done specifically to not copy and instead reuse as much as possible. As we update the store, only the nodes that would be normally written are cloned (keeping the history of the store as well).

radix-engine/src/state_manager/staging.rs
```diff
@@ -1,189 +1,581 @@
...
+            receipt,
+            store,
+        }
+    }
+
+    /// Weight is defined as the number of changes to the ImmutableStore
+    /// done exclusively by this node.
+    fn weight(&self) -> usize {
+        match &self.receipt.result {
+            TransactionResult::Commit(commit) => commit.state_updates.up_substates.len(),
```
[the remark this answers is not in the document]
malexandru-rdx: Currently `down_substates` are not taken into account when applying to a store (i.e no `fn down_substate`). Once the real store supports delete, this needs to change to reflect that as well. Adding comments where modifications are needed at that point.

radix-engine/src/state_manager/staging.rs
malexandru-rdx: We do in `StagedSubstateStoreNode::new` called by `StagedSubstateStoreManager::new_child_node`. Trying to write unit test.

iamyulong: Sounds good!

radix-engine/src/state_manager/staging.rs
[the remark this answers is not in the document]
iamyulong: Cool cool. Can you add a TODO note there in case we forget?

radix-engine/src/state_manager/staging.rs
malexandru-rdx: I've updated the unit test so that it's easy to modify in terms of the structure that it's building and also to check for ALL staged stores that have been built every time a new one is added.

radix-engine/src/state_manager/staging.rs
```diff
@@ -1,189 +1,581 @@
...
-        }
+        dead_weight
     }
-}
 
-pub struct StagedSubstateStore<'t, 's, S: ReadableSubstateStore> {
-    stores: &'t mut StagedSubstateStoreManager<'s, S>,
-    id: u64,
-}
+    pub fn reparent<V: StagedSubstateStoreVisitor>(
```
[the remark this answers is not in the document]
malexandru-rdx: Done. Need more feedback though :)

talekhinezh (approved):

iamyulong (approved):

PR #676, opened 2022-12-09
Fix/powi overflow and tighter integers

arthurvinci:
This PR solves overflow errors in the `powi` function. The following code would overflow at every line involving  `powi`:
```Rust
    #[test]
    fn test_powi_max_decimal() {
        let _max_bug = Decimal::MAX.powi(1);
        let _max_sqrt = Decimal::MAX.sqrt().unwrap();
        let _max_cbrt = Decimal::MAX.cbrt();
        let _max_dec_2 = _max_sqrt.powi(2);
        let _max_dec_3 = _max_cbrt.powi(3);
    }
    ```
    
    The PR also changes integer used in root function of `Decimal` and `PreciseDecimal` to tighter integers (I384 and I768 as suggested by @r001)

PR #693, opened 2022-12-20
Fix/powi overflow and tighter integers betanet

arthurvinci:
Implemented changes from #676 but coming from betanet branch as suggested by @r001.


r001:
These changes are okay. Thank you for your work.

r001 (approved):
All good.

PR #700, opened 2022-12-23
Feature: Genesis Validator Set

talekhinezh:


radix-engine/src/engine/node_properties.rs
```diff
@@ -17,8 +17,13 @@ impl VisibilityProperties {
...
         node_id: RENodeId,
     ) -> bool {
-        if !mode.eq(&ExecutionMode::Application) {
-            return false;
+        match mode {
+            ExecutionMode::LoggerModule => match node_id {
+                RENodeId::Logger => return true,
+                _ => return false,
+            },
+            ExecutionMode::Application => {}
```
iamyulong, line 25: This fall-through is not read-friendly. Possible to combine with the following match?

iamyulong (approved):
LGTM with minor comments.

radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -42,6 +41,8 @@ pub struct CommitResult {
     pub state_updates: StateDiff,
     pub entity_changes: EntityChanges,
     pub resource_changes: Vec<ResourceChange>,
+    pub application_logs: Vec<(Level, String)>,
```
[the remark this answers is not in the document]
talekhinezh: Yup agreed. I think we'll need to figure out the right abstraction between logging and events as logs for debugging shouldn't necessarily be part of persisted event stream.

radix-engine/src/engine/node_properties.rs
talekhinezh: hrm, it's not a fall-through as this match exits the match block whereas the following match returns immediately. Unless you mean something else?

talekhinezh: ah i got you. Yup sure will update this.

PR #701, opened 2022-12-23
Bugfix: Fix Deterministic hash set sbor

talekhinezh:


0xOmarA (approved):
Looks good. We might want to rerun that one canceled test before merging :)

PR #703, opened 2022-12-26
Change NonFungibleAddress representation in resim to canonical

0xOmarA:
This PR moves the `call_method_with_abi` and `call_function_with_abi` logic from the `ManifestBuilder` to being functions in the simulator crate. Looking at the source code, the simulator crate (and some tests) is the only place that used these functions.

With this change:
1. The `ManifestBuilder` no longer requires a `NetworkDefinition` on instantiation. 
2. We can now make strong assumptions about the environment that the code is run in when the `*_with_abi` functions are called. Since they only exist in the simulator crate, then we know that the code is running in the simulator.

Since we can now assume that all `*_with_abi` calls originate from the simulator, this PR changes how `NonFungibleAddress`es are parsed by resim so that they are now parsed using their canonical format. The type of the `NonFungibleId` is determined through a ledger lookup.


simulator/src/ledger/lookup.rs
```diff
@@ -0,0 +1,91 @@
...
+use radix_engine::types::{
+    GlobalAddress, GlobalOffset, NonFungibleIdType, RENodeId, ResourceAddress, ResourceManagerId,
+    ResourceManagerOffset, ResourceType, SubstateId, SubstateOffset,
+};
+use radix_engine::wasm::DefaultWasmEngine;
+use radix_engine_stores::rocks_db::RadixEngineDB;
+
+use crate::resim::get_data_dir;
+
+pub fn lookup_id_type(
```
iamyulong, line 13: Minor: suggest renaming to `lookup_non_fungible_id_type`

simulator/src/ledger/lookup.rs
```diff
@@ -0,0 +1,91 @@
...
+            "A global resource address can not point to anything other than a resource manager"
+        ),
+    };
+
+    // Reading the resource manager substate from the substate store
+    let node_id = RENodeId::ResourceManager(resource_manager_id);
+    let offset = SubstateOffset::ResourceManager(ResourceManagerOffset::ResourceManager);
+    let substate_id = SubstateId(node_id, offset);
+    let resource_manager_substate = substate_store.get_substate(&substate_id).map_or(
+        Err(LedgerLookupError::ResourceManagerNotFound(
```
iamyulong, line 44: Should panic due to illegal state?

simulator/src/resim/addressing.rs
```diff
@@ -147,11 +149,44 @@ impl FromStr for SimulatorNonFungibleAddress {
     type Err = ParseNonFungibleAddressError;
 
     fn from_str(address: &str) -> Result<Self, Self::Err> {
```
iamyulong, line 151: Possibly some unit test to show case the syntax?

simulator/src/utils/prepare_instruction.rs
```diff
@@ -0,0 +1,519 @@
...
+
+    /// Account is required but not provided.
+    AccountNotProvided,
+}
+
+impl From<BuildArgsError> for BuildCallWithAbiError {
+    fn from(error: BuildArgsError) -> Self {
+        Self::FailedToBuildArgs(error)
+    }
+}
```
iamyulong, line 519: This should've been done a long time ago. Would appreciate it if we can add some unit tests for the supported types here.

iamyulong (approved):
LGTM with minor comments.

simulator/src/ledger/lookup.rs
0xOmarA: Yes you're right. If the global address was found then the resource manager should be found as well. will change this to a panic.

simulator/src/utils/prepare_instruction.rs
0xOmarA: Sounds good, will do that!

simulator/src/resim/addressing.rs
0xOmarA: Will add that!

0xOmarA: I have added that. Here is what the syntax for `NonFungibleAddress`es in resim looks like now: https://github.com/radixdlt/radixdlt-scrypto/blob/8ebe250c4ba3ec33b6bcc3a2aab014949bffe50a/simulator/src/utils/prepare_instruction.rs#L853

PR #706, opened 2023-01-02
use cargo nextest in CI - DO NOT MERGE

lrubasze:
for tests only


iamyulong:
While you're on this, I think the windows job slowdown might be related to caching not working properly. It compile `rocksdb` every execution, which takes about 7 plus minutes each time.

We previously appplied this change https://github.com/radixdlt/radixdlt-scrypto/blob/ff6b6312481c8ee2ad1b989aecff83ffc1e60859/.github/workflows/ci.yml#L281 to make it work for ubuntu.

Might need something similar for windows.

PR #711, opened 2023-01-03
Feature: Virtual package badge

talekhinezh:
* Add concept of "native packages" such that every invocation (scrypto+native) has a package identifier associated with it, imilar to the Rust "crate" concept.
* Add virtual "package token" which is included as a virtual proof in the AuthZone, one may now only allow certain functions/methods to be called only if in the same package, similar to Rust  'pub(crate)' visibility flags.


radix-engine-interface/src/api/types/actor.rs
```diff
@@ -6,23 +6,58 @@ use crate::model::*;
...
+    Component,
+    Package,
+    Metadata,
+    EpochManager,
+    Resource,
+    Clock,
+    Logger,
+    TransactionRuntime,
+    TransactionProcessor,
+}
```
iamyulong, line 29: They seem to be blueprint equivalent rather than package, no?

radix-engine-interface/src/api/types/actor.rs
```diff
@@ -119,7 +144,7 @@ impl Into<FnIdentifier> for NativeFunction {
 )]
 #[scrypto(TypeId, Encode, Decode, Describe)]
 #[strum(serialize_all = "snake_case")]
-pub enum AccessRulesChainMethod {
+pub enum AccessRulesChainFn {
```
iamyulong, line 147: Haha, another split merge cycle! The resulting code appears to be cleaner.

radix-engine/src/engine/modules/auth_module.rs
```diff
@@ -282,13 +231,66 @@ impl AuthModule {
...
+        let auth_zone_id = refed
+            .into_iter()
+            .find(|e| matches!(e, RENodeId::AuthZoneStack(..)))
+            .unwrap();
+        call_frame_update.node_refs_to_copy.insert(auth_zone_id);
+
+        if !matches!(
+            actor.identifier,
+            FnIdentifier::Native(NativeFn::AuthZoneStack(..))
+        ) {
```
iamyulong, line 249: For my understanding, why not auth zone stack actor?

radix-engine-interface/src/api/types/actor.rs
talekhinezh: Should be package, for example, `Resource` maintains all Vaults/Bucket/Proof logic.

radix-engine/src/engine/modules/auth_module.rs
talekhinezh: This is self-reference problem. From what I recall, it's easiest to manage self-reference at this layer e.g. so that we can directly reference the last AuthZone in the stack when needing to access the caller's AuthZone rather than always needing to second to last AuthZone.

radix-engine-interface/src/api/types/actor.rs
iamyulong: Cool, fine with me for now. I'm just a bit concerned if we really need divide native nodes into multiple packages. It seems to me that we may need a "blueprint" virtual badge, on top of package badge, to provide finer granularity in access control.

iamyulong (approved):

radix-engine-interface/src/api/types/actor.rs
talekhinezh: Yup, good point but I'm assuming we won't ever need a "blueprint" badge. The main value of such a badge is to not allow outside code apart from the "owner" of the code to be able to make certain calls. For internal protection, one could just use rust separation of files to protect the code owner from himself.

PR #727, opened 2023-01-06
Feature/vault in royalty accumulator

iamyulong:
Adds a vault in package and component royalty accumulator substate.


radix-engine/src/model/royalty/module.rs
```diff
@@ -34,6 +34,24 @@ impl Default for RoyaltyModule {
     }
 }
 
+macro_rules! pre_load_vault {
```
talekhinezh, line 37: Can we add TODO to remove this at some point, definitely a code smell I think that we require this.

radix-engine/tests/transaction_commit_rollback.rs
```diff
@@ -30,10 +30,10 @@ fn test_state_track_success() {
     // Assert
     receipt.expect_commit_success();
     assert_eq!(
-        22,
+        25,
```
talekhinezh, line 33: This feels like a fragile (non-resistant to refactoring) test. Can we just remove this assertion or replace with something more resistant to refactoring?

talekhinezh (approved):

radix-engine/src/model/royalty/module.rs
iamyulong: Yes!

radix-engine/tests/transaction_commit_rollback.rs
iamyulong: Definitely.

PR #729, opened 2023-01-07
Architecture: Worktop buckets

talekhinezh:
* Replace use of `Resource` in `Worktop` with `Bucket`


iamyulong (approved):
LGTM. Thanks for cleaning up worktop!

PR #732, opened 2023-01-10
Feature: Some Non-fungible Id Cleanup

talekhinezh:
* Type checked ids of NonFungible Resource creation by using `NonFungibleType` trait
* Remove `U32` and `U64` types of NonFungibleId and replaced with `Number` which is currently just a `u64`
* Separated out invocations for `UUID` NonFungibleType which auto-generates UUIDs rather than requiring user to specify


radix-engine/src/model/resources/executables/resource_manager.rs
```diff
@@ -506,6 +515,15 @@ impl Executor for ResourceManagerCreateNonFungibleWithInitialSupplyInvocation {
...
         let resource_address: ResourceAddress = global_node_id.into();
 
+        // TODO: Do this check in a better way (e.g. via type check)
+        if self.id_type == NonFungibleIdTypeId::UUID {
+            return Err(RuntimeError::ApplicationError(
+                ApplicationError::ResourceManagerError(
+                    ResourceManagerError::InvalidNonFungibleIdTypeId,
+                ),
+            ));
+        }
```
iamyulong, line 525: Why don't we allow user-defined UUID type?

radix-engine/src/model/resources/executables/resource_manager.rs
```diff
@@ -1031,15 +1129,52 @@ impl Executor for ResourceManagerMintNonFungibleExecutable {
...
+                }
+            };
+
+            if id_type == NonFungibleIdTypeId::UUID {
+                return Err(RuntimeError::ApplicationError(
+                    ApplicationError::ResourceManagerError(
+                        ResourceManagerError::InvalidNonFungibleIdTypeId,
+                    ),
+                ));
+            }
```
iamyulong, line 1150: Same

scrypto/src/resource/resource_builder.rs
```diff
@@ -22,8 +23,8 @@ impl ResourceBuilder {
     }
 
     /// Starts a new builder to create non-fungible resource.
-    pub fn new_non_fungible(non_fungible_id_type: NonFungibleIdType) -> NonFungibleResourceBuilder {
-        NonFungibleResourceBuilder::new(non_fungible_id_type)
+    pub fn new_non_fungible<Y: NonFungibleIdType>() -> NonFungibleResourceBuilder<Y> {
```
iamyulong, line 26: This is nice!

iamyulong (approved):
As discussed offline, if we intend to rely on UUID for uniqueness, we need to evluate the security impact of possible collision of `u128`.

PR #733, opened 2023-01-10
Feature/reduce blob scope

iamyulong:
Reduces the scope of the blob to transaction only, so kernel no longer needs to manage the lifecycle of blobs.


radix-engine/src/model/transaction_processor/executables.rs
```diff
@@ -251,6 +256,11 @@ impl<'a> Executor for TransactionProcessorRunInvocation<'a> {
             RENode::TransactionRuntime(runtime_substate),
         )?;
 
+        let mut blobs_by_hash = HashMap::new();
+        for blob in self.blobs.as_ref() {
+            blobs_by_hash.insert(hash(blob), blob);
```
dhedey, line 261: Hashing's quite expensive - what are your thoughts on preventing any blob hashing / blob commands till after the fee loan has been paid back? (and calculating the hashes lazily on first use).

Not in this PR necessarily 👍

dhedey: That said, I guess we don't delay sig verification... although hashing 1MB is possibly slower.

radix-engine/src/model/transaction_processor/executables.rs
```diff
@@ -433,9 +443,24 @@ impl<'a> Executor for TransactionProcessorRunInvocation<'a> {
...
+                            ),
+                        ))?;
+                    let abi = blobs_by_hash
+                        .get(&abi.0)
+                        .ok_or(RuntimeError::ApplicationError(
+                            ApplicationError::TransactionProcessorError(
+                                TransactionProcessorError::BlobNotFound(abi.clone()),
+                            ),
+                        ))?;
+                    // TODO: remove clone by allowing invocation to have references, like in TransactionProcessorRunInvocation.
```
dhedey, line 460: I imagine this should be pretty trivial to change in this PR to save adding a TODO? 👍 

Also - I'm surprised we need a double clone here, but I might be missing something! Can we avoid one with a deref?

radix-engine-interface/src/data/types/manifest_blob.rs
```diff
@@ -0,0 +1,56 @@
...
+use sbor::rust::vec::Vec;
+use sbor::*;
+use utils::copy_u8_array;
+
+use crate::crypto::Hash;
+use crate::data::*;
+use crate::scrypto_type;
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ManifestBlob(pub Hash);
```
dhedey, line 13: What do you think about `ManifestBlobHash` or `ManifestBlobRef` instead? Because it's not technically the blob itself?

dhedey (approved):
Nice changes. A few thoughts / suggestions

radix-engine/src/model/transaction_processor/executables.rs
dhedey: > I imagine this should be pretty trivial to change in this PR to save adding a TODO? 👍

Ahh maybe not. I guess we need to then include lifetime information up the whole tree of instructions etc...
Okay, ignore this :D

iamyulong: Haha, I tried to update. 

The whole invocation enum hierachiy needs to be updated if we're not separating "transient" invocation and seralized invocation. 

There is very little performance loss from my benchmark, so I didn't pursue it for now.

radix-engine/src/model/transaction_processor/executables.rs
iamyulong: That's a good callout. Let me add a TODO comment here.

radix-engine-interface/src/data/types/manifest_blob.rs
iamyulong: Happy to change to `ManifestBlobRef`. Will do as part of this PR.

PR #737, opened 2023-01-10
Feature: Validator Component

talekhinezh:


radix-engine/src/model/epoch_manager/executables.rs
```diff
@@ -416,3 +456,100 @@ impl EpochManager {
...
+        let substate = api.get_ref(handle)?;
+        let validator = substate.validator();
+        let invocation = EpochManagerUpdateValidatorInvocation {
+            receiver: validator.manager,
+            validator_address: validator.address,
+            key: validator.key,
+            register: true,
+        };
+
+        api.invoke(invocation)?;
```
iamyulong, line 503: Answered my previous question.

I wonder if we really need these methods inside the engine though. Now, we need to configure auth in two places.

More like a client-side abstraction.

iamyulong (approved):
LGTM with question about the necessity of `Validator::Register/Unregister`

PR #744, opened 2023-01-12
Feature/refactor scrypto macros

iamyulong:
Changes:
- Removed `#[scrypto()]`
- Renamed derive `Describe` to `LegacyDescribe`
- Added derive `ScryptoEncode`, `ScryptoDecode`, `ScryptoCategorize`, `ScryptoDescribe`

TODO: Add `Sbor` and `ScryptoSbor` in next PR


radix-engine-derive/src/lib.rs
```diff
@@ -12,32 +15,89 @@ use proc_macro::TokenStream;
...
+/// ```ignore
+/// use scrypto::prelude::*;
+///
+/// #[derive(ScryptoDescribe)]
+/// pub struct MyStruct {
+///     pub field_1: u32,
+///     pub field_2: String,
+/// }
+/// ```
+#[proc_macro_derive(ScryptoDescribe, attributes(sbor))]
```
dhedey, line 98: Did you decide not to add `ScryptoSbor` here in the end?

dhedey: I'd also be interested if ScryptoSbor could include both ScryptoDescribe and LegacyDescribe for now, just to check everything compiles okay? :)

radix-engine-interface/src/api/types/actor.rs
```diff
@@ -1,20 +1,15 @@
...
 use crate::api::types::*;
 use crate::model::*;
-use crate::scrypto;
-use crate::Describe;
+use crate::*;
+use sbor::rust::string::String;
 
-#[derive(Debug, Clone, Eq, PartialEq)]
-#[scrypto(Categorize, Encode, Decode)]
+#[derive(Debug, Clone, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
```
dhedey, line 6: Should all of these just default to `ScryptoSbor` unless we have a good reason not to implement Schema? I guess it depends on how it affects compile time.

Perhaps to save wasted work if compile time is awful, we could try cheekily making eg `ScryptoCategorize` also output `ScryptoDescribe` and see how much all the hashing in ScryptoDescribe slows down compilation.

dhedey (approved):
Really nice work, thanks :D

radix-engine-derive/src/lib.rs
iamyulong: I will add in another PR, as I considered these "combo" derives as new feature.

radix-engine-interface/src/api/types/actor.rs
iamyulong: I don't think it would save compile time, because the whole macro package is combined into a shared libraray. The only difference is ONE or TWO external function calls.

dhedey: I meant more whether adding the new `Describe` trait to all of the types in the system causes too much slow down in compilation because of its internal uses of `const_sha1` (which can't be very efficient!). Hopefully not :).

I expect we'll want `Describe` on many types anyway so they can form parts of wider schemas (eg the schema for `Instruction`)

radix-engine-derive/src/lib.rs
dhedey: Ahh sorry, missed that part of the description :)

PR #750, opened 2023-01-14
Fix: indexmap and indexset now work happily with no-std

dhedey:
IndexMap and IndexSet now work without std being present.

Note that you should use the `index_map_new()` method I added (or `IndexMap::default()`) instead of `IndexMap::new()`; and similarly for `IndexSet` There are complicated reasons for why `new` doesn't work when you override the default type - see this thread for details: https://github.com/bluss/indexmap/pull/207#issuecomment-1382623923

Note: If you use `new` then it gives you a compile error in CI now, when run with the `no-std` / `alloc` feature - previously you'd only get an error if this weird [build.rs script](https://github.com/bluss/indexmap/blob/indexmap-1.x/build.rs) detected no std was actually present (eg WASM builds or something).

This PR also:
* Makes it easier to use `use sbor::rust::collections::*` to pull in all the types and macros you desire
* Fixes a minor bug from the `Map` change and how it interfaces with the `Describe` impl (it was still pulling in `(K, V)` as the dependent type instead of `K` and `V` separately, which was adding an extra type to the schema unnecessarily)


iamyulong:
Unfortunately, it's still not fixed.  If you turn on `sbor/indexmap` in `example/no-std` and run `scrypto build`, you will see compilation error.

```
the lang item is first defined in crate `std` (which `indexmap` depends on)
```  

`std` has been introduced by some implementation of the `indexmap` crate.

Look forward to seeing if you can identify the root cause.

dhedey:
Ohh... Looks like the `autocfg` in their build script in 1.x claims no prisoners: https://github.com/bluss/indexmap/blob/indexmap-1.x/build.rs

Thankfully, they've removed this and replaced it with the more standard `std` feature for their 2.0 release, but it's not quite out yet. However, we can happily pull it in via github at a fixed revision, which is what I've done. (see here for their v2 roadmap https://github.com/bluss/indexmap/issues/135 )

iamyulong (approved):

sbor/Cargo.toml
```diff
@@ -5,15 +5,19 @@ edition = "2021"
 
 [dependencies]
 sbor-derive = { path = "../sbor-derive" }
-hashbrown = { version = "0.12.1", optional = true }
+hashbrown = { version = "0.13.2", optional = true } # Align with the version indexmap pulls in to reduce the dependency tree
```
iamyulong, line 8: This comment is slightly misleading, as cargo select versions to de-duplicate.

dhedey: Based on the simulator's lock file, it seemed to do that with patch versions, but not with minor versions -- but maybe I just needed to delete it and recreate it 🤷

PR #752, opened 2023-01-15
feat: Transaction can abort when fee loan paid back

dhedey:
Transaction can abort when fee loan paid back.

This feature can be used by the node at mempool submission time, to reduce work done.

It also sets an easy attach-point for other kinds of abortion:
* Abort due to reaching some kind of execution/state limit at a block level (note - the transaction itself isn't necessarily invalid, it just won't fit in the block)
* Abort due to a requested abort from another thread (eg via a `abort_flag` AtomicBool which is checked when cost units are consumed) - this could be used for a fail-safe abort in the node if transaction execution takes too long.

This PR also includes a mini refactor to `InvokeError` to make it easier to use.


radix-engine/src/fee/fee_reserve.rs
```diff
@@ -209,6 +239,25 @@ impl SystemLoanFeeReserve {
 }
 
 impl FeeReserve for SystemLoanFeeReserve {
+    fn consume_deferred(
```
iamyulong, line 242: Why do need a dedicated method for this wit duplicated code?

radix-engine/src/engine/errors.rs
```diff
@@ -161,25 +206,76 @@ pub enum ModuleError {
...
-pub enum InvokeError<E> {
+pub enum InvokeError<E: InvokeWrappedError> {
     Error(E),
     Downstream(RuntimeError),
 }
 
-impl<E> From<RuntimeError> for InvokeError<E> {
+/// This is a trait for the non-Downstream part of [`InvokeError`]
+/// We can't use `Into<RuntimeError>` because we need `RuntimeError` _not_ to implement it.
+pub trait InvokeWrappedError {
```
iamyulong, line 256: How about renaming to `NonDownstreamError` or `SelfError`?

radix-engine/src/fee/fee_reserve.rs
dhedey: Because:
* At one point, I wanted to give them different semantics -- ie create a precondition that you weren't allowed to call `consume_deferred` after any calls to `consume` -- before realising it wasn't quite necessary. But I preferred the clarity of the API of the separate methods so kept it.
* The true/false flag was a bit confusing  - it seemed to suggest that it was a choice. But `consume_deferred` is not called from inside the engine after original set-up, and really, shouldn't be. 
* `consume` is on a hot code path so reducing the logic in the method is reasonable
* Boolean parameters also are best avoided as a very general rule: https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3 - and `consume` and `consume_deferred` didn't actually share much code

radix-engine/src/engine/errors.rs
dhedey: Sounds good! I struggled with finding a good name

radix-engine/src/fee/fee_reserve.rs
iamyulong: Is it just because of missing documentation? 

I'm more worried that we might forget to update one place when updating another, especially for anyone who's new to codebase. 

Not convinced we would save much from a single branching instruction.

dhedey: I think the methods are fundamentally different from an abstraction perspective - compare `consume_deferred`:
```
        if amount == 0 {
            return Ok(());
        }

        let units_consumed = checked_multiply(amount, multiplier)?;

        checked_assign_add(
            self.execution_deferred.entry(reason).or_default(),
            units_consumed,
        )?;

        Ok(())
```

With `consume_execution`:

```
        if amount == 0 {
            return Ok(());
        }

        let units_consumed = checked_multiply(amount, multiplier)?;

        self.consume(units_consumed, self.execution_price())?;
        checked_assign_add(self.execution.entry(reason).or_default(), units_consumed)?;

        if self.cost_units_consumed >= self.check_point && !self.fully_repaid() {
            self.repay_all()?;
        }

        Ok(())
```

They're doing very different things - changing different entries, and having different logic around checking for repayment.

iamyulong: To me, they both concern `execution` metering, with different time delays. Function bodies are implementation details.

iamyulong: This is why we want `Decimal::round(mode)` rather than `round_to_zero`, `round_up` and `round_down`.

iamyulong: As discussed, @dhedey's motivation is to disallow clients call `consume_deferred` during execution time, which otherwise may cause illegal state.

He is going to try to split the methods into two traits, to prevent kernel from accidentally calling it.

iamyulong (approved):
LGTM. Thanks for hardening the `FeeReserve` interface!

PR #755, opened 2023-01-17
make transaction receipts clone-able

malexandru-rdx:


radix-engine/src/model/transaction_processor/executables.rs
```diff
@@ -46,10 +47,12 @@ pub enum TransactionProcessorError {
     ResolveError(ResolveError),
 }
 
-pub trait NativeOutput: ScryptoEncode + Debug {}
-impl<T: ScryptoEncode + Debug> NativeOutput for T {}
+pub trait NativeOutput: ScryptoEncode + Debug + DynClone {}
+impl<T: ScryptoEncode + Debug + Clone> NativeOutput for T {}
 
-#[derive(Debug)]
+dyn_clone::clone_trait_object!(NativeOutput);
```
0xOmarA, line 53: I would recommend doing the following here instead of using `DynClone`. `InstructionOutput` can be made clonable on the virtue that objects that implement it also implement `ScryptoEncode`. Alternatively, for a more efficient solution, we might want to use an `enum` for `NativeOutput` instead of it being a trait..

```suggestion
impl Clone for InstructionOutput {
    fn clone(&self) -> Self {
        match self {
            InstructionOutput::Scrypto(output) => InstructionOutput::Scrypto(output.clone()),
            InstructionOutput::Native(output) => {
                // SBOR Encode the output
                let encoded_output = scrypto_encode(&**output)
                    .expect("Impossible Case! Instruction output is not SBOR encodable!");

                // Decode to a ScryptoValue
                let decoded = scrypto_decode::<ScryptoValue>(&encoded_output)
                    .expect("Impossible Case! We literally just encoded this above");

                InstructionOutput::Native(Box::new(decoded))
            }
        }
    }
}
```

malexandru-rdx: That's a nice solution, Omar. Thanks a lot :D. I've updated it.

0xOmarA (approved):

PR #756, opened 2023-01-17
Radix Engine WASM Interface Overhaul

iamyulong:
Updates:
- Replaced `RadixEngineInput`/`radix_engine` with more granular low-level WASM API
- Removed `ScryptoBuffer`, `scrypto_alloc` and `scrypto_free`
- Added Engine-side buffer pool, which can be pulled from Scrypto

New Scrypto function export signatures:
```rust
pub type BufferId = u32;
pub type Buffer = u64; // (id, len)
pub type Slice = u64; // (ptr, len)

extern "C"  {
    // component_id only if it's a method
    pub fn Test_x(component_id: Buffer, args: Buffer) -> Slice;
    
    pub fn Test_abi() -> Slice;
}
```

New engine WASM API (for `wasm32-unknown-unknown`):
```rust
extern "C" {
    //===============
    // Buffer API
    //===============

    /// Consumes a buffer by copying the contents into the specified destination.
    pub fn consume_buffer(buffer_id: BufferId, destination_ptr: *mut u8);

    //===============
    // Invocation API
    //===============

    /// Invokes a method on a component.
    pub fn invoke_method(
        receiver_ptr: *const u8,
        receive_len: usize,
        ident_ptr: *const u8,
        ident_len: usize,
        args_ptr: *const u8,
        args_len: usize,
    ) -> Buffer;

    /// Invokes any function, either scrypto or native.
    pub fn invoke(invocation_ptr: *const u8, invocation_len: usize) -> Buffer;

    //===============
    // Node API
    //===============

    /// Creates a node with the given initial data.
    pub fn create_node(node_ptr: *const u8, node_len: usize) -> Buffer;

    /// Retrieves IDs of visible nodes.
    pub fn get_visible_nodes() -> Buffer;

    /// Destroys a node.
    pub fn drop_node(node_id_ptr: *const u8, node_id_len: usize);

    //===============
    // Substate API
    //===============

    // Locks a substate
    pub fn lock_substate(
        node_id: *const u8,
        node_id_len: usize,
        offset: *const u8,
        offset_len: usize,
        mutable: bool,
    ) -> u32;

    // Reads a substate
    pub fn read_substate(handle: u32) -> Buffer;

    // Writes into a substate
    pub fn write_substate(handle: u32, data_ptr: *const u8, data_len: usize);

    // Releases a lock
    pub fn unlock_substate(handle: u32);

    //===============
    // Actor API
    //===============

    // Returns the current actor.
    pub fn get_actor() -> Buffer;
}
```


radix-engine/src/engine/interpreters/wasm_runtime.rs
```diff
@@ -20,23 +17,23 @@ where
     Y: SystemApi + EngineApi<RuntimeError> + Invokable<ScryptoInvocation, RuntimeError>,
 {
     api: &'y mut Y,
+    buffers: BTreeMap<BufferId, Vec<u8>>,
```
talekhinezh, line 20: What are your thoughts on this being managed by the runtime vs. kernel? I have been moving towards the direction where the Runtime/Intepreter were as stateless as possible.

radix-engine-interface/src/api/wasm.rs
```diff
@@ -0,0 +1,49 @@
...
+    Into<CallTableInvocation> + Invocation<Output = Self::ScryptoOutput>
+{
+    type ScryptoOutput: ScryptoDecode;
+}
+
+pub type BufferId = u32;
+pub type Buffer = u64;
+
+#[macro_export]
+macro_rules! buffer_id {
```
talekhinezh, line 15: Should these be inline functions rather than macros instead to preserve type checking?

radix-engine/src/engine/interpreters/wasm_runtime.rs
iamyulong: So far, it's only used and needed for WASM shim (between kernel/interpreter and wasm module). It seems wrong to move the abstraction to kernel.

In terms stateless, `WasmRuntime` is alredy stateful (the behavior depends on the current state `api`). I don't see a harm for components to maintain well-defined internal state.

radix-engine-interface/src/api/wasm.rs
iamyulong: There is on way to define methods on `u64`, unless we use `Buffer(u64)` which is not compitable with Rust FFI. It's the lesser evil. :)

iamyulong: Maybe I can manually reproduce C layout and alignment. Let me try.

iamyulong: Applied the change

radix-engine-interface/src/api/wasm.rs
```diff
@@ -0,0 +1,61 @@
...
+
+pub trait SerializableInvocation:
+    Into<CallTableInvocation> + Invocation<Output = Self::ScryptoOutput>
+{
+    type ScryptoOutput: ScryptoDecode;
+}
+
+pub type BufferId = u32;
+
+#[repr(C)]
```
talekhinezh, line 13: Fancy!

talekhinezh (approved):

PR #758, opened 2023-01-18
Rename Non-fungible Related Types

iamyulong:
As agreed by team, we're renaming:
- `NonFungibleAddress` -> `NonFungibleGlobalId`
- `NonFungibleId` -> `NonFungibleLocalId`
- `NonFungibleIdTypeId` -> `NonFungibleIdType`
- `NonFungibleIdType` (trait) -> `IsNonFungibleLocalId`


radix-engine-interface/src/model/resource/non_fungible_local_id.rs
```diff
@@ -15,62 +15,67 @@ use crate::abi::*;
 use crate::data::*;
 use crate::*;
 
-pub trait NonAutoGeneratedNonFungibleIdType: NonFungibleIdType {}
-
-pub trait NonFungibleIdType: Into<NonFungibleId> {
-    fn id() -> NonFungibleIdTypeId;
+/// Trait for converting into a `NonFungibleLocalId` of any type (i.e., Number, String, Bytes and UUID).
+pub trait IntoNonFungibleLocalId: Into<NonFungibleLocalId> {
```
talekhinezh, line 19: I think the `Into` prefix is a bit weird, since this is more than just a standard `Into`. Anything wrong with the original `Type` suffix?

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
```diff
@@ -15,62 +15,67 @@ use crate::abi::*;
...
+    }
+}
+impl IntoNonFungibleLocalId for u128 {
+    fn id_type() -> NonFungibleIdType {
+        NonFungibleIdType::UUID
+    }
+}
+
+/// Trait for converting into a `NonFungibleLocalId` of non-auto-generated types (i.e., Number, String and Bytes).
+pub trait IntoManualNonFungibleLocalId: IntoNonFungibleLocalId {}
```
talekhinezh, line 45: Same comment as above.

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
iamyulong: It's a bit confusing with the type of the non-fungible ID.

What this trait does is to mark a Rust type that **A) can be converted into `NonFungibleLocalId`** and **B) can statically reveal the type of the non-fungible ID that it's going to use**.

How about `NonFungibleLocalIdEquivalent` and `ManualNonFungibleLocalIdEquivalent` or `LikeNonFungibleLocalId` and `LikeManualNonFungibleLocalId`?  Explanation will be like "`u8/u16/u32` is a `NonFungibleLocalId`-equivalent and it converts into a `NonFungibleLocalId::Number(_)`"




_Slightly related to this, David is proposing to rename `NonFungibleIdType` (enum) to `NonFungibleIdKind` for consistency._

talekhinezh: Well, really `NonFungibleIdType` enum is used for serialization/deserialization type checking. We don't really use it as a language construct (closed set of a type). To me the trait version is the more important version. Let's chat about this.

iamyulong: Hah, let's chat about this. I think both are important, but I'd rather keep the nonce for struct/enum rather than trait. :)

PR #759, opened 2023-01-18
Canonical Representation for NonFungibleLocalId and NonFungibleGlobalId 

iamyulong:
```
resource_rdx1qj9rjw9funtw958enuv895euwnv:<my_cool_nft_32423>
resource_rdx1qj9rjw9funtw958enuv895euwnv:#203478274#
resource_rdx1qj9rjw9funtw958enuv895euwnv:[deadbeef]
resource_rdx1qj9rjw9funtw958enuv895euwnv:{00000000-0000-4000-8000-000000000000}
```


radix-engine-interface/src/model/resource/non_fungible_global_id.rs
```diff
@@ -204,56 +159,51 @@ mod tests {
...
-            "resource_sim1qzntya3nlyju8zsj8h86fz8ma5yl8smwjlg9tckkqvrs520k2p:10"
+            "resource_sim1qzntya3nlyju8zsj8h86fz8ma5yl8smwjlg9tckkqvrs520k2p:#123#"
         );
 
         assert_eq!(
             NonFungibleGlobalId::try_from_canonical_string(
                 &dec,
-                NonFungibleIdType::UUID,
-                "resource_sim1qzntya3nlyju8zsj8h86fz8ma5yl8smwjlg9tckkqvrs520k2p:1234567890",
+                "resource_sim1qzntya3nlyju8zsj8h86fz8ma5yl8smwjlg9tckkqvrs520k2p:{1234567890}",
```
dhedey, line 182: I don't think we should support such non-conformant IDs

radix-engine-interface/src/model/resource/non_fungible_id_type.rs
```diff
@@ -0,0 +1,13 @@
+use crate::*;
+use sbor::*;
+
+/// Represents type of non-fungible id
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Encode, Decode, LegacyDescribe,
+)]
+pub enum NonFungibleIdType {
```
dhedey, line 8: Would `NonFungibleIdKind` be more consistent?

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
```diff
@@ -73,177 +67,74 @@ impl From<u128> for NonFungibleLocalId {
...
-/// Represents a key for a non-fungible resource
+/// Represents the local id of a non-fungible.
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum NonFungibleLocalId {
+    String(String),
     Number(u64),
-    UUID(u128),
     Bytes(Vec<u8>),
-    String(String),
+    UUID(u128),
```
dhedey, line 76: I wonder if we should make a UUID actually be a `[u8; 16]`?

Also, in the slack thread, we discussed enforcing that it was _actually_ validated as a UUID, ie that it conforms to the UUID spec for randomly generated UUIDs. I think one example of such an id is this:

```
{00000000-0000-4000-8000-000000000000}
```

NB - I think based on my reading that the UUID is the above, with  the 4 nibble fixed, and the top 2 bits of the 8 are fixed (the bottom 2 are random - ie that hex nibble can be any from 8..b) and the other 0s are all 4-bits of randomness. IE we have (128-4-2 = 122 bits of randomness).

What do you think?

dhedey (commented):
Really nice work, a couple of thoughts about UUIDs which I'd like discussed before merge :)

radix-engine-interface/src/model/resource/non_fungible_global_id.rs
iamyulong: Initially thought the same way, but found `uuid` (crate) is taking a mroe lenient approach, https://docs.rs/uuid/1.2.2/uuid/struct.Uuid.html#method.parse_str
> Any of the formats generated by this module (simple, hyphenated, urn, Microsoft GUID) are supported by this parsing function.

Happy to change it to allow braced form only.

dhedey: I feel it'd probably be better to be stricter, ideally so there's a unique representation.

We may even want to enforce eg lower case - what do you think?

iamyulong: [RFC 4122 Section 3](https://tools.ietf.org/html/rfc4122#section-3) requires that the characters be generated in lower case, while being case-insensitive on input.

I figure we'd limit the format but be case-insensitive, for the parsing part.

dhedey: Sounds reasonable to me 👍

radix-engine-interface/src/model/resource/non_fungible_id_type.rs
iamyulong: Decision reverted. Will keep `NonFungibleIdType` to be consistent with doc.

radix-engine-interface/src/model/resource/non_fungible_global_id.rs
iamyulong: Strengthened validation for textual representation.

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
iamyulong: Now stick to UUID v4, variant 1.  Don't think it's necessary to use `[u8; 16]` which is the network bytes abstraction of UUID.

dhedey: 👍 . Fair enough. It's just that u128 to me suggests it's a number, which it most definitely isn't. But that's really semantics I guess!

radix-engine-interface/src/data/value_serializer.rs
```diff
@@ -940,7 +940,7 @@ mod tests {
                 { "type": "NonFungibleLocalId", "value": "hello" },
                 { "type": "NonFungibleLocalId", "value": "123" },
                 { "type": "NonFungibleLocalId", "value": { "hex": "2345" } },
-                { "type": "NonFungibleLocalId", "value": "371" },
+                { "type": "NonFungibleLocalId", "value": "41635955959434943928015079245517041341" },
```
dhedey, line 943: I think I'd rather the value for all of these is just their string encoded form - what do you think?

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
```diff
@@ -70,55 +70,71 @@ impl From<u128> for NonFungibleLocalId {
...
+                    return Err(ContentValidationError::TooLong);
                 }
                 Ok(())
             }
-            NonFungibleLocalId::Number(_) | NonFungibleLocalId::UUID(_) => Ok(()),
+            NonFungibleLocalId::Integer(_) => Ok(()),
+            NonFungibleLocalId::UUID(v) => {
+                // 0100 - v4
+                // 10 - variant 1
+                if (v & 0x00000000_0000_f000_C000_000000000000u128)
```
dhedey, line 130: Super picky, feel free to ignore - but should we have consistent casing? so `f` and `c` perhaps?

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
```diff
@@ -172,9 +188,9 @@ impl<E: Encoder<ScryptoCustomValueKind>> Encode<ScryptoCustomValueKind, E> for N
                 encoder.write_size(v.len())?;
                 encoder.write_slice(v.as_bytes())?;
             }
-            NonFungibleLocalId::Number(v) => {
+            NonFungibleLocalId::Integer(v) => {
                 encoder.write_byte(1)?;
-                encoder.write_slice(&v.to_le_bytes())?;
+                encoder.write_slice(&v.to_be_bytes())?; // TODO: variable length encoding?
```
dhedey, line 193: Out of interest, why are we using `be_bytes` here but `le_bytes` for all other numeric types in sbor?

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
```diff
@@ -238,32 +254,38 @@ impl FromStr for NonFungibleLocalId {
...
-            )
+            let chars: Vec<char> = s[1..s.len() - 1].chars().collect();
+            if chars.len() == 32 + 4
+                && chars[8] == '-'
+                && chars[13] == '-'
+                && chars[18] == '-'
+                && chars[23] == '-'
+            {
+                let hyphen_stripped: String = chars.into_iter().filter(|c| *c != '-').collect();
+                NonFungibleLocalId::UUID(
```
dhedey, line 275: I think we need to check that hyphen_stripped.len() equals 32, otherwise eg `{----------..--0}` would be valid I think?

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
```diff
@@ -70,55 +70,71 @@ impl From<u128> for NonFungibleLocalId {
...
             NonFungibleLocalId::String(..) => NonFungibleIdType::String,
-            NonFungibleLocalId::Number(..) => NonFungibleIdType::Number,
+            NonFungibleLocalId::Integer(..) => NonFungibleIdType::Integer,
             NonFungibleLocalId::Bytes(..) => NonFungibleIdType::Bytes,
             NonFungibleLocalId::UUID(..) => NonFungibleIdType::UUID,
         }
     }
 
-    pub fn validate_contents(&self) -> Result<(), IdValidationError> {
+    pub fn validate_contents(&self) -> Result<(), ContentValidationError> {
```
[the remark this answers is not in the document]
dhedey: I actually have a note to myself to go back to NFs and ensure we add a validate in all sensible places (eg when extracting from a manifest, or in the schema checker) - but I think it'd be best to ensure this with the compiler.

radix-engine/src/model/transaction_runtime/executables.rs
```diff
@@ -96,19 +96,38 @@ impl Executor for TransactionRuntimeGenerateUuidInvocation {
...
+    #[test]
+    fn test_uuid_gen() {
+        let hash =
+            Hash::from_str("71f26aab5eec6679f67c71211aba9a3486cc8d24194d339385ee91ee5ca7b30d")
+                .unwrap();
+        let id = generate_uuid(&hash, 5);
+        assert_eq!(
+            NonFungibleLocalId::UUID(id).to_string(),
+            "{86cc8d24-194d-4393-85ee-91ee00000005}"
+        );
```
dhedey, line 131: Can we add a test to check that all 0s becomes `{00000000-0000-4000-8000-000000000000}` and all fs becomes `{ffffffff-ffff-4fff-bfff-ffffffffffff}`?

radix-engine-interface/src/data/value_serializer.rs
iamyulong: Yeah, we can change to that.

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
iamyulong: Shame!

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
iamyulong: `u128::from_str_radix` would catch that.  I will double check as well.

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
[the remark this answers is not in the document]
iamyulong: Feels like over engineering, essentially to represent different states of non-fungible id: `ValidatedNonFungibleLocalId` and `NonFungibleLocalId`. 

Do we really think clients need such protection? or is it making construction cumbersome?

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
iamyulong: No strong reason, but to make the endian-ness (within non-fungible id) consistent.

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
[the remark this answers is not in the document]
dhedey: I'm just aware that I missed some cases where I didn't validate the ID in my original PR before Christmas, and I imagine there might well be others (eg sys-calls / decode impl / schema validation) where it might be easy to miss it too.

I don't think we'd want a `ValidatedNonFungibleLocalId`, I think we'd probably want `NonFungibleLocalId` (the struct that people see) and `NonFungibleLocalIdContent` (the enum which they'd rarely need to see), where constructing a `NonFungibleLocalId` takes the content (or has methods for each id kind) and ensures we validate it after construction.

radix-engine-interface/src/data/value_serializer.rs
```diff
@@ -527,7 +527,7 @@ pub fn serialize_custom_value<S: Serializer>(
             serializer,
             context,
             ScryptoCustomValueKind::NonFungibleLocalId,
-            &value.serializable(*context),
+            &format!("{}", value),
```
dhedey, line 530: We need to remove `ContextualSerialize<ScryptoValueFormattingContext<'a>> for NonFungibleLocalId` and `ContextualSerialize<ScryptoValueFormattingContext<'a>> for NonFungibleGlobalId` below I think?

dhedey: And to update `ScryptoCustomValue::NonFungibleGlobalId` to use `format!` as well?

radix-engine-interface/src/model/resource/non_fungible_local_id.rs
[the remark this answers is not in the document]
iamyulong: I'd suggest we reconsider it at another time. The `IntoNonFungibleLocalId` (to be renamed) trait family heavily depends on error-free non-fungible id construction.

radix-engine-interface/src/data/value_serializer.rs
iamyulong: `NonFungibleGlobalId` still needs context for the resource address contained.

dhedey (approved):
Approving 👍. Will add the "enforce validation" to the work tracker

PR #761, opened 2023-01-20
Feature: Simple Identity Implementation

talekhinezh:


radix-engine-interface/src/address/hrpset.rs
```diff
@@ -13,8 +13,9 @@ pub struct HrpSet {
 
     normal_component: String,
     account_component: String,
+    identity_component: String,
 
-    system: String,
+    system_component: String,
```
iamyulong, line 18: Should we just remove this?

radix-engine/src/engine/kernel.rs
```diff
@@ -86,7 +86,7 @@ where
                     auth_zone_params.initial_proofs.into_iter().collect(),
                 );
                 let node_id = api.allocate_node_id(RENodeType::AuthZoneStack)?;
-                api.create_node(node_id, RENode::AuthZoneStack(auth_zone))?;
+                api.create_node(node_id, RENodeInit::AuthZoneStack(auth_zone))?;
```
iamyulong, line 89: Liked this rename very much!

iamyulong (approved):
LGT

radix-engine-interface/src/address/hrpset.rs
talekhinezh: Yup good point!

talekhinezh: Actually, we are still using this for `EpochManager` and `Clock` and `Validator`. I will delay updating this until we get agreement with product side what's the right hrp to use for these.

PR #762, opened 2023-01-20
wasmi update to v0.23.0

lrubasze:
Update from wasmi `0.11.0` to `0.23.0` for better performance.

Performance gain observed:
Transaction time: `15-20%`
WASM instantiation: `-40%`
WASM instantiation (pre-loaded): `-80%` - it is mostly caused by the fact, that  'template' WASM instance is cached (not WASM module as so far) and just cloned upon `WasmiModule` instantiation.

Despite `wasmi` switched from `parity-wasm` to `wasmparser`, we are still using it to deserialize the WASM code and instrument it.


Cargo.toml
```diff
@@ -31,3 +31,10 @@ members = [
     "radix-engine-constants",
     "utils",
 ]
+
+# Recommended settings for wasmi for best performance
+#  see: https://github.com/paritytech/wasmi/blob/master/CHANGELOG.md#0140---2022-07-26
+[profile.release]
+lto = "fat"
+codegen-units = 1
```
iamyulong, line 39: Not sure if it will be respected, since this crate is only a library.

May need to inform Network Team to do so.

radix-engine/src/wasm/wasmi.rs
```diff
@@ -11,279 +14,478 @@ use crate::wasm::constants::*;
...
 use crate::wasm::traits::*;
 
+type HostState = WasmiInstanceEnv;
+
+/// A `WasmiModule` defines a parsed WASM module "template" Instance (with imports already defined) and Store,
+/// which keeps user data.
+/// "Template" (Store, Instance) tuple are cached together, and never to be invoked.
+/// Upon instantiation Instance and Store are cloned, so the state is not shared between instances.
+/// It is correctly `Send + Sync` - which is good, because this is the thing which is cached in the
+/// ScryptoInterpreter caches.
```
iamyulong, line 24: It's correctly `Send + Sync` under the assumption that the `data` in the `Store` is set to a valid value unpon invocation.

It should not be sent from one thread to another, because the pointer may point to volatile data.

radix-engine/src/wasm/wasmi.rs
```diff
@@ -11,279 +14,478 @@ use crate::wasm::constants::*;
...
-            CONSUME_COST_UNITS_FUNCTION_NAME => Ok(FuncInstance::alloc_host(
-                signature.clone(),
-                CONSUME_COST_UNITS_FUNCTION_ID,
-            )),
-            _ => Err(Error::Instantiation(format!(
-                "Function {} not found",
-                field_name
-            ))),
-        }
+impl WasmiInstanceEnv {
```
iamyulong, line 53: Can we unimplement `Sync` and `Send` for `WasmiInstanceEnv`.

radix-engine/src/wasm/wasmi.rs
```diff
@@ -11,279 +14,478 @@ use crate::wasm::constants::*;
...
+    fn from(err: Error) -> Self {
+        let e_str = format!("{:?}", err);
+        match err {
+            Error::Trap(trap) => {
+                let invoke_err = trap
+                    .downcast_ref::<InvokeError<WasmRuntimeError>>()
+                    .unwrap_or(&InvokeError::SelfError(
+                        WasmRuntimeError::InvalidExportReturn,
+                    ));
+                invoke_err.clone()
```
iamyulong, line 486: It sucks that we have to clone the error. Would be nice if we can simply downcast and own the error.

radix-engine/src/wasm/wasmi.rs
```diff
@@ -295,31 +497,35 @@ impl WasmInstance for WasmiInstance {
         args: Vec<Buffer>,
         runtime: &mut Box<dyn WasmRuntime + 'r>,
     ) -> Result<Vec<u8>, InvokeError<WasmRuntimeError>> {
-        let mut externals = WasmiExternals {
-            instance: self,
-            runtime,
-        };
+        {
+            // set up runtime pointer
+            self.store.data_mut().runtime_ptr = runtime as *mut _ as usize;
```
iamyulong, line 502: Out of curiosity, how much penalty do we get if we add `Mutex` on the pointer.

iamyulong (commented):
LGTM. Would like @talekhinezh to comment on the pointer usage though.

Cargo.toml
lrubasze: With standard settings we have ~10% reduction in Transaction time benchmark. 
So if 10% is acceptable I can revert it.

radix-engine/src/wasm/wasmi.rs
lrubasze: agree

radix-engine/src/wasm/wasmi.rs
lrubasze: no penalty observed (the results we observed the other day were false)

Cargo.toml
iamyulong: Let's keep it and ask Network Team to do the same.

radix-engine/src/wasm/wasmi.rs
lrubasze: I am afraid we cannot since a `Store`  (which keeps`WasmiInstanceEnv` ) shall be `Sync` and `Send` being a part of `WasmiModule` which is cached by `moka::sync::Cache::builder` which requires `Sync` and `Send`.

radix-engine/src/wasm/wasmi.rs
dhedey: Why does `WasmiModule` contain an `Instance` or `State`? Doesn't this mean that it shouldn't be cached?

Can we explain why it is safe to cache the `WasmiModule` even if it contains `store` and `instance`?

radix-engine/src/wasm/wasmi.rs
dhedey: Hmm this is all scaring me too - can we add a detailed comment explaining the hacks we're using:
* https://github.com/paritytech/wasmi/issues/631
* https://github.com/radixdlt/wasmi/commit/84085f82bd4a18e6ecee530e1cb16675c7334df7

Notably it appears that:
* We're duplicating "State" (and with it, instance) from our cache

I'm slightly worried by the reply:
> Yes, as per the Wasm specification the Store is the owner of all instantiated Wasm related state.
An instance only stores references into its associated Store. The Store even owns all Wasm instances associated to it.

If instance contains pointers into `Store`, will cloning `Instance` and `Store` actually work? Or will `Instance` point into the old `Store`? 

Could we have a test which helps check this - I dunno, maybe it EG has a custom blueprint which has a static variable, and saves to that variable - and then calls the same blueprint in a deeper stack frame and check that its static variable has the original state?

dhedey: Actually - I've taken a look, and it's all fine - it doesn't use pointers, it uses a (very simplistic) Arena allocator.

So an `Instance` is just a `Instance(Stored<InstanceIdx>)` which is `Instance(<offset into Store's instance vec>)` which is basically a paired "arena" index into the store and the instance id in that store.

Our cloned stores actually clone everything (all the arenas and the stored id itself) - so they're essentially the same store - but just different data.

So sharing the Store looks fine from a pointer perspective.

And I've also double-checked the wider use of (eg) the runtime_ptr and it looks good.

dhedey: That said, I'd quite like a significantly longer comment on `WasmiModule` explaining why it's fine 👍

radix-engine/src/wasm/wasmi.rs
```diff
@@ -11,279 +14,478 @@ use crate::wasm::constants::*;
...
+/// A `WasmiModule` defines a parsed WASM module "template" Instance (with imports already defined) and Store,
+/// which keeps user data.
+/// "Template" (Store, Instance) tuple are cached together, and never to be invoked.
+/// Upon instantiation Instance and Store are cloned, so the state is not shared between instances.
+/// It is correctly `Send + Sync` - which is good, because this is the thing which is cached in the
+/// ScryptoInterpreter caches.
 pub struct WasmiModule {
-    module: Module,
+    store: Store<HostState>,
+    instance: Instance,
```
dhedey, line 27: Could we call this eg `template_store` and `template_instance` to indicate that this is the template which should be cloned? 👍

iamyulong (approved):

Cargo.toml
lrubasze: PR: https://github.com/radixdlt/babylon-node/pull/289

PR #767, opened 2023-01-21
Add AccessController as a native component

0xOmarA:


radix-engine/src/model/access_controller/executables.rs
```diff
@@ -395,42 +502,42 @@ impl Executor for AccessControllerQuickConfirmRecoveryExecutable {
...
                     Err(AccessControllerError::NoValidProposedRuleSetExists),
                     |(_, (rule_set, _))| Ok(rule_set.clone()),
-                )?;
+                )?
+        };
 
-            access_controller.ongoing_recoveries = None;
-            access_controller.active_rule_set = new_rule_set;
+        // Update the access rules substate
+        // TODO: We need something better here. Update once we have an AccessRulesChain invocation
```
0xOmarA, line 527: Input is very much welcome on this

radix-engine/src/model/access_controller/substates.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// A vault where the asset controlled by the access controller lives.
+    pub controlled_asset: VaultId,
+
+    /// The amount of time (in minutes) that it takes for timed recovery to be done. Maximum is
+    /// 4,294,967,295 minutes which is 8171.5511700913 years. When this is [`None`], then timed
+    /// recovery can not be performed through this access controller.
+    pub timed_recovery_delay_in_minutes: Option<u32>,
+
+    /// The states of the Access Controller.
+    pub state: (PrimaryRoleState, OperationState),
```
talekhinezh, line 15: It would be more clear to have the P and R recovery states be separate so that the FSM is enforced by structure rather than relying on correct logic, e.g.

```
struct AccessControllerSubstate {
    ...
    pub state: (PrimaryRoleState, PState, RState),
    ...
}
    
pub enum PState {
    Normal,
    Recovery(Ruleset),
}

pub enum RState {
    Normal,
    Recovery(RRecovery),
}

pub enum RRecovery {
    Timed {
        pub timed_recovery_allowed_after: Instant,
        pub rule_set: RuleSet,
    },
    NoTimer(RuleSet),
}
```

radix-engine-interface/src/address/entity.rs
```diff
@@ -27,6 +27,9 @@ pub const ECDSA_SECP_256K1_VIRTUAL_ACCOUNT_COMPONENT_ADDRESS_ENTITY_ID: u8 = 0x0
 /// A unique identifier used in the addressing of a virtual Account Component Addresses.
 pub const EDDSA_ED_25519_VIRTUAL_ACCOUNT_COMPONENT_ADDRESS_ENTITY_ID: u8 = 0x08;
 
+/// A unique identifier used in the addressing of a access controller components.
+pub const ACCESS_CONTROLLER_COMPONENT_ADDRESS_ENTITY_ID: u8 = 0x09;
```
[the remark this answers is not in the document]
0xOmarA: True, I think we don't need it. Will remove it

radix-engine/src/model/access_controller/substates.rs
0xOmarA: Sounds good, will split them up.

radix-engine-constants/src/lib.rs
```diff
@@ -38,7 +38,7 @@ pub const MAX_TRANSACTION_SIZE: usize = 1 * 1024 * 1024;
 pub const DEFAULT_SYSTEM_LOAN: u32 = 1_000_000;
 
 /// The default max call depth, used by transaction executor.
-pub const DEFAULT_MAX_CALL_DEPTH: usize = 12;
+pub const DEFAULT_MAX_CALL_DEPTH: usize = 10;
```
0xOmarA, line 41: @talekhinezh @iamyulong  A little unsure as to why this was needed, or whether there is a larger problem here. Without reducing this, it seemed like the execution environment stack was overflowing before we reach the `MAX_CALL_DEPTH` and therefore it was erroring out. I should note that this issue goes away when running the tests with `--release`.

The tests that were failing are these two:
* [`recursion_bomb_to_failure`](https://github.com/radixdlt/radixdlt-scrypto/blob/7d1030689f5a0e46e5fc8cafc9434f4cc7b1c958/radix-engine/tests/local_component.rs#L164-L203)
* [`recursion_bomb_2_to_failure`](https://github.com/radixdlt/radixdlt-scrypto/blob/7d1030689f5a0e46e5fc8cafc9434f4cc7b1c958/radix-engine/tests/local_component.rs#L240-L279)

iamyulong: Thanks! Will look into this at a later time.

radix-engine/src/model/access_controller/state_machine.rs
```diff
@@ -0,0 +1,421 @@
...
+        _api: &mut Y,
+        input: AccessControllerQuickConfirmRecoveryStateMachineInput,
+    ) -> Result<Self::Output, RuntimeError>
+    where
+        Y: SystemApi + EngineApi<RuntimeError> + InvokableModel<RuntimeError>,
+    {
+        // This transition can not be performed when the confirmor and the proposer are the same.
+        // As in, Role X can't perform a quick confirm on a recovery initiate by Role X, it has to
+        // be different.
+        if input.confirmor == input.proposer.into() {
```
talekhinezh, line 154: Rather than this being implemented as logic, this part could also be cleaned up if the quick_confirm transitions were implemented separately.

talekhinezh: This would also allow us to remove the need for the `QuickConfirmAs[Role]` pattern we currently have.

radix-engine/src/model/access_controller/state_machine.rs
```diff
@@ -0,0 +1,421 @@
...
+        // A recovery attempt can only be canceled when we're in recovery mode regardless of whether
+        // primary is locked or unlocked
+        match self.state {
+            (_, PrimaryOperationState::Recovery(ref proposal), _)
+                if input.proposer == Proposer::Primary =>
+            {
+                let proposal = proposal.clone();
+
+                // Ensure that the caller has passed in the expected proposal
+                validate_recovery_proposal(&proposal, &input.proposal_to_cancel)?;
```
talekhinezh, line 282: Is this necessary for cancelling?

radix-engine-constants/src/lib.rs
talekhinezh: Yup, I've run into this problem as well when running in debug mode

radix-engine/src/model/access_controller/state_machine.rs
0xOmarA: I guess it isn't. CAP-14 primarily mentions passing the proposal again for quick confirm recovery but not for cancellation of a recovery proposal, so this might not be necessary.

radix-engine/src/model/access_controller/state_machine.rs
0xOmarA: Could you expand on this a bit? I think that somewhere in the logic we would need to check to ensure that role X isn't confirming recovery initiated by role X. Not sure how implementing the transitions separately achieves that.

0xOmarA: Yeah, we might want to separate this out into multiple state machine inputs (and probably also do the same for other inputs as well). Only concern here is that it seems like we've formed an hour glass. Multiple invocations go to the same executor which then map out to different state machine inputs. 

```
╲      ╱     Different Invocations
 ╲    ╱
  ╲  ╱
  │  │       Same Executor
  ╱  ╲
 ╱    ╲
╱      ╲     Different State Machine Inputs
```

We could switch to one executor and one input per invocation to clean up the logic a little bit but having a single executor is nice for stuff like updating the access rules and having less duplicated logic.

talekhinezh: I think one input per invocation makes more sense. Less duplicated logic can be solved by using functions/methods encapsulating the duplicated logic.

0xOmarA: Sounds good, I've refactored the code to use one executor per invocation and one input per executor (i.e. no more hourglass)

talekhinezh (approved):

PR #769, opened 2023-01-24
Radix Engine Codebase Restructure

iamyulong:
No logical change, but to restructure the `radix-engine` and `radix-engine-interface`  code base.

This is to reveal the interdependency between `Kernel`, `System`, `KernelModule`, `NodeModule`, `Blueprints` (`Native`/`WASM`) and facilitate follow-up refactoring.

Suggest review: check out this PR and open the `radix-engine` and `radix-engine-interface` folders in VSCode.


talekhinezh:
Rather than separate folders by type (e.g. `kernel_module` vs. `node_module`), can we separate by top-level package which contains related code which need to work together (e.g. `Auth` + `Fee`).

iamyulong:
I definitely see the common dependency between kernel modules and node modules, especially the substate offsets. And they're related from a feature perspective. 

But, on the other side, I think they're technically very different. They work with different Api, and serve different purpose, and may even be compiled independently.

Let's chat about this.

talekhinezh (approved):

PR #772, opened 2023-01-25
Test util for asserting on state hash + Example test

jakrawcz-rdx:
Note: this relies on https://github.com/radixdlt/radixdlt-scrypto/pull/763.

### DIFF
--- a/radix-engine/tests/determinism.rs
+++ b/radix-engine/tests/determinism.rs
@@ -1,10 +1,17 @@
+use radix_engine::types::*;
+use radix_engine_interface::args;
+use radix_engine_interface::constants::FAUCET_COMPONENT;
+use radix_engine_interface::model::{FromPublicKey, NonFungibleGlobalId};
+use scrypto::resource::DIVISIBILITY_MAXIMUM;
 use scrypto_unit::*;
+use transaction::builder::ManifestBuilder;
+use transaction::model::BasicInstruction;
 
 #[test]
 fn test_simple_deterministic_execution() {
     // Arrange
-    let mut test_runner0 = TestRunner::new(true);
-    let mut test_runner1 = TestRunner::new(true);
+    let mut test_runner0 = TestRunner::builder().with_state_hashing().build();
+    let mut test_runner1 = TestRunner::builder().with_state_hashing().build();
 
     // Act
     let (public_key0, _, account0) = test_runner0.new_allocated_account();
@@ -16,4 +23,59 @@ fn test_simple_deterministic_execution() {
     test_runner0
         .substate_store()
         .assert_eq(test_runner1.substate_store());
+}
+
+#[test]
+fn same_executions_result_in_same_final_state_hash() {
+    let state_hashes = (0..5)
+        .map(|_| create_and_pass_multiple_proofs())
+        .collect::<HashSet<Hash>>();
+    assert_eq!(
+        state_hashes.len(),
+        1,
+        "non-deterministic final state hash: {:?}",
+        state_hashes
+    );
+}
+
+/// Simulates a complete "test" which creates multiple proofs and passes them to a method.
+/// Such operation is supposed to trigger non-determinism bugs in the engine.
+/// Returns the root hash of the system's final state.
+fn create_and_pass_multiple_proofs() -> Hash {
+    // Arrange
+    let mut test_runner = TestRunner::builder().with_state_hashing().build();
+    let (public_key, _, account) = test_runner.new_allocated_account();
+    let resource_address =
+        test_runner.create_fungible_resource(100.into(), DIVISIBILITY_MAXIMUM, account);
+    let package_address = test_runner.compile_and_publish("./tests/blueprints/proof");
+
+    // Act
+    let mut builder = ManifestBuilder::new();
+    builder.lock_fee(FAUCET_COMPONENT, 10.into());
+    let proof_ids = (0..20)
+        .map(|_| {
+            builder
+                .create_proof_from_account_by_amount(account, 1.into(), resource_address)
+                .add_instruction(BasicInstruction::PopFromAuthZone)
+                .2
+                .unwrap()
+        })
+        .collect::<Vec<ManifestProof>>();
+    let manifest = builder
+        .call_function(
+            package_address,
+            "VaultProof",
+            "receive_proofs",
+            args!(proof_ids),
+        )
+        .build();
+    let receipt = test_runner.execute_manifest(
+        manifest,
+        vec![NonFungibleGlobalId::from_public_key(&public_key)],
+    );
+
+    // Assert
+    receipt.expect_commit_success();
+
+    test_runner.get_state_hash()
 }
--- a/radix-engine/tests/epoch_manager.rs
+++ b/radix-engine/tests/epoch_manager.rs
@@ -11,7 +11,7 @@ use transaction::signing::EcdsaSecp256k1PrivateKey;
 #[test]
 fn get_epoch_should_succeed() {
     // Arrange
-    let mut test_runner = TestRunner::new(true);
+    let mut test_runner = TestRunner::builder().build();
     let package_address = test_runner.compile_and_publish("./tests/blueprints/epoch_manager");
 
     // Act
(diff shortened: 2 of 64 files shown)


radix-engine/tests/determinism.rs
```diff
@@ -17,3 +24,58 @@ fn test_simple_deterministic_execution() {
...
+        .collect::<HashSet<Hash>>();
+    assert_eq!(
+        state_hashes.len(),
+        1,
+        "non-deterministic final state hash: {:?}",
+        state_hashes
+    );
+}
+
+/// Simulates a complete "test" which creates multiple proofs and passes them to a method.
```
jakrawcz-rdx, line 41: @talekhinezh plz see the impl and let me know if this is what you meant:
image

scrypto-unit/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 radix-engine = { path = "../radix-engine", default-features = false }
 radix-engine-interface = { path = "../radix-engine-interface", default-features = false }
 radix-engine-constants = { path = "../radix-engine-constants" }
+radix-engine-stores = { path = "../radix-engine-stores", default-features = false }
```
jakrawcz-rdx, line 10: **Question from author:**

I am pretty sure I am doing something wrong with the deps (or build?) in this PR.
Selective tests run fine, but a full `./build.sh` is giving me:

```
Building the engine in different configurations...
+ cd radix-engine
+ cargo build --features wasmer
    Finished dev [unoptimized + debuginfo] target(s) in 3.03s
+ cd radix-engine
+ cargo build --no-default-features --features alloc
   Compiling sbor v0.8.0 (/Users/jakrawcz/repo1/radixdlt-scrypto/sbor)
error: Feature `std` and `alloc` can't be enabled at the same time.
 --> sbor/src/lib.rs:6:1
  |
6 | compile_error!("Feature `std` and `alloc` can't be enabled at the same time.");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `sbor` due to previous error
```

(as you can see, I tried copy-pasting various bits of code from other crates .toml, which didn't help; plz advice)

scrypto-unit/src/test_runner.rs
```diff
@@ -103,53 +104,81 @@ impl Compile {
     }
 }
 
-pub struct TestRunner {
-    scrypto_interpreter: ScryptoInterpreter<DefaultWasmEngine>,
-    staged_substate_store_manager: StagedSubstateStoreManager<TypedInMemorySubstateStore>,
-    intent_hash_manager: TestIntentHashManager,
-    next_private_key: u64,
-    next_transaction_nonce: u64,
+pub struct TestRunnerBuilder {
```
jakrawcz-rdx, line 107: **Note from author:**

I introduced a builder (instead of combinatorial number of factory methods).
This is the only reason so many test files are touched by this PR.

scrypto-unit/src/test_runner.rs
```diff
@@ -103,53 +104,81 @@ impl Compile {
...
     }
 
     pub fn substate_store(&self) -> &TypedInMemorySubstateStore {
-        &self.staged_substate_store_manager.root
+        &self.substate_store
     }
 
     pub fn substate_store_mut(&mut self) -> &mut TypedInMemorySubstateStore {
-        &mut self.staged_substate_store_manager.root
+        &mut self.substate_store
```
jakrawcz-rdx, line 181: **Note from author:**

I noticed that "staging" was used here in a pretty degenerate way, so I refactored it into an exact equivalent (to avoid complexity); please shout if I missed something!

scrypto-unit/src/test_runner.rs
dhedey: Looks nice! 👍

scrypto-unit/src/test_runner.rs
dhedey: Yeah, this makes sense. If we were previously just using the `root` property on it directly, then it didn't give it any effective testing.

radix-engine/tests/determinism.rs
```diff
@@ -17,3 +24,58 @@ fn test_simple_deterministic_execution() {
         .substate_store()
         .assert_eq(test_runner1.substate_store());
```
dhedey, line 25: Should we add an assert here that they have equal state hash?

radix-engine/tests/determinism.rs
dhedey: Does this test fail or succeed?

@talekhinezh  - how do we expect the non-determism to manifest?

I guess this non-determinism might only come from the receipt; so eg a (consensus) receipt hash accumulator might show us it? (but maybe not the state root hash?)

radix-engine/tests/epoch_manager.rs
```diff
@@ -88,8 +88,8 @@ fn next_epoch_with_validator_auth_succeeds() {
     let initial_epoch = 5u64;
     let rounds_per_epoch = 2u64;
     let genesis = create_genesis(BTreeMap::new(), initial_epoch, rounds_per_epoch);
-    let mut test_runner = TestRunner::new_with_genesis(true, genesis);
-
+    let mut test_runner = TestRunner::builder().with_custom_genesis(genesis).build();
+    //TestRunner::builder().with_custom_genesis(genesis).build();
```
dhedey, line 92: Accidental comment added in?

scrypto-unit/Cargo.toml
dhedey: Hmmmm this is a bit weird - something in the crate chain is pulling in sbor with both features --- might need to sit down with you and run through it 👍 

Note that the error happens when building `radix-engine` with no-std, and I don't believe radix engine actually relies on `radix-engine-stores`, so it might be something unrelated to this PR?

dhedey (approved):
Really nice work. Give me a shout if you want to debug the build issues.

scrypto-unit/Cargo.toml
jakrawcz-rdx: It is "related" in a sense: that build succeeds for me locally without the .toml changes seen in this PR.
(but I am ofc not aware of any build caches or whatever)

radix-engine/tests/epoch_manager.rs
jakrawcz-rdx: whoops, yes; reverted.

radix-engine/tests/determinism.rs
jakrawcz-rdx: I can add it if you wish.
A better question (which popped up to me after seeing this line earlier) is: why do we need a test util for comparing state hash, if we have a test util for comparing exactly the entire state? (it seems like even better way of asserting determinism, and the performance is probably acceptable, given typical test state sizes)

dhedey: Haha I didn't actually know we had this(!). I'm surprised that SubstateStore has equality defined on it!

But yeah, good point :p

scrypto-unit/Cargo.toml
jakrawcz-rdx: Seems like it is not just my machine: CI encounters the same https://github.com/radixdlt/radixdlt-scrypto/actions/runs/4022806568/jobs/6912935321 (i.e. only the run with `no_std` in the name).

I have no idea how to investigate it further; I suspect either some silly mistake of mine, or a super-deep bug within our existing build setup. Any help appreciated.

jakrawcz-rdx: Ok, so after investigating this with @dhedey, this PR now also fixes the deps of the `-stores` crate :)

Ready for a final look!

dhedey (approved):

radix-engine/tests/determinism.rs
jakrawcz-rdx: it succeeds, 100% for me (not even flakey)
maybe the non-det bugs are not so easy to catch? only occur across machines?

anyways, I am treating this test just as an example for "how to use that new test util (for things that you suspect are non-det)".

We can also hope @talekhinezh has some pointers on how to make this test more flakey :) (I am not 100% sure that I implemented his idea correctly)

PR #774, opened 2023-01-25
Add `ClientPackageApi` and `ClientMeteringApi`

iamyulong:
- Add `ClientPackageApi`
- Add `ClientMeteringApi`
- Updated `ScryptoRuntime` to use Client API only


radix-engine-interface/src/api/metering_api.rs
```diff
@@ -0,0 +1,13 @@
+use crate::api::types::*;
+use crate::blueprints::resource::*;
+
+pub trait ClientMeteringApi<E> {
```
talekhinezh, line 4: Do we see this as a long term API or this is just temporary? A bit strange that we need Application layer knowledge (VaultId/Resource) for the API. If temporary, can we add a quick comment.

iamyulong: I'm currently leaning towards making it part of client API, and limits access to native environment.

But I will add a `TODO` note for more thinking.

PR #776, opened 2023-01-26
Use enum for costing reason (+ some optimisation)

iamyulong:
- Use enum for costing reason
- Use `Vec` instead of `HashMap` for statistics (thanks to @dhedey)


radix-engine/src/system/kernel_modules/fee/fee_reserve.rs
```diff
@@ -198,34 +234,42 @@ impl SystemLoanFeeReserve {
...
         }
 
-        // Sort out the amount from system loan
-        let from_loan = min(self.loan_balance, cost_units);
-
-        // Sort out the amount from locked payments
-        let from_locked = price * (cost_units - from_loan) as u128;
-        if self.xrd_balance < from_locked {
-            return Err(FeeReserveError::InsufficientBalance);
+        if self.loan_balance >= cost_units {
```
dhedey, line 237: It took me a little bit of time to understand what was going on here.

What do you think about some renames?
* `self.loan_balance` => `self.remaining_loan_balance`
* `self.cost_units_consumed` => `self.total_cost_units_consumed`
* `cost_units` => `cost_units_to_consume`

And then we:
* Get rid of `from_loan` - just use `self.remaining_loan_balance` on line 246, and just set it to 0 on line 252

Also - what do you think about having three branches?
* If `self.is_loan_repaid` - then it's simple, the loan has been repaid, and we just do a simpler debit from xrd balance.
* Else we have the two branches written here

(NB - we would set the self.is_loan_repaid in the `repay_all` method)

The benefit of this three-branch approach is that we can call `repay_loan` in the last branch; and remove the duplicated check from `consume_royalty` / `consume_execution`. The check also becomes slightly cheaper (just a single boolean check rather than two checks)

It does add the internal dependency between `consume` and `repay_all` that we never add to deferred or `xrd_owed` after we repay the loan, but I think that's a small price to pay for simpler, more efficient code.

dhedey (approved):
Really nice change. I also like the rename `SysCallX` => `KernelApiCallX`

radix-engine/src/system/kernel_modules/fee/fee_reserve.rs
iamyulong: Liked all the renaming suggestions and get rid of `from_loan`. Will apply.

Let me play with the three-branch idea and get back to you.

iamyulong: Made it three-branch but didn't introduce `is_loan_repaid` state. 

The `remaining_loan_balance` is serving the same purpose.

We may potentially benefit from a dedicated state in `loan_repaid()` method, but I didn't notice performance difference.

radix-engine/src/system/kernel_modules/fee/fee_reserve.rs
```diff
@@ -426,7 +426,7 @@ impl FinalizingFeeReserve for SystemLoanFeeReserve {
         // In case the transaction finishes before check point.
         self.attempt_to_repay_all();
 
-        let mut execution_cost_unit_breakdown = HashMap::new();
+        let mut execution_cost_unit_breakdown = BTreeMap::new();
```
dhedey, line 429: This is duplicated here and below

iamyulong: Thanks! Will remove it.

radix-engine/src/system/kernel_modules/fee/fee_reserve.rs
dhedey: Sure, sounds good. One final suggestion then - 

What did you think about getting rid of this code from `consume_royalty` and `consume_execution`:
```
if self.total_cost_units_consumed >= self.check_point && !self.fully_repaid() {
    self.repay_all()?;
}
```
And instead, adding this code in one place at the end of `consume`? That would let us also delete `self.check_point`
```
if self.remaining_loan_balance == 0 && !self.fully_repaid() {
    self.repay_all()?;
}
```

iamyulong: Sounds like a good idea. Will apply

PR #778, opened 2023-01-26
Fix/SCRY-314 transaction size

mstrug-rdx:
Changed maximum transaction size from 4MiB to 1MiB.
Moved constant to `radix-engine-constants`.


dhedey (approved):
Looks good - we actually override this in the Node at the moment anyway, so we will need to update this there too 👍 - unless you fancy also taking a look at updating the code in `babylon-node`? :)

mstrug-rdx:
@dhedey I will do this change for `babylon-node` also.

PR #779, opened 2023-01-26
Refactor WASM native function invocation

iamyulong:
This PR refactors WASM native function invocation to use `fn_identifier` + `invocation_payload` (individual).

Notes:
- `FnIdentifier` is still an enum, but much lighter than `NativeInvocation`;
- Hopefully, we will get another bump when `FnIdentifier` becomes flat.


scrypto/src/engine/wasm_api.rs
```diff
@@ -61,8 +61,13 @@ extern "C" {
         args_len: usize,
     ) -> Buffer;
 
-    /// Invokes any function or method, for both Scrypto and Native blueprints.
-    pub fn invoke(invocation_ptr: *const u8, invocation_len: usize) -> Buffer;
+    /// Invokes a native function or method.
+    pub fn call_native(
```
talekhinezh, line 65: What is call_native again in this case? Is it still the idea that we are moving away from the client knowing the difference between native vs. non-native? ie, this is a preparatory split before merge?

talekhinezh (approved):

scrypto/src/engine/wasm_api.rs
iamyulong: Yes, it's preparatory. I'm incrementally downscaling the `call_native` until the backend is fully prepared.

```
invoke_scrypto (removed in this PR) + invoke_native_blueprint + invoke_node_module
```
Our step is probably to remove `invoke_native_blueprint`, which seems to be the easiest.

PR #781, opened 2023-01-27
add missing conversions between Decimal and PreciseDecimal

lrubasze:


radix-engine-interface/src/math/decimal.rs
```diff
@@ -538,6 +539,16 @@ impl fmt::Display for ParseDecimalError {
     }
 }
 
+impl From<PreciseDecimal> for Decimal {
```
iamyulong, line 542: Should be `TryFrom` no?

lrubasze: By implementing From trait I received try_from() for free.

Initially I wanted to implement `TryFrom` and `From` (`from()` was supposed to use `try_from()`). But rust was complaining with this: 

`conflicting implementations of trait std::convert::TryFrom<math::precise_decimal::PreciseDecimal> for type math::decimal::Decimal`

Thus I decided to keep just From trait

iamyulong: I'm a bit concerned with this approach this. 

It's not consistent with how Rust deals with conversion.  We only implement `From` for infallible conversions.

lrubasze: yes, you are right. That was a bad approach.
Reworked it accordingly (`BnumXYZ` types also required that).

As mentioned before, cannot keep both traits at the same time (most likely because of this issue: https://github.com/rust-lang/rust/issues/50133).

Thus I decided to switch to `TryFrom` for all possible conversions between `BnumXYZ` types, even those infallible ones, eg. `BnumI512::from(BnumI256)`. This is to keep consistency within `BnumXYZ` types. If that is not ok, please let me know, so I will revert infallible ones to `From`.

Didn't change `PreciseDecimal::from(Decimal)` to `try_from()` (similar to Decimal::try_from(PreciseDecimal`)since:
- conversion shall be infallible
- it has been implemented like that for some time already and don't want to impact the users.

iamyulong: Sorry for the churn. 

I like other changes, but why do you change `BnumI512::from(BnumI256)`? This _should_ be an infallible conversion, no? 

We don't need to have both `From` and `TryFrom` in my view.  Conversions can be either infallible or not, NOT BOTH. 

The ticket traces back to 2018, and I don't think Rust is going to change it.

iamyulong: In rust, we do `let a  = u64::from(1u32)` and `let b = u32::try_from(u64).unwrap("or special error handling")`.

lrubasze: Sure. Thanks for feedback! I will restore infallible Bnum conversions to `from() `and leave fallible ones with `try_from()`.

To justify myself, I thought it would be more coherent to have the same conversion method between Bnum types (even if it is not the pure rust approach). Maybe my previous comment was not clear enough.

iamyulong (approved):
LGTM.

@lrubasze has identified a few more inconsistencies in regards to fallible/infallible integer/decimal conversion. 

He will clean them up in a separate PR.

PR #783, opened 2023-01-27
fix: Tweak cost unit multipliers

dhedey:
Rough and dirty to make a spin-loop in a modified `should_be_aborted_when_loan_repaid` test run out of juice in 350ms (on my laptop).

Also adjusted the system loan. There's still a _lot_ of adjustment to do here, which is waiting on a more holistic review.


iamyulong (approved):
Two more tests to fix, then we're good to go.

PR #787, opened 2023-01-29
Feature: Native Accounts

0xOmarA:


simulator/src/resim/mod.rs
```diff
@@ -304,16 +306,237 @@ pub fn export_abi(
...
+        }
+
+        _ => todo!("Unsupported native ABI."),
+    }
+}
+
+// Keeping this function here for the time being as it helps me make stronger assumptions about the
+// context in which it will be used.
+// TODO: Refactor out once we have a clear plan with native component package ABIs
+pub fn export_account_abi() -> abi::BlueprintAbi {
```
0xOmarA, line 336: Would appreciate input on this. How do we plan to deal with package ABIs with native packages?

talekhinezh: We're moving towards "everything has an ABI including native packages"

0xOmarA: Makes sense. I'll keep this here until we have native package ABI implemented.

talekhinezh (approved):

PR #789, opened 2023-01-30
Refactor: Wasmi with references

dhedey:
An example refactor where I've made Wasmi store an actual reference to the runtime - not just a usize pointer in disguise...

Although all the sins get hidden in:
* An implicit understanding that `WasmiInstanceEnv<'static, 'static'>` is the template instance, and a manual implementation of `Send` + `Sync` for that template instance.
* A transmute call mapping `WasmiInstanceEnv<'static, 'static'>` to `WasmiInstanceEnv<'r, 'a'>`
 
So I'm not sure it's considerably better...

Note that this is just food for thought at this stage, and _in particular_ I haven't updated the Wasmer interface to match.

If we want this, we'd need to fix the following before merge:
* Compare performance
* Update wasmer impl to match
* Update the code comments
* Consider using `MaybeUninit` instead of Option

### DIFF
--- a/radix-engine/benches/wasm.rs
+++ b/radix-engine/benches/wasm.rs
@@ -27,7 +27,7 @@ fn bench_wasm_instantiation(c: &mut Criterion) {
     c.bench_function("WASM instantiation", |b| {
         b.iter(|| {
             let engine = DefaultWasmEngine::default();
-            engine.instantiate(&pretend_instrumented_code);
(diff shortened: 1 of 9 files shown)


talekhinezh:
@dhedey Can we close this?

dhedey:
Yup, I'll close now.

The learnings from this got included in updates to https://github.com/radixdlt/radixdlt-scrypto/pull/762

PR #790, opened 2023-01-30
Allocate xrd to accounts in genesis

LukasGasior1:


radix-engine/src/ledger/bootstrap.rs
```diff
@@ -25,6 +25,7 @@ pub struct GenesisReceipt {
 
 pub fn create_genesis(
     validator_set_and_stake_owners: BTreeMap<EcdsaSecp256k1PublicKey, (Decimal, ComponentAddress)>,
+    account_xrd_allocations: Vec<(Decimal, EcdsaSecp256k1PublicKey)>,
```
talekhinezh, line 28: Any reason we can't use `BTreeMap<EcdsaSecp256k1PublicKey, Decimal>` here instead? Would be nice given the Olympia end state to have a canonical genesis transaction which doesn't depend on an arbitrary ordering.

LukasGasior1: sounds good :+1: no particular reason for Vec here

talekhinezh (approved):

PR #791, opened 2023-01-30
Feature: Add hrps for native components

talekhinezh:


radix-engine-interface/src/address/hrpset.rs
```diff
@@ -27,15 +27,14 @@ impl HrpSet {
...
-            EntityType::Clock => &self.system_component,
+            EntityType::EpochManager => &self.epoch_manager_component,
+            EntityType::Validator => &self.validator_component,
+            EntityType::Clock => &self.clock_component,
             EntityType::EcdsaSecp256k1VirtualAccountComponent => &self.account_component,
             EntityType::EddsaEd25519VirtualAccountComponent => &self.account_component,
             EntityType::EcdsaSecp256k1VirtualIdentityComponent => &self.identity_component,
             EntityType::EddsaEd25519VirtualIdentityComponent => &self.identity_component,
-            EntityType::AccessControllerComponent => &self.normal_component,
+            EntityType::AccessControllerComponent => &self.access_controller_component,
```
0xOmarA, line 37: Currently, the access controller doesn't have a unique entity byte (we removed it during the review process when we were unclear about whether it should have its own unique HRP or not). I think that we should add a unique entity byte for it.

0xOmarA (approved):

iamyulong (approved):

PR #793, opened 2023-01-30
Architecture: Node Module Identifier

talekhinezh:


iamyulong (approved):
Nice work!

The only change I'd like to request to rename `Self1` to something more specific and not confusing with `Self`.

The idea is, we should be able to tell the substate layout given a module ID, except for `Self` (where substate layout is defined by blueprints).

PR #796, opened 2023-01-31
Feature/SCRY-310 resources usage tracking

mstrug-rdx:
# Summary

Added resources usage (heap memory and CPU cycles) measurements for transaction execution.

# Detailed description

Extended `TransactionExecution` struct (part of `TransactionReceipt`) with new field of type `ResourcesUsage` which stores heap memory usage (peak and sum) and CPU cycles for transaction execution (it will be extended with additional data: nested calls depth).
Added new struct `ResourcesTracker` which internally uses `InfoCpu` and `InfoAlloc` structs to measure heap and CPU cycles.
Heap allocations metric uses encapsulation of global allocator which counts peak and sum of allocated memory.
CPU cycles implementation uses Performance Counter mechanism which is currently supported only on Linux OS x86 (32 and 64 bits). To use it, it is needed to enable `resource-usage-with-cpu` feature during compilation. New crate is used here: [perfcnt](https://crates.io/crates/perfcnt).

# Testing

Added new bench group `resources_usage` as a placeholder for all test cases which reproduces extreme conditions of transaction execution. Currently only transfer transaction is used (copied from radix_engine bench) with additional `ResourceInfoFramework` which gathers data on each bench iteration and prints out detailed informations about resources usage after bench execution.

# Other info

`ResourcesUsage` struct has optional field: `cpu_cycles` which exists only for `resource-usage-with-cpu` compilation feature, it is implemented as two versions of the whole struct instead of optional struct field due to issue with rust-analyzer which issues 'no-such-field' warning due to some bug (similar issue was submitted: [here](https://github.com/rust-lang/rust-analyzer/issues/11657)).

`InfoAlloc` uses internally signed counters because of possible transient negative values in case of reallocation of memory allocated before memory measurement starts.

Resource measurement is disabled for `alloc` compilation feature because of lack of `System` allocator in `no_std` environment (alloc crate is not providing allocator).


radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -212,10 +217,19 @@ where
             track.finalize(invoke_result, events)
         };
 
+        // Finish resources usage measurement and get results
+        let resources_usage = match () {
+            #[cfg(feature = "std")]
```
iamyulong, line 222: How about we change all conditional compilation to
```
#[cfg(all(target_os = "linux", feature = "std",  feature = "resource-usage-with-cpu"))]
```
So we don't need to examine each invidiual use case.

radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -14,10 +14,25 @@ use radix_engine_interface::data::{IndexedScryptoValue, ScryptoDecode};
...
+    pub heap_allocations_sum: usize,
+    pub heap_peak_memory: usize,
+    pub cpu_cycles: u64,
+}
+
 #[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
 pub struct TransactionExecution {
     pub fee_summary: FeeSummary,
     pub events: Vec<TrackedEvent>,
+    pub resources_usage: ResourcesUsage,
```
iamyulong, line 35: So, we're puting resource usage under the `execution` section of the `TransactionReceipt`, instead of the `result` section.

This is good as I suppose node is not merkelizing `execution`.

iamyulong (commented):
LGTM with minor comments.

radix-engine/src/transaction/transaction_executor.rs
mstrug-rdx: I wanted to separate CPU cycles which will not be used for production builds from other metrics (heap and call depth).  Also CPU cycles can be used only on Linux OS, but we can still use heap metric on Mac OS and Windows (and call depth in future), that is why here is only `std` feature required.

iamyulong: Another use case is use benchmarking different transactions. We will just turn on `resource_usage` and run with `std`

mstrug-rdx: Actually it was request from Network team to have this resource tracking mechanism without `resource_usage` feature (it was implemented like this at the beginning). @malexandru-rdx what do you think about moving this back under `resource_usage` feature?

iamyulong: I would even consider renaming to `cpu_ram_metrics` to be more specific.

malexandru-rdx: From the perspective of what should be captured in the receipt, on the node side we don't really need the memory usage itself in production. We do however need to limit the max memory usage per WASM instance with a separate field in `RejectionError` (i.e `MaxWASMMemoryExceeded`) from RE. Since we need the metering mechanism either way (were the overhead is), exposing the actual collected data (in prod) is optional from my point of view.

malexandru-rdx: I would also prefer it to be only 1 feature which contains everything metrics related not needed for prod.

radix-engine/src/transaction/transaction_receipt.rs
malexandru-rdx: > This is good as I suppose node is not merkelizing execution.
What do you mean by that?

radix-engine/src/transaction/transaction_executor.rs
iamyulong: Yeah, we will do "soft" memory track and limiting, but not the host heap allocation.

Are we all good with the following?
```
#[cfg(all(target_os = "linux", feature = "std",  feature = "cpu_ram_metrics"))]
```

radix-engine/src/transaction/transaction_receipt.rs
iamyulong: I mean the `execution` is not agreed by validators through consensus.

iamyulong (approved):

PR #797, opened 2023-02-01
Feature: Validator owner rule

talekhinezh:


radix-engine/src/ledger/bootstrap.rs
```diff
@@ -190,13 +190,26 @@ pub fn create_genesis(
                 }
                 .into(),
             );
-            validators.insert(key, (bucket, account_address));
+            let validator_account_address = ComponentAddress::virtual_account_from_public_key(&key);
```
dhedey, line 193: When it comes to the Olympia migration, the owner key should be separate from the validator key, and I assume we need multiple amounts staked per validator - but I guess these can come in Lukasz' changes later, and on the `develop` branch 👍

dhedey (approved):
Looks good, thanks 👍

PR #798, opened 2023-02-01
Refactor ClientApi and the Component trait

iamyulong:
- Refactored various Client APIs
- Seperated `KernelInvokeApi` from Client APIs
- Refactored the `Component` trait in Scrypto lib
- Refactored `KeyValueStoreEntrySubstate` representation


radix-engine-interface/src/api/component/kv_store_substates.rs
```diff
@@ -0,0 +1,45 @@
+use crate::api::types::*;
+use crate::data::IndexedScryptoValue;
+use radix_engine_derive::*;
+use radix_engine_interface::data::ScryptoValue;
+use sbor::rust::collections::*;
+
+#[derive(Debug, Clone, ScryptoCategorize, ScryptoEncode, ScryptoDecode, PartialEq, Eq)]
+pub enum KeyValueStoreEntrySubstate {
+    Some(ScryptoValue, ScryptoValue),
```
talekhinezh, line 9: Can we hold back on moving the Key to the value. Would like more discussion/thought on this.

iamyulong: Can we just merge it as-is for now, given it has no user-facing impact?

It might be easier to implement whatever we eventually come up with than reverting. :)

talekhinezh: Sounds good, can we then just add a small comment as I'm definitely leaning to reverting.

talekhinezh (approved):

radix-engine-interface/src/api/component/kv_store_substates.rs
iamyulong: Sure!

PR #800, opened 2023-02-03
Feature/component type info module

iamyulong:


radix-engine-interface/src/api/component_api.rs
```diff
@@ -4,6 +4,11 @@ use sbor::rust::collections::*;
 use sbor::rust::vec::Vec;
 
 pub trait ClientComponentApi<E> {
+    fn lookup_global_component(
```
talekhinezh, line 7: What is the usecase for this? Making sure that this is just a temporary interface as we remove the notion of global deref. If temporary, can we add a comment

iamyulong: For the time being, I'm using it to replace `ClientDerefApi`, by not changing all other methods to accept both `ComponentId` and `ComponentAddress`.

On the long run, I think we're aligned to update all methods to accept `handle: u32`, so this method will not be needed.

Happy to add comment about this.

talekhinezh (approved):

PR #802, opened 2023-02-03
Feature/SCRY-316 transaction limit max WASM memory

mstrug-rdx:
# Summary

Implementation of WASM memory limit per transaction execution.

# Detailed description

Added new event to `KernelWasmApi` which is called after Scrypto interpreter export invocation: `emit_wasm_post_instantiation_event` with consumed WASM invocation memory as parameter (also updated name of  existing `emit_wasm_instantiation_event` to `emit_wasm_pre_instantiation_event`). This event is forwarded to `BaseModule` newly added handler: `post_wasm_instantiation()`. Also `WasmInstance` trait has been extended with new `consumed_memory()` function in which selected WASM engine (wasmi/wasmer) returns consumed memory.

Created new Kernel module: `TransactionLimitsModule` which handles aforementioned new event and tracks current memory consumption, if limit is exceeded then transaction execution breaks with appropriate error. This module is a place for future implementations of transaction limits.

Current transaction memory limit is set to 12MiB instead of 10MiB, because of this test: `test_max_call_depth_success`. Another solution would be to decrease `DEFAULT_MAX_CALL_DEPTH` to 8 from 10, then test will fit in 10MiB limit.

# Testing

Added new test: `test_max_memory_exceeded()`.


mstrug-rdx:
refactoring is required

PR #806, opened 2023-02-06
Moving JMT's codec code to scrypto repo (using SBOR derive)

jakrawcz-rdx:
A refactoring requested at https://github.com/radixdlt/babylon-node/pull/298/files#r1097250673.

PR #807, opened 2023-02-07
feat: Add derive support for sbor(transparent)

dhedey:
I've been meaning to do this for a while - this adds the ability to use "#[sbor(transparent)]" with a single-child [new type](https://doc.rust-lang.org/rust-by-example/generics/new_types.html) - ie it basically encodes itself as its singular child, and decodes itself similarly.

This is a parallel to `repr(transparent)` and [serde(transparent)](https://serde.rs/container-attrs.html)

This will be useful for various helper types, and also for easily creating slightly more performant SBOR encodes (see eg https://github.com/radixdlt/radixdlt-scrypto/pull/806 )

### DIFF
--- a/radix-engine-interface/src/data/custom_schema.rs
+++ b/radix-engine-interface/src/data/custom_schema.rs
@@ -45,6 +45,7 @@ pub enum ScryptoCustomTypeValidation {}
 
 impl CustomTypeValidation for ScryptoCustomTypeValidation {}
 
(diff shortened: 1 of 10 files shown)


jakrawcz-rdx (approved):
Rust macros are still black magic to me, so it's hard to say if anything could be done better there. However, the tests you added are convincing and I can't wait to use this feature 👍

iamyulong (approved):
This is awesome!

PR #809, opened 2023-02-07
Unify kernel modules

iamyulong:
- All kernel modules implement a flattened `KernelModule` trait
- All kernel modules use `KernelModuleApi`
- Kernel and KernelModule integration is through `KernelModuleMixer`
- Removed kernel module states from `Track`


radix-engine-interface/src/api/event_api.rs
```diff
@@ -1,14 +1,16 @@
...
+    fn credit_cost_units(
         &mut self,
         vault_id: VaultId,
-        fee: Resource,
+        locked_fee: Resource,
         contingent: bool,
     ) -> Result<Resource, E>;
 
-    fn consume_cost_units(&mut self, units: u32) -> Result<(), E>;
+    fn update_instruction_index(&mut self, new_index: usize) -> Result<(), E>;
```
talekhinezh, line 15: Seems like a strange api? Is this temporary?

radix-engine/src/kernel/kernel.rs
```diff
@@ -50,30 +40,27 @@ pub struct Kernel<
...
+    pub(super) track: &'g mut Track<'s>,
 
     /// ID allocator
     pub(super) id_allocator: &'g mut IdAllocator,
     /// Interpreter capable of running scrypto programs
     pub(super) scrypto_interpreter: &'g ScryptoInterpreter<W>,
-    /// Kernel module
-    pub(super) module: &'g mut M,
+    /// Kernel module mixer
+    pub(super) module: KernelModuleMixer,
```
talekhinezh, line 50: Didn't want to use generics for this?

radix-engine/src/kernel/kernel.rs
```diff
@@ -50,30 +40,27 @@ pub struct Kernel<
...
 }
 
-impl<'g, 's, W, R, M> Kernel<'g, 's, W, R, M>
+impl<'g, 's, W> Kernel<'g, 's, W>
 where
     W: WasmEngine,
-    R: FeeReserve,
```
talekhinezh, line 66: Finally this is gone!

radix-engine/src/kernel/kernel.rs
```diff
@@ -82,67 +69,35 @@ where
...
-        );
-
-        kernel
+    pub fn initialize(&mut self) -> Result<(), RuntimeError> {
+        self.execute_in_mode::<_, _, RuntimeError>(ExecutionMode::KernelModule, |api| {
+            KernelModuleMixer::on_init(api)
+        })
+    }
+
+    pub fn teardown(mut self) -> (KernelModuleMixer, Option<RuntimeError>) {
```
talekhinezh, line 81: Can we add a TODO to cleanup this interface? Feels a bit weird and my guess is that there's broken abstraction somewhere.

radix-engine/src/kernel/kernel.rs
```diff
@@ -82,67 +69,35 @@ where
...
+        // Rewind call stack
+        loop {
```
talekhinezh, line 83: Do we need to verify that there's only one frame?

radix-engine/src/kernel/kernel.rs
```diff
@@ -406,104 +359,76 @@ where
...
+        let (output, update) = {
+            // Handle execution start
+            self.execute_in_mode(ExecutionMode::KernelModule, |api| {
+                KernelModuleMixer::on_execution_start(api, &caller)
+            })?;
+
+            // Run
+            let (output, mut update) =
+                self.execute_in_mode(ExecutionMode::Client, |api| executor.execute(api))?;
+
```
talekhinezh, line 394: The auto-drop locks should be moved to here. Otherwise, there's two drop locks in a row below.

radix-engine/src/kernel/kernel_api.rs
```diff
@@ -76,6 +73,27 @@ pub trait KernelNodeApi {
...
 }
 
+/// Internal API for kernel modules.
+/// No kernel state changes are expected as of a result of invoking such APIs, except updating returned references.
+pub trait KernelInternalApi {
+    fn get_module_state(&mut self) -> &mut KernelModuleMixer;
+    fn get_current_depth(&self) -> usize;
+    fn get_current_actor(&self) -> ResolvedActor;
+    fn read_bucket(&mut self, bucket_id: BucketId) -> Option<Resource>;
+    fn read_proof(&mut self, proof_id: BucketId) -> Option<ProofSnapshot>;
```
talekhinezh, line 83: Is this temporary? As buckets/proofs should not be a Kernel abstraction.

radix-engine-interface/src/api/event_api.rs
iamyulong: It is strange. 

In my follow-up PR, I rebranded it as `ClientUnsafeApi` (which enables "trusted" clients to interact with kernel modules).

radix-engine/src/kernel/kernel.rs
iamyulong: I tried this but ran into some difficulties with `get_module_state`. 

Essentially, it requires me to implement a generic module storage (which isn't useful in this stage).

radix-engine/src/kernel/kernel.rs
iamyulong: Sure

radix-engine/src/kernel/kernel.rs
iamyulong: It's checked by `if let Some(f) = self.prev_frame_stack.pop()`, unless I missed something.

radix-engine/src/kernel/kernel.rs
iamyulong: Nice catch. I will clean it up!

radix-engine/src/kernel/kernel_api.rs
iamyulong: It's all to support the optional `ExecutionTrace` kernel module. :)

radix-engine-interface/src/api/event_api.rs
talekhinezh: hrm, not sure about that abstraction, still seems a bit weird but we can discuss when we get to it.

radix-engine/src/kernel/kernel.rs
talekhinezh: Sounds good.

radix-engine/src/kernel/kernel.rs
talekhinezh: Ah, what I mean is does `teardown` need to enforce the invariant that there's only one frame left on the stack as was done previously?

radix-engine/src/kernel/kernel_api.rs
talekhinezh: Cool, can we add a comment :)

radix-engine-interface/src/api/event_api.rs
iamyulong: sure

talekhinezh (approved):

radix-engine/src/kernel/kernel.rs
iamyulong: I think it's about the lifecycle of kernel modules, and whether execution should be allowed unpon runtime error.

A typical kernel module, like `auth`/`logger`/`tx_runtime`,  would create some node `on_init` and drop it `on_teardown`.

Here, I've chosen to rewind the call stack and allowed kernel module to teardown.

But looking at my implementation more, the native call stack rewind may not be sufficient to correct the kernel state, so I will be very happy to enforce the invariants 
-  **(A) Only one frame is left on stack** 
-  **(B) No runtime error has ever occurred**.

PR #810, opened 2023-02-07
feat: Schemas can be validated

dhedey:
`Validate` is basically a big sanity-check that the schema is self-consistent. It will be used by the engine to check that user-submitted schemas are valid.


radix-engine-interface/src/data/custom_schema.rs
```diff
@@ -141,6 +143,106 @@ impl CustomTypeExtension for ScryptoCustomTypeExtension {
...
+            | ScryptoCustomTypeKind::EcdsaSecp256k1Signature
+            | ScryptoCustomTypeKind::EddsaEd25519PublicKey
+            | ScryptoCustomTypeKind::EddsaEd25519Signature
+            | ScryptoCustomTypeKind::Decimal
+            | ScryptoCustomTypeKind::PreciseDecimal
+            | ScryptoCustomTypeKind::NonFungibleLocalId
+            | ScryptoCustomTypeKind::KeyValueStore { .. } => {
+                // All other types currently should return early by having a `None` validation at the layer above,
+                // so if they get here then that's an error
+                return Err(SchemaValidationError::TypeValidationMismatch);
```
iamyulong, line 242: Shouldn't this be a `panic!("Unreachable as there is no validation for Scrypto custom types")`

sbor/src/rust.rs
```diff
@@ -1,3 +1,35 @@
...
+    pub use super::iter::FromIterator;
+
+    // And some extra useful additions we use a lot:
+    pub use super::collections::*;
+    pub use super::fmt::{Debug, Display};
+    pub use super::format;
+    pub use super::marker::PhantomData;
+    pub use super::str::FromStr;
+    pub use super::vec;
+}
```
iamyulong, line 31: This is nice!

sbor/src/schema/schema_validation/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        return Err(SchemaValidationError::DuplicateTypeHash);
+    }
+
+    let context = TypeValidationContext {
+        local_types_len: types_len,
+    };
+
+    for i in 0..types_len {
+        let NovelTypeMetadata {
+            type_hash: _, // We have already validated that the type hashes are distinct at the parent level
```
iamyulong, line 65: Could you remind me why we store `type_hash` for the final schema again?

sbor/src/schema/schema_validation/type_metadata_validation.rs
```diff
@@ -0,0 +1,185 @@
...
+        });
+    }
+    for char in name.chars() {
+        if !matches!(char, '0'..='9' | 'A'..='Z' | 'a'..='z' | '_') {
+            return Err(SchemaValidationError::InvalidTypeName {
+                // We need to validate this because we need to generate types from these names
+                // Rust is much less prescriptive for identifier names (see https://doc.rust-lang.org/reference/identifiers.html and https://unicode.org/reports/tr31/)
+                // But we can always loosen this later
+                message: "At present, types names must match [0-9A-Za-z_]+".into(),
+            });
```
iamyulong, line 139: How about not allowing type name to start with [0-9]? Want this for ease of type generation.

sbor/src/schema/schema_validation/type_metadata_validation.rs
```diff
@@ -0,0 +1,185 @@
...
+            message: "Field name cannot be more than 100 characters".into(),
+        });
+    }
+    for char in name.chars() {
+        if !matches!(char, '0'..='9' | 'A'..='Z' | 'a'..='z' | '_') {
+            return Err(SchemaValidationError::InvalidFieldName {
+                message: "At present, field names must match [0-9A-Za-z_]+".into(),
+            });
+        }
+    }
```
iamyulong, line 183: Possible to have a common function like `validate_ident()`?

sbor/src/schema/schema_validation/mod.rs
dhedey: Roughly speaking:
* To be able to recognise types as identical across-schemas
* To be able to concretely identify a type off-ledger

Neither is particularly strong justification, but it just felt sensible to do 👍

sbor/src/schema/schema_validation/type_metadata_validation.rs
dhedey: Sure - I guess they are the same here, although conceptually they could be different. Will commonise them 👍

sbor/src/schema/schema_validation/type_metadata_validation.rs
dhedey: Ahh right, yes indeed 👍. Easing type generation was the whole point of this restriction, and yet I still failed :P.

I might also restrict to avoid `_` as the first character?

radix-engine-interface/src/data/custom_schema.rs
dhedey: Hmm no - we shouldn't panic here.

A Dapp can submit a manifest that claims to be a scrypto type kind, but uses a "String" validation which is invalid.
This means that their schema isn't valid, it doesn't mean that we should panic :D

sbor/src/rust.rs
dhedey: Hopefully will save my sanity a little xD

radix-engine-interface/src/data/custom_schema.rs
iamyulong: I'm confused here.

Doesn't the interface says `_: &SchemaCustomTypeValidation<Self>`. How come a `String` validation is used at the time this function is entered. If so, it should be checked by the layer above already.
```rust
        TypeKind::Custom(custom_type_kind) => {
            let TypeValidation::Custom(custom_type_validation) = type_validation else {
                return Err(SchemaValidationError::TypeValidationMismatch);
            };
            E::validate_type_validation_with_type_kind(
                context,
                custom_type_kind,
                custom_type_validation,
            )?;
        }
```


Maybe explictly match the input validation like
```rust
match validation {
   ScryptoCustomTypeValidation::None => {
           // No-op, no validation works for all scrypto custom types
   }
}
```

sbor/src/schema/schema_validation/mod.rs
iamyulong: I have doubt about this, especially for the way our type hash is generated, but we can leave it as-is.

sbor/src/schema/schema_validation/type_metadata_validation.rs
iamyulong: 100% agree. Just for now all the different idents share the same format. The comment is more from a minimum code perspective.

sbor/src/schema/schema_validation/mod.rs
dhedey: I have it on my list to update the type hash generation as per our prior discussions 👍

radix-engine-interface/src/data/custom_schema.rs
dhedey: All of these validation methods (for consistency) match on the type_kind and then cross-check on the type validation.

Right now, there is effectively no ScryptoCustomTypeValidation - in fact I've just made it an empty enum to be clear. So this method can never be called, and can be optimised away by the compiler.

BUT we will be adding them soon. So, I wrote the method to be forward-compatible.

BUT this confusion suggests to me that we should be passing in `SchemaTypeValidation<SchemaCustomTypeValidation<Self>>` instead perhaps?

dhedey: I've updated the comments and will merge. But happy to adjust if you think we should be passing the full set of validations down to the custom validator 👍

dhedey: In fact, possibly the custom validator should actually check any custom validation against any basic or custom type??

And the basic validator just checks the basic validations against the basic types. Maybe that's the better abstraction, as it allows the extension to add its own custom validations of basic types, which might be powerful?

PR #813, opened 2023-02-08
Architecture: Refactor Native Function Invocations

talekhinezh:


radix-engine-tests/tests/clock.rs
```diff
@@ -1,10 +1,9 @@
 use radix_engine::errors::{ModuleError, RuntimeError};
 use radix_engine::types::*;
 use radix_engine_interface::api::kernel_modules::auth::AuthAddresses;
-use radix_engine_interface::blueprints::clock::ClockCreateInvocation;
 use scrypto_unit::*;
 use transaction::builder::ManifestBuilder;
-use transaction::model::{Instruction, SystemTransaction};
+use transaction::model::{BasicInstruction, Instruction, SystemTransaction};
```
iamyulong, line 6: I see we're close to removing `SystemInstruction` lol

radix-engine/src/blueprints/transaction_processor/executables.rs
```diff
@@ -208,13 +242,8 @@ fn instruction_get_update(instruction: &Instruction, update: &mut CallFrameUpdat
             | BasicInstruction::PublishPackage { .. }
             | BasicInstruction::PublishPackageWithOwner { .. }
             | BasicInstruction::BurnResource { .. }
-            | BasicInstruction::CreateFungibleResource { .. }
             | BasicInstruction::CreateFungibleResourceWithOwner { .. }
```
iamyulong, line 245: Why do we still have `create_with_owner` here?

radix-engine/src/kernel/interpreters/scrypto_interpreter.rs
```diff
@@ -232,6 +280,105 @@ impl Executor for ScryptoExecutor {
...
+                let invocation: AccessControllerCreateGlobalInvocation =
+                    scrypto_decode(&scrypto_encode(&self.args).unwrap()).unwrap();
+                let rtn = invocation.execute(api)?;
+                return Ok((
+                    scrypto_decode(&scrypto_encode(&rtn.0).unwrap()).unwrap(),
+                    rtn.1,
+                ));
+            }
+            RESOURCE_MANAGER_PACKAGE => match self.export_name.as_str() {
+                "create_non_fungible" => {
```
iamyulong, line 331: Any reason for not using enum and derive string?

radix-engine/src/kernel/interpreters/scrypto_interpreter.rs
```diff
@@ -232,6 +280,105 @@ impl Executor for ScryptoExecutor {
             + ClientStaticInvokeApi<RuntimeError>,
         W: WasmEngine,
     {
+        match self.package_address {
+            IDENTITY_PACKAGE => {
+                let invocation: IdentityCreateExecutable =
```
iamyulong, line 285: Feel like we should check the export even if this is only one function.

radix-engine/src/system/kernel_modules/fee/fee_table.rs
```diff
@@ -127,14 +126,15 @@ impl FeeTable {
...
-                ClockFn::Create => self.fixed_low,
                 ClockFn::SetCurrentTime => self.fixed_low,
                 ClockFn::GetCurrentTime => self.fixed_high,
                 ClockFn::CompareCurrentTime => self.fixed_high,
             },
+            /*
             NativeFn::Identity(identity_fn) => match identity_fn {
                 IdentityFn::Create => self.fixed_low,
             },
+             */
```
iamyulong, line 137: How do we cost native function?

iamyulong (commented):
Nice, it's actually simpler.

radix-engine-interface/src/api/types/re_node.rs
```diff
@@ -212,6 +212,7 @@ impl Into<ResourceAddress> for GlobalAddress {
     ScryptoDecode,
 )]
 pub enum NodeModuleId {
+    TypeInfo,
```
iamyulong, line 215: Can we start with specific `PackageTypeInfo`  and consider generalizing later? I'm adding `ComponentTypeInfo` in another PR.

radix-engine-interface/src/blueprints/access_controller/invocations.rs
```diff
@@ -4,36 +4,28 @@ use crate::blueprints::resource::*;
...
 // Access Controller Create Global
 //=================================
 
-#[derive(Debug, Clone, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
-pub struct AccessControllerCreateGlobalInvocation {
-    pub controlled_asset: BucketId,
+pub const ACCESS_CONTROLLER_CREATE_GLOBAL_IDENT: &str = "create_global";
+
+#[derive(Debug, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+pub struct AccessControllerCreateGlobalInput {
```
iamyulong, line 16: `XyzInput` is back lol

radix-engine-interface/src/blueprints/access_controller/invocations.rs
```diff
@@ -4,36 +4,28 @@ use crate::blueprints::resource::*;
...
+pub const ACCESS_CONTROLLER_BLUEPRINT: &str = "AccessController";
+
 //=================================
 // Access Controller Create Global
 //=================================
 
-#[derive(Debug, Clone, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
-pub struct AccessControllerCreateGlobalInvocation {
-    pub controlled_asset: BucketId,
+pub const ACCESS_CONTROLLER_CREATE_GLOBAL_IDENT: &str = "create_global";
```
iamyulong, line 13: Thoughts on using enum and statically derive the string using `strum`? or have an additional enum capturing all these variants.

I'm slightly worried that costing and dispatcher can easily go out of sync.

radix-engine-tests/tests/blueprints/non_fungible/src/lib.rs
```diff
@@ -301,16 +302,22 @@ mod non_fungible_test {
...
 
             // creating non-fungible id with id type set to default (UUID)
-            let (_, bucket) = ScryptoEnv
-                .invoke(
-                    ResourceManagerCreateNonFungibleWithInitialSupplyInvocation {
+            let rtn = ScryptoEnv
+                .call_function(
+                    RESOURCE_MANAGER_PACKAGE,
+                    RESOURCE_MANAGER_BLUEPRINT,
+                    "create_non_fungible_with_initial_supply",
```
iamyulong, line 309: Use the constant?

iamyulong (commented):
It does look nice!

A step closer from `as-a` to `is-a` blueprint.

radix-engine-interface/src/blueprints/access_controller/invocations.rs
iamyulong: Also, how do you think of traits, which has been fantastic for static invocation so far?

```rust
trait NativeFunctionInput: ScryptoEncode + ScryptoDecode {
     const package_address: PackageAddress;
     const fn_ident: &'str;
     type Output;
} 
```

In `PackageApi`, we have two methods 
```rust
trait PackageApi {
    fn call_function(address, ident, args) -> Vec<u8>;
    
    fn call_function_typed<T: NativeFunctionInput>(input) -> T:Output {
         let output = self.call_function(input.package_address, input.fn_ident, scrypto_encode(&input));
         scrypto_decode(&output)
    } 
}
```

iamyulong: For Scrypto blueprint in future, we may automatically implement this trait for the stubs as well.

radix-engine-interface/src/api/types/re_node.rs
talekhinezh: Sounds good

radix-engine-interface/src/blueprints/access_controller/invocations.rs
talekhinezh: lol give it 2 months and `Invocation` will be back :)

radix-engine-interface/src/blueprints/access_controller/invocations.rs
talekhinezh: hrm, I've thought about it but decided not to go down enum path as that would go back to treating the set of native methods as a closed set rather than an open set, which seems to be against the new philosophy. I think the syncing issue between these identifiers should be resolved by having additional state per method rather then relying on an enum. Happy to chat about this though.

talekhinezh: The traits stuff looks cool! Will save that though for a future PR since it doesn't affect the underlying architecture.

radix-engine-tests/tests/blueprints/non_fungible/src/lib.rs
talekhinezh: good catch!

radix-engine-interface/src/blueprints/access_controller/invocations.rs
talekhinezh: On the other hand, rather than relying on traits I think we should think about how this looks like with code generation reliant on abis.

iamyulong (approved):

PR #815, opened 2023-02-09
Clean up Kernel/Client impl

iamyulong:
Some Kernel API clean-up:
- Made all fields in `Kernel` private again 
   - This is to prevent unintended access from System
- `ClientAPI` is now 100% based on `KernelAPI` (except static kernel module invocation, see `ClientEventApi impl`)
- Prefixed all methods under `KernelApi` with `kernel_`
   - This is to mitigate accidental recursive calls in impls
- Untangled mods under `kernel` for clearer dependency graph
    - No more `use crate::kernel::*`


radix-engine/src/system/node_properties.rs
```diff
@@ -103,18 +104,17 @@ impl VisibilityProperties {
 
         // TODO: Cleanup and reduce to least privilege
         match (mode, offset) {
-            (ExecutionMode::Kernel, ..) => false, // Protect ourselves!
+            (ExecutionMode::Kernel, offset) => match offset {
+                SubstateOffset::Global(GlobalOffset::Global) => read_only,
+                _ => false, // Protect ourselves!
+            },
```
iamyulong, line 110: @talekhinezh I removed `Deref` execution mode so to get rid of the two public `deref` methods in `Kernel`, understanding that we're going to ditch the deref idea anyway.

talekhinezh (approved):

PR #827, opened 2023-02-14
Feature/SCRY-316 transaction limit max WASM memory

mstrug-rdx:
# Summary

Implementation of WASM memory limit per transaction execution and per WASM instance.

# Detailed description

Created new Kernel module: `TransactionLimitsModule` which uses `before_push_frame` and `after_pop_frame` events to maintain internal call frame stack with WASM memory consumption for each call frame. Added new event: `after_wasm_instantiation` which provides memory consumption of WASM instance after instance execution is finished. Module tracks current memory consumption, if limit is exceeded (per WASM instance or per whole transaction) then transaction execution breaks with appropriate error.

To properly handle nested WASM calls, `WasmRuntime` trait has been extended with `set_wasm_memory_consumption()` function, which is called just before nested function call (in `call_function()` wasmi/wasmer implementations), this call updates currently consumed WASM memory in `CallFrame` struct through `ClientEventApi`. Then this value is updated in `before_push_frame` event handler in `TransactionLimitsModule`.

Current transaction memory limit is set to 10 MiB, instance limit is set to 3 MiB.

# Testing

Added new tests: `test_max_instance_memory_exceeded()` and `test_max_transaction_memory_exceeded()`.


radix-engine/src/kernel/call_frame.rs
```diff
@@ -100,6 +100,9 @@ pub struct CallFrame {
 
     next_lock_handle: LockHandle,
     locks: HashMap<LockHandle, SubstateLock>,
+
+    /// WASM instance consumed memory for this call frame
+    pub consumed_wasm_memory: usize,
```
iamyulong, line 105: Can we move this to the module state? It shouldn't be accesssed anywhere other than kernel itself.

radix-engine/src/kernel/kernel_client_facade.rs
```diff
@@ -438,6 +438,10 @@ where
     fn update_instruction_index(&mut self, new_index: usize) -> Result<(), RuntimeError> {
         KernelModuleMixer::on_update_instruction_index(self, new_index)
     }
+
+    fn wasm_memory_consumption(&mut self, size: usize) {
+        self.current_frame.consumed_wasm_memory = size;
```
iamyulong, line 443: ClientAPI implementation shouldn't be accessing `current_frame` directly. You will find this stop working shortly after merging https://github.com/radixdlt/radixdlt-scrypto/pull/815

radix-engine/src/kernel/kernel_main_facade.rs
```diff
@@ -415,10 +419,15 @@ where
...
+    fn emit_wasm_pre_instantiation_event(&mut self, code: &[u8]) -> Result<(), RuntimeError> {
+        KernelModuleMixer::before_wasm_instantiation(self, code)
+    }
 
-        Ok(())
+    fn emit_wasm_post_instantiation_event(
+        &mut self,
+        consumed_memory: usize,
+    ) -> Result<(), RuntimeError> {
+        KernelModuleMixer::after_wasm_instantiation(self, consumed_memory)
```
iamyulong, line 430: I intend to remove these wasm-specific events in favor of the `ClientEventApi` (to be renamed to `ClientUnsafeApi`). Why do we need to emit post-instantiation event?

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
```diff
@@ -0,0 +1,114 @@
...
+    MaxWasmInstanceMemoryExceeded(usize),
+}
+
+/// Tracks and verifies transaction limits during transactino execution,
+/// if exceeded breaks execution with appropriate error.
+pub struct TransactionLimitsModule {
+    /// Maximum WASM memory which can be consumed during transaction execution.
+    max_wasm_memory: usize,
+    /// Maximum WASM memory which can be consumed during transaction execution.
+    max_wasm_instance_memory: usize,
```
iamyulong, line 26: Suggest renaming to `max_wasm_memory_per_call_frame`

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Consumed WASM memory for each invocation call.
+    wasm_memory: Vec<usize>,
```
iamyulong, line 28: Suggest renaming to `wasm_memory_usage_stack`

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    fn after_pop_frame<Y: KernelModuleApi<RuntimeError>>(api: &mut Y) -> Result<(), RuntimeError> {
+        // pop from stack
+        api.get_module_state().transaction_limits.wasm_memory.pop();
+
+        Ok(())
+    }
+
+    fn after_wasm_instantiation<Y: KernelModuleApi<RuntimeError>>(
```
iamyulong, line 95: Should we listen to `ExecutionFinish` instead? More memory might have been allocated after wasm execution.

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
```diff
@@ -0,0 +1,114 @@
...
+            max_wasm_instance_memory,
+            wasm_memory: Vec::with_capacity(8),
+        }
+    }
+
+    /// Checks if maximum WASM memory limit for one instance was exceeded and then
+    /// checks if memory limit for all instances was exceeded.
+    fn validate(&self) -> Result<(), RuntimeError> {
+        // check last (current) call frame
+        let current_instance_memory = *self.wasm_memory.last().unwrap();
```
iamyulong, line 44: Worth changing `unwrap` to `exepct` with specific reason/assumption.

radix-engine-tests/tests/wasm_metering.rs
```diff
@@ -185,3 +195,82 @@ fn test_grow_memory_out_of_cost_unit() {
...
+
+    // Assert
+    receipt.expect_specific_failure(|e| {
+        matches!(
+            e,
+            RuntimeError::ModuleError(ModuleError::TransactionLimitsError(
+                TransactionLimitsError::MaxWasmMemoryExceeded(_)
+            ))
+        )
+    })
```
iamyulong, line 275: Can we also add a test to assert the exact number of wasm memory usage? I want to make our memory measurements are implementation-independent and works deterministically between wasmer and wasmi.

iamyulong (commented):
Nice work!

radix-engine/src/wasm/wasmer.rs
```diff
@@ -206,6 +216,11 @@ impl WasmerModule {
             let ident = read_memory(&instance, ident_ptr, ident_len)?;
             let args = read_memory(&instance, args_ptr, args_len)?;
 
+            // Get current memory consumption and set it in current call frame through api.
+            runtime
+                .set_wasm_memory_consumption(get_memory_size(&instance)?)
+                .map_err(|e| RuntimeError::user(Box::new(e)))?;
```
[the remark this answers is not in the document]
mstrug-rdx: We need to store/check what is the current wasm instance memory usage (call frame 1) before starting nested call to the next wasm instance (call frame 2). Without reading current memory before starting next call frame, we will get information of wasm consumed memory in `after_wasm_instantiation` event (for call frame 1) which will be called after nested call finishes (call frame 2). We will have the consumed memory information too late to properly validate the limit and break transaction execution.

radix-engine/src/kernel/call_frame.rs
mstrug-rdx: done

radix-engine/src/kernel/kernel_client_facade.rs
mstrug-rdx: Changed that to use `KernelModuleMixer::on_update_wasm_memory_usage`.

radix-engine/src/kernel/kernel_main_facade.rs
```diff
@@ -206,6 +206,10 @@ where
         self.current_frame.actor.clone()
     }
 
+    fn get_current_wasm_memory_consumption(&self) -> usize {
```
iamyulong, line 209: Ah, looks like I accidentally commited this unneeded file. Pushed an emergency commit to remove it from `develop`.

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
mstrug-rdx: At execution finish stage there is no wasm instance object, so we cannot get information about consumed memory from it. Implemented functionality is counting only wasm memory, so if some memory is allocated outside wasm instance (after it has been executed) then it is not relevant to current implementation.

radix-engine/src/kernel/kernel_main_facade.rs
iamyulong: Cool, thanks!

Can we please have consistent naming for the following? 
- `set_wasm_memory_consumption` -> `update_wasm_memory_usage`
- `wasm_memory_consumption` -> `update_wasm_memory_usage`
- `on_update_wasm_memory_usage` -> `on_update_wasm_memory_usage`  (no change)

radix-engine/src/wasm/wasmer.rs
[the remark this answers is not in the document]
iamyulong: In that case, let's notify the transaction limits module in both `call_method` and `call_function`.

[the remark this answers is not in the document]
iamyulong: Another point for future thinking - do we really need wasm memory usage limiting if we can't apply real-time WASM limiting?

radix-engine/src/kernel/kernel_main_facade.rs
mstrug-rdx: done

radix-engine/src/wasm/wasmer.rs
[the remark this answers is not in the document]
mstrug-rdx: Current implementation is not real-time, but it can verify memory size in these two cases (before nested call and at invocation end) I think node can use this mechanism during preparation step and discard transaction before execution, maybe @malexandru-rdx can say something more about use of this limitation mechanism on node side.

I was looking to the wasm implementation and maybe we can use maximum number of memory pages which wasm `grow` function can extend. In wat files in `radix-engine-test` folder we are declaring memory like this: `(memory $0 1)`, but we can add third parameter to limit maximum count of pages which wasmi implementation (wasmer to check) validates on each call to `grow`: `(memory $0 1 5)` - this lets to extend memory up to 5 pages. I am not sure how we can add this mechanism to Scrypto compiled files, but maybe it is something worth to try.

Another thing which we can do is to create `grow` function instrumentation, similar to instrumentation of wasmi call stack height from `wasm-instrument` crate.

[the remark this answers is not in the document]
mstrug-rdx: In `enforce_memory_limit()` in prepare.rs file we are checking wasm pages limits, but only initial count, we are not checking maximum wasm pages count, so in this place we can add validation if that value is set and has proper value.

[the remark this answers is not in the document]
iamyulong: That's an interesting idea and we should explore more.

If we inject a max memory size during preparation, we can have a nice upper bounds of memory usage per instance, and we don't need to track per-call-frame usage as well.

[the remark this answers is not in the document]
mstrug-rdx: As we discussed I've reverted max wasm memory limit (moved implementation to branch: `experimental/wasm_max_memory_limit`.

radix-engine-tests/tests/wasm_metering.rs
mstrug-rdx: Added exact values of exceeded memory calculation and validation. Same values for wasmi and wasmer.

iamyulong (approved):

PR #829, opened 2023-02-14
Splitting the hash tree into nested trees

jakrawcz-rdx:
This implements a "nested" hash tree - as promised at https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/2939650077/RNP-6+-+State+Transaction+Inclusion+Exclusion+Verification#State-JMT%E2%80%99s-structure, but now taking the newly-added `NodeModuleId` into account.
The upper layer is addressed by `hash(ReNodeId + NodeModuleId)`; the lower layer is addressed by `hash(SubstateOffset)`.

This change does NOT touch the incorporated open-source "core" JMT impl - the split into layers is handled by our facade.


radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
+        current_version,
+        current_version.unwrap_or(0) + 1,
+        nested_root_changes
+            .into_iter()
+            .map(|change| to_re_node_change(change))
+            .collect(),
+    )
+}
+
+// only internals below
```
jakrawcz-rdx, line 84: @talekhinezh and @dhedey please let me know if you prefer to have an extensive doc for every struct/function below (it is not easy to tell for me [anymore] which parts are non-obvious, so I documented the public api and left out the internals).

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
+            re_node_id,
+            node_module_id,
+        }
+    }
+}
+
+fn index_by_re_node_module(
+    changes: Vec<SubstateHashChange>,
+) -> HashMap<ReNodeModule, Vec<IdChange<SubstateOffset, Hash>>> {
+    let mut by_re_node_module = HashMap::<ReNodeModule, Vec<IdChange<SubstateOffset, Hash>>>::new();
```
dhedey, line 106: We need to avoid HashMap because it creates non-deterministic receipts when iterated over - either we should use IndexMap (more performant, and doesn't require `Ord` on the key) or BTreeMap.

dhedey: I guess given that it's in a private function and we don't expose `HashMap` on the interface, we might be confident it's not used in a way which creates non-determinism, but we agreed it wasn't worth the mental overhead and to simply avoid it going forward 👍

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
-                value_hash.map(|value_hash| (value_hash, id.clone())),
-            )
+    let changes_by_re_node_module = index_by_re_node_module(changes);
+    let mut nested_root_changes = Vec::new();
+    for (re_node_module, substate_changes) in changes_by_re_node_module {
+        let nested_root =
+            put_substate_changes(store, current_version, &re_node_module, substate_changes);
+        nested_root_changes.push(IdChange::new(re_node_module, nested_root));
+    }
+    put_changes(
```
dhedey, line 73: Perhaps `put_re_node_changes` might be clearer, for symmetry with `put_substate_changes`? 👍

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
+/// All nodes that became stale precisely due to this (i.e. not any previous) operation will be
+/// reported before the function returns (see `WriteableTreeStore::record_stale_node`).
+/// Returns the hash of the newly-created root (i.e. representing state at version
+/// `current_version` + 1).
 ///
 /// # Panics
-/// Panics if a root node for `current_version` does not exist. The caller
-/// should use `None` to denote an empty, initial state of the tree (i.e.
-/// inserting at version 1).
+/// Panics if a root node for `current_version` does not exist. The caller should use `None` to
```
dhedey, line 59: I think the API would be nicer if we don't pass `current_version: None`, and instead there was a separate method to create the pre-genesis tree at state version 0?

Then we pass `current_version: 0` in here for genesis?

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
+    node: TreeNode<P>,
+}
+
+fn put_substate_changes<S: TreeStore>(
+    store: &mut S,
+    current_version: Option<Version>,
+    re_node_module: &ReNodeModule,
+    changes: Vec<IdChange<SubstateOffset, Hash>>,
+) -> Option<TreeRoot<SubstateOffset>> {
+    let substate_entry = get_substate_entry(store, current_version, re_node_module);
```
dhedey, line 128: I don't understand the name `substate_entry` isn't this more `get_leaf_in_re_node_module_tree`?

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
+    let (substate_version, substate_root) = break_tuple(substate_entry);
```
dhedey, line 129: Ditto - this isn't the substate version? I'd suggest perhaps `(subtree_last_update_state_version, substates_subtree_root)`

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
+        current_version.unwrap_or(0) + 1,
+        changes
+            .into_iter()
+            .map(|change| to_substate_change(change))
+            .collect(),
+    );
+    let substate_root_node = substate_tree_store
+        .extract_new_root()
+        .map_payload(|payload| match payload {
+            Payload::ReNodeModule(_) => panic!("unexpected payload in substate layer"),
```
dhedey, line 143: I'm a little confused about why we have this `Payload` enum? In the "top" tree, I imagine we just store `ReNodeModuleRoot`s, and in the bottom tree, I imagine we just store `Substate`s - so could this type not be fixed at compile time (as eg a generic parameter of the ReNodeModuleTreeStore / SubstateSubtreeStore)

dhedey: Or do we store both types in the same tree?

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
+    next_version: Version,
+    changes: Vec<LeafChange>,
+) -> Hash {
     let (root_hash, update_result) = JellyfishMerkleTree::new(store)
         .batch_put_value_set(
-            value_set
+            changes
                 .iter()
-                .map(|(x, y)| (x.clone(), y.as_ref()))
+                .map(|change| (change.key_hash.clone(), change.new_payload.as_ref()))
```
dhedey, line 192: If you use `into_iter()` in the line above, you should be able to avoid the clone.

dhedey: Or is the reference required for the new payload?

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -20,55 +23,289 @@ mod test;
...
+        } else {
+            self.underlying.insert_node(self.prefixed(&key), node);
+        }
+    }
+
+    fn record_stale_node(&mut self, key: NodeKey) {
+        if key.nibble_path().is_empty() {
+            self.current_root = None;
+        } else {
+            self.underlying.record_stale_node(self.prefixed(&key));
```
dhedey, line 308: Hmmm this is something I didn't understand initially -- I assumed we'd have both a leaf node in the parent tree, and a child root node - but it sounds like maybe the leaf node in the parent tree is a wrapper around the root of the child tree? In which case, this is awesome as it's more efficient from a node storage perspective :D.

But perhaps it's something worth adding to the Rust docs more explicitly?

radix-engine-stores/src/hash_tree/mod.rs
dhedey: Ahh wait - I guess this is because we wish to serialize both types in the same underlying node DB family?

Potentially an enum is still not required, and we can just know which leaf we're expecting from which tree we're in?

But I guess having an enum is useful anyway in case we want to change the payload in the future. So I don't feel super strongly 👍

radix-engine-stores/src/hash_tree/tree_store.rs
```diff
@@ -41,25 +41,45 @@ pub struct TreeChildEntry {
...
+    ReNodeModule(ReNodeModulePayload),
+    /// Payload specific to the Substate layer.
+    Substate(SubstateOffset),
+}
+
 /// The "read" part of a physical tree node storage SPI.
 pub trait ReadableTreeStore {
     /// Gets node by key, if it exists.
-    fn get_node(&self, key: &NodeKey) -> Option<TreeNode>;
+    fn get_node(&self, key: &NodeKey) -> Option<TreeNode<Payload>>;
```
dhedey, line 76: Just to explain what I meant earlier, this could be a `ReadableTreeStore<P: Payload>` and it could return `TreeNode<P>`, and then:

With eg:
```
trait Payload: Clone + PartialEq + Eq + Hash + Debug + Categorize + Encode + Decode {}
```

And eg
```
impl<P: Payload> ReadableTreeStore<P> for X {
...
}
```

dhedey (approved):
Really nice work! A few minor suggestions, but this is a really neat solution, and surprisingly little code to achieve it.

radix-engine-stores/src/hash_tree/tree_store.rs
jakrawcz-rdx: I see what you mean, but then the `put_at_next_version` would have to take 2 `TreeStore` args in: one for ReNodePayload, one for SubstateOffset. Or a single one with a very weird type constraints? 🤔 Is this what you mean?

radix-engine-stores/src/hash_tree/mod.rs
jakrawcz-rdx: > Potentially an enum is still not required, and we can just know which leaf we're expecting from which tree we're in?

But how to do it ("expect") without an enum?

radix-engine-stores/src/hash_tree/mod.rs
jakrawcz-rdx: > in a private function and we don't expose HashMap on the interface, we might be confident it's not used in a way which creates non-determinism

That gives 0% guarantee on non-det :( (imagine rewriting something from that hashmap into a vec before returning - effectively it shuffles the results)
So I agree, if we are on a war against non-det, then let me switch to `IndexMap` everywhere 👍

radix-engine-stores/src/hash_tree/mod.rs
jakrawcz-rdx: hmm, not that easy... a "pre-genesis tree" is literally "empty TreeStore" (not `NullNode`, nothing).

What you ask for is to "treat version=0 as pre-genesis".
This makes sense for the ReNode layer (because versions only ever advance there by 1), but please note it is no longer true for SubstateOffset layer (some substate may start at v=8, and it doesn't say anything about what previous version exists, if any). To me `None` is a very good indication of "no previous version exists". `0` would be more of a "hacky marker/sentinel" for that. Also, the original opensource authors use `None`, if that means anything (yea their API is not an ideal example, but still).

radix-engine-stores/src/hash_tree/mod.rs
jakrawcz-rdx: you are right, `substate_entry` is ambiguous; let's go for `get_re_node_module_leaf_entry` (because I am getting the leaf node + its version = "an entry").

radix-engine-stores/src/hash_tree/mod.rs
jakrawcz-rdx: if you are talking about "just rename": `put_changes()` is called both to put the ReNode changes and the substate changes (it is called by `put_substate_changes`).

But actually I can introduce the mid-level `put_re_node_changes()` (to call `put_changes()`); PTAL if it simplified anything.

radix-engine-stores/src/hash_tree/mod.rs
jakrawcz-rdx: the reference is required by the ultra-weird API of that low-level, opensource JMT method: `batch_put_value_set(..., value_set: Vec<(Hash, Option<&(Hash, K)>)>, ...)`.
Maybe they had a reason, maybe their `K` can weight megabytes - I don't know and I am not touching that :p
If it helps visually, that `.clone()` call can actually be skipped :P (and maybe the compiler optimizes it away?)

radix-engine-stores/src/hash_tree/mod.rs
jakrawcz-rdx: > but it sounds like maybe the leaf node in the parent tree is a wrapper around the root of the child tree? In which case, this is awesome as it's more efficient from a node storage perspective :D.

Yes, it's exactly like that.
It saves ~0.1% of storage _(number not actually measured)_ and also saves 1 hop during tree traversal _(number actually exact)_.

But I did it actually to avoid a possible awkward DB key clash:
Imagine we have a ReNode leaf of hash(ReNodeId+ModuleId) = `0xa1b2c3`.
And unfortunately a sibling hash also exists (let's say `0xa1b2c4`).
It means the ReNode leaf will have to use a all nibbles of the hash in its DB key, like `key=v7|a1b2c3`.
And guess what is the DB key of a root node of the substates' subtree of that ReNode? yup, it is `key=v7|a1b2c3` (i.e. an empty nibble path [as for every root], prefixed by its ReNode's full nibble path). 
This has very low probability, but it hints that conceptually the ReNode leaf has the same identity as the Substate's subtree root. So I adopted that concept.

_(of course the unlikely key clash could be battled e.g. by adding some separator nibble, or a ReNode/Substate discriminator nibble at the beginning, but these are solutions from the "ugly sentinel" territory)._

jakrawcz-rdx: I added Rust doc (on `ReNodeModulePayload`) for it, PTAL.

dhedey: It totally makes sense that they're the same thing, I really like it tbh :D

radix-engine-stores/src/hash_tree/mod.rs
dhedey: See example PR: https://github.com/radixdlt/radixdlt-scrypto/pull/835/files - feel free to merge it in if you like it

radix-engine-stores/src/hash_tree/mod.rs
dhedey: I agree it makes sense for the subtrees 👍. I'm happy to keep it for the root tree for ease/consistency, it just gave me a slight wtf before I read the Rust Docs :)

radix-engine-stores/src/hash_tree/tree_store.rs
dhedey: https://github.com/radixdlt/radixdlt-scrypto/pull/835/files is what I meant :)

jakrawcz-rdx:
@dhedey I like your PR (merged into this one already) really, a lot. No `panic!("unexpected payload type")` is gold.
The only thing that I found weird was the sheer existence of `Payload` trait. I mean: in Java it could be considered nice, as a "marker interface". But in Rust - as I understand it - it is not even "an interface", but a set of constraints, and with constraints I prefer to be more selective (i.e. apply them when they matter).
Please see the newest commit (https://github.com/radixdlt/radixdlt-scrypto/pull/829/commits/555247b2738b5cb055e35a6365977797f32450ba) - I hope you are fine with this.

dhedey:
> @dhedey I like your PR (merged into this one already) really, a lot. No `panic!("unexpected payload type")` is gold.
> The only thing that I found weird was the sheer existence of `Payload` trait. I mean: in Java it could be considered nice, as a "marker interface". But in Rust - as I understand it - it is not even "an interface", but a set of constraints, and with constraints I prefer to be more selective (i.e. apply them when they matter).
> Please see the newest commit ([555247b](https://github.com/radixdlt/radixdlt-scrypto/commit/555247b2738b5cb055e35a6365977797f32450ba)) - I hope you are fine with this.

You're totally right with this being a standard design philosophy in rust, particularly with respect to making things easy to consume as a library. EG std is full of impl blocks as you write them, and we've worked very hard in SBOR to also do this for flexibility's sake/

Having said that, quite often, I find having an explicit trait (quite similar to Rust marker traits in core) can provide benefits to semantics, and can improve code readability. It can be a mechanism of explicitly labelling a type as "for a particular purpose", and can also be a way of wrapping up expected type bounds and avoid confusing issues - eg it says "hey - if you want to be used for this, you need to implement all these traits" - in the more flexible "constraints on impls" case, you just end up missing lots of trait impls, and then need to spend time digging through code to realise that you're eg missing ScryptoEncode, which is why you can't be written to a database.

Sure, it's less flexible, but it can be easier to reason about for domain-specific implementations. In this case, we have created a code facade _just for our Radix use case_, and I think it's reasonable to be very clear about the two types that we expect to be payloads in our merkle trees.

Just wanted to explain my reasoning. I don't feel strongly about it at all in this case, so I'm happy for us to ditch it, and merge as-is :D.

jakrawcz-rdx:
> Just wanted to explain my reasoning in case you're curious.

Ok, I can live with that `Payload` if Rust does "marker traits" too 👍 
(ofc it will be the first line I remove if it causes me any problems in future 😈 )

jakrawcz-rdx:
I brought back the `Payload` and I switched to `IndexMap` (instead of `BTreeMap`) 👍 

Let me merge this before another update from develop comes :P

PR #830, opened 2023-02-15
Change default hashing algorithm to Blake2b

lrubasze:
- switch to `blake2b` as a default hash
- sign message hash for both `EdsaSecp256k1` and `EddsaEd25519`
in case of `EdsaSecp256k1` double `sha256` has been removed
- removed `sha2` and `sha3` dependencies from the repository
- tests aligned to the above changes


transaction/src/signing/signer.rs
```diff
@@ -1,18 +1,21 @@
 use super::{EcdsaSecp256k1PrivateKey, EddsaEd25519PrivateKey};
 use crate::model::SignatureWithPublicKey;
+use radix_engine_interface::crypto::hash;
 
 pub trait Signer {
     fn sign(&self, message: &[u8]) -> SignatureWithPublicKey;
```
iamyulong, line 6: Change interface to `message_hash: &Hash` instead?

transaction/src/signing/ecdsa_secp256k1.rs
```diff
@@ -11,9 +11,8 @@ impl EcdsaSecp256k1PrivateKey {
         EcdsaSecp256k1PublicKey(PublicKey::from_secret_key_global(&self.0).serialize())
     }
 
-    pub fn sign(&self, msg: &[u8]) -> EcdsaSecp256k1Signature {
-        let h = sha256(sha256(msg));
-        let m = Message::from_slice(&h.0).expect("Hash is always a valid message");
+    pub fn sign(&self, msg_hash: &[u8]) -> EcdsaSecp256k1Signature {
```
iamyulong, line 14: Change interface to `message_hash: &Hash` instead?

transaction/src/validation/signature_validator.rs
```diff
@@ -3,25 +3,26 @@ use radix_engine_interface::crypto::*;
...
-            recover_ecdsa_secp256k1(message, signature).map(Into::into)
+            recover_ecdsa_secp256k1(&message_hash, signature).map(Into::into)
         }
         SignatureWithPublicKey::EddsaEd25519 { public_key, .. } => Some(public_key.clone().into()),
     }
 }
 
 pub fn recover_ecdsa_secp256k1(
-    message: &[u8],
+    message_hash: &[u8],
```
iamyulong, line 17: Same for all in this file.

iamyulong (commented):

transaction/src/signing/ecdsa_secp256k1.rs
0xOmarA: To be consistent with the interface of `EddsaEd25519PrivateKey` I would recommend that the passed argument is the message and not the message hash.

lrubasze: > To be consistent with the interface of EddsaEd25519PrivateKey I would recommend that the passed argument is the message and not the message hash.
Actually, it should be message hash in `EcdsaSecp256k1Signature`.  Thanks for spotting this @0xOmarA

transaction/src/validation/signature_validator.rs
lrubasze: Sure, will change it accordingly

iamyulong (approved):

PR #831, opened 2023-02-16
Architecture: Refactor transient component methods

talekhinezh:


radix-engine-constants/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 //==========================
 
 /// The default cost unit limit.
-pub const DEFAULT_COST_UNIT_LIMIT: u32 = 100_000_000;
+pub const DEFAULT_COST_UNIT_LIMIT: u32 = 500_000_000;
```
iamyulong, line 6: Downstream has been relying on it being 100M for now. Is there a particular reason that we want to increase it even more, before metering fine-tuning?

radix-engine-interface/src/api/node_modules/mod.rs
```diff
@@ -1,2 +1,3 @@
 pub mod auth;
 pub mod metadata;
+pub mod royalty;
```
iamyulong, line 3: Thanks for moving royalty to the proper place!

radix-engine-interface/src/api/package/substates.rs
```diff
@@ -11,6 +11,9 @@ pub const EPOCH_MANAGER_PACKAGE_CODE_ID: u8 = 2u8;
 pub const CLOCK_PACKAGE_CODE_ID: u8 = 3u8;
 pub const ACCOUNT_PACKAGE_CODE_ID: u8 = 4u8;
 pub const ACCESS_CONTROLLER_PACKAGE_CODE_ID: u8 = 5u8;
+pub const LOGGER_CODE_ID: u8 = 6u8;
+pub const TRANSACTION_RUNTIME_CODE_ID: u8 = 7u8;
+pub const AUTH_ZONE_CODE_ID: u8 = 8u8;
```
iamyulong, line 16: We don't have a `_PACKAGE_CODE_ID` because they're not a package?

radix-engine-interface/src/api/types/actor.rs
```diff
@@ -87,31 +83,20 @@ impl ScryptoFnIdentifier {
 )]
 pub enum NativeFn {
     AccessRulesChain(AccessRulesChainFn),
-    Component(ComponentFn), // TODO: investigate whether to make royalty universal and take any "receiver".
+    ComponentRoyalty(ComponentRoyaltyFn),
```
iamyulong, line 86: Feel so relieved with such a short list.

radix-engine-interface/src/blueprints/resource/bucket.rs
```diff
@@ -8,206 +8,56 @@ use crate::*;
...
-        NativeFn::Bucket(BucketFn::GetNonFungibleLocalIds)
-    }
-}
-
-impl Into<CallTableInvocation> for BucketGetNonFungibleLocalIdsInvocation {
-    fn into(self) -> CallTableInvocation {
-        NativeInvocation::Bucket(BucketInvocation::GetNonFungibleLocalIds(self)).into()
-    }
-}
+pub const BUCKET_GET_AMOUNT_IDENT: &str = "Bucket_get_amount";
```
iamyulong, line 47: Wait, why do we start adding `Bucket_` prefix here? Grouping exports of different blueprints within the same package?

I feel like we might want consistent `export_name` for all our native blueprints, or split it into `blueprint_ident + fn_ident`.

radix-engine-constants/src/lib.rs
talekhinezh: Yes, due to the increased number of substate locks/reads our `large_package` test is failing. I can fix this by reducing the cost units of those.

radix-engine-interface/src/api/package/substates.rs
talekhinezh: Yup, different abstraction. This ID refers to just code.

radix-engine-interface/src/blueprints/resource/bucket.rs
talekhinezh: Yup agreed.

radix-engine-interface/src/api/types/actor.rs
talekhinezh: hehe

radix-engine-interface/src/api/package/substates.rs
talekhinezh: ah you mean the suffix, yup will update this.

radix-engine-interface/src/blueprints/resource/bucket.rs
talekhinezh: Can we push this to a later PR as would like to put this together into a large PR.

iamyulong (approved):
Approve as-is, since most comments will be addressed in follow-up PRs.

radix-engine/src/system/kernel_modules/costing/fee_table.rs
```diff
@@ -68,6 +71,17 @@ impl FeeTable {
             identifier.package_address,
             identifier.blueprint_name.as_str(),
         ) {
+            (LOGGER_PACKAGE, RESOURCE_MANAGER_BLUEPRINT) => match identifier.ident.as_str() {
```
0xOmarA, line 74: Should be `LOGGER_BLUEPRINT`?

talekhinezh: good catch!

PR #839, opened 2023-02-17
Feature/SCRY-319 transaction limit substate read

mstrug-rdx:
## Summary

Implementation of transaction execution substate reads limit.

## Details

Functionality implemented in Kernel module: `TransactionLimitsModule` - added `on_read_substate` event handler with  implementation of counting and validation functionalities.

Additionally introduced `TransactionLimitsConfig` struct for better encapsulation of the limits values.

Default value of substate read limit is set to 20000, it was chosen basing on tests execution - highest value which causes tests execution fail (`should_be_able_run_large_manifest`) is around 16700 reads.

## Testing

Added new test by just modification of `should_be_able_run_large_manifest`. I was unable to create other test using some blueprint function with high count of substate reads because earlier error: `FeeReserveError(LimitExceeded)` was arising.

## Update Recommendations

As new transaction execution limit has been added, it may cause transactions working so far to fail with an error: `TransactionLimitsError(MaxSubstateReadsCountExceeded)`.
For internal discussion how we should handle such a situation (options: increase limit or refactor transaction to use less substate reads).

### DIFF
--- a/radix-engine/src/system/kernel_modules/transaction_limits/module.rs
+++ b/radix-engine/src/system/kernel_modules/transaction_limits/module.rs
@@ -24,13 +24,11 @@ pub enum TransactionLimitsError {
     MaxSubstateReadsCountExceeded,
 }
 
-/// Representatino of data which needs to be limited for each call frame.
+/// Representation of data which needs to be limited for each call frame.
 #[derive(Default)]
-struct CallFrameInfo {
+struct CallFrameLimitInfo {
     /// Consumed WASM memory.
     wasm_memory_usage: usize,
-    /// Substate store read count.
-    substate_store_read: usize,
 }
 
 pub struct TransactionLimitsConfig {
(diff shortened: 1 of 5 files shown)


radix-engine/src/system/kernel_modules/transaction_limits/module.rs
```diff
@@ -82,20 +122,34 @@ impl KernelModule for TransactionLimitsModule {
...
+        _size: usize,
+    ) -> Result<(), RuntimeError> {
+        let tlimit = &mut api.kernel_get_module_state().transaction_limits;
+
+        // Increase read coutner.
+        tlimit.substate_store_read += 1;
+
+        // Validate
+        tlimit.validate_substates()
+    }
```
iamyulong, line 151: Do we want two limits? one for READ and one for WRITE?

iamyulong (commented):
Really nice work!

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
mstrug-rdx: I though of separate limits for read and write. Write implemented in similar way as read - using `on_write_substate` event.

iamyulong: Haha, didn't read the PR title carefully. It's about substate read only in this PR.

iamyulong (approved):

PR #840, opened 2023-02-18
Feature: `Sbor`/`ScryptoSbor` switchover

dhedey:
## Summary
* Fixes a few issues with Describe implementations
* Main change: Change all derive impls in the engine to use `Sbor` / `ScryptoSbor` instead of their separate types.

## Details
This replaces the separate derives with a single derive `Sbor` / `ScryptoSbor` across the codebase. This proves out that the Categorize implementation is stable across a large range of use cases.

After this change, I'm slightly worried that the Describe impls may cause lag in the IDE on load/edit. We should be on the look-out for this, and report issues in case we wish to slow down / stop changes - but at least to me, it doesn't appear to be a minor cause of slow-down on my machine after this change.

## Testing
It compiling is the main test.

### For Dapp Developers
Deriving `ScryptoEncode`, `ScryptoDecode`, `ScryptoCategorize` and `ScryptoDescribe` can be replaced with deriving `ScryptoSbor`. Similarly, `Encode`, `Decode`, `Categorize` and `Descrive` can be replaced with deriving `Sbor`.

### For Internal Integrators
As for Dapp Developers.

### DIFF
--- a/native-sdk/src/lib.rs
+++ b/native-sdk/src/lib.rs
@@ -15,6 +15,6 @@ pub use macros::*;
 
 // Re-export radix engine derives
 pub extern crate radix_engine_derive;
-pub use radix_engine_derive::{LegacyDescribe, ScryptoCategorize, ScryptoDecode, ScryptoEncode};
+pub use radix_engine_derive::{LegacyDescribe, ScryptoSbor};
 
 pub extern crate radix_engine_interface;
--- a/native-sdk/src/lib.rs
+++ b/native-sdk/src/lib.rs
@@ -15,6 +15,8 @@ pub use macros::*;
 
 // Re-export radix engine derives
 pub extern crate radix_engine_derive;
-pub use radix_engine_derive::{LegacyDescribe, ScryptoSbor};
+pub use radix_engine_derive::{
+    LegacyDescribe, ScryptoCategorize, ScryptoDecode, ScryptoDescribe, ScryptoEncode, ScryptoSbor,
+};
 
 pub extern crate radix_engine_interface;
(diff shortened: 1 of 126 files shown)


iamyulong (commented):
Still prefer composable, individual derives in engine implementation, and leave `Sbor/ScryptoSbor` as a shortcut for Scrypto developers. This is similarly to how `serde-derive` works.

Force adding all derives on all types, in engine implementation, for purpose of testing doesn't seem to be justified.

dhedey:
> Still prefer composable, individual derives in engine implementation, and leave `Sbor/ScryptoSbor` as a shortcut for Scrypto developers. This is similarly to how `serde-derive` works.
> 
> Force adding all derives on all types, without verifying the results, for purpose of testing doesn't seem to be justified.
> 
> When schema is ready, we can simply replace `LegacyDerive` with `ScryptoDerive`.

Nearly all types will want a describe implementation though… eg many substates and native method signatures which I assume will get an ABI along with native packages?

Would you still want all four types derived separately? I’m happy to stick with keeping them separate if you’d rather!

Out of interest, why do you prefer having them separate? To save on compile time? Or to make things a little more explicit?

iamyulong:
> Out of interest, why do you prefer having them separate? To save on compile time? Or to make things a little more explicit? I don’t necessarily see the benefit if you’ll almost always want all 4 anyway, but it’s not something I feel particularly strongly about

Mainly from a "add what is needed" and composability perspective. We don't always need 4, epsecially for substates and function identifiers. And there might be edge cases where not all derives can be used. This should be allowed.

But on the other side, maybe having every type implement all 4 traits will increase interoperability, due to reduced variation.

iamyulong:
Like I said before, it's personal engineering preference. There is nothing technically wrong about the approach taken in this PR (if we're confident the added complexity isn't causing noticeable slowdown).

As long as we have one/two example of using invidual derives and it works, I'm fine.

iamyulong:
As discussed offline, ABIs for substates will be needed when native blueprints are fully unified with WASM blueprints.

Individual macros can still be imported through `radix_engine_derive` and are tested there as well.

Happy to merge as-is.

iamyulong (approved):

PR #844, opened 2023-02-20
Fix/fix scrypto categorize with generic parameters

dhedey:
## Summary
Fixes an issue spotted by @jakrawcz-rdx 

## Details
We needed to make sure that the use of the associated type was fully qualified, because otherwise the compiler complains :(

## Testing
Added tested that this code now compiles:
```
#[derive(Clone, PartialEq, Eq, Hash, Debug, ScryptoSbor)]
pub struct Thing<T> {
    pub field: T,
}
```


jakrawcz-rdx (approved):
the rust macro impl is as readable as expected, so I am glad there is a UT :D

jakrawcz-rdx:
@dhedey the `test_describe_enum()` fails, but I believe it is only about adjusting the assert data.

PR #846, opened 2023-02-21
Feature/SCRY-320 transaction limit substate write

mstrug-rdx:
## Summary

Implementation of transaction execution substate writes limit.

## Details

Functionality implemented in Kernel module: `TransactionLimitsModule` - added `on_write_substate` event handler with implementation of counting and validation functionality.

Default value of substate write limit is set to 5000, it was chosen basing on tests execution, especially this one: `should_be_able_run_large_manifest`.

## Testing

Added new test: `transaction_limit_exceeded_substate_writes_should_fail` and refactored all tests related to transaction limits - moved to new file and added new blueprint used for that tests.

## Update Recommendations

As new transaction execution limit has been added, it may cause transactions working so far to fail with an error: `TransactionLimitsError(MaxSubstateWritesCountExceeded)`.
For internal discussion how we should handle such a situation (options: increase limit or refactor transaction to use less substate writes).


iamyulong (approved):
LGTM

PR #849, opened 2023-02-22
Features/SCRY-321 transaction limit invocation payload size

mstrug-rdx:
## Summary

Implementation of function and method invocation payload size limit.

## Details

Extended `ExecutableInvocation` trait with `payload_size()` method with appropriate implementation in `MethodInvocation`, `FunctionInvocation` and `TransactionProcessorRunInvocation` (this one is filtered out during limit validation by checking call depth > 0).

Current limit is set to 4MiB.

## Testing

Added new test dedicated to this functionality. Updated some tests fee tables (as now payload size passed to the `CostingModule` has proper value instead of 0), updated test `test_publish_large_package` to fit within the limit (it needed to adjust code size by 87 bytes) and updated consumed total cost units - now it is in range 60M-70M (was 20M-30M).

## Update Recommendations

As new transaction execution limit has been added, it may cause transactions working so far to fail with an error: `TransactionLimitsError(MaxInvokePayloadSizeExceeded(value))`.
For internal discussion how we should handle such a situation (options: increase the limit or refactor transactions to reduce payload size).


radix-engine/src/blueprints/transaction_processor/executables.rs
```diff
@@ -267,6 +267,10 @@ impl<'a> ExecutableInvocation for TransactionProcessorRunInvocation<'a> {
 
         Ok((actor, call_frame_update, self))
     }
+
+    fn payload_size(&self) -> usize {
+        self.instructions.len() + self.blobs.len()
```
iamyulong, line 272: Transaction size is checked by `TransacitonValidator`. Maybe we just return `0` and add a comment 

```
The payload size of TransactionProcessor invocation (allowed for ROOT only) is not limited
```

radix-engine-tests/tests/transaction_limits.rs
```diff
@@ -168,3 +168,42 @@ fn transaction_limit_exceeded_substate_writes_should_fail() {
...
+    let receipt = test_runner.execute_manifest(manifest, vec![]);
+
+    // Assert
+    receipt.expect_specific_failure(|e| match e {
+        RuntimeError::ModuleError(ModuleError::TransactionLimitsError(
+            TransactionLimitsError::MaxInvokePayloadSizeExceeded(x),
+        )) => *x == DEFAULT_MAX_INVOKE_INPUT_SIZE + 87,
+        _ => false,
+    })
+}
```
iamyulong, line 209: Can we please add a test for invocation that is not `TransactionProcessorRun`?

iamyulong (commented):

radix-engine/src/kernel/interpreters/scrypto_interpreter.rs
```diff
@@ -217,6 +217,10 @@ impl ExecutableInvocation for MethodInvocation {
             executor,
         ))
     }
+
+    fn payload_size(&self) -> usize {
+        self.args.len()
```
iamyulong, line 222: Can we limit everything in the invocation?

radix-engine/src/blueprints/transaction_processor/executables.rs
mstrug-rdx: done

radix-engine/src/kernel/interpreters/scrypto_interpreter.rs
mstrug-rdx: Added method name length to method invocation and function name, blueprint name, package address lengths to function invocation.

radix-engine-tests/tests/transaction_limits.rs
mstrug-rdx: I've created new test: `transaction_limit_exceeded_direct_invoke_input_size_should_fail` which uses direct call to `kernel::invoke_function` to test it without `TransactionProcessorRun`. This required to create the kernel object with some empty/default values.

iamyulong (approved):

PR #850, opened 2023-02-22
Refactor Resource Native Blueprints

iamyulong:
## Summary

This PR refactors the implementation of native blueprints: `Bucket`, `Vault`, `Proof` and `AuthZoneStack`.

## Details

- Replaced `Rc<RefCell<Resource>>` with `LocalRef`
    - Original resource container is the single owner of locked resource 
- Refactored fungible and non-fungible resource backend (while maintaining the same Blueprint interface)
- Refactored `CallFrame` w.r.t. reference management and location inference
- New substate layout of `Bucket`, `Proof` and `Vault`.
    - Vault: `VaultInfo`, `VaultLiquidFungible`, `VaultLiquidNonFungible`, `VaultLockedFungible`, `VaultLockedNonFungible`
    - Bucket: `BucketInfo`, `BucketLiquidFungible`, `BucketLiquidNonFungible`, `BucketLockedFungible`, `BucketLockedNonFungible`
    - Proof: `ProofInfo`, `FungibleProof`, `NonFungibleProof`
- Updated `AuthZone` implementation, proof composition and authorisation verification
    - Method auth verification requires `ClientApi` 
    - Proof composition requires `ClientApi`
- Updated `Worktop::drain` implementation

## Known Issues

Limitations or poor implementations:
- The new `lock`/`unlock` methods are not protected yet
- There isn't a proper `LocalRef` representation in Scrypto value model
- Heavy `AuthZoneStack` clone because of borrow check on `api`
- Drop of partially locked bucket

Will be fixed by follow-up PRs.

## Testing

Existing tests.

### For Internal Integrators

This `Vault` substate layout has been updated. Depending on the fungiblibity of resouce, different subset of offsets will be used.

In addition, the `resource_address` in `ResourceChange` has been removed (lookup panics for failed transactions).  Gateway may use get the resource address from vault ID, using its internal database.


radix-engine-interface/src/blueprints/resource/bucket.rs
```diff
@@ -60,8 +60,40 @@ pub const BUCKET_CREATE_PROOF_IDENT: &str = "Bucket_create_proof";
 #[derive(Debug, Clone, Eq, PartialEq, ScryptoSbor)]
 pub struct BucketCreateProofInput {}
 
+#[derive(Debug, Clone, Eq, PartialEq, ScryptoSbor)]
+pub struct BucketLockAmountInput {
```
talekhinezh, line 64: Do we really want BucketLockAmount? I think conceptually easier if we just don't allow this. For example, this can be done at the application layer by `bucket.take().lock()`.

radix-engine-tests/tests/account.rs
```diff
@@ -170,7 +170,5 @@ fn assert_resource_changes_for_transfer(
     assert_eq!(2, resource_changes.len()); // Two transfers (withdraw + fee, deposit)
     assert!(resource_changes
         .iter()
-        .any(|r| r.resource_address == resource_address
-            && r.component_id == target_account
-            && r.amount == Decimal::from(transfer_amount)));
+        .any(|r| r.component_id == target_account && r.amount == Decimal::from(transfer_amount)));
```
talekhinezh, line 173: Why removed?

radix-engine/src/kernel/call_frame.rs
```diff
@@ -117,55 +120,67 @@ impl CallFrame {
...
-                    .map_err(KernelError::TrackError),
-                RENodeLocation::Heap => {
-                    if flags.contains(LockFlags::UNMODIFIED_BASE) {
-                        Err(KernelError::TrackError(
-                            TrackError::LockUnmodifiedBaseOnNewSubstate(substate_id),
-                        ))
-                    } else {
-                        Ok(())
-                    }
+            if heap.contains_node(&node_id) {
```
talekhinezh, line 131: Not sure about this interface where we check heap first since location of node should be well known from the time of it's creation.

talekhinezh (approved):

PR #852, opened 2023-02-23
Architecture: Use Client SubstateLock API

talekhinezh:
## Summary
* Use Client SubstateLock API for blueprints
* Use LockFlags for Client Substate Lock API
* Refactor Kernel SubstateRef and SubstateRefMut calls to return typed reference
* Cleaned up AccessRule self mutability circular dependency issue


iamyulong (approved):
LGTM.

PR #853, opened 2023-02-23
Architecture: Remove package deref

talekhinezh:
## Summary
* Remove ResourceManager and Package from Global Substate RENode


radix-engine-interface/src/api/types/re_node.rs
```diff
@@ -94,11 +92,10 @@ impl fmt::Debug for RENodeId {
...
             Self::Vault(id) => f.debug_tuple("Vault").field(&hex::encode(id)).finish(),
-            Self::ResourceManager(id) => f
-                .debug_tuple("ResourceManager")
-                .field(&hex::encode(id))
-                .finish(),
-            Self::Package(id) => f.debug_tuple("Package").field(&hex::encode(id)).finish(),
+            Self::GlobalResourceManager(address) => {
+                f.debug_tuple("ResourceManager").field(&address).finish()
+            }
+            Self::GlobalPackage(address) => f.debug_tuple("Package").field(&address).finish(),
```
iamyulong, line 98: Minor: Can we change the debug message to align with the variant name?

radix-engine/src/system/kernel_modules/kernel_debug/module.rs
```diff
@@ -109,6 +109,7 @@ impl KernelModule for KernelDebugModule {
         offset: &SubstateOffset,
         flags: &LockFlags,
     ) -> Result<(), RuntimeError> {
+        /*
```
iamyulong, line 112: Any reason we disable it? I've been relying on them for debugging.

radix-engine/src/system/node.rs
```diff
@@ -88,7 +88,13 @@ impl RENodeModuleInit {
...
 #[derive(Debug)]
 pub enum RENodeInit {
-    Global(GlobalSubstate),
+    GlobalComponent(GlobalSubstate),
+    Package(
+        PackageInfoSubstate,
+        PackageCodeTypeSubstate,
+        PackageCodeSubstate,
+    ),
+    ResourceManager(ResourceManagerSubstate),
```
iamyulong, line 97: Suggest we rename to `GlobalXXXX` as well

iamyulong (commented):
LTGM with some minor comments.

radix-engine/src/system/kernel_modules/kernel_debug/module.rs
talekhinezh: Yes, I've been disabling Substate Lock/Read/Write/Unlock logs as it's taken up too much log space. But since you're still using it I'll reenable these back and we can have discussion on best strategy here.

iamyulong (approved):

PR #858, opened 2023-02-26
Feature: Application Events (Part 1)

0xOmarA:
## Summary

This pull is the first in a three-part series of PRs implementing application events in the Radix Engine as seen in REP-62. 

## Details

This is the initial implementation of application events in the Radix Engine as seen in REP-62. This PR adds application events for a number of the native blueprints as well as the framework for storing these events at runtime, including them in receipts, displaying them when receipts are printed, and some basic costing for events.

In this PR, event schemas are neither stored nor enforced. This is planned for the next PR for application events. 

### Future Improvements and Followup PRs

As mentioned earlier, this is the first in a series of three PRs for application events. The following are next two planned PRs and what you can expect to see in them:

1. Moving events away from `LegacyDescribe` and using the new `Describe` trait. Additionally, this PR is when event schemas will be stored, checked, and enforced. So, this PR is all about event schemas
2. Proper support of events in Scrypto. This PR will add ways for registering events through Scrypto WASM packages and will add extensive testing for events. 

## Update Recommendations

For external and internal integrators, the following is some changes which you might need to be aware of

### Transactions

The items below impact transaction receipts

* This PR adds a new section to transactions called the `application_events` section which contains all of the events emitted during the transaction execution.
* All application events and logs are now obtained from the `EventsModule` and `LoggerModule` kernel module to be included in receipts. This follows a similar approach to execution traces.

### Scrypto Lib

* Although this is still not fully functional, this PR adds a new method to the Scrypto `Runtime` struct to emit events.

### Substates

* This PR adds a new RENode type which is the `EventStore` RENode as well as an `EventStoreSubstate`. Clients such as the Gateway API _probably_ don't have a use for this RENode and Substate since this is a transient RENode which is never persisted on-ledger. 

### For Dapp Developers

* Events can now be emitted in Scrypto through the `Runtime::emit_event` function.


radix-engine-interface/src/api/events_api.rs
```diff
@@ -0,0 +1,11 @@
+use crate::abi::LegacyDescribe;
+use crate::data::ScryptoEncode;
+use crate::sbor::rust::vec::Vec;
+
+use super::types::Hash;
+
+pub trait ClientEventsApi<E> {
+    fn emit_event<T: ScryptoEncode + LegacyDescribe>(&mut self, event: T) -> Result<(), E>;
+
+    fn emit_raw_event(&mut self, schema_hash: Hash, event_data: Vec<u8>) -> Result<(), E>;
```
talekhinezh, line 10: Can we choose one or the other interface?

radix-engine-interface/src/events/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+use crate::ScryptoSbor;
+
+/// Identifies a specific event schema emitter by some emitter RENode.
+///
+/// This type is an identifier uses to identify the schema of events emitted by an RENode of some
+/// [`RENodeId`]. With this identifier, the schema for an event can be queried, obtained, and with
+/// it, the SBOR encoded event data can be decoded and understood.
+///
+/// It is important to note that application events are always emitted by an RENode, meaning that
+/// there is always an emitter of some [`RENodeId`].
```
talekhinezh, line 11: Is this not taking into consideration events emitted by functions yet?

radix-engine-tests/tests/events.rs
```diff
@@ -0,0 +1,92 @@
...
+                args,
+            )
+            .build();
+        test_runner.execute_manifest_ignoring_fee(manifest, vec![])
+    };
+    receipt
+}
+
+/// Tests that Scrypto code can't lock the event store. This is to ensure that Scrypto code can't
+/// add arbitrary events by manually locking the substate and adding events into it.
```
talekhinezh, line 29: I'd add as a TODO that we don't even expose the EventStore RENode to the client. But this will only be possible after a bit more refactors in the system layer.

radix-engine/src/blueprints/access_controller/package.rs
```diff
@@ -318,22 +315,28 @@ impl AccessControllerNativePackage {
...
             AccessControllerInitiateRecoveryAsPrimaryStateMachineInput {
-                proposal: RecoveryProposal {
-                    rule_set: input.rule_set,
-                    timed_recovery_delay_in_minutes: input.timed_recovery_delay_in_minutes,
-                },
+                proposal: proposal.clone(),
             },
         )?;
 
+        api.emit_event(InitiateRecoveryEvent {
```
talekhinezh, line 331: Looks very cool!

radix-engine/src/system/client_api_impl.rs
```diff
@@ -375,4 +380,85 @@ where
...
+            let handle = self.kernel_lock_substate(
+                RENodeId::EventStore,
+                NodeModuleId::SELF,
+                SubstateOffset::EventStore(EventStoreOffset::EventStore),
+                LockFlags::MUTABLE,
+            )?;
+            let event_store = self.kernel_get_substate_ref_mut::<EventStoreSubstate>(handle)?;
+            event_store.0.push((event_type_id, event_data));
+            self.kernel_drop_lock(handle)?;
+        };
```
talekhinezh, line 458: Would be cleaner if this code was in a separate `EventBlueprint` which gets called directly (without going through an invocation).

talekhinezh (approved):

radix-engine-interface/src/api/mod.rs
```diff
@@ -28,5 +30,6 @@ pub trait ClientApi<E: sbor::rust::fmt::Debug>:
     + ClientNodeApi<E>
     + ClientSubstateApi<E>
     + ClientUnsafeApi<E>
+    + ClientEventsApi<E>
```
iamyulong, line 33: Minor: How about `ClientEventApi`?

radix-engine-interface/src/api/events_api.rs
0xOmarA: Yup, that's cool with me.

radix-engine-interface/src/events/mod.rs
0xOmarA: Not at the current moment of time. This will change with my next PR where we introduce Schemas into the mix.

radix-engine/src/system/client_api_impl.rs
0xOmarA: Yup that's cool with me. Will make that refactor.

radix-engine-tests/tests/events.rs
0xOmarA: That would be cool. In that model, how would Scrypto blueprints emit events?

radix-engine/src/system/client_api_impl.rs
```diff
@@ -375,4 +380,85 @@ where
...
+
+        Ok(())
```
iamyulong, line 460: Just wonder why we don't store the events in the kernel module? What functionalities do we need from `RENode`?

radix-engine/src/kernel/track.rs
```diff
@@ -587,47 +589,38 @@ impl<'s> FinalizingTrack<'s> {
         invoke_result: Result<Vec<InstructionOutput>, RuntimeError>,
         fee_summary: &mut FeeSummary,
         vault_ops: Vec<(TraceActor, VaultId, VaultOp)>,
+        application_events: Vec<(EventTypeIdentifier, Vec<u8>)>,
+        application_logs: Vec<(Level, String)>,
```
iamyulong, line 593: Should we revert events in case of transaction failures?

radix-engine-interface/src/api/mod.rs
0xOmarA: Sure, I've went ahead and renamed it.

radix-engine/src/system/client_api_impl.rs
0xOmarA: I think that storing them in the kernel module is a good idea. It also makes it simpler to reason about whether Scrypto can read the RENode or not and also fixes some issues that I ran into. Will make that change for application events. Thoughts on making that change for logs too?

radix-engine/src/kernel/track.rs
0xOmarA: Hmm... since these events have already been costed and paid for, why should we revert them on commit failures?

radix-engine/src/system/client_api_impl.rs
iamyulong: Moving logs too would be a good idea.

radix-engine/src/kernel/track.rs
0xOmarA: I was leaning more on them being discarded and ignored by the gateway for debugging reasons but I think that there might be better ways to solve the simulator experience.

radix-engine/src/system/client_api_impl.rs
0xOmarA: Cool, will probably have that in a separate PR.

iamyulong (approved):

PR #859, opened 2023-02-27
Feature: Drop empty bucket

talekhinezh:
## Summary

* Add ability to drop empty buckets
* Remove Account `withdraw_all` method
* Refactored object dropping

## Details

Add the ability to drop empty bucket in Scrypto:
```
fn(empty_bucket: Bucket) {
  bucket.drop_empty()
}
```

### For Dapp Developers

Added `bucket.drop_empty()`


radix-engine/src/kernel/call_frame.rs
```diff
@@ -507,7 +507,16 @@ impl CallFrame {
         self.take_node_internal(node_id)?;
         let node = heap.remove_node(node_id)?;
         for (_, substate) in &node.substates {
-            let (_, child_nodes) = substate.to_ref().references_and_owned_nodes();
+            let (refs, child_nodes) = substate.to_ref().references_and_owned_nodes();
+            for node_ref in refs {
+                self.immortal_node_refs.insert(
```
iamyulong, line 512: Could you elaborate a bit more why we treat references in substate immortal? 

Also, I'd suggest we add a common method that deals with references in substate, for logic here and what's in `acquire_lock`, to make future refactoring easier.

iamyulong (commented):
Nice refactoring about drop!

Also, I had a comment about `ResourceManager::burn_bucket` over Slack.

radix-engine/src/kernel/call_frame.rs
talekhinezh: This is effectively removing the owning substate of the reference so there is no chance to remove it if it were "temporary". Temporary references are only used for mutable substates in case the reference gets removed by another locker.

iamyulong: Not sure I'm following.  What's effectively done here is putting all references in the substate in the call frame permanently. These references may become invalid, possibly due to the node being dropped.

For instance, after 1) create a bucket; 2) create a proof; 3) drop the proof; and 4) drop the bucket, we will have an invalid reference to a bucket in the call frame?

talekhinezh: Ah! I failed to take into account LocalRefs.

talekhinezh: Will add a test and fix this

iamyulong (approved):

radix-engine/src/kernel/call_frame.rs
talekhinezh: As discussed, we will need to do a deep look into fixing/designing LocalRefs and garbage collecting of references.

PR #861, opened 2023-02-28
Features/SCRY-325 transaction limit substate read write size

mstrug-rdx:
## Summary

First part of implementation of transaction limits for substate read size and write size. Second part will be implemented after refactoring of track implementation.

## Details

Added validation of read/write substate size in `TransactionLimitsModule` basing on `on_read_substate()` and `on_write_substate()` events.

As it was pointed in the summary this is partial implementation due to borrow checker restrictions in `kernel_get_substate_ref()` and `kernel_get_substate_ref_mut()` and needs to be updated later. 

Currently only read substate size is notified from `kernel_read_substate()` and `kernel_create_wasm_instance()` methods. Write substate size is not notified at all, only logic in `TransactionLimitsModule` is ready.

Updated FeeTable Kernel API cost `CostingEntry::ReadSubstate` to `10 * size` (from 100).

Default maximum substate read size is set to 500KiB (lowest round value which passes all test). Maximum default write size is set to 10KiB, but it is not used for a now.

## Update Recommendations

As a new transaction execution limit has been added, it may cause transactions working so far to fail with an error: TransactionLimitsError(MaxSubstateReadSizeExceeded(value)).
For internal discussion how we should handle such a situation (options: increase the limit or advice refactoring of transaction logic).


radix-engine-constants/src/lib.rs
```diff
@@ -55,5 +55,11 @@ pub const DEFAULT_MAX_SUBSTATE_READS_PER_TRANSACTION: usize = 20_000;
 /// The default maximum substates writes count per transaction.
 pub const DEFAULT_MAX_SUBSTATE_WRITES_PER_TRANSACTION: usize = 5_000;
 
+/// The default maximum substates read size.
+pub const DEFAULT_MAX_SUBSTATE_READ_SIZE: usize = 500 * 1024;
+
+/// The default maximum substates write size.
+pub const DEFAULT_MAX_SUBSTATE_WRITE_SIZE: usize = 10 * 1024;
```
iamyulong, line 62: Can we add a `TODO: Apply this limit in create_node too`?

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
```diff
@@ -45,9 +49,13 @@ pub struct TransactionLimitsConfig {
     /// Maximum WASM memory which can be consumed in one call frame.
     pub max_wasm_memory_per_call_frame: usize,
     /// Maximum Substates reads for a transaction.
-    pub max_substate_reads: usize,
+    pub max_substate_reads_count: usize,
     /// Maximum Substates writes for a transaction.
-    pub max_substate_writes: usize,
+    pub max_substate_writes_count: usize,
```
iamyulong, line 54: Minor: `read_count` and `write_count`.

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
```diff
@@ -115,15 +123,38 @@ impl TransactionLimitsModule {
...
+            if size > self.limits_config.max_substate_read_size {
+                return Err(RuntimeError::ModuleError(
+                    ModuleError::TransactionLimitsError(
+                        TransactionLimitsError::MaxSubstateReadSizeExceeded(size),
+                    ),
+                ));
+            }
+        }
+        if let Some(size) = write_size {
+            if size > self.limits_config.max_substate_read_size {
```
iamyulong, line 142: Comparing to wrong parameter.

radix-engine-constants/src/lib.rs
```diff
@@ -55,5 +55,11 @@ pub const DEFAULT_MAX_SUBSTATE_READS_PER_TRANSACTION: usize = 20_000;
 /// The default maximum substates writes count per transaction.
 pub const DEFAULT_MAX_SUBSTATE_WRITES_PER_TRANSACTION: usize = 5_000;
 
+/// The default maximum substates read size.
+pub const DEFAULT_MAX_SUBSTATE_READ_SIZE: usize = 500 * 1024;
```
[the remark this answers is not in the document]
mstrug-rdx: done

radix-engine-constants/src/lib.rs
mstrug-rdx: done

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
mstrug-rdx: Fixed also in other places.

radix-engine/src/system/kernel_modules/transaction_limits/module.rs
mstrug-rdx: Changed read and write validation to `max_substate_size`, left separate errors for read and write to provide more info to the user.

iamyulong (approved):

PR #864, opened 2023-02-28
Architecture: Remove internal node modules

talekhinezh:
## Summary
* Removed node modules from local objects
* Removed chain from `AccessRulesSubstate`
* Refactored node module creation
* Add global flag to TypeInfo

### For Dapp Developers
* Instantiated local components no longer have metadata/access_rules methods
* Several `globalize` variants have been added. Note though that a better interface for these still needs to be implemented

### For Internal Integrators
* AccessRulesSubstate has been changed from a vector of `AccessRules` to only a single `AccessRules`


iamyulong:
Sorry, I pushed my commit to the wrong branch...Force reverted

native-sdk/src/access_rules/access_rules.rs
```diff
@@ -0,0 +1,32 @@
...
+use radix_engine_interface::api::ClientApi;
+use radix_engine_interface::blueprints::resource::AccessRules;
+use radix_engine_interface::constants::ACCESS_RULES_PACKAGE;
+use radix_engine_interface::data::model::Own;
+use radix_engine_interface::data::{scrypto_decode, scrypto_encode, ScryptoDecode};
+use sbor::rust::fmt::Debug;
+
+pub struct AccessRulesObject;
+
+impl AccessRulesObject {
```
iamyulong, line 13: Rename to `AccessRules`?

native-sdk/src/component/component.rs
```diff
@@ -1,37 +1,15 @@
...
-            RENodeId::Component(self.0),
-            NodeModuleId::AccessRules,
-            ACCESS_RULES_ADD_ACCESS_CHECK_IDENT,
-            scrypto_encode(&AccessRulesAddAccessCheckInput { access_rules }).unwrap(),
-        )?;
-
-        Ok(self)
-    }
-
     pub fn sys_set_royalty_config<Y, E: Debug + ScryptoDecode>(
```
iamyulong, line 13: Should royalty be removed as well?

radix-engine-tests/tests/abi.rs
```diff
@@ -151,6 +151,7 @@ fn test_export_abi() {
 }
 
 #[test]
+#[ignore]
```
iamyulong, line 154: Why is this ignored? Are we supposed to fix it later, or remove it?

radix-engine/src/system/node.rs
```diff
@@ -30,7 +31,7 @@ pub enum RENodeModuleInit {
     Metadata(MetadataSubstate),
 
     /* Access rules */
-    ObjectAccessRulesChain(MethodAccessRulesChainSubstate),
+    ObjectAccessRulesChain(MethodAccessRulesSubstate),
```
iamyulong, line 34: Can we rename `ObjectAccessRulesChain` to `ObjectAccessRules`?

native-sdk/src/access_rules/access_rules.rs
talekhinezh: Will save the renaming to a future PR since `AccessRules` conflicts with the `AccessRules` rust object used in scrypto.

native-sdk/src/component/component.rs
talekhinezh: good point

radix-engine-tests/tests/abi.rs
talekhinezh: Ah, yes, this is to be removed.

radix-engine/src/system/node.rs
talekhinezh: yup good catch

PR #870, opened 2023-03-02
Architecture: Use CreateObject Client API + RENodeId cleanup

talekhinezh:
## Summary
* Reduce most `RENodeId` types to `RENodeId::Object`. The existing RENode types are kernel module and global related nodes which will be refactored out in later PR
* Use the `ClientObjectApi` for creating new objects (RENode) for most blueprints

### For Internal Integrators
Biggest change is the removal of most types of `RENodeId`. One should use the type info stored with the node rather than the Node identifier for type information.


radix-engine/src/blueprints/resource/resource_manager.rs
```diff
@@ -988,19 +970,18 @@ impl ResourceManagerBlueprint {
...
-                    LiquidNonFungibleResource::new(ids.clone()),
+            let info = BucketInfoSubstate {
+                resource_address,
+                resource_type: ResourceType::NonFungible { id_type },
+            };
+            let bucket_id = api.new_object(
+                BUCKET_BLUEPRINT,
+                btreemap!(
+                    0 => scrypto_encode(&info).unwrap(),
+                    1 => scrypto_encode(&LiquidNonFungibleResource::new(ids.clone())).unwrap()
```
iamyulong, line 981: Offset `1` is used for both fungible and non-fungible resource. 

Before future refactoring and engine does not consider application-level type info (fungibility, divisibility and local id type), can we please have the following schema for buckets and vaults?
```
0 - info
1 - liquid fungible
2 - locked fungible
3 - liquid non-fungible
4 - locked non-fungible
```

Would also like to have named constants for all substate offsets.

talekhinezh: Based on discussion, we won't go with this approach and will use vector (tuple) of substates rather a map.

iamyulong (approved):

iamyulong:
Stack overflow again, https://github.com/radixdlt/radixdlt-scrypto/actions/runs/4330233724/jobs/7561334097.

It's under investigation by Rubin.

Merging...

PR #873, opened 2023-03-06
Feature/application events schema

0xOmarA:
## Summary
This PR is the second in a three part series implementing application events in the Engine and Scrypto. This PR is a followup to #858 implements event schemas such that packages store event schema of the events that its blueprints can emit and this schema gets validated when an event is emitted. 

### For Internal Integrators
* A new node module has been added which is the `PackageEventSchema` node module. This module can only be on packages and it stores all of the event schemas for the blueprints of that package

## Next Steps

In the next PR I plan on tackling exposing the event registration to Scrypto. Once that is implemented, I plan on adding a number of tests which are only possible with Scrypto. 

## Known Limitations:

The following are known limitations:
* Scrypto blueprints can't register events. This will be addressed in the next PR since it needs a larger refactor.


radix-engine/src/kernel/interpreters/scrypto_runtime.rs
```diff
@@ -148,10 +148,19 @@ where
...
         let metadata = scrypto_decode::<BTreeMap<String, String>>(&metadata)
             .map_err(WasmRuntimeError::InvalidMetadata)?;
-
-        let package_address =
-            self.api
-                .new_package(code, schema, access_rules, royalty_config, metadata)?;
+        let event_schema = scrypto_decode::<
+            BTreeMap<String, Vec<(LocalTypeIndex, Schema<ScryptoCustomTypeExtension>)>>,
+        >(&event_schema)
+        .map_err(WasmRuntimeError::InvalidMetadata)?;
```
iamyulong, line 154: Minor: `InvalidSchema`

radix-engine/src/kernel/track.rs
```diff
@@ -595,17 +595,33 @@ impl<'s> FinalizingTrack<'s> {
...
+                            RENodeId::GlobalPackage(EPOCH_MANAGER_PACKAGE),
+                            NodeModuleId::SELF,
+                            ..,
+                        )
+                        | Emitter::Method(
+                            RENodeId::GlobalComponent(ComponentAddress::EpochManager(..)),
+                            NodeModuleId::SELF,
+                        ),
+                        event_name,
+                    ) if *event_name == expected_event_name => true,
```
iamyulong, line 624: I feel like we should completely remove this logic from engine and let consumers find the events in `application_events` of the receipt.

radix-engine/src/system/package/package.rs
```diff
@@ -111,6 +113,37 @@ impl Package {
...
+                            .map_or(
+                                Err(RuntimeError::ApplicationError(
+                                    ApplicationError::EventError(EventError::InvalidEventSchema),
+                                )),
+                                Ok,
+                            )?
+                            .type_name)
+                            .to_owned()
+                    };
+                    blueprint_schema.insert(event_name, (local_type_index, schema));
```
iamyulong, line 136: Missing duplicate check?

iamyulong (commented):
LTGM with some minor comments.

radix-engine/src/kernel/interpreters/scrypto_runtime.rs
0xOmarA: Yes, you're right. Will update :D

radix-engine/src/kernel/track.rs
0xOmarA: So, do you propose that the commit result should no longer have the `next_epoch` section? Instead, to know if they should advance over to the next epoch they look for this event in the receipt.

iamyulong: That's what I would like, althought it will require some change in node integration.

radix-engine/src/system/package/package.rs
iamyulong: I'm a bit confused. We scan an array of `(LocalIndexId, ScryptoSchema)` and extract the event names in the schema. Wouldn't the `insert` overwrite existing entry and automatically de-duplicate?

```
if blueprint_event_schemas.len()
                    != blueprint_event_schemas
                        .keys()
                        .collect::<BTreeSet<_>>()
                        .len()
```

Also, isn't the above check always false?

0xOmarA: > I'm a bit confused. We scan an array of (LocalIndexId, ScryptoSchema) and extract the event names in the schema.

Ooof, yes you're right. Good catch! Will change.

> Also, isn't the following check always false?

Why would it always be false? If there aren't as many events as unique events names then we should fail. Am i missing something here?

iamyulong: `blueprint_event_schemas` is a `BTreeMap` and we're comparing the number of entries with the number of keys.

iamyulong (approved):

PR #883, opened 2023-03-08
Update criterion benchmark action to use the same comment

muzuke:


iamyulong:
FYI @muzuke, I noticed that the bot keeps adding more comments as new commits are pushed to a PR. 

Possible to configure it to reuse the same comment or just delete old ones?

iamyulong (approved):

iamyulong:
Thank you @muzuke so much for adding the extra feature! Much appreciated it.

You may consider making a pull request to the upstream repo as well, but it's totally up to you. :)

PR #885, opened 2023-03-08
Refactor/remove transaction invocation

iamyulong:
## Summary

This PR removes the special `TransactionProcessorRunInvocation` in favor of `call_function`/`call_method`.

## Details

- Removed `TransactionProcessorRunInvocation`
- Added `(Local)Reference` to Scrypto value model


radix-engine-interface/src/constants.rs
```diff
@@ -18,19 +18,41 @@ pub const PACKAGE_TOKEN: ResourceAddress = vanity_address!(EntityType::Resource,
 pub const OLYMPIA_VALIDATOR_TOKEN: ResourceAddress = vanity_address!(EntityType::Resource, 5);
 
 /// The address of the faucet package.
-pub const PACKAGE_LOADER: PackageAddress = vanity_address!(EntityType::Package, 0);
+pub const PACKAGE_PACKAGE: PackageAddress = vanity_address!(EntityType::Package, 0);
```
talekhinezh, line 21: Package Package!

radix-engine/src/blueprints/transaction_processor/package.rs
```diff
@@ -0,0 +1,75 @@
...
+use super::TransactionProcessorBlueprint;
+
+pub struct TransactionProcessorNativePackage;
+
+impl TransactionProcessorNativePackage {
+    pub fn schema() -> PackageSchema {
+        let mut aggregator = TypeAggregator::<ScryptoCustomTypeKind>::new();
+
+        let mut substates = Vec::new();
+        substates.push(aggregator.add_child_type_and_descendents::<()>());
```
talekhinezh, line 24: I didn't question this in the last PR but why do you think we should have atleast one substate per blueprint?

radix-engine-common/src/data/scrypto/custom_schema.rs
```diff
@@ -24,6 +24,8 @@ pub enum ScryptoCustomTypeKind {
     Decimal,
     PreciseDecimal,
     NonFungibleLocalId,
+
+    Reference,
```
talekhinezh, line 28: Can we rename this to `InternalRef` since `Address` is a reference in itself as well.

talekhinezh (commented):
Maybe I missed it but do we have any checks to prevent `Reference` from being stored in any substate?

radix-engine/src/blueprints/transaction_processor/package.rs
iamyulong: It's changed in later PRs. `SELF` substates can be empty.

The original motivation of making `[0]` mandatory is for state-based access rules.

Changed auth to tolerate missing `[0]` as well.

radix-engine-common/src/data/scrypto/custom_schema.rs
iamyulong: Sure

iamyulong:
Good point about reference in storage!

I originally planed to do that as part of the reference overhaul story, but let me add some basic check.

talekhinezh (approved):

PR #889, opened 2023-03-10
Drop support for PackageAddress, ComponentAddress and ResourceAddress in Manifest

iamyulong:
## Summary

Drop support of PackageAddress, ComponentAddress and ResourceAddress in Manifest

### For Dapp Developers

You can use `Address("bech32_address")` instead.


dhedey (approved):
Looks good!

PR #890, opened 2023-03-10
Feature/drop all virtual proofs

iamyulong:
## Summary

- Added `CLEAR_SIGNATURE_PROOFS` manifest instruction
- Fixed `AuthZone::clear` not clearing signature proofs
- Fixed `CLEAR_AUTH_ZONE` resetting named proofs

### For Dapp Developers

To reiterate,
- `CLEAR_SIGNATURE_PROOFS` removes all signature proofs in the auth zone;
- `CLEAR_AUTH_ZONE` removes all signature proofs and regular proofs in the auth zone;
- `DROP_ALL_PROOFS` removes all named proofs in the call frame, and then calls `CLEAR_AUTH_ZONE`.

### DIFF
--- a/simulator/src/utils/common_instructions.rs
+++ b/simulator/src/utils/common_instructions.rs
@@ -194,8 +194,9 @@ fn build_call_arguments<'a>(
         _ => panic!("Inconsistent schema"),
     }
 
-    let manifest_value: ManifestValue = to_manifest_value(&ManifestValue::Tuple { fields: built_args })
-        .expect("Failed to transcode ManifestValue");
+    let manifest_value: ManifestValue =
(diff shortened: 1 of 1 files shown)


talekhinezh:
It seems to me that this might be the wrong abstraction. As an example, I think dropping the package virtual proof doesn't make sense. I think main usecase for this instruction is to support dropping the virtual public keys. Let's chat tomorrow :)

iamyulong:
Right, I was also hesitant to clear the non-extending virtual proofs.

Maybe we change it to `DROP_NON_EXTENDING_VIRTUAL_PROOFS`?

talekhinezh (approved):

PR #901, opened 2023-03-15
feat: Improved schema validation error message

dhedey:
## Summary
Adds a more detailed error message when schema validation fails, which includes the value path (including logical type names, where we have them!) to where the error occurred.

## Details
Example path:

"MyStruct.hello[0]->MyEnum::Option2{1}.inner[0]->MyEnum::Option1{0}.[0]->Map[0].Value->Array[0]->Tuple.[0]->Enum::{6}.[0]->Tuple.[1]->Map[0].Key"

## Testing
Have added a test in `payload_validator.rs`


iamyulong (commented):
Nice!

iamyulong (approved):

PR #903, opened 2023-03-16
Feature/remove renode auth zone stack

iamyulong:
## Summary

This PR removes the `AuthZoneStack` RENode abstraction.

### For Dapp Developers

Access rule assertion has been relocated to `Runtime::assert_access_rule()`.


native-sdk/src/component/package.rs
```diff
@@ -22,7 +22,7 @@ impl BorrowedPackage {
         Y: ClientApi<E>,
     {
         api.call_module_method(
-            RENodeId::GlobalObject(self.0.into()),
+            &RENodeId::GlobalObject(self.0.into()),
```
talekhinezh, line 25: Would rather avoid this reference if `1.68.0` is working.

radix-engine-interface/src/api/kernel_modules/auth_api.rs
```diff
@@ -0,0 +1,8 @@
+use crate::{api::types::*, blueprints::resource::AccessRule};
+use sbor::rust::fmt::Debug;
+
+pub trait ClientAuthApi<E: Debug> {
+    fn get_auth_zone(&mut self) -> Result<ObjectId, E>;
```
talekhinezh, line 5: Interesting approach, do you see this as the long term approach to getting the object id? If so, does it make sense to have our other "transient" methods to use the same approach?

radix-engine-interface/src/api/kernel_modules/auth_api.rs
```diff
@@ -0,0 +1,8 @@
...
+
+    fn assert_access_rule(&mut self, rule: AccessRule) -> Result<(), E>;
```
talekhinezh, line 7: Now that I see it in code, this method does seem a bit awkward. Why isn't it a method on an auth zone again?

radix-engine-interface/src/api/types/re_node.rs
```diff
@@ -55,7 +52,6 @@ impl From<RENodeId> for [u8; OBJECT_ID_LENGTH] {
         match value {
             RENodeId::KeyValueStore(id) => id,
             RENodeId::Object(id) => id,
-            RENodeId::AuthZoneStack => [5u8; OBJECT_ID_LENGTH], // TODO: Remove, this is here to preserve receiver in invocation for now
```
talekhinezh, line 58: Sweet!

radix-engine-interface/src/blueprints/resource/auth_zone.rs
```diff
@@ -6,14 +6,14 @@ use sbor::rust::prelude::*;
 
 pub const AUTH_ZONE_BLUEPRINT: &str = "AuthZone";
 
-pub const AUTH_ZONE_POP_IDENT: &str = "pop";
+pub const AUTH_ZONE_POP_IDENT: &str = "AuthZone_pop";
```
talekhinezh, line 9: If this prefix is to be used, should probably use the export name.

radix-engine-interface/src/api/kernel_modules/auth_api.rs
talekhinezh: Ah is this because, we are asserting access_rule on the previous boundary authzone?

native-sdk/src/component/package.rs
iamyulong: It's not working with `1.68` on `x86_64`.

Why would prefer a copy of RENodeId?

radix-engine-interface/src/api/kernel_modules/auth_api.rs
iamyulong: I found it the only sensible way of exposing nodes created by kernel module. We can try to extend to other methods if node abstraction is needed.

radix-engine-interface/src/api/kernel_modules/auth_api.rs
iamyulong: Exactly, it's more a method of the stack, rather than an individual authzone.

native-sdk/src/component/package.rs
iamyulong: CI run: https://github.com/radixdlt/radixdlt-scrypto/actions/runs/4471594998/jobs/7856685400

talekhinezh (approved):

.github/workflows/ci.yml
```diff
@@ -51,13 +51,13 @@ jobs:
       with:
         toolchain: stable
     - name: Run tests
-      run: cargo test --features serde
+      run: RUST_MIN_STACK=4194304 cargo test --features serde
```
lrubasze, line 54: Alternatively to avoid duplication you can set env per step/job or for whole workflow.
see https://docs.github.com/en/actions/learn-github-actions/variables

PR #906, opened 2023-03-17
Feature/application events in scrypto

0xOmarA:
## Summary

This pull request adds support for application events in Scrypto and cleans up the overall architecture of application events using `LocalTypeIndex` wherever appropriate to reference event schemas.

## Details

The following is a list of the changes made in this PR:

* `Blueprint` macro now expects an optional attribute `#[events(...)]` for the event registration. Multiple `#[events(...)]` attributes could be provided to add more events.
```rust
#[blueprint]
#[event(NewUserEvent, UserIsNoMoreEvent, UserIsUserEvent)]
mod blueprint {
	struct Club {}
	impl Club {}
}
```
* Event schemas are no longer stored in the `PackageEventSchema` NodeModule. Instead, they're stored alongside for each blueprint in the `PackageInfoSubstate` of the `SELF` node module of packages.
* Optimized the event schemas so that each one is no longer self contained. Instead, they're just local indices into the larger blueprint schema.
* The `EventTypeIdentifier`s in receipts no longer refer to event schemas through their names, instead they refer to them through their LocalTypeIndex.
* The `BlueprintSchema` struct now contains a new field for event schemas of the type `BTreeMap<String, LocalTypeIndex>` which maps the event names to their respective `LocalTypeIndex` in the blueprint schema. 

### For Dapp Developers

You can now register and emit events directly from Scrypto. The following is an example of what the Scrypto syntax for event registration and event emission looks like:

```rust
use scrypto::prelude::*;

#[derive(ScryptoSbor, ScryptoEvent)]
struct RegisteredEvent {
    number: u64,
}

#[blueprint]
#[events(RegisteredEvent)]
mod scrypto_events {
    struct ScryptoEvents;

    impl ScryptoEvents {
        pub fn emit_registered_event(number: u64) {
            Runtime::emit_event(RegisteredEvent { number });
        }        
    }
}

```

### For Internal Integrators

* The `BlueprintSchema`s stored in the `PackageInfoSubstate`s now have an additional field for the `event_schema`.
* The `PackageEventSchema` node module has been removed as well as all of its associated substates.
* The `EventTypeIdentifier` now contains a `LocalTypeIndex` instead of the event name.


iamyulong (approved):
It looks good to me. 

Will need to delete event_schema again, because it has been moved to `SELF`. :)

0xOmarA:
@iamyulong Thanks the review! will make the change + remove merge conflicts tomorrow

PR #909, opened 2023-03-18
feat: Can serialize SBOR with schema to simple and invertible JSON

dhedey:
## Summary
Can serialize SBOR with schema to simple and invertible JSON.

## Details

This replaces the old `value_serializer` with an even nicer JSON representation - particularly if you have a schema handy. With this we should finally be able to map out JSON nicely in the Core API - and Gateway API once the Toolkit is hooked up. For now, I've only implemented serialization - but it essentially 1:1 mimics the serialisation format, so it's trivially invertible.

This PR also makes a couple of other tweaks / fixes:
* The type name is now optional on TypeMetadata - if your type name is just repeating the ValueKind, then instead you should leave the type name as None. This works out better.
* IndexMap no longer requires keys to be Ord in SBOR - ie the serialization of entries is by insertion order, not by the key order (like HashMap) - this makes it more efficient, and the serialization consistent.

## Testing
Check out the tests in `serialization.rs` for some example simple and invertible payloads.

### For Internal Integrators
Core API needs to be updated to use the new method to convert to JSON available in this PR.


radix-engine-interface/src/api/types/indexed_value.rs
```diff
@@ -14,7 +14,7 @@ use utils::ContextualDisplay;
 #[derive(Clone, PartialEq, Eq)]
 pub struct IndexedScryptoValue {
     bytes: Vec<u8>,
-    references: HashSet<RENodeId>,
```
talekhinezh, line 17: Can we remove this change to `BTreeSet` for now since it isn't relevant to PR. From what I recall, Yulong changed these to `HashSet` due to really bad performance of `BTreeSet`, or it might have been somewhere else. Regardless, let's revert this for now.

radix-engine/src/kernel/call_frame.rs
```diff
@@ -15,26 +15,26 @@ use super::track::{Track, TrackError};
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CallFrameUpdate {
     pub nodes_to_move: Vec<RENodeId>,
-    pub node_refs_to_copy: HashSet<RENodeId>,
```
talekhinezh, line 18: Same thing here, is this relevant to this PR?

radix-engine-interface/src/api/types/indexed_value.rs
dhedey: Sure - can remove it from this PR if you want.

radix-engine-common/src/data/scrypto/custom_serde.rs
```diff
@@ -0,0 +1,286 @@
...
+                    "value": "00000000000000000000000000000000000000000000000000000000000000"
+                }
+            ],
+            "kind": "Tuple"
+        });
+
+        let context = ScryptoValueDisplayContext::with_optional_bench32(Some(&encoder));
+
+        assert_simple_json_matches(&value, context, expected_simple);
+        assert_invertible_json_matches(&value, context, expected_invertible);
```
iamyulong, line 244: (Personal) I like `invertible_json` more than `simple_json`. :)

sbor/src/serde_serialization/value_map_aggregator.rs
```diff
@@ -0,0 +1,244 @@
...
+        enum_metadata: Option<&'a TypeMetadata>,
+    ) -> Option<&'a ChildNames> {
+        self.fields
+            .push(("variant_id", SerializableType::U8(variant_id)));
+        enum_metadata.and_then(|metadata| {
+            if let Some(ChildNames::EnumVariants(variants)) = &metadata.child_names {
+                variants.get(&variant_id).and_then(|variant_metadata| {
+                    variant_metadata.get_name().map(|variant_name| {
+                        self.fields
+                            .push(("variant", SerializableType::Str(variant_name)))
```
iamyulong, line 219: `variant_name`?

iamyulong (commented):
A few comments from a quick scan.

Will find a good time (when mind is clear) to review `serialization.rs` and `value_map_aggregator.rs`.

talekhinezh (approved):

.vscode/settings.json
```diff
@@ -3,9 +3,13 @@
...
-    "rust-analyzer.cargo.features": [
-        "serde",
-    ],
+    // > Enable the below temporarily to work on certain features in VS Code.
+    // > But note that this can cause problems: https://github.com/rust-lang/rust-analyzer/issues/10298#issuecomment-1474923721
+    // > In particular, dummy features might need adding to the simulator/Cargo.toml for rust-analyzer to start.
+    // > And with these features enabled, all "Run Test" commands in projects without this feature will break - so disable this after work.
+    // "rust-analyzer.cargo.features": [
+    //     "serde",
+    // ],
```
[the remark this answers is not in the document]
dhedey: Whenever I work on these files that require a feature flag I have to go hunting for this flag again in google, so having this here saves me time.

I assume I'm not alone with this - out of interest, how do you work on these files?

sbor/src/serde_serialization/value_map_aggregator.rs
dhedey: Happy to change this if you'd rather 👍

.vscode/settings.json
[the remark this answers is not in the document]
iamyulong: I typically just add the feature to the `default` when developing feature-related stuff.

The above comment making it an invalid JSON file is my biggest concern.

radix-engine-common/src/data/scrypto/custom_serde.rs
dhedey: Based on your thoughts I'll default the API to using the invertible format then 👍

PR #910, opened 2023-03-19
Fix/Minor determinism tweaks

dhedey:
## Summary
(Split from #909)

Just a few replacings of HashSet with BTreeSet - for references. This was motivated by wanting the Core API iteration to be deterministic.

Also considered IndexSet (which doesn't need to do comparisons on insert), which might be more performant if this is performance sensitive. Josh implied it might be, but will wait for @iamyulong input?

### DIFF
--- a/radix-engine-interface/src/api/types/indexed_value.rs
+++ b/radix-engine-interface/src/api/types/indexed_value.rs
@@ -14,7 +14,7 @@ use utils::ContextualDisplay;
 #[derive(Clone, PartialEq, Eq)]
 pub struct IndexedScryptoValue {
     bytes: Vec<u8>,
-    references: HashSet<RENodeId>,
+    references: BTreeSet<RENodeId>,
     owned_nodes: Vec<RENodeId>,
     scrypto_value: RefCell<Option<ScryptoValue>>,
 }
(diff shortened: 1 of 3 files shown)


iamyulong:
Did you run into any specific problem as a result of `HashMap/HashSet` used in this PR?

dhedey:
> Did you run into any specific problem as a result of HashMap/HashSet used in this PR?
> Want to make sure BTreeMap/IndexMap does not obscure the underlying problem, i.e. certain engine logic depends on the order of elements.

I just needed to include the IndexedScryptoValue references in the returned receipt from the Core API, and I wanted the references to be returned deterministically... I could just sort in the Core API, but thought I may as well be a good citizen and just fix it in the engine.

To be fair, in the context of indexing a ScryptoValue, IndexMap makes more sense. But as this was used directly with the call frame, including with manual inserts, I thought you/Josh might prefer BTreeMap.

iamyulong:
> I just needed to include the IndexedScryptoValue references in the returned details from the Core API, and I wanted the references to be returned deterministically... I could just sort in the Core API, but thought I may as well be a good citizen and just fix it in the engine.
> 
> To be fair, in the context of indexing a ScryptoValue, IndexMap makes more sense. But as this was used directly with the call frame, including with manual inserts, I thought you/Josh might prefer BTreeMap.

My opinion is that we need to understand each hashset/hashmap use case, 
- Whether the engine logic depends on the order of elements
- Whether it has a side-effect on client side
- What's the performance requirement

and then decide the right replacement. Ideally, we would also add justification to field definition.

PR #911, opened 2023-03-19
Feature/separate fungible non fungible entity ids

talekhinezh:
## Summary
Separate Fungible and NonFungible Entity Ids

### For Internal Integrators
A separate entity ID for both Fungible and NonFungible Resources have been added. The Entity IDs have also been modified to more align with REP-60.


radix-engine-common/src/data/scrypto/model/address_resource_manager.rs
```diff
@@ -12,7 +12,8 @@ use utils::{copy_u8_array, ContextualDisplay};
 /// Represents a resource address.
 #[derive(Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub enum ResourceAddress {
-    Normal([u8; ADDRESS_HASH_LENGTH]), // TODO: change to fungible & non-fungible
+    NonFungible([u8; ADDRESS_HASH_LENGTH]),
```
dhedey, line 15: Very niggly, but can we re-order them as Fungible then NonFungible for consistency?

radix-engine-tests/tests/account.rs
```diff
@@ -43,6 +43,10 @@ fn can_withdraw_from_my_account_internal(use_virtual: bool) {
         other_account,
         transfer_amount,
     );
+    println!(
```
dhedey, line 46: I think this was unintentional?

radix-engine-tests/tests/execution_trace.rs
```diff
@@ -574,6 +573,7 @@ fn test_worktop_changes() {
                     ]
                     .into()
                 )),
+                WorktopChange::Take(ResourceSpecifier::Amount(fungible_resource, 100.into())),
```
dhedey, line 576: Out of interest - why has this order changed? Shouldn't it be deterministic if we're using a vec?

dhedey: I guess it might be deterministic based on some BTreeMap of some objectid hash which has changed due to the addressing changes?

dhedey (approved):
Oh dang. Bet updating all those addresses was painful...

Thanks for fighting through this!

We should probably do the rest of REP-60 in one batch (and leave some spare ids?) to save on wasted overhead updating all the addresses again.

Separately, tagging @siy and @KrzysztofLabus-RDX  as this change in address prefix might affect Ledger/Gateway respectively. (see address/entity.rs - 4th file in the PR)

radix-engine-common/src/data/scrypto/model/address_resource_manager.rs
talekhinezh: yup!

radix-engine-tests/tests/execution_trace.rs
talekhinezh: Not sure, this seems to keep on changing every once in a while and haven't had a chance to dig into why.

PR #922, opened 2023-03-24
Feature: Pretty print events in receipts with schema context

0xOmarA:
# Summary

Adds the ability for events to be pretty printed in receipts instead of the old logic which printed out the `LocalTypeIndex` which didn't provide users with much useful information

Pretty Print


dhedey (approved):
Really nice, thanks Omar! 👍

PR #927, opened 2023-03-28
Feature: Global Address API + Virtual Global Component Proof

talekhinezh:
## Summary
Add API to retrieve GlobalAddress and add virtual Global Component Proof

## Details
* Add `get_global_address` api to `ClientActorApi`
* Add `GLOBAL_OBJECT_TOKEN` which gets added to global object Authzones
* Refactored `Actor` to remove duplicated info

### For Dapp Developers
* Added `Runtime::global_address` which retrieves the global address of the component where in a local or global context


assets/radiswap.wasm
iamyulong, line 1: Great improvement over code size!

radix-engine-tests/tests/access_rules.rs
```diff
@@ -297,6 +297,7 @@ fn assert_access_rule_through_manifest_when_not_fulfilled_fails() {
 }
 
 #[test]
+#[ignore]
```
iamyulong, line 300: Why are the tests ignored?

radix-engine/src/kernel/interpreters/scrypto_runtime.rs
```diff
@@ -206,8 +206,15 @@ where
...
     }
 
+    fn get_global_address(&mut self) -> Result<Buffer, InvokeError<WasmRuntimeError>> {
+        let address = self.api.get_global_address()?;
+
+        let buffer = scrypto_encode(&address).expect("Failed to encode address");
+        self.allocate_buffer(buffer)
+    }
+
     fn get_actor(&mut self) -> Result<Buffer, InvokeError<WasmRuntimeError>> {
```
iamyulong, line 216: Maybe rename to `get_blueprint`?

radix-engine/src/kernel/kernel.rs
```diff
@@ -495,8 +492,24 @@ where
...
+            match actor {
+                Actor::Method {
+                    global_address: Some(address),
+                    ..
+                } => {
+                    self.current_frame.add_ref(
+                        RENodeId::GlobalObject(*address),
+                        RENodeVisibilityOrigin::Normal,
+                    );
+                }
```
iamyulong, line 507: Should we take the approach regarding package address?

radix-engine/src/wasm/traits.rs
```diff
@@ -75,6 +75,8 @@ pub trait WasmRuntime {
 
     fn drop_lock(&mut self, handle: LockHandle) -> Result<(), InvokeError<WasmRuntimeError>>;
 
+    fn get_global_address(&mut self) -> Result<Buffer, InvokeError<WasmRuntimeError>>;
+
     fn get_actor(&mut self) -> Result<Buffer, InvokeError<WasmRuntimeError>>;
```
iamyulong, line 80: Suggest rename

radix-engine-tests/tests/address.rs
```diff
@@ -0,0 +1,259 @@
+use radix_engine::errors::{RuntimeError, SystemError};
+use radix_engine::types::*;
+use scrypto_unit::*;
+use transaction::builder::ManifestBuilder;
+
+#[test]
+fn get_global_address_in_local_should_fail() {
```
iamyulong, line 7: Are we sure we don't want to return an `Option<Address>`?

radix-engine-tests/tests/address.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    // Assert
+    receipt.expect_specific_failure(|e| {
+        matches!(
+            e,
+            RuntimeError::SystemError(SystemError::GlobalAddressDoesNotExist)
+        )
+    });
+}
+
```
iamyulong, line 43: Am I right in understanding your rules?

1. `Function/NonSelfMethod` -> `LocalComponent` -> `Runtime::get_global_address`   ===> Failure
2. `GlobalComponent::Method` -> `LocalComponent` -> `Runtime::get_global_address` ===> Success
3. `GlobalComponent::Method` -> `OwnedComponent` -> `Runtime::get_global_address` ===> Success

I'm not feeling comfortable reading 1 & 2 at the same time.

radix-engine-tests/tests/access_rules.rs
talekhinezh: This is still pending some decisions to be made about auth zone.

radix-engine-tests/tests/address.rs
talekhinezh: Ah good point! 2 SHOULD fail but I think current implementation will let it pass. Let me verify.

radix-engine/src/kernel/kernel.rs
talekhinezh: good point

talekhinezh: We can then remove some of the hacks in `scrypto_interpreter`.

radix-engine-tests/tests/address.rs
iamyulong: Good point, although we can `unwrap` in the `Runtime::get_global_address`.

iamyulong (approved):

PR #931, opened 2023-03-28
Getting rid of plain HashMap usage from prod code

jakrawcz-rdx:
This is the first, almost-100%-mechanical step.

The base idea was to introduce a new `NonIterMap` type - a wrapper for `HashMap` which does not expose ANY operation that has non-deterministic effects (in practice: iteration like `iter()`, `drain()` & co).

Here is the algorithm I used to apply this new type throughout the codebase:
- Define `NonIterMap` in `utils/src/rust.rs` and export it so that it is available with a classic `sbor::rust::collections` import.
- For every _applicable_* usage of `HashMap` that I found:
   - If that usage needs to iterate over that `HashMap`:
     - Migrate from `HashMap` to `IndexMap`
   - Else (if that usage only does simple put/get & co.):
     - Migrate from `HashMap` to `NonIterMap`
     - _NOTE: there were like 2 cases where the caller also wanted to do `entry()`, which was hard for me to provide using a code that must work in "std XOR hashbrown" env; I migrated these `entry()` calls to `get_mut()` and marked them in PR comments here_

Which files were not _applicable_ for this migration?
- tests,
- test utils
    - In some non-obvious (for me) places, I added a comment about the thing being used just in tests
- `sbor-derive-common`
    - I had problems introducing the dep on `../utils` there (as always, std vs alloc); but in principle we could switch from `HashMap` to `NoIterMap` there
- JMT impl copied from another open-source project
    - The hashmap used there is safe: it mostly does put/get, and when the logic wants to iterates over these entries, they are explicitly sorted. We do not plan to touch this code.
    - Since that copy is verbatim, I did not want to add my own inline comments about that fact.


radix-engine/src/blueprints/resource/auth_zone_composition.rs
```diff
@@ -211,22 +212,17 @@ fn max_ids_locked<Y: KernelSubstateApi + ClientApi<RuntimeError>>(
...
-                    max.insert(container_id.clone(), new_ids);
+                    per_container.insert(container_id.clone(), locked_ids.clone());
                 }
             }
         } else {
             api.sys_drop_lock(handle)?;
         }
     }
-    let mut total = BTreeSet::<NonFungibleLocalId>::new();
-    for value in max.values() {
```
jakrawcz-rdx, line 226: _Note from author:_

AFAIU, this post-iteration can be avoided (by collecting on the way), and I did that - please review carefully.

radix-engine/src/blueprints/resource/auth_zone_composition.rs
```diff
@@ -211,22 +212,17 @@ fn max_ids_locked<Y: KernelSubstateApi + ClientApi<RuntimeError>>(
...
-        total.extend(value.clone());
-    }
-    let per_container = max.into_iter().collect();
```
jakrawcz-rdx, line 229: _Note from author:_

I could not figure out why this was needed, so I removed it to make this one `NonIterMap`. Was I, as often, terribly wrong?

radix-engine/src/kernel/call_frame.rs
```diff
@@ -159,10 +159,11 @@ impl CallFrame {
...
         for node_id in &temp_references {
-            self.temp_node_refs
-                .entry(node_id.clone())
-                .or_default()
-                .add_assign(1);
+            if let Some(count) = self.temp_node_refs.get_mut(&node_id) {
+                *count += 1;
+            } else {
+                self.temp_node_refs.insert(node_id.clone(), 1);
+            }
```
jakrawcz-rdx, line 166: _Note from author:_

Here is one of the places where I had to replace `entry()` with an equivalent.
Please see if it is correct AND acceptable from a performance PoV.
(as an alternative, we can switch to an `IndexMap`)

radix-engine/src/kernel/call_frame.rs
```diff
@@ -424,14 +425,13 @@ impl CallFrame {
...
-                    }
-                })
-                .or_insert(ref_data.clone());
+            if let Some(existing_ref_data) = to.immortal_node_refs.get_mut(&node_id) {
+                if existing_ref_data.visibility == RENodeVisibilityOrigin::DirectAccess {
+                    existing_ref_data.visibility = ref_data.visibility;
+                }
+            } else {
+                to.immortal_node_refs.insert(node_id, ref_data.clone());
+            }
```
jakrawcz-rdx, line 434: _Note from author:_

Here is the other place where I had to replace `entry()` with an equivalent.
Please see if it is correct AND acceptable from a performance PoV.
(as an alternative, we can switch to an `IndexMap`)

radix-engine/src/state_manager/deprecated_staging.rs
```diff
@@ -1,185 +0,0 @@
-use crate::ledger::*;
```
jakrawcz-rdx, line 1: _Note from author:_

This file was using `HashMap`s, and I figured it is easier to delete the deprecated thing than migrate it.
I myself am 100% sure that this staging implementation is no longer usable with state manager (so even keeping it as some "reference impl" for test purposes seems no longer appropriate).

jakrawcz-rdx:
One more interesting thing, which maybe you guys (@iamyulong ?) can help me understand.

Why do we even _support_ using a `HashMap` in scrypto, and serializing it?
I see some code that does it (and even tries to fix its non-det) e.g. here:
https://github.com/radixdlt/radixdlt-scrypto/blob/b852078159135fabc03fcfbfc53702cfaf03651d/sbor/src/codec/collection.rs#L131
I think that this ⬆️ still allows for non-deterministic behaviors (you can persist something in a deterministic order, but then you will deserialize it into some effectively-random new order, and run logic dependent on the iteration order - is this right?).
Wouldn't we be better off simply banning `HashMap`?

jakrawcz-rdx:
The latest commit migrates macro impls to `BTreeMap` instead, for simplicity (see https://rdxworks.slack.com/archives/C01M90Y2448/p1680038403504169?thread_ts=1680028951.916639&cid=C01M90Y2448).

radix-engine/src/kernel/call_frame.rs
jakrawcz-rdx: N/A.
_(I had some difficulties with importing `.entry()` the right way, but today I learned rust and did it)_

radix-engine/src/kernel/call_frame.rs
jakrawcz-rdx: N/A.
_(I had some difficulties with importing `.entry()` the right way, but today I learned rust and did it)_

radix-engine/src/blueprints/resource/auth_zone_composition.rs
```diff
@@ -188,12 +188,13 @@ fn max_ids_locked<Y: KernelSubstateApi + ClientApi<RuntimeError>>(
 ) -> Result<
     (
         BTreeSet<NonFungibleLocalId>,
-        HashMap<LocalRef, BTreeSet<NonFungibleLocalId>>,
+        NonIterMap<LocalRef, BTreeSet<NonFungibleLocalId>>,
```
iamyulong, line 191: This is an interesting one. I would expect we have a dependency over the iteration order, but it actually just sort the total locked non-fungible local ids.

radix-engine/src/blueprints/transaction_processor/tx_processor.rs
```diff
@@ -614,17 +614,17 @@ impl TransactionProcessorBlueprint {
 }
 
 struct TransactionProcessor<'blob> {
-    proof_id_mapping: HashMap<ManifestProof, ObjectId>,
-    bucket_id_mapping: HashMap<ManifestBucket, ObjectId>,
+    proof_id_mapping: IndexMap<ManifestProof, ObjectId>,
```
iamyulong, line 617: Why do we need need `IndexMap` here?

radix-engine/src/kernel/call_frame.rs
```diff
@@ -96,17 +96,17 @@ pub struct CallFrame {
...
+    temp_node_refs: NonIterMap<RENodeId, u32>,
 
     /// Owned nodes which by definition must live on heap
     /// Also keeps track of number of locks on this node
-    owned_root_nodes: HashMap<RENodeId, u32>,
+    owned_root_nodes: IndexMap<RENodeId, u32>,
 
     next_lock_handle: LockHandle,
-    locks: HashMap<LockHandle, SubstateLock>,
+    locks: IndexMap<LockHandle, SubstateLock>,
```
iamyulong, line 109: We can use `NoIterMap` for `owned_root_nodes` and `locks`?

radix-engine/src/system/node.rs
```diff
@@ -66,15 +66,12 @@ pub enum RENodeInit {
 }
 
 impl RENodeInit {
-    pub fn to_substates(self) -> HashMap<SubstateOffset, RuntimeSubstate> {
-        let mut substates = HashMap::<SubstateOffset, RuntimeSubstate>::new();
+    pub fn to_substates(self) -> BTreeMap<SubstateOffset, RuntimeSubstate> {
```
iamyulong, line 69: Why do we use different map for `RENodeInit` and not for `RENodeModuleInit`?

sbor-derive-common/src/utils.rs
```diff
@@ -41,13 +41,13 @@ pub fn print_generated_code<S: ToString>(kind: &str, code: S) {
 pub fn extract_attributes(
     attrs: &[Attribute],
     name: &str,
-) -> Option<HashMap<String, Option<String>>> {
+) -> Option<BTreeMap<String, Option<String>>> {
```
iamyulong, line 44: Why not `NonIterMap`?

scrypto-derive/src/non_fungible_data.rs
```diff
@@ -1,6 +1,6 @@
 use proc_macro2::{Span, TokenStream};
 use quote::quote;
-use std::collections::HashMap;
+use std::collections::BTreeMap;
```
iamyulong, line 3: Same

iamyulong (commented):
Good work indeed!

radix-engine/src/blueprints/transaction_processor/tx_processor.rs
jakrawcz-rdx: Because we iterate over this one at:
https://github.com/radixdlt/radixdlt-scrypto/blob/ca80a43d162461c2bed48f501aa9440d30f6fcad/radix-engine/src/blueprints/transaction_processor/tx_processor.rs#L216

(I was not analyzing whether the particular iteration has potential of causing determinism or not; I simply followed the migration algo described at https://github.com/radixdlt/radixdlt-scrypto/pull/931#issue-1644373644)

radix-engine/src/kernel/call_frame.rs
jakrawcz-rdx: We iterate over `owned_root_nodes` at:
https://github.com/radixdlt/radixdlt-scrypto/blob/fe73820af3c6709aba58ed75ac371b09fd9c4bf7/radix-engine/src/kernel/call_frame.rs#L550-L551

We iterate over `locks` at:
https://github.com/radixdlt/radixdlt-scrypto/blob/fe73820af3c6709aba58ed75ac371b09fd9c4bf7/radix-engine/src/kernel/call_frame.rs#L445-L447

(I was not analyzing whether the particular iteration has potential of causing determinism or not; I simply followed the migration algo described at https://github.com/radixdlt/radixdlt-scrypto/pull/931#issue-1644373644)

radix-engine/src/system/node.rs
jakrawcz-rdx: I notice an opportunity to avoid the rewriting (from map to map) here, in `RENodeInit`.
Interestingly, I have missed the very same opportunity in `RENodeModuleInit` :D

Thanks for bringing this up - I now applied the same idea there 👍 
_(we are better off letting the `BTreeMap` stay here, because an instance is already available)_

sbor-derive-common/src/utils.rs
jakrawcz-rdx: This is a macro impl, and I figured that performance does not matter at all. While pulling in a dependency to `NonIterMap` could be problematic, says David (see https://rdxworks.slack.com/archives/C01M90Y2448/p1680038448642949?thread_ts=1680028951.916639&cid=C01M90Y2448).

jakrawcz-rdx:
> Question: with the introduction of NonIterMap, possible to use only IndexMap or BTreeMap rather than both, for remaining use cases?

@iamyulong It is possible...

... BUT:
- if we do it, we should probably choose `IndexMap`, because: 
   - for large enough collection the O(logn) aspect of `BTreeMap` must finally show (and destroy our performance);
   - there might be cases where we have keys that are `Hash` but are not `Ord`
- however, there may still be places where we actually enjoy the natural ordering of keys that `BTreeMap` provides:
   - in such places, `IndexMap` would need a `sort()` before each iteration, potentially being slower than `BTreeMap`
   - I don't believe I can reliably guess which of our current `BTreeMap`s usages rely on keys being ordered, and which do not; hence I was not touching existing `BTreeMap`s in this PR.

Besides "consistency" (which I value), are there any reasons to NOT just allow ourselves to use both `BTreeMap` and `IndexMap`, depending on whether the author thinks that keys have "meaningful"/potentially-useful natural order?

iamyulong:
> Besides "consistency" (which I value), are there any reasons to NOT just allow ourselves to use both BTreeMap and IndexMap, depending on whether the author thinks that keys have "meaningful"/potentially-useful natural order?

I would like to trust author to make the right choice. In fact, I'm in support of careful examiniation of every use case as previously stated [here](https://github.com/radixdlt/radixdlt-scrypto/pull/910#issuecomment-1477482328).

Just wondered if there is some principle that we should follow, as it has implication on the interface between components/teams.

From a reviewer's perspective, if the author uses `BTreeMap`/`IndexMap`, may I right to assume the implementaiton depends on the sorting/inserting order correspondingly? Or it just happens that the implementation requires "some" iteration.

PR #934, opened 2023-03-29
Bugfix: compiler performance

0xOmarA:
This PR converts the `start` and `end` coordinates of `Token` `Span`s from being a tuple of `(usize, usize)` to being a single `usize` each and removed the `index_to_coordinate` method which was slowing down transaction compilation dramatically for large manifests.

### DIFF
--- a/transaction/src/manifest/lexer.rs
+++ b/transaction/src/manifest/lexer.rs
@@ -3,9 +3,9 @@ use sbor::rust::str::FromStr;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Span {
     /// The start of the span, inclusive
-    pub start: (usize, usize),
+    pub start: usize,
     /// The end of the span, inclusive
-    pub end: (usize, usize),
+    pub end: usize,
 }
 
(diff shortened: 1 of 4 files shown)


iamyulong (approved):

transaction/src/manifest/lexer.rs
```diff
@@ -3,9 +3,9 @@ use sbor::rust::str::FromStr;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Span {
     /// The start of the span, inclusive
```
dhedey, line 5: As mentioned in the thread, can we document _what_ the index is?

In this case, it's the index in terms of rust characters - ie  Unicode code points / _UTF-32 characters_.

transaction/src/manifest/lexer.rs
```diff
@@ -3,9 +3,9 @@ use sbor::rust::str::FromStr;
...
-    pub start: (usize, usize),
+    pub start: usize,
     /// The end of the span, inclusive
```
dhedey, line 7: We used to do self.current - 1, but we don't any more. So this should probably be `exclusive`.

PR #935, opened 2023-03-29
Getting rid of HashSet from prod code

jakrawcz-rdx:
A follow-up to https://github.com/radixdlt/radixdlt-scrypto/pull/931.

In the Set's case, we literally have no occurrence of "I need a HashSet but promise to not iterate over it".
Hence, I did not have to introduce `NonIterSet<>`.
I simply either migrated to `IndexSet<>`, or removed the need for allocating a set altogether.

### DIFF
--- a/radix-engine/src/kernel/track.rs
+++ b/radix-engine/src/kernel/track.rs
@@ -949,10 +949,7 @@ impl<'a, 'b> BalanceChangeAccounting<'a, 'b> {
                 };
                 let mut new_balance = substate.vault_liquid_non_fungible().ids().clone();
 
-                let intersection: HashSet<NonFungibleLocalId> =
-                    new_balance.intersection(&old_balance).cloned().collect();
-                new_balance.retain(|x| !intersection.contains(x));
-                old_balance.retain(|x| !intersection.contains(x));
+                remove_intersection(&mut new_balance, &mut old_balance);
 
                 Some(BalanceChange::NonFungible {
                     added: new_balance,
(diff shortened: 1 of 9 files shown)


radix-engine/src/kernel/track.rs
```diff
@@ -992,3 +989,14 @@ impl<'a, 'b> BalanceChangeAccounting<'a, 'b> {
...
+
+/// Removes the `left.intersection(right)` from both `left` and `right`, in place, without
+/// computing (or allocating) the intersection itself.
+/// Implementation note: since Rust has no "iterator with delete" capabilities, the implementation
+/// uses a (normally frowned-upon) side-effect of a lambda inside `.retain()`.
+/// Performance note: since the `BTreeSet`s are inherently sorted, the implementation _could_ have
+/// an `O(n+m)` runtime (i.e. traversing 2 iterators). However, it would then contain significantly
+/// more bugs than the `O(n * log(m))` one-liner below.
+fn remove_intersection<T: Ord>(left: &mut BTreeSet<T>, right: &mut BTreeSet<T>) {
+    left.retain(|id| !right.remove(id));
```
jakrawcz-rdx, line 1001: _Note from author:_

Still worth reviewing! :p

sbor-derive-common/src/utils.rs
```diff
@@ -192,16 +192,7 @@ pub fn get_hash_of_code(input: &TokenStream) -> [u8; 20] {
...
-    let mut out = Vec::<&syn::Type>::new();
-    for t in types {
-        if seen.contains(t) {
-            continue;
-        }
-        seen.insert(t);
-        out.push(t);
-    }
-    out
+    types.into_iter().unique().cloned().collect()
```
jakrawcz-rdx, line 195: _Note from author:_

I hope itertools is not a problematic dep (we do use it in other places already).
The `.unique()` maintains the original order (which is what we care about, I guess).

radix-engine/src/kernel/track.rs
iamyulong: As you said, merge-sort would be optimal. 

A minor improvement to the one-liner would be to compare the size before `retain`.

iamyulong (approved):

PR #939, opened 2023-03-31
Use different power schedules per fuzzing session

lrubasze:
## Summary

Try different power schedules to diversify fuzzing mutations


iamyulong (approved):
I feel bad that I have to blindly approve this PR. If you've tested it and it works, go ahead!

PR #942, opened 2023-04-01
Architecture/remove access rules 1

talekhinezh:
## Summary
Add notion of system-level blueprint child and remove `NodeModuleId::AccessRules1`

## Details
* Blueprints (Native only for now) now have the option of specifying a child blueprint
* `Vault` Blueprint is implemented as a child of `ResourceManager` Blueprint
* AccessRules for an object may include AccessRules for blueprint children of the object
* Removed `get_resource_address` method and replaced with system query to retrieve blueprint parent

### For Internal Integrators
* AccessRules substate updated
* TypeInfo substate updated


iamyulong (approved):
It looks good, although can we have an update on the PR description?

talekhinezh:
> It looks good, although can we have an update on the PR description?

Lol done

PR #945, opened 2023-04-04
Add resources and stakes to genesis; add GenesisHelper utility package

LukasGasior1:
## Summary

Extended genesis functionality:
- resource creation (/w initial allocation)
- validators with multiple stakers

This is handled in an auxiliary scrypto blueprint: GenesisHelper.


assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+#[blueprint]
+mod genesis_helper {
+    struct GenesisHelper;
+
+    impl GenesisHelper {
+        pub fn init(
+            network: NetworkDefinition,
+            mut genesis_data: GenesisData,
+            mut whole_lotta_xrd: Bucket,
+            validator_owner_token: [u8; 26], // TODO: Clean this up
```
LukasGasior1, line 49: not really sure what's the desired clean up here, but carried this comment over from epoch manager :)

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+                    scrypto_encode(&AccountDepositInput { bucket }).unwrap(),
+                );
+            }
+
+            // return the remainder
+            whole_lotta_xrd
+        }
+
+        fn create_resource(
+            bech32_encoder: &Bech32Encoder,
```
talekhinezh, line 155: I don't think we should have the bech32 encoder in here. This should be handled by frontends rather than in the engine since we've agreed in the past that the engine should be abstracted from Bech32 + NetworkDefinition.

LukasGasior1: Okay, sure :+1: This was only used to provide a detailed description of the owner badge
```
 .metadata(
    "description",
    format!(
        "Owner badge for resource {}",
        bech32_encoder.encode_resource_address_to_string(&resource_address)
    ),
)
```

LukasGasior1: although arguably "the engine should be abstracted from Bech32 + NetworkDefinition" still holds as the NetworkDefinition is provided as a parameter

talekhinezh: It looks like it was only added a parameter for the purpose of Bech32?

LukasGasior1: yes

LukasGasior1: @talekhinezh I've removed it

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+            if resource.symbol.eq_ignore_ascii_case("XRD") {
+                panic!("XRD shouldn't be included in genesis resources");
+            }
+
+            let mut metadata = BTreeMap::new();
+            metadata.insert("symbol".to_owned(), resource.symbol.clone());
+            metadata.insert("name".to_owned(), resource.name);
+            metadata.insert("description".to_owned(), resource.description);
+            metadata.insert("url".to_owned(), resource.url);
+            metadata.insert("icon_url".to_owned(), resource.icon_url);
```
talekhinezh, line 161: Can we add a TODO to update this to use the metadata Url value once that is implemented?

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+
+            let mut access_rules = BTreeMap::new();
+            access_rules.insert(Deposit, (rule!(allow_all), rule!(deny_all)));
+            access_rules.insert(Withdraw, (rule!(allow_all), rule!(deny_all)));
+
+            if let Some(owner) = owner_with_mint_and_burn_rights {
```
talekhinezh, line 167: What about update_metadata?

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+                // Note that we also set "tags" metadata later on
+                let owner_badge = ResourceBuilder::new_fungible()
```
talekhinezh, line 169: I think a non_fungible would be more appropriate and probably use of a single resource.

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, ScryptoSbor)]
+pub struct GenesisResource {
+    pub symbol: String,
+    pub name: String,
+    pub description: String,
+    pub url: String,
+    pub icon_url: String,
+    pub address_bytes: [u8; 26],
```
talekhinezh, line 34: Is this a straight mapping from Olympia?

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+            access_rules.insert(Withdraw, (rule!(allow_all), rule!(deny_all)));
+
+            if let Some(owner) = owner_with_mint_and_burn_rights {
+                // Note that we also set "tags" metadata later on
+                let owner_badge = ResourceBuilder::new_fungible()
+                    .divisibility(DIVISIBILITY_NONE)
+                    .metadata(
+                        "name",
+                        format!("Resource Owner Badge ({})", resource.symbol),
+                    )
```
talekhinezh, line 174: Perhaps add a metadata field for the `address` of the resource

radix-engine/src/ledger/bootstrap.rs
```diff
@@ -518,6 +587,28 @@ pub fn create_genesis(
         });
     }
 
+    // Genesis helper package
+    {
+        let genesis_helper_code = include_bytes!("../../../assets/genesis_helper.wasm").to_vec();
```
talekhinezh, line 592: Can we add a TODO to add access rules to prevent any other user from calling genesis helper

talekhinezh (approved):

PR #947, opened 2023-04-04
Expose access rules fields; add resolve_entry helper method

LukasGasior1:
Makes `get_default_mutability` public and re-adds pub `get_all_method_mutability` and `get_all_grouped_auth_mutability`.
Also adds a helper method `resolve_entry` which removes some duplication.

### DIFF
--- a/radix-engine-interface/src/blueprints/resource/access_rules.rs
+++ b/radix-engine-interface/src/blueprints/resource/access_rules.rs
@@ -120,12 +120,7 @@ impl AccessRulesConfig {
     pub fn get_access_rule(&self, key: &MethodKey) -> AccessRule {
         match self.method_auth.get(key) {
             None => self.get_default(),
-            Some(AccessRuleEntry::AccessRule(access_rule)) => access_rule.clone(),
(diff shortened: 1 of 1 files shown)


radix-engine-interface/src/blueprints/resource/access_rules.rs
```diff
@@ -159,14 +157,8 @@ impl AccessRulesConfig {
...
-    fn get_default_mutability(&self) -> AccessRule {
-        match &self.default_auth_mutability {
-            AccessRuleEntry::AccessRule(access_rule) => access_rule.clone(),
-            AccessRuleEntry::Group(name) => match self.grouped_auth.get(name) {
-                Some(access_rule) => access_rule.clone(),
-                None => AccessRule::DenyAll,
-            },
-        }
+    pub fn get_default_mutability(&self) -> AccessRule {
+        self.resolve_entry(&self.default_auth_mutability)
```
talekhinezh, line 161: Wow, was surprised rust didn't complain about multiple `self` reference use but realized that these are read-only refs. Magic!

talekhinezh (approved):

PR #948, opened 2023-04-05
Feature: Add remove method to KVStore

talekhinezh:
## Summary
Add Remove method to Scrypto KVStore

### For Dapp Developers
Removing entries from KVStore is now possible (provided that the entry does not contain owned components)


scrypto/src/component/kv_store.rs
```diff
@@ -122,6 +122,34 @@ impl<
...
+        let mut env = ScryptoEnv;
+        let key_payload = scrypto_encode(&key).unwrap();
+        let offset = SubstateOffset::KeyValueStore(KeyValueStoreOffset::Entry(key_payload.clone()));
+        let handle = env
+            .sys_lock_substate(
+                RENodeId::KeyValueStore(self.id),
+                offset.clone(),
+                LockFlags::MUTABLE,
+            )
+            .unwrap();
```
0xOmarA, line 137: So, this panics if there is value associated with the given key, right? I think that instead of a panic, we should return `None` if the key is not present if we want to be consistent with how Rust does it: https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.remove

scrypto/src/component/kv_store.rs
```diff
@@ -122,6 +122,34 @@ impl<
             .unwrap();
         env.sys_drop_lock(handle).unwrap();
     }
+
+    /// Inserts a new key-value pair into this map.
```
0xOmarA, line 126: Minor: comment is inaccurate

scrypto/src/component/kv_store.rs
0xOmarA: I see that you have a test for this exact case so please ignore the above

0xOmarA (approved):

scrypto/src/component/kv_store.rs
0xOmarA: I see that you have a test with this exact case (deleting a key which doesn't exist) and it behaves as expected. So, ignore my above comment

talekhinezh: The `unwrap` is if the substate cannot be locked for any reason (e.g. if the nodeId doesn't exist).

PR #956, opened 2023-04-13
tests: Add tests for distinguishing proof/bucket

dhedey:
## Summary
Adds some tests related to Atom's report in Discord - notably about confusing errors when passing a bucket/proof where the other is expected.

## Details
The relevant tests should be un-ignored when the issue is fixed.


iamyulong:
Thank you David for adding the tests!

Now might be a good time to implement own/refererence type check since the new schema and `Own/Reference/NodeId` clean-up have been merged. :)

iamyulong (approved):

PR #958, opened 2023-04-16
Feature/SCRY-421 costing modification

mstrug-rdx:
## Summary
Implementation of new costing model for Kernel API basing on CPU resources usage.

## Details
Basing on the tooling described [here](https://radixdlt.atlassian.net/wiki/spaces/S/pages/3026190337/Radix+Engine+Resource+tracker) fee table `kernel_api_cost()` function has been modified - it is divided into two parts: 1st one uses CPU resources usage mechanism to apply cost and the other one uses already present cost basing on memory usage.

CPU resources usage costing mechanism uses special build (features: `resource_usage`) executed on Amazon EC2 instance (used as a reference environment for calculations) to gather data by execution of the whole `radix-engine-tests` on release build. Then output files are converted to summary table basing on Kernel API function call and function parameters (where applicable). Summary table is used to create `kernel_api_cost_from_cpu_usage()` function (currently manually). Detailed description how to run the whole process is available [here](https://radixdlt.atlassian.net/wiki/spaces/S/pages/3042115875/Running+CPU+costing+evaluation).
Source table is available [here](https://docs.google.com/spreadsheets/d/1hfhFfafkeeCw1_Aqcil-Tov_7nWAJpOKUByE_g5BtYc/edit?usp=sharing)

Also mechanism to scale easily up/down whole costs is added (by modification `COSTING_COEFFICENT_DIV_BITS_ADDON` constant which is used to shift left/right function output), current value of this constant was adjusted to provide costing on level which lets to pass all tests without modifications (except three tests which validates each cost reason: `test_basic_transfer`, `test_flash_loan`, `test_radiswap`).

Kernel API functions which are currently not covered by this new costing mechanism because of its simplicity (just returning struct field):
- `kernel_get_callback()`
- `kernel_get_current_depth()`
Functions not covered, a little more complicated than two above: 
- `kernel_get_lock_info()`
- `kernel_get_node_info()`
Functions not covered, due to `//todo remove` comment: 
- `kernel_get_current_actor()`
- `kernel_load_common()`
- `kernel_load_package_package_dependencies()'

For `kernel_invoke()` function call of `Package::publish_native` and `Package::publish_wasm_advanced` linear regression was used to calculate cost basing on `input_size` parameter, in linked [spreadsheet](https://docs.google.com/spreadsheets/d/1hfhFfafkeeCw1_Aqcil-Tov_7nWAJpOKUByE_g5BtYc/edit?usp=sharing) charts are available.


radix-engine/src/kernel/kernel.rs
```diff
@@ -319,18 +322,16 @@ where
     M: KernelCallbackObject,
     S: SubstateStore,
 {
```
iamyulong, line 324: Can we remove tracing for `KernelInternalApi`? They are unstable APIs for special needs of some kernel modules.

radix-engine/src/kernel/kernel_callback_api.rs
```diff
@@ -132,4 +132,20 @@ pub trait KernelCallbackObject: Sized {
...
+        api: &mut Y,
+    ) -> Result<(), RuntimeError>
+    where
+        Y: KernelApi<Self>;
+
+    fn on_read_bucket<Y>(api: &mut Y)
+    where
+        Y: KernelApi<Self>;
+
+    fn kernel_read_proof<Y>(api: &mut Y)
```
iamyulong, line 148: No need to add call back for `read_bucket` and `read_proof`.

radix-engine/src/system/system_modules/costing/costing_module.rs
```diff
@@ -370,4 +379,49 @@ impl<V: SystemCallbackObject> SystemModule<SystemCallback<V>> for CostingModule
...
+                        entity_type,
+                        virtual_node,
+                    })
+                },
+                1,
+            )?;
+        Ok(())
+    }
+
+    fn on_read_bucket<Y: KernelApi<SystemCallback<V>>>(api: &mut Y) {
```
iamyulong, line 404: Same

radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -54,17 +81,321 @@ impl FeeTable {
...
+                    ("Validator", "update_key") => 386889,
+                    ("Worktop", "Worktop_drain") => 80356,
+                    ("Worktop", "Worktop_drop") => 76247,
+                    ("Worktop", "Worktop_put") => 244598,
+                    ("Worktop", "Worktop_take") => 320777,
+                    ("Worktop", "Worktop_take_all") => 68469,
+                    ("Worktop", "Worktop_take_non_fungibles") => 174567,
+                    ("Package", "publish_native") => (input_size * 13 + 10908) >> 2, // calculated using linear regression on gathered data
+                    ("Package", "publish_wasm_advanced") => input_size * 25 + 218153, // calculated using linear regression on gathered data
+                    _ => 454121, // average of above values
```
iamyulong, line 298: At the moment, native blueprints and wasm blueprints are calling `consume_cost_units` directly by themselves. 

We need a way to consolidate the two mechanisms.

radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -54,17 +81,321 @@ impl FeeTable {
...
+                            EntityType::InternalKeyValueStore => 210,
+                            EntityType::InternalNonFungibleVault => 292,
+                            _ => 264, // average of above values
+                        }
+                    } else {
+                        264 // average of above values
+                    }
+                } else {
+                    // virtual_node
+                    106
```
iamyulong, line 115: Interesting that cost for allocating node ID depends on entity type and virtualisation flag.

iamyulong (commented):
Nice work! 

It's fanstatic that costing for Kernel API and native blueprints are now based on real measurements.

iamyulong (approved):
We still need a few iterations over the `FeeTable`, especially to move native blueprints costing to their package and to add size factor.

Approve this PR as-is so we can get the adjusted costing in and evaluate the overall effect.

radix-engine/src/system/system_modules/costing/fee_table.rs
mstrug-rdx: Internally `allocate_node_id()` has additional code which creates new `NodeId` struct, which is missing from `allocate_virtual_node_id()` because it gets `NodeId` as a parameter, so this is why virtual node id takes less instructions.
Also there is variation of the instructions count over EntityType because of `ids.insert(node_id)` call which inserts new item to `BTreeSet` in `allocate_node_id()` function, most probably if memory allocation occurs to extend available capacity it takes longer. It looks that dependency of the instructions count from EntityType is rather random.
Because of above I think we should just take average of the `kernel_allocate_node_id` instruction calls instead of EntityType based match. @iamyulong WDYT?

radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -54,17 +81,321 @@ impl FeeTable {
...
+                Some(EntityType::GlobalValidator) => 2392,
+                Some(EntityType::GlobalVirtualEcdsaAccount) => 1613,
+                Some(EntityType::GlobalVirtualEcdsaIdentity) => 1559,
+                Some(EntityType::InternalAccount) => 959,
+                Some(EntityType::InternalFungibleVault) => 1018,
+                Some(EntityType::InternalGenericComponent) => 961,
+                Some(EntityType::InternalKeyValueStore) => 851,
+                Some(EntityType::InternalNonFungibleVault) => 1035,
+                _ => 1528, // average of above values
+            },
```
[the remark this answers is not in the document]
mstrug-rdx: Do you mean size of `module_init: BTreeMap<SysModuleId, BTreeMap<SubstateKey, IndexedScryptoValue>>` argument of `kernel_create_node()` ?

PR #959, opened 2023-04-17
Architecture: Refactor Non fungible Vault + Many Refactors

talekhinezh:
## Summary
* Add Index API
* Refactor NonFungible vaults to use Index API
* Add 3 variants of `SubstateKey`: `Tuple`, `Map`, `Sorted`
* Refactored Substate lock api to be more specific given the datastructure
* Removed receiver from method invocation and only allow tuple substates to be accessed by method actor

## Details
An Index API has been added to support iterable substates specifically for use with NonFungibles. This lets us get rid of any memory problems associated with using a `BTreeMap` to store non fungibles.

A separation of abstraction between a `DBKey` and a `SubstateKey` has also been added. This is in order to support some constraints which a database may require which we don't want to necessarily be constrained with at the substate layer. An example of this is JMTs requirement that keys are spread evenly.

### For Internal Integrators
The substate layout of NonFungible vaults has been updated to use an index.


radix-engine/src/track/track.rs
```diff
@@ -114,22 +110,106 @@ impl TrackedSubstateKey {
...
+            | TrackedSubstateKey::ReadExistAndWrite(_, Write::Delete)
+            | TrackedSubstateKey::ReadOnly(ReadOnly::NonExistent)
+            | TrackedSubstateKey::Garbage => None,
+        }
+    }
+
+    pub fn set(&mut self, value: IndexedScryptoValue) {
+        match self {
+            TrackedSubstateKey::Garbage => {
+                *self = TrackedSubstateKey::WriteOnly(Write::Update(RuntimeSubstate::new(value)));
```
iamyulong, line 139: Can it be `ReadAndWrite` here?

radix-engine/src/system/system.rs
```diff
@@ -665,24 +582,200 @@ where
...
+            .kernel_remove_substate(node_id, module_id, &substate_key)?
+            .map(|v| v.into());
+
+        Ok(rtn)
+    }
+
+    fn scan_index(&mut self, node_id: &NodeId, count: u32) -> Result<Vec<Vec<u8>>, RuntimeError> {
+        let type_info = TypeInfoBlueprint::get_type(&node_id, self.api)?;
+        match type_info {
+            TypeInfoSubstate::IterableStore => {}
```
iamyulong, line 733: Should we rename all `IterableStore` to `Index`, or the other way, for consistency?

radix-engine/src/system/system.rs
```diff
@@ -699,7 +792,7 @@ where
     }
 
     #[trace_resources]
-    fn insert_into_sorted_store(
+    fn insert_into_sorted_index(
```
iamyulong, line 795: Same for `SortedStore`

iamyulong (approved):

radix-engine/src/system/system.rs
talekhinezh: Yup good catch

radix-engine/src/track/track.rs
talekhinezh: Nope, this should never be `ReadAndWrite` as `Garbage` is only created from `WriteOnly`

simulator/src/ledger/dumper.rs
```diff
@@ -114,19 +111,6 @@ pub fn dump_component<T: SubstateDatabase, O: std::io::Write>(
         );
     }
 
-    writeln!(output, "{}", "Non-fungible Resources".green().bold());
```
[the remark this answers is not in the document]
talekhinezh: Added back in

PR #962, opened 2023-04-19
Split genesis into multiple transactions

LukasGasior1:
There are now three kinds of genesis transaction:
- a single system bootstrap transaction: creates the system packages / resources / etc
- 0 or more genesis data ingestion transactions: add initial validators/stakes/resources/resource-balances/xrd-balances (must be executed in order, i.e. a transaction that creates a given validator must come before a transaction that allocates its stakes and similarly resource creation must come before resource allocation) 
- a single genesis wrap up transaction: finishes up the genesis data ingestion; creates a faucet


assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -4,178 +4,204 @@ use scrypto::blueprints::epoch_manager::*;
...
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, ScryptoSbor)]
+pub struct GenesisResourceAllocation {
+    pub account_index: u32,
+    pub amount: Decimal,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, ScryptoSbor)]
+pub enum GenesisDataChunk {
```
talekhinezh, line 37: Would it make better sense if all accounts were stored in a single index rather than per chunk?

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -4,178 +4,204 @@ use scrypto::blueprints::epoch_manager::*;
...
-                scrypto_encode(&EpochManagerCreateInput {
-                    validator_owner_token,
-                    component_address: epoch_manager_component_address,
-                    validator_set: validators_with_initial_stake,
-                    initial_epoch,
-                    rounds_per_epoch,
-                    num_unstake_epochs,
+        fn create_validator(&mut self, validator: GenesisValidator) {
+            let (validator_address, owner_token_bucket): (ComponentAddress, Bucket) =
+                Runtime::call_method(
```
talekhinezh, line 102: Can we write/use sdks for epoch_manager and accounts rather than using `Runtime::call_method`

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -4,178 +4,204 @@ use scrypto::blueprints::epoch_manager::*;
...
+            for (validator_key, stake_allocations) in allocations.into_iter() {
+                let validator_address = self.validators.get(&validator_key).unwrap();
+                for GenesisStakeAllocation {
+                    account_index,
+                    xrd_amount,
+                } in stake_allocations.into_iter()
+                {
+                    let staker_account_address = accounts[account_index as usize].clone();
+                    let stake_bucket = self.xrd_vault.take(xrd_amount);
+                    let lp_bucket: Bucket = Runtime::call_method(
```
talekhinezh, line 153: Same here, should use an sdk

assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -217,31 +243,71 @@ mod genesis_helper {
...
+                    scrypto_encode(&AccountDepositInput { bucket }).unwrap(),
                 );
             }
-            bucket.drop_empty();
+        }
+
+        pub fn wrap_up(&mut self) -> Bucket {
+            // A little hack to move to the next epoch,
+            // which also updates the validator set
+            // TODO: clean this up (add a dedicated method in epoch manager?)
```
talekhinezh, line 296: Yes, epoch_manager now has a method called `start` which may be the right method to start the epoch_manager

assets/blueprints/genesis_helper/src/lib.rs
LukasGasior1: hmh, but then we'd need a single transaction that lists all accounts? which may be quite sizeable

LukasGasior1: ah, but we could split the index into multiple txns too, but then we'd need to store it in the genesis helper component (in a kv map?) which may ruin the whole benefit

LukasGasior1: I did a quick size test with the current format:
a single genesis txns that was previously ~28.5 MiB, when turned into ~600 txns (granularity tbd) totalled ~37 MiB. Looks quite reasonable to me

talekhinezh: Ah got it, so each GenesisDataChunk is supposed to correspond to another txn. Makes sense!

LukasGasior1: > so each GenesisDataChunk is supposed to correspond to another txn
yup

assets/blueprints/genesis_helper/src/lib.rs
LukasGasior1: done

radix-engine/src/system/bootstrap.rs
```diff
@@ -664,109 +848,100 @@ pub fn create_genesis(
...
             }
             .into(),
         );
-
-        let bucket = id_allocator.new_bucket_id().unwrap();
         instructions.push(Instruction::CallFunction {
-            package_address: FAUCET_PACKAGE,
-            blueprint_name: FAUCET_BLUEPRINT.to_string(),
+            package_address: GENESIS_HELPER_PACKAGE,
+            blueprint_name: "GenesisHelper".to_string(),
```
talekhinezh, line 862: Can we use `GENESIS_HELPER_BLUEPRINT` variable here

talekhinezh (approved):

PR #965, opened 2023-04-20
Fight non-deterministic instrumentation output when fuzzing

lrubasze:
## Summary

Introduced few changes to improve consistency across runs for the same input data.

## Details

Following changes introduced to increase the consistency:
* Use stub hasher (instead of random one) for hash and index maps/sets
* disabled cache in VM-related stuff (most likely the cache uses some hashing inside)
* initialize static objects outside fuzzing loop
* save TestRunner state before fuzzing loop and restore it upon each fuzzing loop step
* use "resettable" lazy_static when fuzzing

Above changes increased the stability to `~99.5%`, which is good enough (for more info see _Nerd details_ below)

Additionally:
* got rid of `unsafe` when initializing well_known_lookup
It was suspected to impact consistency, but it turned out it does not. But it is always good to get rid of `unsafe` if possible.
* reformat Fuzzing summary (include coverage and stability)
* improve Fuzzing processing
  * process even if failed or cancelled
  * quit AFL instances before processing

## Nerd details

Inconsistency is indicated with following:
* warning message `WARNING: Instrumentation output varies across runs` when AFL starts and calibrates itself using provided input data
The more messages like that the less stable is the fuzzing process.
image
* stability field in AFL UI (more details [Path geometry](https://github.com/AFLplusplus/AFLplusplus/blob/3b6fcd911a860a8c823c912c4b08b423734e4cfe/docs/afl-fuzz_approach.md#path-geometry)
before changes:
image
after changes:
image

According to the linked above stability description as long as the stability is 

> shown in purple the fuzzing process is unlikely to be negatively affected

### DIFF
--- a/fuzz-tests/Cargo.toml
+++ b/fuzz-tests/Cargo.toml
@@ -29,7 +29,7 @@ scrypto-unit = { path = "../scrypto-unit", default-features = false }
 # - warnings such as "Instrumentation output varies across runs"
 # - low stability reports
 # More details: https://github.com/rust-fuzz/afl.rs#lazy_static-variables
-lazy_static = { git = "https://github.com/rust-fuzz/resettable-lazy-static.rs" }
+lazy_static = { git = "https://github.com/rust-fuzz/resettable-lazy-static.rs", rev = "c5eb91f" }
 
 [workspace]
 members = ["."]
--- a/fuzz-tests/Cargo.toml
+++ b/fuzz-tests/Cargo.toml
@@ -35,7 +35,11 @@ members = ["."]
 
 [profile.release]
 # TODO: check if really this required
-debug = 1
+# Commented out out since due to following error, when building AFLplusplus
+#  - in debug mode (see GNUmakefile:142 - Werror is set)
+#  - with some modern gcc toolchain: clang 16.0, llvm-gcc 14
+# src/afl-fuzz-redqueen.c:1603:20: warning: variable 'cons_0' set but not used [-Wunused-but-set-variable]
+#debug = 1
 
 [[bin]]
 name = "transaction"
--- a/fuzz-tests/Cargo.toml
+++ b/fuzz-tests/Cargo.toml
@@ -19,6 +19,7 @@ sbor = { path = "../sbor", default-features = false }
 radix-engine = { path = "../radix-engine", default-features = false, features = ["radix_engine_fuzzing"] }
 radix-engine-constants = { path = "../radix-engine-constants" }
 radix-engine-interface = { path = "../radix-engine-interface", default-features = false }
+radix-engine-stores = { path = "../radix-engine-stores", default-features = false }
 transaction = { path = "../transaction", default-features = false }
 utils = { path = "../utils", default-features = false, features = ["radix_engine_fuzzing"] }
 scrypto-unit = { path = "../scrypto-unit", default-features = false }
--- a/fuzz-tests/Cargo.toml
+++ b/fuzz-tests/Cargo.toml
@@ -20,7 +20,7 @@ radix-engine = { path = "../radix-engine", default-features = false }
 radix-engine-constants = { path = "../radix-engine-constants" }
 radix-engine-interface = { path = "../radix-engine-interface", default-features = false }
 transaction = { path = "../transaction", default-features = false }
-utils = { path = "../utils", default-features = false }
+utils = { path = "../utils", default-features = false, features = ["radix_engine_fuzzing"] }
 scrypto-unit = { path = "../scrypto-unit", default-features = false }
 
 [workspace]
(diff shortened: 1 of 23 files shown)


sbor/src/schema/well_known_types.rs
```diff
@@ -24,54 +24,47 @@ pub mod basic_well_known_types {
...
-                    let mut lookup: [sbor::rust::mem::MaybeUninit<Option<TypeData<$custom_type_kind, LocalTypeIndex>>>; 256] = unsafe {
-                        sbor::rust::mem::MaybeUninit::uninit().assume_init()
-                    };
+        paste::paste! {
+            const [<$lookup_name:upper _INIT>]: Option<TypeData<$custom_type_kind, LocalTypeIndex>> = None;
 
-                    for elem in &mut lookup[..] {
-                        unsafe { sbor::rust::ptr::write(elem.as_mut_ptr(), None); }
-                    }
+            lazy_static::lazy_static! {
```
lrubasze, line 30: This part is not strictly related to Fuzzing stability. But I thought it would make sense to get rid `unsafe` if possible.
Unless there is some particular reason to keep `unsafe` there... If yes, please let me know.

radix-engine/src/vm/wasm/wasm_instrumenter.rs
```diff
@@ -4,10 +4,13 @@ use crate::vm::wasm::{WasmMeteringConfig, WasmModule};
 use sbor::rust::sync::Arc;
 
 pub struct WasmInstrumenter {
-    #[cfg(not(feature = "moka"))]
+    #[cfg(all(not(feature = "radix_engine_fuzzing"), not(feature = "moka")))]
```
talekhinezh, line 7: Can we add some comments to why this configuration is required

radix-engine/src/vm/wasm/wasmi.rs
```diff
@@ -1,11 +1,13 @@
 use sbor::rust::mem::transmute;
 use sbor::rust::mem::MaybeUninit;
+#[cfg(not(feature = "radix_engine_fuzzing"))]
```
talekhinezh, line 3: Can we add some comments to why this configuration is required

sbor/src/schema/well_known_types.rs
talekhinezh: As mentioned in our chat, let's make sure to have this reviewed by the `git blame` of this code.

scrypto-unit/src/test_runner.rs
```diff
@@ -227,6 +235,22 @@ impl TestRunner {
         }
     }
 
+    pub fn get_snapshot(&self) -> TestRunnerSnapshot {
```
talekhinezh, line 238: Minor: rename to `create_snapshot()`. `get_` seems to imply to me a return of borrowed value.

talekhinezh (approved):

fuzz-tests/Cargo.toml
```diff
@@ -9,26 +9,38 @@ cargo-fuzz = true
...
-utils = { path = "../utils", default-features = false }
+utils = { path = "../utils", default-features = false, features = ["radix_engine_fuzzing"] }
 scrypto-unit = { path = "../scrypto-unit", default-features = false }
 
+[patch.crates-io]
+# Use "resettable" `lazy_static` version to prevent incorrect AFL behaviour, eg.
+# - warnings such as "Instrumentation output varies across runs"
+# - low stability reports
+# More details: https://github.com/rust-fuzz/afl.rs#lazy_static-variables
+lazy_static = { git = "https://github.com/rust-fuzz/resettable-lazy-static.rs" }
```
dhedey, line 32: If we do this we need to fix the commit hash to prevent injection risks.

sbor/src/schema/well_known_types.rs
dhedey: Ahh nice fix / workaround :).

dhedey (commented):
Just looked at the bits you flagged to me - concern about missing package version fixing, otherwise looks good.

fuzz-tests/Cargo.toml
lrubasze: Good point! Thanks @dhedey

scrypto-unit/src/test_runner.rs
lrubasze: Yup, that makes sense.

lrubasze:
All requested changes applied. Thanks for review!

fuzz-tests/Cargo.toml
dhedey: Is it possible to specify the full git hash? A collision of 7 hex characters is easily brute forcible.

lrubasze: Sorry, missed your comment and merged 😕 
Separate PR created for this: https://github.com/radixdlt/radixdlt-scrypto/pull/971

PR #966, opened 2023-04-20
Feature/add lock unlock auth

iamyulong:
## Summary

This PR protects `lock_amount`, `unlock_amount`, `lock_non_fungibles`, `unlock_non_fungibles` from public access. This is to safeguard the locked resource state.

The approach taken is to leverage the `parent-child` based authorisation (since there are no access rules for local nodes). As a result, I had to split `Bucket` blueprint into `FungibleBucket` and `NonFungibleBucket` due to the single parent constraint.

## Details

- Added auth for  `lock_amount`, `unlock_amount`, `lock_non_fungibles`, `unlock_non_fungibles`;
- Split `Bucket` into `FungibleBucket` and `NonFungibleBucket` (backend)
- Fixed the amount parsing bug in proof composition


radix-engine/src/blueprints/resource/bucket_common.rs
```diff
@@ -0,0 +1,35 @@
...
+pub enum BucketError {
+    ResourceError(ResourceError),
+    ProofError(ProofError),
+    MismatchingResource,
+    InvalidAmount,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
+pub struct BucketInfoSubstate {
+    pub resource_address: ResourceAddress,
```
talekhinezh, line 19: I think we can remove this since we can just get this resource address through the type_info api call. Or can we atleast add a TODO to fix this.

radix-engine/src/blueprints/resource/resource_manager_common.rs
```diff
@@ -221,35 +259,98 @@ fn build_access_rules(
...
+                NonFungibleGlobalId::package_actor(RESOURCE_MANAGER_PACKAGE),
+            )),
+        ))),
         DenyAll,
     );
 
-    (resman_access_rules, vault_access_rules)
+    // Not that if a local reference to a bucket is passed to another actor, the recipient will be able
+    // to take resource from the bucket. This is not what Scrypto lib supports/encourages, but can be done
+    // theoretically.
```
talekhinezh, line 308: I think we will want to prevent local references from being passed through call frames.

talekhinezh: This is probably part of the larger discussion around locks/refs/call frames.

talekhinezh (approved):

radix-engine/src/blueprints/resource/bucket_common.rs
iamyulong: I thought I added a TODO somewhere else. Will update if not

radix-engine/src/blueprints/resource/resource_manager_common.rs
iamyulong: Exactly!

PR #977, opened 2023-04-27
Feature/Add back typed substate layout

dhedey:
## Summary
Adds back typed substate values which we need for the Core API - and captures the various invariants / expectations about valid/possible EntityType/Substate combinations in code form.

Also fixes a few compile issues I saw on `develop`...

## Testing

I've added a test that we can ingest the Genesis receipt and pass the changes through this function without error.

In a future PR I will add a test that verifies Genesis receipts validate against the schemas in the resultant database.

### For Internal Integrators

A few changes:
* `radix-engine-stores` has its higher level stuff split into a `radix-engine-queries` crate - which includes the new `typed_substate_layout` methods.
* Use `typed_substate_layout` along with the receipt to map to typed changes for the Core API - see the bootstrap tests for an example of how to do this.


radix-engine-stores/src/hash_tree/test.rs
```diff
@@ -366,8 +366,15 @@ fn change(
     substate_offset_seed: u8,
     value_hash_seed: Option<u8>,
 ) -> SubstateHashChange {
+    use crate::interface::DatabaseMapper;
+    use crate::jmt_support::JmtMapper;
+    let (node_id, module_id, substate_key) =
+        substate_id(node_id_seed, module_id, substate_offset_seed);
```
dhedey, line 372: This fixes a compile error in these tests... not sure how this wasn't caught on the previous PR?

radix-engine/src/blueprints/package/package.rs
```diff
@@ -209,7 +215,12 @@ impl PackageNativePackage {
     pub fn schema() -> PackageSchema {
         let mut aggregator = TypeAggregator::<ScryptoCustomTypeKind>::new();
 
-        let substates = Vec::new();
+        let mut substates = Vec::new();
+        substates.push(aggregator.add_child_type_and_descendents::<PackageInfoSubstate>());
```
dhedey, line 219: These were somehow missed off the schema... I guess we don't validate these yet?

iamyulong: This is probabaly caused by `PackagePackage` creating packages via `KernelApi` rather than `SystemApi` (which enforces payload-schema check).

dhedey: Very possibly! It also turns out we didn't validate that someone can't write substates at offsets above their schema. Which this PR fixes :).

In a follow-up PR I'm going to add a test that all genesis substates match their respective schemas

radix-engine-queries/src/typed_substate_layout.rs
```diff
@@ -0,0 +1,561 @@
...
+    }
+}
+
+/// Doesn't include non-object modules, nor transient nodes.
+#[derive(Debug, Clone)]
+pub enum TypedObjectModuleSubstateKey {
+    // Objects
+    Package(PackageOffset),
+    FungibleResource(ResourceManagerOffset),
+    NonFungibleResource(ResourceManagerOffset),
```
iamyulong, line 97: Rename to `(Non)FungibleResourceManager`?

iamyulong (approved):
LGTM

PR #978, opened 2023-04-27
Architecture: Track System Updates

talekhinezh:
## Summary
Add System level updates to Track (including `NodeId`/`ModuleId`/`SubstateKey` abstractions)

## Details
Track now keeps track of both DatabaseUpdates and SystemUpdates. SystemUpdates retains information about the system level `SubstateKeys` which are being updated.

### For Internal Integrators
`CommitResult` has been updated to include SystemUpdates


radix-engine-stores/src/interface.rs
```diff
@@ -158,10 +158,14 @@ pub trait SubstateStore {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
-pub struct DatabaseUpdates {
-    pub database_updates: IndexMap<Vec<u8>, IndexMap<Vec<u8>, DatabaseUpdate>>,
+pub struct StateUpdates {
+    pub database_updates: DatabaseUpdates,
```
dhedey, line 162: I don't think we should be returning both. This doubles the size of our receipt for no reason.

Instead, just exposing SystemUpdates is fine and we can map to the Database keys super easily in the client

radix-engine-stores/src/interface.rs
```diff
@@ -191,11 +195,11 @@ pub trait SubstateDatabase {
         &self,
         node_id: &NodeId,
         module_id: ModuleId,
-        substate_key: SubstateKey,
+        substate_key: &SubstateKey,
     ) -> Option<D> {
         self.get_substate(
             &M::map_to_db_index(node_id, module_id),
```
dhedey, line 201: I've suggested to Jakub that we rename some of these and refactor this mapper to not be generic (as the engine _relies_ on the concrete implementation of the mapper for ordering of the index)

radix-engine-stores/src/interface.rs
dhedey: This can be changed in a follow-up PR though

dhedey (approved):

iamyulong (approved):

radix-engine-stores/src/interface.rs
talekhinezh: The engine does rely on this, but the engine relies on Track to uphold this invariant NOT the database. I'd still like the database layer to be abstracted from this.

radix-engine-stores/src/interface.rs
talekhinezh: Maybe, it feels like a weird abstraction to me but lets chat about it when we get to it.

PR #982, opened 2023-04-28
Bug fix: Events emitted by transient nodes

0xOmarA:
## Summary

This PR fixes a bug in the events implementation where the event schema of events emitted by NodeModules prior to globalization could not be looked up. This is fixed by doing the appropriate replacements to the events in the event store when a node is globalized.

### DIFF
--- a/radix-engine/src/system/system.rs
+++ b/radix-engine/src/system/system.rs
@@ -396,6 +396,7 @@ where
 
         let global_node_id = self.api.kernel_allocate_node_id(entity_type)?;
         let global_address = GlobalAddress::new_or_panic(global_node_id.into());
+
         self.globalize_with_address(modules, global_address)?;
         Ok(global_address)
     }
@@ -429,6 +430,10 @@ where
             .ok_or(RuntimeError::SystemError(SystemError::MissingModule(
                 ObjectModuleId::SELF,
             )))?;
+        self.api.kernel_get_callback().modules.events.replace_key(
+            (node_id, ObjectModuleId::SELF),
+            (*address.as_node_id(), ObjectModuleId::SELF),
+        );
         let mut node_substates = self.api.kernel_drop_node(&node_id)?;
 
         // Update the `global` flag of the type info substate.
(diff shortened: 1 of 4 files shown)


radix-engine/src/system/system.rs
```diff
@@ -468,6 +477,15 @@ where
                         )));
                     }
 
+                    self.api
+                        .kernel_get_callback()
+                        .modules
+                        .events
+                        .add_replacement(
+                            (node_id, ObjectModuleId::SELF),
```
dhedey, line 485: Out of interest- what are these three lines for?

You seem to be mapping from `SELF` to `AccessRules` but I imagine you should be mapping from `AccessRules` to `AccessRules`? But an internal node doesn't have access rules (or indeed any module other than self) so this wouldn't make sense.

Relatedly, I'd perhaps expect you to just provide a (node_id => address) mapping (ignoring the module id) and map independently of the module id :)

0xOmarA: This is actually intentional.

So, a bit of a background on this bug and how it occurs. During the account (and identities) secure-ifying process an `AccessRules` node is created. This isn't a node module, no it's an actual node with a node id. 

During the secure-ifying process the "update rules" method is called on this `AccessRules` node which emits an event. This event is then stored in the event store as: "The SELF node module of this AccessRules node emitted this event". It's stored in the event store in this way because at this point this `AccessRules` is a node and not a node module, so we can't associate it with any parent or anything of that sort.

When the parent node (the account) is globalized, all of the nodes associated with the node modules are dropped and the associated substates are stored as node modules under the parent node. So, in the case of the example I provided above, the `AccessRules` node is dropped and its substates are used for the `AccessRules` node module of the account. 

However, the event store still has the event that I mentioned earlier, the "The SELF node module of this AccessRules node emitted this event", but this node has already been dropped and thus its schema can not be looked up. 

So, to fix this, when a node is globalized we replace all of the events emitted by the SELF node module of the node modules to the appropriate parent node module. So, in this case, we're trying to replace all events emitted by the SELF node module of the AccessRules node to say that they've been emitted by the AccessRules node module of the parent node (the account).

dhedey (approved):
Nice! Approving - sorry didn't realise I hadn't done that before!

Will let you decide if you want to do https://github.com/radixdlt/radixdlt-scrypto/pull/982/files#r1182531311 but happy for you to merge regardless

0xOmarA:
@dhedey Thanks! Would rather keep the name the way it is as.

PR #984, opened 2023-04-28
Feature/fixed faucet address

dhedey:
## Summary
Have properly implemented the `globalize_with_fixed_address` in scrypto, and have set the faucet and genesis helper to use a fixed address.

## Details

The fixed address only works if the transaction is run as a SystemTransaction with pre-allocated addresses - so wouldn't work in regular user transactions. I've thus not exposed it as anything on the Scrypto lib-exposed helper types, but you have to directly use ScryptoEnv to make it work.

It was already very partly implemented on ScryptoEnv, but I had to fix it up and connect it up to the engine APIs.

Have also:
* Cleaned up the Genesis Receipts to be more consistent.
* Improved error reporting of CannotInstantiate Prepare errors on scrypto build
* Change the engine to avoid panicking if a package can't be instantiated for some reason (say, after an update to instrumentation)

## Testing
* Tests now used the fixed faucet address, so this works.
* Have manually tested better error reporting (and no panics) when doing `scrypto build` with a WASM file which requires incorrect linking.

## Update Recommendations

NB - this doesn't actually change the transaction - but the change in faucet address might change peoples' manifests, so wanted to flag it.

### For Internal Integrators
The faucet address is fixed again, so can be exposed in the toolkit / Core API easily.


radix-engine/src/vm/wasm/wasmi.rs
```diff
@@ -957,17 +1001,17 @@ impl WasmEngine for WasmiEngine {
...
             #[cfg(feature = "moka")]
             if let Some(cached_module) = self.modules_cache.get(metered_code_key) {
-                return cached_module.as_ref().instantiate();
+                return Ok(cached_module.as_ref().instantiate());
             }
         }
 
         let code = &instrumented_code.code.as_ref()[..];
-        let module = WasmiModule::new(code).unwrap();
+        let module = WasmiModule::new(code)?;
```
talekhinezh, line 1014: I'm not sure about the error propagation here as this not failing I think is an invariant of the system and we don't currently distinguish between invariant-breaking errors vs. non-invariant-breaking errors.

An example of how this may affect things practically: if a validator has some misconfiguration or broken hardware which causes this error to occur they would rather panic rather than vote on a different result on a transaction from the other validators.

simulator/src/resim/mod.rs
```diff
@@ -89,41 +89,6 @@ use transaction::model::TestTransaction;
 use transaction::model::TransactionManifest;
 use utils::ContextualDisplay;
 
-/// The address of the faucet component, test network only.
-/// TODO: remove
-pub const FAUCET_COMPONENT: ComponentAddress = ComponentAddress::new_or_panic([
```
talekhinezh, line 94: Finally!

talekhinezh (approved):
Looks good just a few comments

radix-engine/src/vm/wasm/wasmi.rs
dhedey: Hmm so I would argue that there's nothing really wrong with a single validator misagreeing and voting for a different outcome. They still wouldn't be able to sync past this transaction, they would still end up panicking later when they try to ingest the ledger from ledger sync and get a byzantine quorum exception (or similar) when they get a different merkle hash.

Conversely - consider if there's a bug in Wasmi - or we upgrade the instrumenter - such that after an instrumentation change,  the instrumented code in a particular package couldn't be run any more - and this wasn't caught at testing time.

Would you rather a liveness break? Or validators agree that the transaction failed?

Personally I'd prefer the latter.

dhedey: Or if we really want to panic, we should at least be panicking higher up the stack in `radix-engine/src/vm/vm.rs` rather than here :D

talekhinezh: The concern is not necessarily a liveness break but as a validator I would rather panic than vote incorrectly and get slashed.

But yes panic prolly makes more sense at a higher layer where the invariant is defined.

dhedey: I'm not sure if voting incorrectly is slashable? Maybe it is. I can add the panic back in `radix-engine/src/vm/vm.rs` if we want.

More generally perhaps in future we need to have "invariant broken" kinds of errors which would let us return a different transaction result (and move the choice to panic or not back into the node) -- very similar to how we have the Aborted transaction result.

PR #985, opened 2023-04-29
Feature/Meaningful entity types and addresses

dhedey:
## Summary
Updated entity types and addresses as per REP-71, to create vanity system addresses which are easy to read/interpret by-eye.

## Details
Also features some renames of constants, in line with REP-71.

## Testing
Added tests to check that the addresses are as per REP-71, and added good comments (including rust docs) to all the constants.

### For dApp Developers
You can now just used `XRD` and `FAUCET` constants in your tests. The old constants (`NATIVE_TOKEN` and `FAUCET_COMPONENT`) are still there as aliases to minimize breakages.

### For Internal Integrators
Note that entity types and addresses all changes.

This will particularly affect the ledger app (cc @siy).

The addresses will change - but this should be for the last time. REP-71 fixes the addresses and entity types for good.


radix-engine-common/src/types/entity_type.rs
```diff
@@ -2,33 +2,108 @@ use strum::FromRepr;
...
+    //=========================================================================
+    // System Components (start with char s for system)
+    //=========================================================================
+    /// The global epoch manager entity. Gives Bech32 prefix: `s` followed by one of `c`, `e`, `6` or `m`.
+    GlobalEpochManager = 0b10000110, //-------------- 10000 => s, 110xx => ce6m [se vanity prefix]
+
+    /// A global validator entity. Gives Bech32 prefix: `s` followed by one of `v`, `d`, `w` or `0`.
+    GlobalValidator = 0b10000010, //----------------- 10000 => s, 110xx => vdw0
+
+    /// The global clock entity. Gives Bech32 prefix: `s` followed by one of `c`, `e`, `6` or `m`.
```
lrubasze, line 27: Tiny thing, but I believe it should be “`5`, `4`, `k` or `h`” here

dhedey: Spot on, thanks! Fixed :)

radix-engine-interface/src/constants.rs
```diff
@@ -1,67 +1,429 @@
...
+//=========================================================================
+// FUNGIBLES
+//=========================================================================
+
+/// XRD is the native token of the Radix ledger.
+/// It is a fungible token, measured in attos (`10^-18`).
+///
+/// It is used for paying fees and staking.
+pub const XRD: ResourceAddress = ResourceAddress::new_or_panic([
+    93, 166, 99, 24, 198, 49, 140, 97, 245, 166, 27, 76, 99, 24, 198, 49, 140, 247, 148, 170, 141,
```
lrubasze, line 17: I’m wondering if it would be possible to use somehow the values from `EntityType` enum instead of hardcoding (since enum cannot be used in constant, then maybe replace enum with constants… I know it seems pretty cumbersome).
Alternatively maybe we could just add some comment here or in entity _types.rs what is the `EntityType` in decimal.

By replacing enum with constant I mean something like this
```rust
pub mod EntityType {
pub const GlobalPackage: u32 = 0b00001101,
  …
}
```

dhedey: So this actually fails at compile time (due to the const functions) if there's an issue -- and I also assert that it's of the right type in the tests below.

I quite like having the address in this format because it's what the generation script outputs; and I quite like having the entity types in that format because it's easier to sanity-check against the Bech32.

I'm happy to add the decimal as a comment to EntityType though :)

lrubasze: Sure, fair enough :) thanks!

iamyulong (approved):

PR #986, opened 2023-04-29
Refined Track vs Database interface

jakrawcz-rdx:
As discussed on our call (https://whimsical.com/substate-keys-and-structure-S3tN6qJ7n9d7MtVvfoHM86).


radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -141,32 +124,32 @@ fn put_substate_changes<S: TreeStore<IndexPayload> + TreeStore<()>>(
...
         changes
             .into_iter()
-            .map(|change| to_index_change(change))
+            .map(|change| to_partition_change(change))
             .collect(),
     )
 }
 
-fn get_index_leaf_entry<S: ReadableTreeStore<IndexPayload>>(
+fn get_parition_leaf_entry<S: ReadableTreeStore<PartitionPayload>>(
```
dhedey, line 143: `get_partition_leaf_entry`

radix-engine-stores/src/lib.rs
```diff
@@ -1,13 +1,12 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
+extern crate core;
```
dhedey, line 3: Why is this needed?

radix-engine-stores/src/memory_db.rs
```diff
@@ -1,61 +1,63 @@
...
+use radix_engine_store_interface::interface::*;
 use sbor::rust::prelude::*;
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub struct InMemorySubstateDatabase {
-    substates: BTreeMap<Vec<u8>, Vec<u8>>,
+    partitions: IndexMap<DbPartitionKey, BTreeMap<DbSortKey, DbSubstateValue>>,
 }
 
 impl InMemorySubstateDatabase {
```
dhedey, line 9: Much cleaner :)

radix-engine-stores/src/rocks_db.rs
```diff
@@ -16,52 +17,64 @@ impl RocksdbSubstateStore {
...
-                let (_, substate_key) =
-                    decode_substate_id(key.as_ref()).expect("Failed to decode substate ID");
-                let value = value.as_ref().to_vec();
-                (substate_key, value)
-            });
+                let (iter_key_bytes, iter_value) = kv.as_ref().unwrap();
+                let iter_key = decode_from_rocksdb_bytes(iter_key_bytes);
+                (iter_key, iter_value.to_vec())
+            })
+            .take_while(move |((iter_partition_key, _), _)| *iter_partition_key == partition_key)
```
dhedey, line 43: This is slightly less efficient than the old approach, as we're to vec'ing the full substate in the next DB key - which could potentially be a full package substate - I wonder if we could do the take while first by checking the key prefix, like the old approach?

radix-engine-stores/src/rocks_db.rs
```diff
@@ -16,52 +17,64 @@ impl RocksdbSubstateStore {
...
+                };
+                result.expect("IO error");
             }
         }
     }
 }
+
+fn encode_to_rocksdb_bytes(partition_key: &DbPartitionKey, sort_key: &DbSortKey) -> Vec<u8> {
+    let mut buffer = Vec::new();
+    buffer.extend(u32::try_from(partition_key.0.len()).unwrap().to_be_bytes());
```
dhedey, line 67: Hmm I was fairly certain the partition key is fixed length. (unlike the sort key)

Is this prefix-based encoding because you didn't realise that? Or because you're wanting to be forward-compatible?

radix-engine-stores/src/rocks_db.rs
dhedey: (We have a separate implementation of this in the node anyway, so this is not a big deal)

radix-engine/README.md
```diff
@@ -10,12 +10,12 @@ Each layer provides an API to the layer above.
...
+| Track                                                                       | Dependency and Update Tracking<br>Proof and Receipt Creation                       | [Substate Store](../radix-engine-store-interface/src/interface.rs)                                                                                                                                                                                                               |                                                          | [Track](src/track)                                                                                         |
```
dhedey, line 20: I imagine the link on line 20 might not want to link to the typed substate store, as we have now separated that abstraction?

radix-engine/src/track/db_key_mapper.rs
```diff
@@ -0,0 +1,88 @@
...
+/// A mapper between the business RE Node / Module / Substate IDs and database keys.
+pub trait DatabaseKeyMapper {
+    /// Converts the given RE Node and Module ID to the database partition's key.
+    fn to_db_partition_key(node_id: &NodeId, module_id: ModuleId) -> DbPartitionKey;
+
+    /// Converts the given Substate key to the database's sort key.
+    fn to_db_sort_key(key: &SubstateKey) -> DbSortKey;
+}
+
+/// A [`DatabaseKeyMapper`] tailored for databases which cannot tolerate long common prefixes
```
dhedey, line 18: Do you think it's worth mentioning that this is the "real" (protocol-enforced) key mapper that we're using for our protocol?

radix-engine/src/track/db_key_mapper.rs
```diff
@@ -0,0 +1,88 @@
...
+                bytes.extend(hash(key).0[12..Hash::LENGTH].to_vec()); // 20 bytes
+                bytes
+            }
+        };
+        DbSortKey(bytes)
+    }
+}
+
+/// Convenience methods for direct `SubstateDatabase` readers.
+pub trait MappedSubstateDatabase {
```
dhedey, line 52: I still feel weird about us needing to specify `SpreadPrefixKeyMapper` in every function call - it keeps making me feel like the mapper choice is arbitrary when it's really not.

I wonder if we specify it as a generic parameter to the trait and remove it from the methods?
EG: `MappedSubstateDatabase<M: DatabaseKeyMapper>` and then only implement specifically `impl<S: SubstateDatabase> MappedSubstateDatabase<SpreadPrefixKeyMapper> for S` then I imagine we can drop the specification of M everywhere - what do you think? Rust might complain about it though, but I don't think it would.

My suggestion feels a bit weird though.

Alternatively:
* We can just fix `pub use SpreadPrefixKeyMapper as ProtocolKeyMapper`, use that in the `MappedSubstateDatabase` impl and let it be changed at compile time if people wanted to 🤷 .
* We can have a `get_protocol_mapped_substate` method on a new trait which fixes `M = ProtocolKeyMapper = SpreadPrefixKeyMapper` and just use that method everywhere :')

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -187,7 +188,7 @@ where
             crate::kernel::resources_tracker::ResourcesTracker::start_measurement();
 
         // Prepare
-        let mut track = Track::<_, JmtMapper>::new(self.substate_db);
+        let mut track = Track::<_, SpreadPrefixKeyMapper>::new(self.substate_db);
```
dhedey, line 191: Maybe this is the answer inside the engine - we make Track a trait - and we instantiate an implementation of `Track::<_, SpreadPrefixKeyMapper>` in the transaction executor when we create it - and then things like `to_state_updates` are functions on the track interface, rather than requiring the mapper to be provided.

Track then save a `TypedSubstateStore<DatabaseKeyMapper>` in its struct so that it knows _which_ typed substate store it is, and so we don't have to specify at each call-site.

That would at least get rid of most of the `SpreadPrefixKeyMapper` generics around the `radix-engine` codebase.

dhedey: (I imagine we can do a similar trick in `scrypto-unit` and `radix-engine-queries`)

dhedey (approved):
Really nice!

I think there is still some changes we can do regarding `SubstateLocalKey` and multiple types for `TupleKey` / `SortedKey` / `MappedKey`, but that can be done in a different PR.

jakrawcz-rdx:
> I think there is still some changes we can do regarding SubstateLocalKey and multiple types for TupleKey / SortedKey / MappedKey, but that can be done in a different PR.

```
if pr.len > 300 && stuff.is_unrelated_enough() && stuff.can_be_done_in_another_pr() {
  stuff.do_in_separate_pr();
}
```
(I literally never regretted splitting work into separated PRs - less scope to review, less things to discuss, easier to revert in case of problems)

+ we probably have to discuss these SubstateLocalKey changes a bit more.

radix-engine-queries/src/query/traverse.rs
```diff
@@ -46,7 +47,7 @@ pub trait StateTreeVisitor {
 
     fn visit_node_id(
         &mut self,
-        _parent_id: Option<&(NodeId, ModuleId, Vec<u8>)>,
+        _parent_id: Option<&(NodeId, ModuleId, DbSortKey)>,
```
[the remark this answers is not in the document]
jakrawcz-rdx: yes, I noticed this inconsistency and thought "this should totally be `SubstateLocalKey`". But the reversible mapping is a prerequisite (which I prefer to do in a separate PR).

radix-engine-stores/src/lib.rs
jakrawcz-rdx: whaaaaaaaat who put it here? Seriously, not me.
(either a weird "update from develop" conflict artifact, or some random IntelliJ "auto-fix" action 🤦 )

radix-engine-stores/src/rocks_db.rs
jakrawcz-rdx: AFAIU, this impl (in scrypto repo) is for tests and local experiments, so let's go for simplicity/readability.

We'll do the performance-driven raw-byte-prefix comparisons in our Node impl, as you say 👍

radix-engine-stores/src/rocks_db.rs
jakrawcz-rdx: I know the partition key is fixed-length, but I don't like the hardcoded `26` (and I can't reach to consts defined in `radix-engine[-interface]` which we no longer depend on here).

Ofc the key encoding scheme used for this test-only rocksdb is irrelevant (as long as it is valid and reversible).
I ack, though, that we will have the same problem in Node's rocksdb impl.

Looks like the best way would be to have a `DbPartitionKey::LENGTH` constant, do you agree?

radix-engine/README.md
jakrawcz-rdx: ahhhh, I forgot to re-adjust these docs in the end, thanks for catching 👍

radix-engine/src/track/db_key_mapper.rs
jakrawcz-rdx: It's the 7th time I regret that we have a trait for this. (without the trait, it would be obvious that precisely this logic is used for real in the protocol)...

radix-engine/src/track/db_key_mapper.rs
jakrawcz-rdx: hah, look, [the 8th time](https://github.com/radixdlt/radixdlt-scrypto/pull/986#discussion_r1181195620) :p

let's drop the trait?

radix-engine/src/transaction/transaction_executor.rs
jakrawcz-rdx: and [the 9th time](https://github.com/radixdlt/radixdlt-scrypto/pull/986#discussion_r1181195620).

radix-engine-stores/src/rocks_db.rs
jakrawcz-rdx: > Looks like the best way would be to have a DbPartitionKey::LENGTH constant

I tried it and now I think I should have said "best" = "easiest to do and will work fine for us ~forever, but actually very lame in terms of database api". So I am not so quick to apply that without hearing other opinions.

Alternatives I see:
A) `RocksdbSubstateStore::new(root_path, partition_key_length)` (and then db actually checks that on save)
B) actually keep the `len || value` encoding, with slight tweaks for performance: make that length u8 and move it to the end of the key (if rocksdb cares about key distribution at all).

jakrawcz-rdx:
Let me merge this one - I addressed the follow-ups from our discussions in the now-revised https://github.com/radixdlt/radixdlt-scrypto/pull/989.

PR #988, opened 2023-05-01
Fix worktop drop vulnerability 

iamyulong:
## Summary

This PR fixes an issue with `Worktop::drop` function.

Without it, Radix Engine will panic when a mismatched `Own` is passed. 

Also checked how `Own`/`Reference` is used in substates and found no issue.

### DIFF
--- a/radix-engine-interface/src/blueprints/resource/worktop.rs
+++ b/radix-engine-interface/src/blueprints/resource/worktop.rs
@@ -1,9 +1,12 @@
 use crate::blueprints::resource::*;
+use crate::constants::RESOURCE_MANAGER_PACKAGE;
 use crate::data::scrypto::model::*;
 use crate::math::Decimal;
 use crate::*;
+use radix_engine_common::data::scrypto::*;
 use radix_engine_common::types::*;
 use sbor::rust::prelude::*;
+use sbor::*;
 
 pub const WORKTOP_BLUEPRINT: &str = "Worktop";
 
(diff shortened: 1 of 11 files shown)


radix-engine-interface/src/blueprints/resource/worktop.rs
```diff
@@ -1,17 +1,44 @@
...
+
+#[derive(Debug, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+#[sbor(transparent)]
+pub struct OwnedWorktop(pub Own);
+
+impl Describe<ScryptoCustomTypeKind> for OwnedWorktop {
+    const TYPE_ID: GlobalTypeId =
+        GlobalTypeId::Novel(const_sha1::sha1("OwnedWorktop".as_bytes()).as_bytes());
+
+    fn type_data() -> Option<TypeData<ScryptoCustomTypeKind, GlobalTypeId>> {
```
dhedey, line 28: Ahh I see what you mean about this - at some point I wanted to add an `#[sbor(validation_getter = get_validation_method_name)]` attribute to the schema generator for the Describe trait 👍

radix-engine-interface/src/blueprints/resource/worktop.rs
```diff
@@ -1,17 +1,44 @@
...
-    pub worktop: Own,
+    pub worktop: OwnedWorktop,
+}
+
+#[derive(Debug, Eq, PartialEq, ScryptoCategorize, ScryptoEncode, ScryptoDecode)]
+#[sbor(transparent)]
+pub struct OwnedWorktop(pub Own);
+
+impl Describe<ScryptoCustomTypeKind> for OwnedWorktop {
+    const TYPE_ID: GlobalTypeId =
```
dhedey, line 25: We could make it well-known if you like! Would save some size in the ResourceManager schema, although don't know how much that matters.

dhedey (approved):
Nice 👍

PR #989, opened 2023-05-01
Making DatabaseKeyMapper reversible + using this in MappedSubstateDatabase

jakrawcz-rdx:
As requested in: https://github.com/radixdlt/radixdlt-scrypto/pull/986#discussion_r1181126693

This implements the idea described at https://rdxworks.slack.com/archives/C01HK4QFXNY/p1682680003738499, with hash length = 20 (as discussed).


radix-engine/src/track/db_key_mapper.rs
```diff
@@ -1,71 +1,147 @@
...
+/// This implementation is the actual, protocol-enforced one, to be used in public Radix networks.
+///
+/// This implementation achieves the prefix-spreading by:
+/// - using a (long, ~100% unique) hash instead of plain RE Node and Module ID (please note that it
+///   makes this mapping effectively irreversible);
+/// - using a (shorter, but hard to crack) hash prefix for Substate key.
 pub struct SpreadPrefixKeyMapper;
 
 impl DatabaseKeyMapper for SpreadPrefixKeyMapper {
     fn to_db_partition_key(node_id: &NodeId, module_id: ModuleId) -> DbPartitionKey {
```
dhedey, line 39: I know we've left the call now - but I think I do want there to be some way to invert this...

Notably if we need to migrate to a new key-ing paradigm (say, a move to verkle trees), being able to migrate all the data is super useful. Having _some way_ of inverting these would prove useful no doubt.

Ditto - it would be _super useful_ for trustless clients (eg wallets) to know the actual address of something -- or indeed at Xi'an when we're using merkle trees a lot - having this data would be really useful.

Maybe we add the "own address" to TypeInfo @talekhinezh ?

radix-engine/src/track/db_key_mapper.rs
```diff
@@ -1,71 +1,147 @@
...
+                Self::SORTED_DISCRIMINATOR,
+                [
+                    bucket.to_be_bytes().as_slice(),
+                    &Self::to_hash_prefixed(key),
+                ]
+                .concat(),
+            ),
         };
-        DbSortKey(bytes)
+        DbSortKey(combine(discriminator_byte, &rest_bytes))
```
dhedey, line 60: Putting the discriminator byte first feels wrong for JMT performance, and feels weird/wasteful for the "sorted key" abstraction (even if it works) - maybe we could put it at the end?

As it is fixed for a given partition anyway, putting at the end makes most sense I think. It enables invertibility without getting in the way at all.

radix-engine/src/track/db_key_mapper.rs
```diff
@@ -1,71 +1,147 @@
...
+    }
+}
+
+impl SpreadPrefixKeyMapper {
+    /// Length of hashes that are used in the database *instead* of plain RE Node and Module IDs.
+    const PARTITION_KEY_HASH_LENGTH: usize = 26;
+
+    /// Discriminator bytes for telling apart different types of sort keys (needed for reversible
+    /// encoding).
+    const TUPLE_DISCRIMINATOR: u8 = 1;
```
dhedey, line 86: Slightly interesting choice to start discriminators from 0 but I don't mind it too much :D

dhedey (commented):
A few thoughts, happy to approve when resolved 👍

radix-engine/src/track/db_key_mapper.rs
```diff
@@ -1,71 +1,147 @@
...
+                Self::from_hash_prefixed(&rest_bytes[2..]).to_vec(),
+            )),
+            unexpected_byte => panic!("unexpected discriminator byte: {}", unexpected_byte),
+        }
+    }
+}
+
+impl SpreadPrefixKeyMapper {
+    /// Length of hashes that are used in the database *instead* of plain RE Node and Module IDs.
+    const PARTITION_KEY_HASH_LENGTH: usize = 26;
```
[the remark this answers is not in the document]
jakrawcz-rdx: It prevoiusly used 26 and I kept that.
I am a fan of 32.

radix-engine/src/track/db_key_mapper.rs
talekhinezh: Remember a separate index could always be used for both trustless clients and large tree migrations.

I'd prefer optimizing and keeping this execution store for it's main purpose, ie represents proveable state on top of which transactions can be deterministically run on top off.

Other things like querying/traversing can be much better served by separate indices/data structures.

radix-engine/src/track/db_key_mapper.rs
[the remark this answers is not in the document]
dhedey: Given that our addresses are only 30 using 32 isn't any safer

radix-engine/src/track/db_key_mapper.rs
```diff
@@ -1,71 +1,147 @@
...
-            SubstateKey::Map(key) => {
-                hash(key).0[12..Hash::LENGTH].to_vec() // 20 bytes
-            }
-            SubstateKey::Sorted((bucket, key)) => {
-                let mut bytes = bucket.to_be_bytes().to_vec();
-                bytes.extend(hash(key).0[12..Hash::LENGTH].to_vec()); // 20 bytes
-                bytes
-            }
+        let (discriminator_byte, rest_bytes) = match key {
+            SubstateKey::Tuple(field) => (Self::TUPLE_DISCRIMINATOR, vec![*field]),
```
talekhinezh, line 49: ooo, wait a moment, was not aware of this needed discriminator key. Let's have a quick chat on this.

radix-engine/src/track/db_key_mapper.rs
jakrawcz-rdx: > As it is fixed for a given partition anyway (...)

wowowow, I was not aware of this!
In such case, maybe we should actually have API like `list<K>(NodeId, ModuleId) -> Iterator<K>` (where `K` can be `TupleKey`, `MapKey`, `SortedU16Key`) ?
(no discriminator byte)

If the above is not feasible, then I can definitely put the discriminator at the end 👍

radix-engine/src/track/db_key_mapper.rs
jakrawcz-rdx: (coming from C/Java): it is a very good thing to see "illegal discriminator byte: 0" (instead of silent invalid result) when you try to parse a clear (zeroed) buffer :P

radix-engine/src/track/db_key_mapper.rs
jakrawcz-rdx: maybe it's not actually needed? see https://github.com/radixdlt/radixdlt-scrypto/pull/989#discussion_r1181718054

jakrawcz-rdx: I mean: it is NOT needed if all places that call `list()` know what flavor of substate key they expect.
is it a valid assumption? (e.g. in this line https://github.com/radixdlt/radixdlt-scrypto/blob/2b66d33162f4e132d1199762b7cd90d360246c31/radix-engine-queries/src/query/traverse.rs#L181)

talekhinezh: Yes, I think we can just remove this as part of the agreement with a client is that the client will only use a single SubstateKey type for a given partition. Otherwise, the result is undefined.

radix-engine/src/track/db_key_mapper.rs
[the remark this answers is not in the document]
dhedey: Actually maybe 32 does make sense, as it will be the length of address + reserved byte + partition byte according to REP-60.

radix-engine/src/track/db_key_mapper.rs
jakrawcz-rdx: I tried to apply the "client will only use a single SubstateKey type for a given partition" rule (see the latest commit) and there are just 2 cases where it is not obvious; hopefully I will understand more after our call 👍

dhedey: If we agreed to add the length to the RocksDB key to allow separating it into parts, I feel we perhaps should eat an extra byte here (at the end?) to allow easy decoding?

jakrawcz-rdx: > a typeinfo lookup

o_O do you mean like "another db read"? is it THAT bad?
I thought that we have (or easily can have) a helper like `resolveSubstateKeyType(node_id, module_id)`...?

dhedey: In the Core API we're mapping historic transaction receipts where we have no way of reading historic state.

jakrawcz-rdx: > I disagree with removing the discriminator UNLESS we make the key kind statically derivable from the DbPartitionKey.

To an untrained eye (i.e. mine), it looks like we can statically derive this from `NodeId.entity_type() + ModuleId` (see `decode_substate_key()` in the latest commit).
⚠️ I may be wrong - I copy-pasted from [to_typed_substate_key()](https://github.com/radixdlt/radixdlt-scrypto/blob/cf1ee4170d244acfed74b855897ad60ba0a4b993/radix-engine-queries/src/typed_substate_layout.rs#L119) and filled in the gaps with wild guesses).

But if that mapping is right, and we commit to preserving this property in future, then I am actually FOR dropping the discriminator byte from substate key DB encoding. Would you be able to verify this, @talekhinezh ?

jakrawcz-rdx: huh, so looking at the tests, clearly my >2 pages of static substateKey type resolution is imperfect :D
the question is - is it fixable?

dhedey: You're right actually - I had this conversation with Josh last night in fact, and made it a requirement.

At least for now, the Entity Id + Module ID (soon to be Partition Number - see REP-60) will tell us the PartitionKey. This won't necessarily be true in future if we allow scrypto modules to define their partition - instead it will need a schema lookup from a substate. That's probably okay though :)

radix-engine/src/transaction/state_update_summary.rs
```diff
@@ -386,7 +386,7 @@ impl<'a, S: SubstateDatabase> BalanceAccounter<'a, S> {
         self.tracked
             .get(node_id)
             .and_then(|tracked_node| tracked_node.tracked_modules.get(&module_id))
-            .and_then(|tracked_module| tracked_module.substates.get(&M::to_db_sort_key(key)))
+            .and_then(|tracked_module| tracked_module.substates.get(&key.to_db_sort_key()))
```
jakrawcz-rdx, line 389: in this line (in particular) it becomes very visible that introduction of `SortKey` trait defeats the decoupling provided by `mapper` (and also defeats the purpose of mapper trait's existence).

I am open to considering such move, but:
- we should go all the way with it (i.e. do the same to partition key mapping logic!)
- it really is a scope of some next PR, not this one.

radix-engine-tests/tests/balance_changes.rs
```diff
@@ -220,12 +220,12 @@ fn test_balance_changes_when_transferring_non_fungibles() {
             account.into() => indexmap!(
                 resource_address => BalanceChange::NonFungible {
                     added: BTreeSet::new(),
-                    removed: btreeset!(NonFungibleLocalId::integer(2))
+                    removed: btreeset!(NonFungibleLocalId::integer(3))
```
jakrawcz-rdx, line 223: I am quite surprised by this change o_O

(was it definitely wrong before? or is maybe the current impl wrong?)

radix-engine-queries/src/query/traverse.rs
```diff
@@ -175,24 +174,58 @@ impl<'s, 'v, S: SubstateDatabase, V: StateTreeVisitor> StateTreeTraverser<'s, 'v
...
+                                    }
+                                }
+                            }
+
+                            SysModuleId::Metadata | SysModuleId::Virtualized => {
+                                for (map_key, value) in self
+                                    .substate_db
+                                    .list_mapped::<SpreadPrefixKeyMapper, ScryptoValue, MapKey>(
+                                        &node_id,
+                                        SysModuleId::Virtualized.into(),
```
jakrawcz-rdx, line 212: previously, this line (effectively) said `t.into(),`
and right now this is the only difference between the 2 introduced "branches" (well, apart from the `TupleKey` vs `MapKey`).

Just checking - is this intentional?

radix-engine/src/track/db_key_mapper.rs
jakrawcz-rdx: > this mapping should be statically decodable.

even if it is possible, it is quite complex, and (most importantly) it reaches quite deep into business knowledge about node/module/substate key types. I am not sure that a database mapping layer should really have it.

radix-engine-queries/src/query/traverse.rs
talekhinezh: Not intentional this should be `t.into()`

radix-engine-tests/tests/balance_changes.rs
dhedey: It looks like somehow the ordering of the keys in the vault has changed - which suggests that the DbMapper for sorted keys has changed? Which I guess it has?

talekhinezh: It's because the iteration over the index has changed due to the new sort key. Or do you think this should have remained the same?

radix-engine-queries/src/query/traverse.rs
jakrawcz-rdx: alright, I turned this duplicated code into a shared method (parameterized with `<K>` for tuple vs map aspect).

radix-engine-tests/tests/balance_changes.rs
jakrawcz-rdx: if the "index" you refer to is _not_ a sorted index, then yeah, totally - we changed the DB key serialization (hash prefix, discriminator, whatever), so the iteration order might have changed 👍 

AFAIU, this test should not care about which exact NFLI was transferred, right? (it just should assert that account A lost it and account B got it). I adjusted these asserts.

jakrawcz-rdx:
Ok, we seem to have reached the following conclusion:

We go for **R2**.

David:
> Haha I have a uniform distribution on both sides, I can see benefits to both models.
On the R1 side - as we’re adding a hash anyway, I don’t see why it matters to add a discriminator at the end as well :sweat_smile:.
On the R2 side - but I can see that double-specifying the information maybe feels wrong, and as long as the type of the key can (for now) be statically decoded from EntityByte + Partition Number (and such a method is implemented inside scrypto, and there are tests to ensure it is correct inside the genesis transactions), I’m happy with R2.

Josh:
> Would prefer R2 as it more aligns with actual protocol architecture, even at the expense of a little more awkward execution layers
> to me IF [the discriminator byte] was to be stored in the database, it should be stored in a per-partition way

Me:
> both R1 and R2 have valid (conceptual) interpretations, but R1 is more convenient to use (makes the DB's key representations self-contained), while costing just 1 byte.

It happens that current state of this PR is **R2**, achieved by:
- static type annotations [[by Josh](https://github.com/radixdlt/radixdlt-scrypto/pull/989/commits/4d52ab8bec38afd58f16110d7f888557035567ce)]
- removed interference with **D** aspect [[by me](https://github.com/radixdlt/radixdlt-scrypto/pull/989/commits/82612bbf8e381dba641db6b792e00f4bd7846161)]
- my minor fixes of problems noticed on the way [[by me](https://github.com/radixdlt/radixdlt-scrypto/pull/989/commits/6f88ad2ca349bd586dc6871351e7b4a24998495a)]

I suggest we have a final look at these 3 new commits, and merge when approved.

dhedey (approved):
👍  Let's get this merged! Thanks @jakrawcz-rdx  for pushing through :D

jakrawcz-rdx:
Big thanks to everyone involved!

PR #993, opened 2023-05-03
Feature/manifest sbor json + SBOR refactors

dhedey:
## Summary
The goal of this PR was to allow Omar to pick up SBOR JSON integration in the toolkit - which also (apparently) required getting Manifest JSON encoding working.

Ideally though Manifest JSON should work with **Scrypto schemas** - so we can actually statically validate the Manifest against well-known schemas for things like accounts in the toolkit ✨.

It turned out with a few refactors, this was actually remarkably easy to do, and feels like a much neater abstraction and naming scheme. The key changes are:
* Removal of CustomValueKind from the TypeKind generic parameters - it turns out it wasn't actually needed anywhere in the end! Must have been some relic from an early implementation that I never realised I could remove.
* Separation of CustomTypeExtension into CustomSchema and CustomExtension.
* Some changes to the validate methods to work with `Expression` and `Blob` needing to match `Vec<OwnBucket>`/`Vec<OwnProof>` and `Vec<u8>` respectively.

And then the following implementations:
* Typed traversal and payload validation for ManifestSbor (against the Scrypto schema) - see custom_validation.rs for some nice tests.
* (Rustlike) Display and JSON encode for ManifestSbor
* Refactor to how the engine payload validation works to clean up the code a little and avoid needing to pass a &mut reference. The use of a trait also paves the way for adding validation in the tests after execution ends, when we don't have access to a SystemService, just a test runner.

### For Internal Integrators

After this change, the toolkit will be able to implement functionality such as:
1. Validate manifest payloads of sub-payloads against a Scrypto schema
2. Encode manifest payloads or sub-payloads to JSON against a Scrypto schema
3. Encode manifest payloads or sub-payloads to RustLike string using a Scrypto schema

Part 1. can be used to add static validations or CALL_METHOD / CALL_FUNCTION parameters against well-known schemas (eg account, access controller) - or indeed any method if the relevant schemas are provided by the toolkit caller. Probably they first need to ask the toolkit "what schemas do I need" and get them and then they can be passed to the toolkit in a second call.

Part 2/3. can be used for a nicer manifest display of CALL_METHOD / CALL_FUNCTION parameters in the wallet - EG we could display the actual access rules enum names - rather than just their integer values.


radix-engine-common/src/data/manifest/custom_extension.rs
```diff
@@ -0,0 +1,58 @@
...
+                TypeKind::Custom(ScryptoCustomTypeKind::Reference)
+            ),
+            ManifestCustomValueKind::Bucket => {
+                matches!(type_kind, TypeKind::Custom(ScryptoCustomTypeKind::Own))
+            }
+            ManifestCustomValueKind::Proof => {
+                matches!(type_kind, TypeKind::Custom(ScryptoCustomTypeKind::Own))
+            }
+            // Vec<Own> buckets - will check more in the payload_validator
+            ManifestCustomValueKind::Expression => matches!(type_kind, TypeKind::Array { .. }),
```
iamyulong, line 34: Can we at least verify the `element_type`, so anything built on top of `TypedTraverser` can have valid assumption about value-type-matching.

Similary for `Blob`.

radix-engine-common/src/data/scrypto/custom_validation.rs
```diff
@@ -1,15 +1,185 @@
...
+pub trait TypeInfoLookup {
+    fn get_node_type_info(&self, node_id: &NodeId) -> Option<TypeInfoForValidation>;
+    // radix-engine-common is not privy to specifics of engine blueprints
+    // - so instead, we delegate this to the TypeInfoLookup implementation
+    fn is_bucket(&self, type_info: &TypeInfoForValidation) -> bool;
+    fn is_proof(&self, type_info: &TypeInfoForValidation) -> bool;
+}
+
+#[derive(Debug, Clone)]
+pub enum TypeInfoForValidation {
```
iamyulong, line 17: I started with this approach when implementing stateful custom type validation, but realised it's better, from a dependency perspective, if we move stateful validation into system, so `radix-engine-common` has no dependency on type info enum, address/blueprint constants and node property definition.

sbor/src/payload_validation/traits.rs
```diff
@@ -1,13 +1,20 @@
...
+pub trait ValidatableCustomExtension<T>: CustomExtension {
     // Note that the current SBOR extension only supports terminal custom type,
     // i.e., no custom value can be container.
 
-    fn apply_custom_type_validation<'de>(
-        custom_type_validation: &Self::CustomTypeValidation,
-        value: &TerminalValueRef<'de, Self::CustomTraversal>,
-        context: &mut T,
-    ) -> Result<(), ValidationError>;
+    fn apply_validation_for_custom_value<'de>(
```
iamyulong, line 8: Glad we moved away from `TypeValidation`-driven validation to `Value`-driven validation.

radix-engine-common/src/data/manifest/custom_extension.rs
dhedey: Yes - I agree with you on second thoughts.

It requires changing this method to take the schema reference - but I think that's the correct thing! :)

sbor/src/payload_validation/traits.rs
dhedey: Haha yes, it's kinda a little more intuitive, I guess?

The value-driven validation slightly hides the fact that anything passes `TypeValidation::None` and the fact that a mismatch between TypeValidation and TypeKind is a SchemaValidation

BUT I think it's just generally clearer/more intuitive, I agree. And actually in this Manifest case, it's required :)

radix-engine-common/src/data/scrypto/custom_validation.rs
iamyulong: Let's stick with what's implemented in this PR now. Don't think splitting the validation gives us much benefit now.

dhedey: Ahh sorry I didn't see this comment in time, and it was easy to do so I've separated it back out, PR ready for re-review.

iamyulong: No problem

iamyulong (approved):

PR #994, opened 2023-05-04
Architecture: Add Index to Blueprint Schema

talekhinezh:
## Summary
Add Index and SortedIndex into blueprint schema. This is to allow blueprint schema to define these datastructures statically rather than rely on dynamically created structures. 

## Details
* Added Index and SortedIndex into blueprint schema
* Refactored `EpochManager` and `NonFungibleVault` to use the static schema
* Refactored `IndexedBlueprintSchema`
* Rename `ModuleNumber` to `PartitionNumber`
* Add notion of `ObjectHandle` and `CollectionIndex` to API

### For Internal Integrators
* `EpochManager`, `NonFungibleVault` schemas have been updated to use statically defined indices
* Updated `IndexedBlueprintSchema` part of `PackageInfoSubstate`


radix-engine-interface/src/api/object_api.rs
```diff
@@ -35,12 +35,12 @@ pub enum ObjectModuleId {
...
+    pub fn base_partition_num(&self) -> PartitionNumber {
         match self {
-            ObjectModuleId::Metadata => METADATA_BASE_MODULE,
-            ObjectModuleId::Royalty => ROYALTY_BASE_MODULE,
-            ObjectModuleId::AccessRules => ACCESS_RULES_BASE_MODULE,
-            ObjectModuleId::SELF => OBJECT_BASE_MODULE,
+            ObjectModuleId::Metadata => METADATA_BASE_PARTITION,
+            ObjectModuleId::Royalty => ROYALTY_BASE_PARTITION,
+            ObjectModuleId::AccessRules => ACCESS_RULES_BASE_PARTITION,
+            ObjectModuleId::SELF => OBJECT_BASE_PARTITION,
```
dhedey, line 43: To align with REP-60, should we change this to `ObjectModuleId::Main`?

radix-engine-interface/src/blueprints/resource/resource.rs
```diff
@@ -61,7 +61,7 @@ impl LiquidFungibleResource {
 #[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
 pub struct LiquidNonFungibleVault {
     pub amount: Decimal,
-    pub ids: Own,
+    //pub ids: Own,
```
dhedey, line 64: I assume this should be removed? :)

radix-engine-interface/src/types/node_layout.rs
```diff
@@ -6,40 +6,40 @@ use sbor::rust::prelude::*;
 // Please update REP-60 after updating types/configs defined in this file!
 //=========================================================================
 
-pub const TYPE_INFO_BASE_MODULE: ModuleNumber = ModuleNumber(0u8);
-pub const METADATA_BASE_MODULE: ModuleNumber = ModuleNumber(1u8);
-pub const ROYALTY_BASE_MODULE: ModuleNumber = ModuleNumber(2u8);
-pub const ACCESS_RULES_BASE_MODULE: ModuleNumber = ModuleNumber(3u8);
-pub const OBJECT_BASE_MODULE: ModuleNumber = ModuleNumber(32u8);
+pub const TYPE_INFO_BASE_PARTITION: PartitionNumber = PartitionNumber(0u8);
```
dhedey, line 9: What does `BASE` mean here? The start of the partition counts? But I thought as per REP-60 these partitions were fixed by the system?

Perhaps the following might be more accurate?
* `TYPE_INFO_FIELD_PARTITION`
* `METADATA_KVSTORE_PARTITION`
* `ROYALTY_FIELD_PARTITION`
* `ACCESS_RULES_FIELD_PARTITION`
* `MAIN_FIELD_PARTITION`
* `MAIN_ADDITIONAL_PARTITIONS_START_FROM`

I see in this PR you've basically used "FIELD" instead of "TUPLE" to describe these partitions - which I much prefer (to save confusion with the SBOR tuple value kind) - would you be up for changing REP-60 to use the name "FIELD" instead of "TUPLE" for this partition type?

radix-engine-queries/src/typed_substate_layout.rs
```diff
@@ -167,46 +167,58 @@ pub fn to_typed_object_module_substate_key(
...
+                NonFungibleResourceManagerPartitionOffset::ResourceManager => {
                     TypedObjectModuleSubstateKey::NonFungibleResource(
-                        NonFungibleResourceManagerOffset::try_from(substate_key)?,
+                        NonFungibleResourceManagerField::try_from(substate_key)?,
                     )
                 }
-                NonFungibleResourceManagerModuleOffset::NonFungibleData => {
+                NonFungibleResourceManagerPartitionOffset::NonFungibleData => {
                     let key = substate_key.for_map().ok_or(())?;
                     TypedObjectModuleSubstateKey::GenericKeyValueStore(key.clone())
```
dhedey, line 196: For Core API purposes, it would be great if we could instead map this into the well known substate type `TypedObjectModuleSubstateKey::NonFungibleResourceData(x)` and possibly rename the other `TypedObjectModuleSubstateKey::NonFungibleResource` to `NonFungibleResourceField` etc.

radix-engine-queries/src/typed_substate_layout.rs
```diff
@@ -167,46 +167,58 @@ pub fn to_typed_object_module_substate_key(
...
+            let partition_offset = EpochManagerPartitionOffset::try_from(partition_offset)?;
+            match partition_offset {
+                EpochManagerPartitionOffset::EpochManager => {
+                    TypedObjectModuleSubstateKey::EpochManager(EpochManagerField::try_from(
+                        substate_key,
+                    )?)
+                }
+                EpochManagerPartitionOffset::SecondaryIndex => {
+                    let key = substate_key.for_sorted().ok_or(())?;
+                    TypedObjectModuleSubstateKey::GenericSortedU16Index(key.clone())
```
dhedey, line 210: Similarly here for the Epoch Manager 👍

radix-engine-queries/src/typed_substate_layout.rs
```diff
@@ -218,12 +230,24 @@ fn to_typed_object_substate_key_internal(
...
+
+            match partition_offset {
+                NonFungibleVaultPartitionOffset::Balance => {
+                    TypedObjectModuleSubstateKey::NonFungibleVault(NonFungibleVaultField::try_from(
+                        substate_key,
+                    )?)
+                }
+                NonFungibleVaultPartitionOffset::NonFungibles => {
+                    let key = substate_key.for_map().ok_or(())?;
+                    TypedObjectModuleSubstateKey::GenericIndex(key.clone())
```
dhedey, line 247: And here for the NF Vault :)

radix-engine/src/blueprints/resource/auth_zone.rs
```diff
@@ -31,8 +31,11 @@ impl AuthZoneBlueprint {
             RuntimeError::SystemUpstreamError(SystemUpstreamError::InputDecodeError(e))
         })?;
 
-        let auth_zone_handle =
-            api.actor_lock_field(AuthZoneOffset::AuthZone.into(), LockFlags::MUTABLE)?;
+        let auth_zone_handle = api.actor_lock_field(
+            OBJECT_HANDLE_SELF,
```
dhedey, line 35: I wonder if we should add `actor_lock_self_field` etc to the API to make it slightly nicer to use, so we don't have to add `OBJECT_HANDLE_SELF` everywhere :)

radix-engine/src/blueprints/resource/non_fungible/non_fungible_resource_manager.rs
```diff
@@ -35,6 +35,8 @@ pub struct NonFungibleResourceManagerMutableFieldsSubstate {
 
 pub type NonFungibleResourceManagerTotalSupplySubstate = Decimal;
 
+pub const NON_FUNGIBLE_RESOURCE_MANAGER_DATA_STORE: CollectionIndex = 0u8;
```
dhedey, line 38: What's a collection index? Is this effectively the partition offset?

Or the partition offset - 1? (to save room for the tuple partition?)

radix-engine/src/blueprints/resource/non_fungible/non_fungible_vault.rs
```diff
@@ -256,15 +260,12 @@ impl NonFungibleVault {
     where
         Y: ClientApi<RuntimeError>,
     {
-        let handle = api.actor_lock_field(
-            NonFungibleVaultOffset::LiquidNonFungible.into(),
-            LockFlags::read_only(),
+        let items: Vec<NonFungibleLocalId> = api.actor_index_scan_typed(
```
dhedey, line 263: Just to check - am I right in thinking we're going to change this API at some point to only return the first N for some N?

radix-engine/src/system/payload_validation.rs
```diff
@@ -68,8 +68,6 @@ impl<'s, 'a, Y: KernelApi<SystemConfig<V>>, V: SystemCallbackObject> TypeInfoLoo
                 }
             }
             TypeInfoSubstate::KeyValueStore(_) => TypeInfoForValidation::KeyValueStore,
-            TypeInfoSubstate::Index => TypeInfoForValidation::Index,
```
dhedey, line 71: Can we now no longer have these as separate "Owned" things?

radix-engine/src/system/system.rs
```diff
@@ -49,6 +50,22 @@ pub struct SystemService<'a, Y: KernelApi<SystemConfig<V>>, V: SystemCallbackObj
...
+enum ActorObjectType {
+    SELF,
+    OuterObject,
+}
+
+impl TryFrom<ObjectHandle> for ActorObjectType {
+    type Error = RuntimeError;
+    fn try_from(value: ObjectHandle) -> Result<Self, Self::Error> {
+        match value {
+            OBJECT_HANDLE_SELF => Ok(ActorObjectType::SELF),
```
dhedey, line 62: I'm happy to keep using self here - this makes sense :)

radix-engine/src/system/system.rs
```diff
@@ -49,6 +50,22 @@ pub struct SystemService<'a, Y: KernelApi<SystemConfig<V>>, V: SystemCallbackObj
     pub phantom: PhantomData<V>,
 }
 
+enum ActorObjectType {
+    SELF,
```
dhedey, line 54: Out of interest, why is `SELF` capitalised? Can we just do `Self`?

radix-engine/src/transaction/state_update_summary.rs
```diff
@@ -233,7 +231,7 @@ impl<'a, S: SubstateDatabase> BalanceAccounter<'a, S> {
                 }
             } else {
                 // Scan loaded substates to find children
-                for tracked_module in tracked_node.tracked_modules.values() {
+                for tracked_module in tracked_node.tracked_parititions.values() {
```
dhedey, line 234: `tracked_partitions`*

scrypto-schema/src/lib.rs
```diff
@@ -136,28 +152,44 @@ pub struct IndexedBlueprintSchema {
...
+
+        let mut fields = None;
+        if !schema.fields.is_empty() {
+            fields = Some((PartitionOffset(partition_offset), schema.fields));
+            partition_offset += 1;
         };
 
-        let mut key_value_stores = Vec::new();
+        let mut collections = Vec::new();
+        for kv_schema in schema.kv_stores {
```
dhedey, line 164: I've got some pretty grave concerns about this strategy for auto-allocation of partition numbers when it comes to upgradability.

Adding an additional KVStore would change the partition numbers of the indices / sorted indices.

I think instead we probably need to explicitly claim indices for the relevant partitions in the schema itself (and validate that they make sense and won't overflow).

What do you think?

scrypto-schema/src/lib.rs
```diff
@@ -184,24 +216,64 @@ impl From<PackageSchema> for IndexedPackageSchema {
...
+    pub fn key_value_store_partition(
+        mut self,
+        collection_index: u8,
+    ) -> Option<(PartitionOffset, ScryptoSchema, BlueprintKeyValueStoreSchema)> {
+        let index = collection_index as usize;
+        if index >= self.collections.len() {
+            return None;
+        }
+
+        match self.collections.swap_remove(index) {
```
dhedey, line 247: I'm very surprised by `swap_remove`? Doesn't this change the collection indices, and would break on subsequent calls to `key_value_store_partition`?

dhedey (approved):
Really nice - I like how this is turning out.

A few thoughts, and some minor concerns around how we map collection index <=> module offsets.

radix-engine-interface/src/api/actor_index_api.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// Removes an entry from an index
+    fn actor_index_remove_typed<V: ScryptoDecode>(
+        &mut self,
+        object_handle: ObjectHandle,
+        collection_index: CollectionIndex,
+        key: Vec<u8>,
+    ) -> Result<Option<V>, E> {
+        let rtn = self
+            .actor_index_remove(object_handle, collection_index, key)?
+            .map(|e| scrypto_decode(&e).unwrap());
```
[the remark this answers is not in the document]
talekhinezh: Yup but also if it was exposed to scrypto this particular code would live in WASM and would panic in WASM if decode failed.

radix-engine-interface/src/blueprints/resource/resource.rs
talekhinezh: yup!

radix-engine/src/system/payload_validation.rs
talekhinezh: Yup exactly. If needed in the future, could always add it.

radix-engine-queries/src/typed_substate_layout.rs
talekhinezh: Yup makes sense!

radix-engine-queries/src/typed_substate_layout.rs
talekhinezh: Yup makes sense!

radix-engine/src/system/system.rs
talekhinezh: `Self` is reserved by rust

scrypto-schema/src/lib.rs
talekhinezh: Adding an additional KVStore wouldn't use the same interface and the idea would be for the backend to manage the mapping as new collections are added (the reason why offset is included in the collection vector).

I much prefer the user layer not needing to manage indices on it's own. All the user layer sees is a vector of collections (similar to fields).

PR #995, opened 2023-05-04
Extending the `EpochManager.next_round()` API to track leader reliability

jakrawcz-rdx:
This is for _future_ emissions penalty computation purposes - part of https://radixdlt.atlassian.net/browse/SCRY-435.

This PR only modifies the API. (I want to get it in first, in order to simultaneously work on the integration in Node)
The impl of the actual emissions will come in another PR.

- The leader reliability information "schema" - as discussed with @dhedey and @LukasGasior1.
- The channel (i.e. via `EpochManager.next_round()` input) - as discussed with @talekhinezh.

Wow ☝️ I don't even feel like an author of this PR anymore...


radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
```diff
@@ -50,9 +51,62 @@ pub const EPOCH_MANAGER_NEXT_ROUND_IDENT: &str = "next_round";
...
+        }
+    }
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, Sbor)]
+pub struct LeaderProposalMissHistory {
+    /// The validators which were leaders of the "gap" rounds (i.e. those that were not reported to
+    /// the epoch manager since the previous call; see `EpochManagerNextRoundInput::round`).
+    /// This list will contain exactly `current_call.round - previous_call.round - 1` elements; in
+    /// theory, this makes `EpochManagerNextRoundInput::round` field redundant (i.e. computable),
```
jakrawcz-rdx, line 85: _(note from author)_

I strongly consider removing that `round` field.
Look how it would simplify the `::successful()` constructor...
Wdyt, reviewers?

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
```diff
@@ -50,9 +51,62 @@ pub const EPOCH_MANAGER_NEXT_ROUND_IDENT: &str = "next_round";
...
 #[derive(Debug, Clone, Eq, PartialEq, Sbor)]
 pub struct EpochManagerNextRoundInput {
+    /// Current round number.
+    /// Please note that in case of liveness breaks, this number may be different than previous
+    /// reported `round + 1`. Such gaps are considered "round leader's fault" and are penalized
+    /// on emission, according to leader reliability statistics (see `LeaderProposalMissHistory`).
     pub round: u64,
+
+    /// A captured history of leader proposal misses since the previous reported round.
+    pub leader_proposal_misses: LeaderProposalMissHistory,
```
talekhinezh, line 61: Can we add a comment that we should change this at some point so that the engine itself knows the leaders who missed the proposals. (ie we shouldn't be relying on the consensus layer to tell us this otherwise a quorum of validators may potentially gang up on other validators).

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
talekhinezh: See my comment above regarding this :)

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
```diff
@@ -50,9 +51,62 @@ pub const EPOCH_MANAGER_NEXT_ROUND_IDENT: &str = "next_round";
...
+
+    /// An indication of the current round operating in a "fallback" mode (i.e. indicating a fault
+    /// of the current leader).
+    /// A `None` denotes a regular round (not a fallback).
+    /// A `Some(current_leader)` specifies the current round's leader and indicates that current
+    /// round is a fallback one (and that validator should be penalized during emissions, same as
+    /// `gap_round_leaders`).
+    /// Please note that this effect is independent from the gap rounds (i.e. the value of this
+    /// field does not affect the invariants of `gap_round_leaders` or the round numbers).
+    pub current_round_fallback_leader: Option<ValidatorIndex>,
```
talekhinezh, line 99: Again, we should add a comment that in the long term we should make it a goal that the engine does not rely on consensus to give the right leader for the same reasons above.

talekhinezh (approved):

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
dhedey: From discussing this a few days ago, even in that set-up, a byzantine quorum could just pretend that any validators they don't like missed their proposals anyway. So the attack surface differential in these two scenarios is rather small.

I agree it would be nice-to-have, but I'm not sure how important it is.

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
```diff
@@ -50,9 +51,62 @@ pub const EPOCH_MANAGER_NEXT_ROUND_IDENT: &str = "next_round";
...
+    /// This list will contain exactly `current_call.round - previous_call.round - 1` elements; in
+    /// theory, this makes `EpochManagerNextRoundInput::round` field redundant (i.e. computable),
+    /// but this relation can be used for an extra consistency check.
+    /// The validators on this list should be penalized during emissions at the end of the current
+    /// epoch.
+    pub gap_round_leaders: Vec<ValidatorIndex>,
+
+    /// An indication of the current round operating in a "fallback" mode (i.e. indicating a fault
+    /// of the current leader).
+    /// A `None` denotes a regular round (not a fallback).
```
dhedey, line 93: This doesn't look quite right. If Java is in charge of the rotation, then Rust needs to know who the leader is even in a non-fallback scenario - because rust needs to be incrementing the proposals made counter.

dhedey: And so maybe this should be `LeaderProposalHistory` and we separate `current_round_leader` from `is_fallback`

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
```diff
@@ -50,9 +51,62 @@ pub const EPOCH_MANAGER_NEXT_ROUND_IDENT: &str = "next_round";
...
+    pub current_round_fallback_leader: Option<ValidatorIndex>,
+}
+
+/// An index of a specific validator within the current validator set.
+/// To be exact: a `ValidatorIndex` equal to `k` references the `k-th` element returned by the
+/// iterator of `BTreeMap<ComponentAddress, Validator>`.
+/// This uniquely identifies the validator, while being shorter than `ComponentAddress` (we do care
+/// about the constant factor of the space taken by `LeaderProposalMissHistory` under prolonged
+/// liveness break scenarios).
+pub type ValidatorIndex = usize;
```
dhedey, line 108: We may even wish to consider using a u8 (8 times smaller than a usize = u64), but there are maybe bigger fish to fry (like the fact validators are storing every round's proof at the moment).

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
jakrawcz-rdx: I can add the comment, but more importantly I have this question:

Do we want [hard] to move the leader rotation management to the Engine? 

(I mean: are there maybe more reasons why it makes sense? Is this conceptually very right thing?)

If the answer is "yes", then I suggest we design it and impl it right now, in a separate refactor (before continuing the "Emissions" work).

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
jakrawcz-rdx: if we go this way (i.e. shave any bytes off the very long validator ID) I agree we should go to the limit.
(I assume there is some business rule that ensures validatorSet.len < 256 and that we plan to hold it for a long time)

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
jakrawcz-rdx: ahhhh totally!! I forgot that success counts are also needed for computation... fixed now.

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
jakrawcz-rdx: I hope that my general comment on the higher-level (i.e. entire `leader_proposal_history` section) covers that now.
(and no point repeating this here)

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
jakrawcz-rdx: I got the answer:
image

with this assumption, I guess I can add the requested TODO here and merge :)

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
```diff
@@ -50,9 +51,68 @@ pub const EPOCH_MANAGER_NEXT_ROUND_IDENT: &str = "next_round";
 
 #[derive(Debug, Clone, Eq, PartialEq, Sbor)]
 pub struct EpochManagerNextRoundInput {
+    /// Current round number.
+    /// Please note that in case of liveness breaks, this number may be different than previous
+    /// reported `round + 1`. Such gaps are considered "round leader's fault" and are penalized
+    /// on emission, according to leader reliability statistics (see `LeaderProposalMissHistory`).
```
dhedey, line 57: `LeaderProposalHistory`*

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
```diff
@@ -50,9 +51,68 @@ pub const EPOCH_MANAGER_NEXT_ROUND_IDENT: &str = "next_round";
...
     pub round: u64,
```
dhedey, line 58: To ensure uniqueness of round transactions (which we probably don't actually need any more, given we've changed how the DB is indexed), we include an epoch in their payload on the node. I wonder if we should include an epoch here for sanity checking? Although I guess the fact that the `LeaderProposalHistory` length has to align with the difference in `round` is quite a good sanity check anyway.

dhedey: Separately, we want to move the clock to the EpochManager - so it might be worth sticking the `proposed_timestamp` on here (possibly none on fallback vertices) - @talekhinezh - do you think now's a good time to do that, given Jakub is touching round changes?

dhedey (approved):

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
jakrawcz-rdx: > I wonder if we should include an epoch here for sanity checking?

I am closer to removing the round number :p

> Separately, we want to move the clock to the EpochManager

I am not sure what is the motivation for this, but even if very valid, then definitely a separate refactor 👍

PR #996, opened 2023-05-04
Feature/pre allocated addresses in scrypto

dhedey:
## Summary

Adds support for pre-allocated addresses in scrypto - for (eg) use with the GlobalCaller badges.

Also:
* Fixed an issue I introduced when I allowed creating account at address in scrypto, where the entity type wasn't verified against the blueprint.
* Commonized some of the globalize logic better

## Testing

Added some tests in `address.ts`.

### For dApp developers

You can now use pre-allocation and GlobalCaller badges - but wait for the next PR for nicer syntax and examples (see #997)


radix-engine/src/system/system.rs
```diff
@@ -524,170 +650,64 @@ where
         self.new_object_internal(&blueprint, instance_context, schema, fields, kv_entries)
     }
 
+    #[trace_resources]
+    fn preallocate_global_address(
+        &mut self,
+        entity_type: EntityType,
+    ) -> Result<GlobalAddress, RuntimeError> {
+        if !entity_type.is_global() {
```
iamyulong, line 658: Feel like we should limit it to non-virtualised-global address?

radix-engine/src/system/system.rs
```diff
@@ -524,170 +650,64 @@ where
...
+            return Err(RuntimeError::SystemError(
+                SystemError::InvalidGlobalEntityType,
+            ));
+        }
+        let allocated_node_id = self.api.kernel_allocate_node_id(entity_type)?;
```
iamyulong, line 663: As briefly mentioned before, I think we need to change `preallocate_global_address` interface to accept a blueprint parameter, rather than entity type, and create some placeholder in track to store its state. For instance, we can store
```
<preallocated_address>::<type_info> => TypeInfo::Preallocated(Blueprint)
```

Because now we have a few issues:
- It's an invalid global references, before consumed, which breaks system invariants;
- There is no mechanism to prevent a preallocated address from being used twice (may cause orphaned nodes);
- There is no mechanism to check if a preallocated address has been consumed (if we track its lifecycle, we may enable cross-transaction preallocation. Fun!)

It doesn't have to be done fully this PR, since it concerns mostly the interface.

dhedey: Most of the issues you cover aren't currently possible due to the design of the preallocator - but will add tests for them.

As for the TypeInfo::Preallocated change - I agree this is nice, and should be done alongside the creation of references - and the globally allocated addresses should effectively be equivalent to transaction processor preallocations. I don't think `TypeInfo::Preallocated` necessarily needs tying to a blueprint though? I suggest we leave the above to a separate PR - and maybe it makes sense for you or Josh to pick that up, rather than me ramping up on all the areas where typeinfo is read? :). Let me know your thoughts.

iamyulong: You're right. The call frame based model will prevent most if not all bad behaviours.

Let's just add some tests to verify. 

I will discuss with Josh and see if he likes the idea. Then, either Josh or I can implement it.

talekhinezh: Definitely prefer not exposing entity type to the scrypto layer as it's something scrypto itself shouldn't be concerned about. Will update this in the next PR.

talekhinezh (approved):

PR #999, opened 2023-05-05
Refined Reference Model

iamyulong:
## Summary

This PR introduces a few refactors and fixes to the reference model employed by Radix Engine.

For more technical details, please see REP-79.

## Details

- Disallowed non-stable reference to be passed between frames
    * Stable references are `Global` and `DirectAccess`
- Divided node visibility based on its origin
- Added `call_direct_access_method` Scrypto API
    * You can now recall resources in Scrypto 
- Added node borrow count for `Borrowed` references
   * Borrowed nodes can’t be dropped! 
- Added `actor_node_id` API  to make a "Mini Resource System in Scrypto" possible
- Added various refactors, notably to the node globalisation process

## Testing

New tests has been added for various reference constraints.


radix-engine-tests/tests/min_resource_system.rs
```diff
@@ -0,0 +1,103 @@
...
+                .expect_failure()
+                .to_string();
+            if !error_message.contains(msg) {
+                panic!("Expected {}, but was {}", msg, error_message);
+            }
+        }
+    };
+}
+
+// Note that we may decide to disallow borrowed reference in Scrypto, then
```
talekhinezh, line 46: Can we have these tests which disallow borrowed references in Scrypto so that we can verify they'll fail?

radix-engine-tests/tests/recallable.rs
```diff
@@ -114,3 +115,79 @@ fn can_take_on_recallable_vault() {
         .unwrap();
     assert_eq!(other_amount, Decimal::one());
 }
+
+#[test]
+fn test_recall_on_internal_vault() {
```
talekhinezh, line 120: Can we update name to `owned_internal_vault`

radix-engine-tests/tests/recallable.rs
```diff
@@ -114,3 +115,79 @@ fn can_take_on_recallable_vault() {
...
+    receipt.expect_specific_failure(|e| {
+        matches!(
+            e,
+            RuntimeError::KernelError(KernelError::InvalidInvokeAccess)
+        )
+    });
+}
+
+#[test]
+fn test_recall_on_received_direct_access_reference() {
```
talekhinezh, line 164: Nice!

radix-engine/src/blueprints/resource/auth_zone.rs
```diff
@@ -291,4 +292,43 @@ impl AuthZoneBlueprint {
...
     }
+
+    pub(crate) fn drop<Y>(
+        input: &IndexedScryptoValue,
+        api: &mut Y,
+    ) -> Result<IndexedScryptoValue, RuntimeError>
+    where
+        Y: KernelSubstateApi<SystemLockData> + ClientApi<RuntimeError>,
+    {
+        // TODO: add `drop` callback for drop atomicity, which will remove the necessity of kernel api.
```
talekhinezh, line 303: I think we used to do this but there were some reasons I moved away from this. I think it was to be able to capture the owned nodes in the dropped object, which a Rust style like `Drop` interface wouldn't really allow.

radix-engine/src/blueprints/resource/auth_zone.rs
```diff
@@ -291,4 +292,43 @@ impl AuthZoneBlueprint {
...
+
+        let input: AuthZoneDropInput = input.as_typed().map_err(|e| {
+            RuntimeError::SystemUpstreamError(SystemUpstreamError::InputDecodeError(e))
+        })?;
+
+        // Detach proofs from the auth zone
+        let handle = api.kernel_lock_substate(
```
talekhinezh, line 310: Instead of locking, you could just use the substates returned from `drop_object` no?

radix-engine/src/blueprints/resource/fungible/fungible_proof.rs
```diff
@@ -135,18 +137,24 @@ impl FungibleProofBlueprint {
...
-            .unwrap();
-
-        api.call_method(
-            parent.as_node_id(),
-            RESOURCE_MANAGER_DROP_PROOF_IDENT,
-            scrypto_encode(&ResourceManagerDropProofInput { proof }).unwrap(),
+        // TODO: add `drop` callback for drop atomicity, which will remove the necessity of kernel api.
+
+        // Notify underlying buckets/vaults
+        let handle = api.kernel_lock_substate(
```
talekhinezh, line 145: Similarly here you could just use the substates returned from `drop_object`

radix-engine/src/blueprints/resource/non_fungible/non_fungible_proof.rs
```diff
@@ -156,18 +158,25 @@ impl NonFungibleProofBlueprint {
...
-            .outer_object
-            .unwrap();
-
-        api.call_method(
-            parent.as_node_id(),
-            RESOURCE_MANAGER_DROP_PROOF_IDENT,
-            scrypto_encode(&ResourceManagerDropProofInput { proof }).unwrap(),
+        // TODO: add `drop` callback for drop atomicity, which will remove the necessity of kernel api.
+
+        // Notify underlying buckets/vaults
```
talekhinezh, line 165: Also here

radix-engine/src/blueprints/resource/worktop.rs
```diff
@@ -38,24 +39,37 @@ impl WorktopBlueprint {
...
-        let mut node_substates = api.kernel_drop_node(input.worktop.0.as_node_id())?;
-        let substate = node_substates
-            .remove(&OBJECT_BASE_PARTITION)
-            .unwrap()
-            .remove(&WorktopField::Worktop.into())
-            .unwrap();
-        let worktop: WorktopSubstate = substate.as_typed().unwrap();
-        for (_, bucket) in worktop.resources {
+        // Detach buckets from worktop
+        let handle = api.kernel_lock_substate(
```
talekhinezh, line 51: Why not the old style here where we use the substates returned from the dropped node?

radix-engine/src/kernel/call_frame.rs
```diff
@@ -10,139 +10,221 @@ use radix_engine_interface::blueprints::resource::{
...
 
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct CallFrameUpdate {
-    pub nodes_to_move: Vec<NodeId>,
-    pub node_refs_to_copy: IndexSet<NodeId>,
+/// A message used for communication between call frames.
+///
+/// Note that it's just an intent, not checked/allowed by kernel yet.
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+pub struct Message {
```
talekhinezh, line 21: Can we change name, `Message` definitely does not give impression that this includes state movement?

radix-engine/src/kernel/kernel.rs
```diff
@@ -190,53 +220,51 @@ where
...
+            M::on_execution_finish(&mut message, self)?;
+
+            (output, message)
         };
 
-        // Pop call frame
+        // Move
         {
-            let mut parent = self.prev_frame_stack.pop().unwrap();
+            let parent = self.prev_frame_stack.last_mut().unwrap();
```
talekhinezh, line 239: Is there a reason you didn't want to pop the parent yet?

talekhinezh (approved):

radix-engine-interface/src/types/indexed_value.rs
```diff
@@ -12,7 +12,7 @@ use utils::ContextualDisplay;
 #[derive(Clone, PartialEq, Eq)]
 pub struct IndexedScryptoValue {
     bytes: Vec<u8>,
-    references: IndexSet<NodeId>,
+    references: Vec<NodeId>,
```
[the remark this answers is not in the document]
iamyulong: Yes, when handling references in `CallFrame`, we apply deduplication.

radix-engine-tests/tests/recallable.rs
iamyulong: No problem.

radix-engine-tests/tests/min_resource_system.rs
iamyulong: When disallowing borrowed reference in `create_object/unlock_substate`, we will have dedicated tests to verify failure and disable these tests.

radix-engine/src/blueprints/resource/auth_zone.rs
iamyulong: A general direction I'd like to push for is to move drop customisation to clients - they may decide to detach owned objects and let system attempt to auto drop; or they may decide to clean up everything without system assumption.

Even in `rust`, [the `Drop` interface](https://doc.rust-lang.org/std/ops/trait.Drop.html) exposes a `&mut self`, so we can detach owned objects too.

radix-engine/src/blueprints/resource/non_fungible/non_fungible_proof.rs
iamyulong: Not here. If we drop, we lose visibility to the buckets/vaults.

radix-engine/src/blueprints/resource/worktop.rs
iamyulong: Ditto, primarily to avoid visiblity loss.

radix-engine/src/kernel/kernel.rs
iamyulong: I think we shouldn't change the frame stack structure at the time of processing returns.

As you can tell, we have a dedicate stage for "pop call frame".

radix-engine/src/kernel/call_frame.rs
iamyulong: The reason I renamed `CallFrameUpdate` is to make it clear that it's an itent, not a result.

But, renaming suggestion welcome.

radix-engine/src/blueprints/resource/fungible/fungible_proof.rs
iamyulong: In the long run, I'd like us to consider removing return of `drop_object`, since it has no semantic meaning (but pure bytes).

Like in Rust, if a struct is destructed, we don't have access to its internal.

radix-engine/src/blueprints/resource/auth_zone.rs
iamyulong: Commented in another thread.

native-sdk/src/resource/proof.rs
```diff
@@ -110,7 +110,7 @@ impl SysProof for Proof {
 
     fn sys_drop<Y, E: Debug + ScryptoCategorize + ScryptoDecode>(self, api: &mut Y) -> Result<(), E>
     where
-        Y: ClientApi<E>,
+        Y: ClientObjectApi<E> + ClientBlueprintApi<E>,
```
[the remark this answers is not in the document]
iamyulong: I believe it's to make it usable in system modules where not the full `ClientApi` is allowed.

PR #1001, opened 2023-05-05
Feature/SCRY-428 substate db costing part 2

mstrug-rdx:
## Summary

Added mechanism for recognition of the first database access for substate scan and take.

## Details

If particular pair of Node Id and Module number is not tracked (in `Track` module) and database interaction is done using Partition key based on that pair, it means that it is the first time database is accessed for that pair. Such logic has been implemented for `Track` module functions: `scan_substates()`, `take_substates()`, `scan_sorted_substates()`.

Also updated `take_substate()` function regarding first database access basing on return value of `get_tracked_substate_virtualize()` (implemented in PR #974).

Added Kernel callbacks to pass information about first time database accesses to the Costing module: `on_take_substates()` and `on_scan_substates()`.


radix-engine/src/track/track.rs
```diff
@@ -655,9 +655,11 @@ impl<'s, S: SubstateDatabase, M: DatabaseKeyMapper> SubstateStore for Track<'s,
 
         // Optimization, no need to go into database if the node is just created
         if is_new {
-            return items;
+            return (items, false);
```
iamyulong, line 658: You might've realised it - a single `boolean` isn't sufficient anyway; during a scan, some of the entries will be from the `Track` and additional entries may/may not be loaded from database.

In the long run, we'll need to change the boolean to a richer data structure to record additional info, e.g. how much is read from track, how much is read from database, any database delete we will have to rebase, and any database write that we will have to charge more.

mstrug-rdx: So far I added new type instead of `boolean`, still I need to look into described cases.

mstrug-rdx: I extended the data type.

radix-engine/src/track/interface.rs
```diff
@@ -23,6 +23,29 @@ pub enum TakeSubstateError {
...
+        first_time_read_records_count: usize,
+        read_from_track_count: usize,
+        update_count: usize,
+    },
+    ScanSorted {
+        first_time_read_records_count: usize,
+        read_from_track_count: usize,
+    },
+    #[default]
+    NoAccess,
```
iamyulong, line 46: We might want to tie DB access to primitive operations, rather than action.
```rust
pub struct StoreAccessInfo(Vec<StoreAccessInfo>);

// Per substate
pub enum StoreAccess {
    // When store invokes `SubstateDatabase::get_substate()`.
    // size might be zero when the entry does not exist.
    ReadFromDb(usize),
    // When store reads from the `tracked_nodes`.
    ReadFromTrack(usize),
    // When store writes into the `tracked_nodes`.
    // This will eventually be flushed into database.
    Write(usize),
    // When store updates an entry for the second time
    // Need to report both the previous size and new size as partial refund may be required
    // (Reason: only one write will be applied when the transaction finishes, despite substate being updated twice)
    Rewrite(usize, usize),

    // In future, we may want to introduce something like `Clear` to rebate behaviours that 
    // help reduce the state size.
}
```

In case you missed it, the stack looks like:
```
===========
Engine
===========
Track (Store)
===========
Database
===========
````

File: radix-engine/src/system/system_modules/costing/fee_table.rs [hunk 3]
```diff
@@ -41,6 +41,8 @@ pub enum CostingEntry<'a> {
     WriteSubstate {
         size: u32,
     },
+    ScanSubstateFirstTime,
+    TakeSubstatesFirstTime,
```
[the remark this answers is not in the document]
mstrug-rdx: Don't we need to know if particular substate was loaded for the first time also? As I remember for first time read we need to load database which should incur additional cost.

mstrug-rdx:
WIP

mstrug-rdx:
I've reimplemented `Track` store access using proposed `StoreAccessInfo` struct. Now all methods from `SubstateStore` interface trait returns `StoreAccessInfo` which is filled with items depending on utilised database or internal `tracked_nodes`. Then this struct is passed through `CallFrame` to the `Kernel` which passes it to the `KernelCallbackObject` and then to modules, especially `CostingModule` which applies costing basing on vector items.

I've also updated `FeeTable` module `kernel_api_cost_storage_usage()` function which handles all new costing entries related to `SubstateStore`. I've left `ReadSubstate`, `WriteSubstate` and `Invoke` costing entries which uses ordinary argument size of passed value which we can use to additionally apply cost 'per byte' of argument size. I added similar scaling mechanism as in CPU usage section.

Adding new costing entries required updating tests, which I tweaked to just pass for a now, but I think we need to adjust fee table together with other values from `radix-engine-constants`.

radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -299,30 +314,49 @@ impl FeeTable {
...
+    fn kernel_api_cost_storage_usage(&self, entry: &CostingEntry) -> u32 {
+        ((match entry {
             CostingEntry::Invoke {
                 input_size,
                 actor: _,
             } => 10 * input_size,
-            CostingEntry::ReadSubstate { size } => 10 * size,
+            CostingEntry::ReadSubstate { size } => 100 * size,
             CostingEntry::WriteSubstate { size } => 1000 * size,
+            CostingEntry::SubstateReadFromDb { size } => 100 * size, // todo: determine correct value
```
iamyulong, line 346: We should move `ReadSubstate` and `WriteSubstate` to computation cost, as the compuation cost is "linear" to size because of reading/copying/transforming memory chunks?

radix-engine/src/track/interface.rs
```diff
@@ -102,24 +107,98 @@ pub trait SubstateStore {
...
+
+    pub fn with_vector(vector: Vec<StoreAccess>) -> Self {
+        Self(vector)
+    }
+
+    pub fn data(&self) -> &Vec<StoreAccess> {
+        &self.0
+    }
+
+    pub fn get_whole_size(&self) -> usize {
```
iamyulong, line 152: Can we split into `total_read_size` and `total_write_size`? In the future, we may also need to relate costing to `number_of_keys` depending on our database characteristics.

iamyulong (approved):
It looks really good. Thanks for working with my comments.

PR #1002, opened 2023-05-06
Feature/metadata additions

dhedey:
## Summary
* Adds `Origin` and `PublicKeyHash` and metadata values, in time for the ash internal release.
* Also slightly adjusted `Instant` and the virtualization process uses `PublicKeyHash` for slight better alignment / semantics.
* Added `owner_keys` metadata on virtual account/identity instantiation, in preparation for ROLA

## Details

`Origin` is supposed to represent a website origin (ie scheme, host + port) in the web-browser sense, and will be used for ROLA. Validation has not (yet) been added to Origin. This will follow in a later PR when I also fix up other issues with the metadata behaviour.

The `PublicKeyHash` is the final 29 bytes of the Blake2b-256 hash of the public key bytes (in the canonical Radix encoding).

### BREAKING CHANGE

Instant is now serialized in the manifest as _just_ `Xi64` instead of a `Tuple(Xi64)` (for X = seconds since the unix epoch).

```
# Set Instant (or the value in seconds since unix epoch) Metadata on Resource
SET_METADATA
    Address("${resource_address}")
    "field_name"
    Enum(0u8, # Single item
        Enum(12u8, # Instant - ie Timestamp
            # Value in seconds since Unix Epoch
            10000i64
        )
    );
```

### New metadata value types - Origin and PublicKeyHash

The following commands can be used to set `Origin` and `PublicKeyHash` - both of which are being used for ROLA.

I've also updated `transaction/examples/metadata/metadata.rtm` with more comments to help understand how to set metadata.

```
# Set Origin Metadata on Resource
SET_METADATA
    Address("${resource_address}")
    "field_name"
    Enum(0u8, # Single item
        Enum(14u8, # Origin
            "https://radixdlt.com"
        )
    );

# Set PublicKeyHash Metadata on Resource
SET_METADATA
    Address("${resource_address}")
    "field_name"
    Enum(0u8, # Single item
        Enum(15u8, # PublicKeyHash
            Enum(0u8, # 0 = Secp256k1, 1 = Ed25519
                # The hex-encoded final 29 bytes of the Blake2b-256 hash of the public key bytes (in the canonical Radix encoding)
                Bytes("0000000000000000000000000000000000000000000000000000000000")
            )
        )
    );

```


radix-engine/src/blueprints/account/package.rs
```diff
@@ -245,11 +245,14 @@ impl AccountNativePackage {
                 let input: VirtualLazyLoadInput = input.as_typed().map_err(|e| {
                     RuntimeError::SystemUpstreamError(SystemUpstreamError::InputDecodeError(e))
                 })?;
-                let rtn = AccountBlueprint::create_virtual_ecdsa_256k1(input.id, api)?;
+
+                let public_key_hash =
```
talekhinezh, line 249: Minor: can we move the conversion from `Vec<u8>` to `PublicKeyHash` into the `create_virtual` function as the input for virtualization should be a generic byte array (only the function should know what it actually represents).

radix-engine/src/blueprints/identity/package.rs
```diff
@@ -168,7 +171,10 @@ impl IdentityNativePackage {
                     RuntimeError::SystemUpstreamError(SystemUpstreamError::InputDecodeError(e))
                 })?;
 
-                let rtn = IdentityBlueprint::create_eddsa_virtual(input.id, api)?;
+                let public_key_hash =
```
talekhinezh, line 174: ditto here

radix-engine/src/blueprints/identity/package.rs
```diff
@@ -221,34 +227,15 @@ impl IdentityBlueprint {
         Ok((address, bucket))
     }
 
-    pub fn create_ecdsa_virtual<Y>(
-        id: [u8; NodeId::UUID_LENGTH],
+    pub fn create_virtual<Y>(
```
talekhinezh, line 230: As mentioned can we revert this or to clean it up create a separate internal function which combines the two.

talekhinezh (approved):

radix-engine/src/blueprints/account/package.rs
dhedey: Can do. I viewed the AccountBlueprint as wanting to talk in terms of abstractions it understands (ie `PublicKeyHash`) and felt that handling NodeIds and how they translate to hashes was the remit of this virtualization layer? But happy to move it :)

PR #1003, opened 2023-05-07
Feature: Restricted Proof Movement

talekhinezh:
## Summary
* Allow restricted proofs to be passed between internal call frames.
* Implement Function calls as a barrier

## Details
* Allow restricted proofs to be passed between internal call frames.
* Implement Function calls as a barrier
* Due to the function calls being implemented as a barrier, tests needed to be updated to do things in the manifest rather than in a utility function
* Added `SET_GROUP_ACCESS_RULE` and `SET_GROUP_MUTABILITY` manifest instructions
* Update ResourceManager to use groups for most auth

### For Dapp Developers
* Manifest has been added with set group auth functions
* Restricted proofs may now be passed through internal call frames
* Functions now act as a barrier


radix-engine/src/system/system_modules/node_move/node_move_module.rs
```diff
@@ -99,39 +104,44 @@ impl NodeMoveModule {
...
-                    let type_info = TypeInfoBlueprint::get_type(&method.node_id, api)?;
-                    if let TypeInfoSubstate::Object(ObjectInfo { blueprint, .. }) = type_info {
-                        if blueprint.eq(&Blueprint::new(&RESOURCE_PACKAGE, AUTH_ZONE_BLUEPRINT)) {
-                            changed_to_restricted = false;
-                        }
+                if callee.is_barrier() {
+                    let handle = api.kernel_lock_substate(
+                        &node_id,
+                        OBJECT_BASE_PARTITION,
+                        &FungibleProofField::Moveable.into(),
```
dhedey, line 111: This should technically be `NonFungibleProofField::Moveable`, and same on line 131.

radix-engine/src/kernel/actor.rs
```diff
@@ -28,6 +29,27 @@ pub enum Actor {
...
+            }
+            Actor::Function { .. } => false,
+            Actor::VirtualLazyLoad { .. } => false,
+            Actor::Root { .. } => false,
+        }
+    }
+
+    pub fn is_barrier(&self) -> bool {
+        match self {
+            Actor::Method(MethodActor { object_info, .. }) => object_info.global,
```
dhedey, line 46: I assume this also includes calls to modules:

I might be tempted to make that more explicit, ie: `module_id != Main || object_info.global`

radix-engine-tests/tests/clock.rs
```diff
@@ -157,11 +156,10 @@ fn test_clock_comparison_methods_against_the_current_time() {
     // Act
     let manifest = ManifestBuilder::new()
         .lock_fee(test_runner.faucet_component(), 10.into())
-        .call_function(
-            package_address,
-            "ClockTest",
-            "set_current_time",
-            manifest_args!(CLOCK, 1669663688996 as i64),
+        .call_method(
+            CLOCK,
```
dhedey, line 160: I assume we should delete `ClockTest`'s `set_current_time` method?

dhedey (approved):
Nice! A few small thoughts :)

transaction/src/validation/transaction_validator.rs
```diff
@@ -292,6 +292,8 @@ impl NotarizedTransactionValidator {
                 | Instruction::ClaimPackageRoyalty { .. }
                 | Instruction::ClaimComponentRoyalty { .. }
                 | Instruction::SetMethodAccessRule { .. }
```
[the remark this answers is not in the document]
talekhinezh: Yup, exactly why I didn't add it :)

transaction/src/model/instruction.rs
```diff
@@ -160,6 +162,20 @@ pub enum Instruction {
         rule: AccessRule,
     },
 
+    SetGroupAccessRule {
```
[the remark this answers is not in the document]
talekhinezh: ooof yeah kk

radix-engine/src/system/system_modules/node_move/node_move_module.rs
talekhinezh: Good catch!

radix-engine/src/kernel/actor.rs
talekhinezh: `object_info.global = true` is set for any object module in a global node.

radix-engine-tests/tests/clock.rs
talekhinezh: yup good point

radix-engine/src/kernel/actor.rs
dhedey: Not that we have them, but I imagine it should also be true for non-main modules in an internal node?

PR #1007, opened 2023-05-09
Tracking missed/made proposals of validators

jakrawcz-rdx:
In this PR, I add internal tracking of missed/made proposal counters of validators:
- counting on round change;
- resetting on epoch change.

I follow the approach we discussed on Slack (captured as **Q4-B** at https://radixdlt.atlassian.net/wiki/spaces/S/pages/3058827265/REP-83+Calculating+Emissions+and+Handling+Validator+Fees#Unreliability-Penalty).

These numbers are not yet taken into account during Emissions (since Emissions itself is not implemented yet), and in fact, the counting has no externally observable effect on the engine's behavior (which is why I haven't invested much into tests yet - I just fixed a couple of existing tests that were breaking some newly-introduced sanity checks).

### DIFF
--- a/radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
+++ b/radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
@@ -67,19 +67,12 @@ impl CurrentProposalStatisticSubstate {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
+#[derive(Debug, Clone, PartialEq, Eq, Default, ScryptoSbor)]
 pub struct ProposalStatistic {
     /// A counter of successful proposals made by a specific validator.
     pub made: u64,
     /// A counter of missed proposals (caused both by gap rounds or fallback rounds).
     pub missed: u64,
-}
-
-impl ProposalStatistic {
-    /// Creates a zeroed statistic.
-    pub fn zero() -> Self {
-        Self { made: 0, missed: 0 }
-    }
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Sbor)]
@@ -431,7 +424,7 @@ impl EpochManagerBlueprint {
         // statistics (to be used for unreliability penalty calculation); at the moment we only
         // reset it.
         statistic.validator_statistics = (0..next_validator_set.len())
-            .map(|_index| ProposalStatistic::zero())
+            .map(|_index| ProposalStatistic::default())
             .collect();
         api.field_lock_write_typed(statistic_handle, statistic)?;
         api.field_lock_release(statistic_handle)?;
--- a/radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
+++ b/radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
@@ -40,9 +40,62 @@ pub struct CurrentValidatorSetSubstate {
     pub validator_set: BTreeMap<ComponentAddress, Validator>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
+pub struct CurrentProposalStatisticSubstate {
+    /// A proposal statistic of each validator from the current validator set, in the iteration
+    /// order of [`CurrentValidatorSetSubstate.validator_set`].
+    pub validator_statistics: Vec<ProposalStatistic>,
+}
+
+impl CurrentProposalStatisticSubstate {
+    /// Gets a mutable reference to a proposal statistic tracker of an individual validator.
+    pub fn get_mut_proposal_statistic(
+        &mut self,
+        validator_index: ValidatorIndex,
+    ) -> Result<&mut ProposalStatistic, RuntimeError> {
+        let validator_count = self.validator_statistics.len();
+        self.validator_statistics
+            .get_mut(validator_index as usize)
+            .ok_or_else(|| {
+                RuntimeError::ApplicationError(ApplicationError::EpochManagerError(
+                    EpochManagerError::InvalidaValidatorIndex {
+                        index: validator_index as usize,
+                        count: validator_count,
+                    },
+                ))
+            })
+    }
+}
+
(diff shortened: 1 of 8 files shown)


radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
```diff
@@ -293,15 +398,34 @@ impl EpochManagerBlueprint {
         let next_validator_set: BTreeMap<ComponentAddress, Validator> =
             validators.into_iter().collect();
 
-        let handle = api.actor_lock_field(
+        let validator_set_handle = api.actor_lock_field(
             OBJECT_HANDLE_SELF,
             EpochManagerField::CurrentValidatorSet.into(),
             LockFlags::MUTABLE,
         )?;
-        let mut validator_set: CurrentValidatorSetSubstate = api.field_lock_read_typed(handle)?;
```
jakrawcz-rdx, line 301: _(a note from the author)_

I noticed that this read was not needed, so I deleted it.

But my real question is: do we have some "just overwrite this field with this new value" API operation? (that could avoid some lock costs OR at least turn this into a 1-liner?)

radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
```diff
@@ -293,15 +398,34 @@ impl EpochManagerBlueprint {
...
+        )?;
+        api.field_lock_release(validator_set_handle)?;
+
+        let statistic_handle = api.actor_lock_field(
+            OBJECT_HANDLE_SELF,
+            EpochManagerField::CurrentProposalStatistic.into(),
+            LockFlags::MUTABLE,
+        )?;
+        let mut statistic: CurrentProposalStatisticSubstate =
+            api.field_lock_read_typed(statistic_handle)?;
```
jakrawcz-rdx, line 420: _(a note from the author)_

Here I definitely want to read and write (see the TODO below, to be addressed soon).

jakrawcz-rdx:
hmm, the CI is giving very weird errors:
```
error[E0433]: failed to resolve: use of undeclared type `TypedEpochManagerSubstateValue`
   --> radix-engine-queries/src/typed_substate_layout.rs:521:21
    |
521 |                     TypedEpochManagerSubstateValue::CurrentProposalStatistic(scrypto_decode(data)?)
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                     |
    |                     use of undeclared type `TypedEpochManagerSubstateValue`
    |                     help: an enum with a similar name exists: `TypedEpochManagerFieldValue`
```
(I mean: the line it complains does not look like that on my branch)

talekhinezh:
> hmm, the CI is giving very weird errors:
> 
> ```
> error[E0433]: failed to resolve: use of undeclared type `TypedEpochManagerSubstateValue`
>    --> radix-engine-queries/src/typed_substate_layout.rs:521:21
>     |
> 521 |                     TypedEpochManagerSubstateValue::CurrentProposalStatistic(scrypto_decode(data)?)
>     |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     |                     |
>     |                     use of undeclared type `TypedEpochManagerSubstateValue`
>     |                     help: an enum with a similar name exists: `TypedEpochManagerFieldValue`
> ```
> 
> (I mean: the line it complains does not look like that on my branch)

I'm guessing this may be a merge issue?

radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
talekhinezh: Agreed this would be a nice API but short answer, no we don't have this operation yet.

jakrawcz-rdx:
> I'm guessing this may be a merge issue?

Wuuuuut I can't believe it - I literally did `git pull origin develop` right before my original push, and it was "no changes". What a timing.
_(and then git auto-merged a non-obvious change without marking it as conflict - a rare case, and yeah it merged that wrong :p)_

I've now updated again, let's see.

radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
```diff
@@ -287,6 +351,47 @@ impl EpochManagerBlueprint {
...
 
+    fn update_proposal_statistics<Y>(
+        progressed_rounds: u64,
+        proposal_history: LeaderProposalHistory,
+        api: &mut Y,
+    ) -> Result<(), RuntimeError>
+    where
+        Y: ClientApi<RuntimeError>,
+    {
+        if proposal_history.gap_round_leaders.len() as u64 != progressed_rounds - 1 {
```
talekhinezh, line 362: Minor: feels like a little bit of a code smell that `progressed_rounds` is only used for verification. Maybe makes more sense for the check to occur before this function is called?

radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
```diff
@@ -40,9 +40,62 @@ pub struct CurrentValidatorSetSubstate {
...
+pub struct ProposalStatistic {
+    /// A counter of successful proposals made by a specific validator.
+    pub made: u64,
+    /// A counter of missed proposals (caused both by gap rounds or fallback rounds).
+    pub missed: u64,
+}
+
+impl ProposalStatistic {
+    /// Creates a zeroed statistic.
+    pub fn zero() -> Self {
```
talekhinezh, line 80: Mintor: Maybe makes better sense to use trait `Default` for this?

talekhinezh (approved):

radix-engine/src/blueprints/epoch_manager/epoch_manager.rs
jakrawcz-rdx: I agree it smells a bit.

But at the other hand, doing a `proposal_history` argument validation fits very well inside the `update_proposal_statistics()` function (which is supposed to process the `proposal_history`). I am reluctant to delegate the "argument validation" responsibility to the caller of the function (it would be more visible if we were e.g. calling this function from 2 places).

PR #1009, opened 2023-05-10
Rename ComponentAuthZone to LocalAuthZone

iamyulong:
## Summary

Rename `ComponentAuthZone` to `LocalAuthZone`, since there is an auth zone for every call frame (except the root), including both component and blueprint functions.

### For Dapp Developers

Use `LocalAuthZone` instead of `ComponentAuthZone`


dhedey (approved):
👍 👍

PR #1011, opened 2023-05-10
Grouping {max_validators, rounds_per_epoch, num_unstake_epochs} as EpochManagerInitialConfiguration struct

jakrawcz-rdx:
This is a ~pure refactor, which will allow to easily add more fields into that configuration (planned for emissions and validator fees work stream).


assets/blueprints/genesis_helper/src/lib.rs
```diff
@@ -75,7 +73,6 @@ mod genesis_helper {
 
             Self {
                 epoch_manager,
-                rounds_per_epoch,
```
jakrawcz-rdx, line 78: _(note from author)_

This was not used, so I deleted it, and then had to regenerate assets (`.sh`) and then apply some assert adjustments to unrelated metering tests 🤷‍♂️

dhedey (approved):
👍

PR #1015, opened 2023-05-11
Feature/Explicit SBOR discriminators

dhedey:
## Summary

Adds a new feature to SBOR derive which I've been meaning to add for ages: fixed discriminators.

The primary use case for this is to help align the Instruction discriminators to save @siy so much pain in the ledger app.

Notably:
* If an enum gives all its variants `#[sbor(discriminator(Y))]` OR uses #[repr(X)] and has u8-convertible ` = Y` then SBOR uses these discriminants for encoding, decoding and schema (the former takes precedence over the latter). `Y` can either be a u8 literal, decimal literal, or string containing that literal - or a path to a constant variable.
* I've also refactored/improved more of the SBOR derive codebase - notably I've created a nice 1-level-deep attribute parsing utility, with much better/clearer error reporting (using spans) if sbor attributes are not in the correct format.

## Testing

* Added tests in the `sbor-tests` crate for the happy path.
* Tested lots of the unhappy paths in the IDE to check that error messages are output in nice places.
* If you use a duplicate index, then the code output by `SborDecode` / `SborDescribe` results in a build warning due to unreachable code - this signals to the user that there is a duplicate discriminator problem (on the duplicated discriminator, thanks to the magic of using spans correctly!). I initially tried building this detection into the macro, but it's too flimsy because two different constants can have the same value and the macro has no way to know (because of when it's built) - so instead, I removed this and rely on the IDE.

### For Internal Integrators - Ledger App

* See instructions.rs for the constants to use now.
* Some other models (such as `EntityType`) have automatically had their discriminants updated as part of this change. I don't think this should affect you, but just letting you know.


sbor-tests/tests/enum.rs
```diff
@@ -0,0 +1,95 @@
...
+const CONST_55: u8 = 55;
+
+#[derive(Sbor, PartialEq, Eq, Debug)]
+#[repr(u8)]
+pub enum Mixed {
+    #[sbor(id = 5)]
+    A,
+    #[sbor(id(7))]
+    B,
+    #[sbor(id("8"))]
```
iamyulong, line 27: I'm a bit worried that we support all three syntaxes. 

Possible to limit to a single syntax with value being "any expression that evaluates into u8"?

dhedey: Hah I actually pre-empted this feedback with my latest comment and removed `id` before reading your message :) - we only support `discriminator` now.

This gives us then two core options (repr and `#[sbor(discriminator)]`).

I think if the user specifies a `repr` then it makes sense to use that for the discriminators by default (because I imagine at a low level the compiler can optimise away lookups/branches).

But forcing use of repr feels wrong (as it might affect performance in other ways) - so I wanted to provide an alternative mechanism, hence the sbor attribute.

Does that sound good to you?

iamyulong: I don't think we should use `repr`, because users may `repr` to `non-u8` for other purposes. I feel we should be very SBOR-specific, even if it means users may have to annotate twice.

But I will leave the decision to you :)

iamyulong (approved):

iamyulong (approved):

iamyulong (approved):

sbor-tests/tests/enum.rs
dhedey: I take your point!

I've tweaked it so that we only use the `repr` values if we specify `#[sbor(use_repr_discriminators)]` - this way it won't happen by accident. And they can still be overridden with explicit annotations.

PR #1016, opened 2023-05-11
Feature/rename worktop instructions

iamyulong:
## Summary

Another batch of renames:

| Old Instruction                     | New instruction                         |
|-------------------------------------|-----------------------------------------|
| `TAKE_FROM_WORKTOP_BY_AMOUNT`       | `TAKE_FROM_WORKTOP`                     |
| `TAKE_FROM_WORKTOP_BY_IDS`          | `TAKE_NON_FUNGIBLES_FROM_WORKTOP`       |
| `TAKE_FROM_WORKTOP`                 | `TAKE_ALL_FROM_WORKTOP`                 |
| `ASSERT_WORKTOP_CONTAINS_BY_AMOUNT` | `ASSERT_WORKTOP_CONTAINS`               |
| `ASSERT_WORKTOP_CONTAINS_BY_IDS`    | `ASSERT_WORKTOP_CONTAINS_NON_FUNGIBLES` |
| `ASSERT_WORKTOP_CONTAINS`           | Removed                                 |


transaction/src/model/instruction.rs
```diff
@@ -35,19 +35,14 @@ pub enum Instruction {
...
     },
 
-    /// Asserts worktop contains resource.
-    AssertWorktopContains {
-        resource_address: ResourceAddress,
-    },
-
     /// Asserts worktop contains resource by at least the given amount.
-    AssertWorktopContainsByAmount {
+    AssertWorktopContains {
```
dhedey, line 39: What are your thoughts on also fixing the order of the parameters on all these instructions so that `resource_address` is first, as per the transfer methods and other changes?

I thought generally we agreed a few months ago we were going to change all instructions to have ResourceAddress first, because as you write these instructions, you care about what resource, and then the amount.

(Appreciate you might say "assert I have 5 XRD" / "transfer 5 XRD" but we said in each case specifying Transfer XRD 5 was more logical :) )

dhedey (approved):
Really nice - thanks for working through these changes so rigorously/diligently and fixing up other things around the instructions (eg manifest builder + comments)

transaction/src/model/instruction.rs
iamyulong: As commented in another PR. I don't mint either way. A larger clean-up will be required for the parameter order change.

PR #1017, opened 2023-05-11
Feature/split bucket vault and proof

iamyulong:
## Summary

This PR splits `Bucket`, `Vault` and `Proof` into its fungible and non-fungible version.

| Type                | Methods available                            |
|---------------------|----------------------------------------------|
| `Bucket`            | Common bucket methods                        |
| `FungibleBucket`    | Common bucket methods + Fungible Methods     |
| `NonFungibleBucket` | Common bucket methods + Non-fungible methods |
| `Vault`             | Common vault methods                         |
| `FungibleVault`     | Common vault methods + Fungible Methods      |
| `NonFungibleVault`  | Common vault methods + Non-fungible methods  |
| `Proof`             | Common proof methods                         |
| `FungibleProof`     | Common proof methods + Fungible Methods      |
| `NonFungibleProof`  | Common proof methods + Non-fungible methods  |

To convert a bucket into its fungible/non-fungible variant, use `.as_fungible()`/`.as_non_fungible()`.

To convert back into its common form, use `.into()`.

Example:
```rust
fn take(&self, bucket: Bucket) {
    let non_fungible_bucket = bucket.as_non_fungible();
    // Non-fungible methods are now available on the `non_fungible_bucket` variable
    let bucket: Bucket = non_fungible_bucket.into();
}
```

Refactor has been made to the proof validation process.  To read proof state, you now have to use either of the following methods on a `Proof`, `FungibleProof` or `Proof`:

```rust
let checked_proof = proof.check(some_resource_address);
let checked_proof = proof.skip_checking();

let amount = checked_proof.amount();
let checked_non_fungible_proof = checked_proof.as_non_fungible();
```

### For Dapp Developers

You can now use `(Non)FungibleBucket`/`Vault`/`Proof` as function arguments or returns.

To use non-fungible methods, you will need to convert a `Bucket` into `NonFungibleBucket` using `.as_non_fungible()`. The same applies to `Vault` and `Proof`.

See above for the proof checking change.


native-sdk/src/resource/vault.rs
```diff
@@ -7,13 +7,90 @@ use radix_engine_interface::types::*;
...
-impl Vault {
-    pub fn sys_new<Y, E: Debug + ScryptoDecode>(
+pub trait NativeVault {
+    fn create<Y, E: Debug + ScryptoDecode>(
         resource_address: ResourceAddress,
         api: &mut Y,
-    ) -> Result<Self, E>
+    ) -> Result<Vault, E>
+    where
+        Y: ClientApi<E>;
```
dhedey, line 18: Incidental comment - not suggesting we change this as part of this PR!

For all of these, I'd be tempted to consider ditching the `Y` generic and using `api: &mut impl ClientApi<E>` which would save a few lines on each definition :). Also - I'm really not sure why ClientApi is generic over the error type? It appears E is always RuntimeError, so we could remove that too.

Perhaps if we wanted to change this we could fix them all together - I'm not suggesting it's a big priority though.

radix-engine-interface/src/blueprints/resource/bucket.rs
```diff
@@ -77,9 +77,31 @@ pub type BucketCreateProofOfAllOutput = Proof;
 // Stub
 //========
 
+// TODO: update schema type
```
dhedey, line 80: We could/should probably make this transparent too?

And we can derive `ScryptoCategorize`, `ScryptoEncode` and `ScryptoDecode` automatically? :) - (with Describe manual for now)

radix-engine-interface/src/blueprints/resource/proof.rs
```diff
@@ -42,66 +38,35 @@ pub struct ProofCloneInput {}
...
 
-    /// Specifies that the `Proof` should have its resource address validated against a set of `ResourceAddress`es. If
-    /// the `Proof`'s resource address belongs to the set, then its valid.
-    ValidateResourceAddressBelongsTo(BTreeSet<ResourceAddress>),
+// TODO: update schema type
 
-    /// Specifies that the `Proof` should be validating for containing a specific `NonFungibleGlobalId`.
-    ValidateContainsNonFungible(NonFungibleGlobalId),
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub struct Proof(pub Own);
```
dhedey, line 48: Similarly - I'd suggest we stick on an #[sbor(transparent)] and impl all but Describe manually?

(I've still got a TODO to allow for easier replacement of validations in Describe 👍)

radix-engine-interface/src/blueprints/resource/vault.rs
```diff
@@ -69,9 +69,31 @@ pub type VaultCreateProofOfAmountOutput = Proof;
 // Stub
 //========
 
+// TODO: update schema type
```
dhedey, line 72: Ditto here RE auto-derive?

radix-engine-tests/tests/blueprints/non_fungible/src/lib.rs
```diff
@@ -382,17 +392,17 @@ mod non_fungible_test {
...
-            let _: Sandwich = vault.non_fungible().data();
+            let _: Sandwich = vault.as_non_fungible_vault().non_fungible().data();
 
             // read singleton proof
-            let proof = vault.create_proof();
-            let validated_proof = proof.validate_proof(vault.resource_address()).unwrap();
-            let _: Sandwich = validated_proof.non_fungible().data();
-            validated_proof.drop();
+            let proof = vault.create_proof().skip_checking();
+            assert_eq!(proof.resource_address(), vault.resource_address());
```
dhedey, line 403: Out of interest - why did we remove `validate_proof`? Or why did we change this line if we didn't remove it?

radix-engine-tests/tests/blueprints/proof_creation/src/lib.rs
```diff
@@ -18,63 +18,69 @@ mod pc {
 
         pub fn create_proof_from_fungible_bucket() {
             let bucket = Self::create_fungible_bucket();
-            let proof = bucket.create_proof();
+            let proof = bucket.create_proof().skip_checking();
```
dhedey, line 21: Out of interest, why do we now need to skip checking on proofs created from buckets?

scrypto/src/resource/proof.rs
```diff
@@ -366,7 +387,7 @@ impl ValidatedProof {
     ///
     /// # Panics
     /// Panics if this is not a singleton bucket
-    pub fn non_fungible_local_id(&self) -> NonFungibleLocalId {
+    fn non_fungible_local_id(&self) -> NonFungibleLocalId {
```
dhedey, line 390: I feel that these RustDocs by default should be on the trait? I'm not sure if you can always see them on an impl?

dhedey (approved):
Really nice work - can't have been the funnest of tasks!

scrypto/src/resource/proof.rs
dhedey: And ditto to all these impls actually.

assets/blueprints/faucet/src/lib.rs
```diff
@@ -48,7 +48,7 @@ mod faucet {
         pub fn lock_fee(&mut self, amount: Decimal) {
             // There is MAX_COST_UNIT_LIMIT and COST_UNIT_PRICE which limit how much fee can be spent
             // per transaction, thus no further limitation is applied.
-            self.vault.lock_fee(amount);
+            self.vault.as_fungible_vault().lock_fee(amount);
```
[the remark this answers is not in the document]
iamyulong: I wanted to use shorter names as suggested, but realised `.as_non_fungible()` is very confusing with `.non_fungible()`. This is why I kept full name for both fungible/non-fungible buckets, proofs, and vaults.

radix-engine-tests/tests/blueprints/non_fungible/src/lib.rs
```diff
@@ -59,10 +59,10 @@ mod non_fungible_test {
         }
 
         pub fn update_nft(mint_badge: Bucket, proof: Proof) -> Bucket {
-            let proof = proof.unsafe_skip_proof_validation();
+            let proof = proof.skip_checking();
```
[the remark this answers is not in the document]
iamyulong: Yeah, had a comment there. If we have `Proof<Address>`, we don't need `CheckedProof` etc.

assets/blueprints/faucet/src/lib.rs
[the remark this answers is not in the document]
iamyulong: Maybe I will just remove the suffix and see if it causes real problem.

native-sdk/src/resource/bucket.rs
```diff
@@ -11,118 +11,121 @@ use sbor::rust::fmt::Debug;
 
 use super::ResourceManager;
 
-pub trait SysBucket {
-    fn sys_new<Y, E: Debug + ScryptoCategorize + ScryptoDecode>(
+// TODO: split impl
```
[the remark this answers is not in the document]
iamyulong: Right

assets/blueprints/faucet/src/lib.rs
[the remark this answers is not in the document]
iamyulong: Very good point! 

I think It's worth us doing a final round of example update, after all these Scrypto changes are in.

assets/blueprints/flash_loan/src/lib.rs
```diff
@@ -91,7 +91,7 @@ mod basic_flash_loan {
             );
 
             // Verify we are being sent at least the amount due
-            let terms: LoanDue = loan_terms.non_fungible().data();
+            let terms: LoanDue = loan_terms.as_non_fungible_bucket().non_fungible().data();
```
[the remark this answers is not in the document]
iamyulong: Ditto

native-sdk/src/resource/vault.rs
iamyulong: It's because we use it for both `native-sdk` and `scrypto`. They produce different errors.

radix-engine-tests/tests/blueprints/non_fungible/src/lib.rs
iamyulong: Sorry, I don't really remember this specific line change. It's possibly caused by massive find & replace. :)

radix-engine-tests/tests/blueprints/proof_creation/src/lib.rs
iamyulong: Because if we have a bucket from function arguments and create a proof from it. We may forget to check the address.

scrypto/src/resource/proof.rs
iamyulong: Thought about that. Feel like we need a round of Rustdoc clean-up after all scrypto changes are in.

native-sdk/src/resource/vault.rs
dhedey: Got it - although in `native-sdk`, I think all these generics are quite distracting, and would suggest it would be cleaner if we could either have a concrete type for the whole ClientApi, or a trait "alias" `NativeClientApi` which hardcodes `E = RuntimeError` (similar to eg the ScryptoEncode trait)

PR #1018, opened 2023-05-11
refactor: SBOR type data + type name updates

dhedey:
## Summary

Simplified the Describe trait and well-known types by making Type Data not optional.

This was motivated by wanting to fix the transparent macro so it now replaces the type name with the wrapper type name.

Also fixed the issue that [u8; N] is now not called "Bytes" which Omar noticed.


sbor/src/schema/well_known_types.rs
```diff
@@ -3,22 +3,87 @@ pub const CUSTOM_WELL_KNOWN_TYPE_START: u8 = 0x80;
...
+        })
+    }
+
     pub const UNIT_ID: u8 = 0x42; // `()`
+    pub fn unit_type_data<C: CustomTypeKind<L>, L: SchemaTypeLink>() -> TypeData<C, L> {
+        TypeData::no_child_names(
+            TypeKind::Tuple {
+                field_types: sbor::rust::prelude::vec![],
+            },
+            "None",
```
iamyulong, line 84: Is `None` preferred? I personally like `Unit` more, following Rust's terminologies.

iamyulong (approved):

sbor/src/schema/well_known_types.rs
dhedey: So when I originally envisioned this, the name was going to be shown in (eg) wallets/dashboards etc, in the user display of types. Where Unit would be out of place.

But actually I think I should leave this empty and leave it up to the UI to decide how to display this - so will make it unnamed.

PR #1022, opened 2023-05-12
Add initial time to genesis

LukasGasior1:


radix-engine/src/system/bootstrap.rs
```diff
@@ -171,7 +174,7 @@ where
                 data_ingestion_receipts.push(receipt);
             }
 
-            let genesis_wrap_up_receipt = self.execute_genesis_wrap_up(next_nonce);
+            let genesis_wrap_up_receipt = self.execute_genesis_wrap_up(next_nonce, initial_time_ms);
```
dhedey, line 177: Can we do this when the clock is created - so in the first transaction - probably in the Clock constructor?

One of the main reasons I want this is to avoid the Clock ever being set to 0, because that results in annoying logic in the Gateway DB when doing SQL queries 😆

radix-engine/src/system/bootstrap.rs
```diff
@@ -150,6 +152,7 @@ where
         genesis_data_chunks: Vec<GenesisDataChunk>,
         initial_epoch: u64,
         initial_configuration: EpochManagerInitialConfiguration,
+        initial_time_ms: i64,
```
dhedey, line 155: I possibly expected this to be an `Instant` but I see `Instant` is only to second precision. Even though we only store it at minute precision...

In any case, this is fine I guess :D. At some point soon we'll move Clock into EpochManager (ConsensusManager?) and fix the substates.

radix-engine/src/system/bootstrap.rs
LukasGasior1: Done

dhedey (approved):
Approving subject to tests being fixed before merge

PR #1023, opened 2023-05-12
fix: capture invalid addresses in SBOR values

lrubasze:
Improve error handling for ManifestAddress encoding

## Summary
- Added `ManifestAddress` constructor that checks if given entity type is valid.
- Added `EncodeError::CustomError` for SBOR encoder to handle encoding errors for SBOR custom values
- Added test to verify encoding/decoding of invalid/valid manifest custom addresses


sbor/src/encoder.rs
```diff
@@ -22,6 +22,9 @@ pub enum EncodeError {
         value_value_kind: u8,
         actual_value_kind: u8,
     },
+    InvalidAddressEntityType {
```
dhedey, line 25: So I have a couple of slight concerns re abstractions here:

Firstly - this isn't an error with the model from an SBOR perspective - this is an error with the fact we have an invalid address.

So I'd rather we make it impossible to create the custom value in the first place - eg in the constructor / decoder of manifest address like we do with (eg) `GlobalAddress`.

As a second point, this is a Custom Value - so shouldn't be added to the native SBOR EncodeError if we wanted to capture this, we should probably define some kind of CustomError(String) or something :)

dhedey (commented):
The right idea, but I'd suggest we approach this a little differently :) - have left a suggestion

sbor/src/encoder.rs
lrubasze: Thanks @dhedey for feedback.
I reworked my changes according to your suggestions.

PR #1025, opened 2023-05-14
Feature: Access Controller recovery badge

0xOmarA:
# Summary

* Upon the instantiation of the access controller component a new resource called the "recovery resource" is created.
* The recovery resource has the component address of the access controller in its metadata
* More of the recovery resource can be minted by the primary and recovery roles. 

There is a minor change to substates where the access controller substate now has an additional `recovery_badge` field. 

Note: The access controller uses a post_instantiation method because of the bug with preallocated addresses and references discussed in slack. Without the post instantiation function, we're unable to get the recovery badge to have the access controller's component address in its metadata.

### DIFF
--- a/radix-engine/src/blueprints/access_controller/package.rs
+++ b/radix-engine/src/blueprints/access_controller/package.rs
@@ -5,19 +5,13 @@ use crate::event_schema;
 use crate::kernel::kernel_api::KernelNodeApi;
 use crate::system::system_modules::costing::FIXED_LOW_FEE;
 use crate::types::*;
+use native_sdk::component::BorrowedObject;
 use native_sdk::modules::access_rules::{AccessRules, AccessRulesObject, AttachedAccessRules};
 use native_sdk::modules::metadata::Metadata;
 use native_sdk::modules::royalty::ComponentRoyalty;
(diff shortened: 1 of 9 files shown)


radix-engine/src/blueprints/access_controller/package.rs
```diff
@@ -565,9 +705,134 @@ impl AccessControllerNativePackage {
...
+                value: scrypto_decode(
+                    &scrypto_encode(&MetadataEntry::Value(MetadataValue::String(
+                        "Recovery Badge".into(),
+                    )))
+                    .unwrap(),
+                )
+                .unwrap(),
+            })
+            .unwrap(),
+        )?;
```
iamyulong, line 775: Can leverage the Native SDK to simplify the module method call?

iamyulong (approved):

radix-engine/src/blueprints/access_controller/package.rs
0xOmarA: Sure can do! I initially didn't use the native sdk as I thought we'd soon fix the issue anyway. I've updated the PR to use the native SDK here :)

PR #1037, opened 2023-05-17
Implementing the operations for owner's interaction with his "locked stake unit vault"

jakrawcz-rdx:
This implements the **Q5-B** of https://radixdlt.atlassian.net/wiki/spaces/S/pages/3058827265/REP-83+Calculating+Emissions+and+Handling+Validator+Fees pretty exactly as designed.

Please excuse the PR size - more than half is tests.


radix-engine/src/blueprints/epoch_manager/validator.rs
```diff
@@ -456,6 +457,168 @@ impl ValidatorBlueprint {
...
+
+        api.field_lock_release(handle)?;
+        Ok(())
+    }
+
+    /// Starts the process of unlocking the owner's stake units stored in the internal vault.
+    /// The requested amount of stake units (if available) will be ready for withdrawal after the
+    /// network-configured [`EpochManagerConfigSubstate.num_owner_stake_units_unlock_epochs`] via a
+    /// call to [`finish_unlock_owner_stake_units()`].
+    /// As an optimization, this method also performs the same "finish unlocking" operation, and
```
iamyulong, line 487: I don't quite like the optimisation here. 
- It makes the interface less intuitive and force users to handle the returned bucket
- It may even downgrade the performance due to extra `take` invocations, which can batched in `finish_unlock`.

radix-engine-interface/src/blueprints/epoch_manager/invocations.rs
```diff
@@ -55,6 +55,11 @@ impl EpochManagerInitialConfiguration {
         self.min_validator_reliability = new_value;
         self
     }
+
+    pub fn with_num_owner_stake_units_unlock_epochs(mut self, new_value: u64) -> Self {
```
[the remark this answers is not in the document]
jakrawcz-rdx: I am worried that then David will come and say "let's have long, descriptive names, bring back the `with_num_owner_stake_units_unlock_epochs()`" :P

The field is named `num_owner_stake_units_unlock_epochs` so the "update" method kinda has to be `with_X()`. I prefer to keep the `_epochs` suffix there, since it specifies the "time units" of that unlock delay. (normally I would agree to drop the unit from the name, IF we had a well-named data type representing `epochs`, but currently we use the generic u64)

radix-engine/src/blueprints/epoch_manager/validator.rs
```diff
@@ -456,6 +457,168 @@ impl ValidatorBlueprint {
...
+            .pending_owner_stake_unit_withdrawals
+            .range(..=current_epoch)
+            .map(|(epoch, available_amount)| (epoch.clone(), available_amount.clone()))
+            .collect::<Vec<_>>();
+        let mut total_already_available_amount = Decimal::zero();
+        for (epoch, available_amount) in available_withdrawals {
+            // no batch delete in a BTree
+            substate.pending_owner_stake_unit_withdrawals.remove(&epoch);
+            total_already_available_amount += available_amount;
+        }
```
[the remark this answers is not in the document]
jakrawcz-rdx: yup, `rust` is still so immature :p

radix-engine/src/blueprints/epoch_manager/validator.rs
jakrawcz-rdx: Your arguments are very valid.

I remember that this "optimization" idea came around somewhere on Slack (probably from David/Matt?) in order to mitigate the problem of "max pending withdrawals limit" (i.e. owner should be able to do "continuous withdrawal" easily).

One improvement that I see here would be to NOT return the alread-unlocked bucket, but instead move the amount from the `pending_owner_stake_unit_withdrawals` queue to a new, simple field (like, `already_unlocked_owner_stake_unit_amount: Decimal`). And on `finish_unlock`, the logic would simply take this field into account too.

This should actually address both of your points (easier API + no `.take()` in the `start_unlock`) for a little price of slightly-more-complex `finish_unlock` :) let me try...

radix-engine/src/blueprints/epoch_manager/validator.rs
```diff
@@ -456,6 +457,168 @@ impl ValidatorBlueprint {
...
+        // - insert the requested withdrawal as pending (if possible)
+        let currently_pending_withdrawals = substate.pending_owner_stake_unit_withdrawals.len();
+        if currently_pending_withdrawals >= OWNER_STAKE_UNITS_PENDING_WITHDRAWALS_LIMIT {
+            return Err(RuntimeError::ApplicationError(
+                ApplicationError::ValidatorError(
+                    ValidatorError::PendingOwnerStakeWithdrawalLimitReached,
+                ),
+            ));
+        }
+        substate
```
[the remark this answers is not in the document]
jakrawcz-rdx: From one point of view - yeah, it is arbitrary (we only have it because someone raised concern about iterating over thousands of items).

But at the same time, I think it's more elegant to return "sorry, you have too many pending withdrawals already" than to return "sorry, substate size limit reached".

radix-engine/src/blueprints/epoch_manager/validator.rs
jakrawcz-rdx: I have applied the idea from https://github.com/radixdlt/radixdlt-scrypto/pull/1037#discussion_r1201909487 - now the API is cleaner (no auto-finish-unlock-as-a-drive-by anymore). Please have a look @iamyulong 

I am still waiting for opinions about "let's not have any limit at all" - if more people vote that we can do it, I will do it (and then ofc this code will be ditched).

jakrawcz-rdx: Alright, so I realized that this collection will never reach the 4MB substate limit (in fact, it won't grow larger than a number of owner stake withdrawal epochs delay - because we stack withdrawals scheduled for the same epoch! see `multiple_pending_owner_stake_unit_withdrawals_stack_up()`).

This means we can remove the explicit limit too, without much risk of a performance hit. Done 👍

iamyulong (approved):
Good to go!

jakrawcz-rdx:
🙇 Thanks for the thorough review, @iamyulong !
Let's go to https://github.com/radixdlt/radixdlt-scrypto/pull/1046 then 🎉

PR #1043, opened 2023-05-18
Fix a bug in wasm prepare step

iamyulong:
## Summary

Fix a bug in the wasm prepare step. 

Because of instrumentation, we can't assume the module is still valid post transformation (despite serializable).


dhedey (approved):
Approving, conditional on us adding validation to the schema generation in the Scrypto CLI - OR reverting the panic back to a Result, and putting the expect back in `vm.rs`

PR #1044, opened 2023-05-18
fix: Let 'resim show-ledger' list items as before

lrubasze:
## Summary

This PR fixes `resim show-ledger` command and let it display ledger entries as before.

## Details

- Added support for reverse mapping for the `DbPartitionKey` in order to retrieve `NodeId` used to restore ledger addresses.
It was agreed to change `DbPartitionKey` format to:
`HashPrefix(NodeId + ParitionNumber) + NodeId + PartitionNumber`
 It extends its size from `32` to `51` bytes, which is the price wa are willing to pay to have the reverse mapping available.
 
- Implemented `list_partition_keys()` for `InMemorySubstateDatabase` and `RocksdbSubstateStore`

## Testing

Added test to `resim.sh` to check if 'resim show-ledger' dumps entities.

### For Internal Integrators

Node shall be updated to align to the new `DbPartitionKey` format. Database needs to be reset.

### DIFF
--- a/radix-engine-stores/src/rocks_db.rs
+++ b/radix-engine-stores/src/rocks_db.rs
@@ -64,18 +64,16 @@ impl CommittableSubstateDatabase for RocksdbSubstateStore {
 
 impl ListableSubstateDatabase for RocksdbSubstateStore {
     fn list_partition_keys(&self) -> Box<dyn Iterator<Item = DbPartitionKey> + '_> {
-        let mut partition_keys: Vec<DbPartitionKey> = self
-            .db
-            .iterator(IteratorMode::Start)
-            .map(|kv| {
-                let (iter_key_bytes, _) = kv.as_ref().unwrap();
-                let (iter_key, _) = decode_from_rocksdb_bytes(iter_key_bytes);
-                iter_key
-            })
-            .collect();
-        // Eliminate duplicated keys (if any). Rocksdb iterator returns sorted entries, so we can
-        // dedup() directly.
-        partition_keys.dedup();
+        let mut partition_keys: Vec<DbPartitionKey> = vec![];
+
+        self.db.iterator(IteratorMode::Start).for_each(|kv| {
+            let (iter_key_bytes, _) = kv.as_ref().unwrap();
+            let (iter_key, _) = decode_from_rocksdb_bytes(iter_key_bytes);
+            // Eliminate duplicated keys (if any).
+            if !partition_keys.contains(&iter_key) {
+                partition_keys.push(iter_key)
+            }
+        });
 
         Box::new(partition_keys.into_iter())
     }
--- a/radix-engine-stores/src/rocks_db.rs
+++ b/radix-engine-stores/src/rocks_db.rs
@@ -62,6 +62,22 @@ impl CommittableSubstateDatabase for RocksdbSubstateStore {
     }
 }
 
+impl ListableSubstateDatabase for RocksdbSubstateStore {
+    fn list_partition_keys(&self) -> Box<dyn Iterator<Item = DbPartitionKey> + '_> {
+        let partition_keys: Vec<DbPartitionKey> = self
+            .db
+            .iterator(IteratorMode::Start)
+            .map(|kv| {
+                let (iter_key_bytes, _) = kv.as_ref().unwrap();
+                let (iter_key, _) = decode_from_rocksdb_bytes(iter_key_bytes);
+                iter_key
+            })
+            .collect();
+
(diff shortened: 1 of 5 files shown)


radix-engine/src/track/db_key_mapper.rs
```diff
@@ -61,18 +61,27 @@ pub trait DatabaseKeyMapper {
 ///
 /// This implementation is the actual, protocol-enforced one, to be used in public Radix networks.
 ///
-/// This implementation achieves the prefix-spreading by:
-/// - using a (long, ~100% unique) hash instead of plain RE Node and Module ID (please note that it
-///   makes this mapping effectively irreversible);
-/// - using a (shorter, but hard to crack) hash prefix for Substate key.
+/// This implementation achieves the prefix-spreading by adding a hash prefix (shortened hash for
+/// performance reasons, but still hard to crach) to:
```
jakrawcz-rdx, line 65:
```suggestion
/// performance reasons, but still hard to crack) to:
```

radix-engine-stores/src/rocks_db.rs
```diff
@@ -62,6 +62,22 @@ impl CommittableSubstateDatabase for RocksdbSubstateStore {
...
 }
 
+impl ListableSubstateDatabase for RocksdbSubstateStore {
+    fn list_partition_keys(&self) -> Box<dyn Iterator<Item = DbPartitionKey> + '_> {
+        let partition_keys: Vec<DbPartitionKey> = self
+            .db
+            .iterator(IteratorMode::Start)
+            .map(|kv| {
+                let (iter_key_bytes, _) = kv.as_ref().unwrap();
+                let (iter_key, _) = decode_from_rocksdb_bytes(iter_key_bytes);
```
jakrawcz-rdx, line 72: hah, funny that `ListableSubstateDatabase` existed but had no impl :)

Anyway, its contract is not clear, but I assume it wants to list every partition key once. AFAIU, `InMemorySubstateDatabase` does it this way, but the rocksdb here lists duplicates?

radix-engine/src/track/db_key_mapper.rs
lrubasze: good catch! thanks 👍

radix-engine-stores/src/rocks_db.rs
lrubasze: Indeed keys might be duplicated. Fixed.
Decided not to use `dedup()` to not change the order of the keys.

radix-engine-stores/src/rocks_db.rs
```diff
@@ -62,6 +62,22 @@ impl CommittableSubstateDatabase for RocksdbSubstateStore {
     }
 }
 
+impl ListableSubstateDatabase for RocksdbSubstateStore {
+    fn list_partition_keys(&self) -> Box<dyn Iterator<Item = DbPartitionKey> + '_> {
+        let mut partition_keys: Vec<DbPartitionKey> = vec![];
+        // Eliminate duplicated keys (if any)
```
jakrawcz-rdx, line 68: The duplicates are kinda inherent, since a single partition is meant to hold potentially unbounded number of substates under different sort keys (by design).

Rocksdb iterates sorted entries (by design), so `.dedup()` would actually be very appropriate 👍

lrubasze: Alright. Will change to `dedup()`

radix-engine-stores/src/rocks_db.rs
```diff
@@ -62,6 +62,25 @@ impl CommittableSubstateDatabase for RocksdbSubstateStore {
...
+            .iterator(IteratorMode::Start)
+            .map(|kv| {
+                let (iter_key_bytes, _) = kv.as_ref().unwrap();
+                let (iter_key, _) = decode_from_rocksdb_bytes(iter_key_bytes);
+                iter_key
+            })
+            .collect();
+        // Eliminate duplicated keys (if any). Rocksdb iterator returns sorted entries, so we can
+        // dedup() directly.
+        partition_keys.dedup();
```
jakrawcz-rdx, line 78: I know this impl is just for test purposes, so we don't need to worry about e.g. runtime memory consumption... but still I'd really recommend putting the `.dedup()` in the iterator chain (right before `.collect()`).

jakrawcz-rdx (approved):
Very nice 👍 
_(I will assume that `list_entries()` renders stuff exactly as you like)_

lrubasze:
@jakrawcz-rdx thanks for review and hints. All works as expected 👍

PR #1045, opened 2023-05-18
Feature/expose fee reserve states

iamyulong:
## Summary

This PR introduces 4 APIs for querying the states of fee reserve:
- `Runtime::cost_unit_limit()`: Returns the cost unit limit, set by transaction
- `Runtime::cost_unit_price()`: Returns the current cost unit price in XRD, defined by protocol
- `Runtime::tip_percentage()`: Returns the tip percentage, set by transaction
- `Runtime::fee_balance()`: Returns the remaining locked fee balance


radix-engine/src/system/system_modules/costing/fee_reserve.rs
```diff
@@ -214,8 +214,8 @@ impl SystemLoanFeeReserve {
             effective_royalty_price,
 
             // System loan is used for both execution and royalty
-            xrd_balance: cost_unit_price * system_loan as u128,
-            xrd_owed: cost_unit_price * system_loan as u128,
+            xrd_balance: effective_execution_price * system_loan as u128,
+            xrd_owed: effective_execution_price * system_loan as u128,
```
iamyulong, line 218: Use `effective_execution_price` to avoid situations where system loan is not enough for a `lock_fee` instruction due to high tip percentage.

talekhinezh (approved):

PR #1048, opened 2023-05-21
Feature: Account Deposit Modes Revamped 

0xOmarA:
## Summary

This is the implementation of account deposit modes as described in REP-89. The account component now has the following partitions:

0x00: Account Substate
0x01: Account Vaults: `KeyValueStore<ResourceAddress, Own>` 
0x02: Account Resource Deposit Configurations: `KeyValueStore<ResourceAddress, Allowed | Disallowed>`

## Update Recommendations

For internal and external integrators, account deposits should now be done through `try_deposit*` or `try_deposit_unsafe`. If the signature or badge of the owner of the account can be produced then `deposit` can be used. 

### Internal Integrators

Accounts now have an additional partition. Please see REP-60 for the up to date state model.


radix-engine/src/blueprints/account/blueprint.rs
```diff
@@ -350,77 +348,104 @@ impl AccountBlueprint {
...
         for bucket in buckets {
             Self::deposit(bucket, api)?;
         }
-
         Ok(())
     }
 
-    pub fn safe_deposit<Y>(bucket: Bucket, api: &mut Y) -> Result<Option<Bucket>, RuntimeError>
+    /// Method is public to all - if the resource can't be deposited it is returned.
+    pub fn try_deposit<Y>(bucket: Bucket, api: &mut Y) -> Result<Option<Bucket>, RuntimeError>
```
iamyulong, line 378: For consistency, `try_deposit_return_on_failure`?

radix-engine/src/blueprints/account/blueprint.rs
```diff
@@ -350,77 +348,104 @@ impl AccountBlueprint {
...
         for bucket in buckets {
-            let rtn = Self::safe_deposit(bucket, api)?;
-            if let Some(bucket) = rtn {
-                undeposited_buckets.push(bucket)
-            }
+            Self::try_deposit_abort_on_failure(bucket, api)?;
         }
-
-        Ok(undeposited_buckets)
+        Ok(())
```
iamyulong, line 448: It would be efficient to call `try_deposit_batch_return_on_failure` here, because it check deposit rules upfront?

radix-engine/src/blueprints/account/blueprint.rs
```diff
@@ -762,52 +710,112 @@ impl AccountBlueprint {
...
-            AccountDepositsMode::AllowExisting => {
-                // Case: Deposit of XRD is always allowed in `AllowExisting` mode.
-                if *resource_address == RADIX_TOKEN {
-                    Ok(true)
+        // Read the account's current deposit operation mode
+        let default_deposit_rule = Self::get_current_default_deposit_rule(api)?;
+        let resource_deposits_configuration =
+            Self::get_resource_deposit_configuration(resource_address, api)?;
+
+        let is_deposit_allowed = match (default_deposit_rule, resource_deposits_configuration) {
```
iamyulong, line 724: I would consider refactoring the implementation here, by using nested matches instead of flattened matches, because loading the default deposit rule substate is optional.

```rust
match Self::get_resource_deposit_configuration(resource_address, api)? {
    Allowed => yes,
    Disallowed = no,
    Neither => match Self::get_current_default_deposit_rule(api)? {
            Accept => yes,
            Reject => no,
            AllowExisting => additional_check()
    }
}
```

iamyulong (approved):
LGTM with minor comments on performance and tidiness.

radix-engine/src/blueprints/account/blueprint.rs
0xOmarA: Yup, sorry did not notice with all the renames.

radix-engine/src/blueprints/account/blueprint.rs
0xOmarA: Agreed. I think that would be a nice clean up.

PR #1050, opened 2023-05-22
Adds indent to canonical manifest + some decompiler refactor

iamyulong:
## Summary

This PR adds indent to canonical manifest representation, plus some refactors to the manifest decompiler.

Example:
```rust
SET_COMPONENT_ROYALTY_CONFIG
    Address("${account_address}")
    Tuple(
        Map<String, U32>(
            "method",
            1u32
        ),
        0u32
    );
```

In addition, the syntax for enum has been changed to
```rust
Enum<0u8>()

Enum<1u8>(
    "String",
    100u32
)

Enum<"Well::Known">()

Enum<"Well::Known">(
    "String",
    100u32
)
```

### For Dapp Developers

See above for the new syntax for enum variants.

### DIFF
--- a/transaction/examples/account/deposit_modes.rtm
+++ b/transaction/examples/account/deposit_modes.rtm
@@ -28,8 +28,7 @@ CALL_METHOD
 CALL_METHOD 
     Address("${account_address}") 
     "change_allowed_deposits_mode" 
-    Enum(
-        "AccountDepositsMode::AllowList", 
+    Enum<"AccountDepositsMode::AllowList">(
         Array<Address>(
             Address("${first_resource_address}")
         )
@@ -51,8 +50,7 @@ CALL_METHOD
 CALL_METHOD 
     Address("${account_address}") 
     "change_allowed_deposits_mode" 
-    Enum(
-        "AccountDepositsMode::DisallowList", 
+    Enum<"AccountDepositsMode::DisallowList">(
         Array<Address>(
             Address("${first_resource_address}")
         )


transaction/src/manifest/e2e.rs
```diff
@@ -240,16 +248,34 @@ CALL_METHOD
     "aliases"
     Enum(0u8)
     Enum(0u8)
-    Enum(1u8, "hello")
-    Enum(1u8, "hello")
-    Enum(0u8, "test")
-    Enum(0u8, "test")
-    Enum(1u8, "test123")
-    Enum(1u8, "test123")
+    Enum(1u8,
```
0xOmarA, line 251: I think that it would be nice if the indented formatting for enums with fields was:
```rust
Enum(
    0u8,
    "hello",
    "world",
)
```

what do you think?

iamyulong: I think the enum discriminator is conceptually different from the associated values, and more coupled with enum variant, so was put besides `Enum(`. And you've been using this syntax for many of the examples, which I quite like.

But I'm happy to change it if you think the other way is more appropriate. 

_(Another crazy idea is to make it `Enum<5u8>()`)_

0xOmarA: Agreed on the variant id being different from the fields of the enum. Maybe one way to sort of distinguish these two things is by:

```
Enum(
    0u8, // Variant id
    [    // fields
        "hello",
        "world",
    ]
)
```

Also, I like crazy ideas and would never say no to them :D

iamyulong: I like the proposed from a structure wise, but I feel it's not very concise (depth increases by 2, but in fact is 1).

```
Enum(
    0u8,
    []
)

Enum(
    0u8,
    [
         A,
         B,
         C
    ]
)
```

Will give the crazy idea a try.

```
Enum<0u8>()

Enum<1u8>(
    A,
    B,
    C
)
```

0xOmarA: Yup happy with the "generics" syntax for enum variant ids and also happy with the way that enums are currently formatted :)

iamyulong: Updated. Please re-review when you're free. :)

PR #1052, opened 2023-05-23
Refactor: merge EpochManager + Clock -> ConsensusManager

jakrawcz-rdx:
~~⚠️ This builds on top of https://github.com/radixdlt/radixdlt-scrypto/pull/1046~~

As described at https://rdxworks.slack.com/archives/C01HK4QFXNY/p1684760243274809.


jakrawcz-rdx:
The dependency https://github.com/radixdlt/radixdlt-scrypto/pull/1046 got merged; I switched the target branch.

radix-engine-interface/src/types/node_layout.rs
```diff
@@ -123,11 +123,13 @@ impl TryFrom<u8> for EpochManagerPartitionOffset {
...
 #[derive(Debug, Clone, Sbor, PartialEq, Eq, Hash, PartialOrd, Ord, FromRepr)]
-pub enum EpochManagerField {
+pub enum ConsensusManagerField {
     Config,
-    EpochManager,
+    ConsensusManager,
     CurrentValidatorSet,
     CurrentProposalStatistic,
+    CurrentTimeRoundedToMinutes,
+    CurrentTime,
```
dhedey, line 132: I wonder if we should merge `CurrentTime` into `CurrentProposalStatistic` for efficiency? It's not super important, as they're both under the same node, the JMT doesn't care too much.

radix-engine-tests/tests/bootstrap.rs
```diff
@@ -380,27 +379,24 @@ fn test_genesis_time() {
...
+            dummy_consensus_manager_configuration(),
             123 * 60 * 1000 + 22, // 123 full minutes + 22 ms (which should be rounded down)
         )
         .unwrap();
 
-    let current_time_substate = substate_db
-        .get_mapped::<SpreadPrefixKeyMapper, ClockSubstate>(
-            CLOCK.as_node_id(),
+    let current_time_rounded_to_minutes_ms = substate_db
+        .get_mapped::<SpreadPrefixKeyMapper, i64>(
```
dhedey, line 388: I have a slight preference for this being an explicit named type - even if it's just a `type` alias.

radix-engine/src/blueprints/consensus_manager/package.rs
```diff
@@ -16,18 +16,20 @@ use resources_tracker_macro::trace_resources;
...
 
         let mut fields = Vec::new();
-        fields.push(aggregator.add_child_type_and_descendents::<EpochManagerConfigSubstate>());
-        fields.push(aggregator.add_child_type_and_descendents::<EpochManagerSubstate>());
+        fields.push(aggregator.add_child_type_and_descendents::<ConsensusManagerConfigSubstate>());
+        fields.push(aggregator.add_child_type_and_descendents::<ConsensusManagerSubstate>());
         fields.push(aggregator.add_child_type_and_descendents::<CurrentValidatorSetSubstate>());
         fields
             .push(aggregator.add_child_type_and_descendents::<CurrentProposalStatisticSubstate>());
+        fields.push(aggregator.add_child_type_and_descendents::<i64>());
```
dhedey, line 31: Would you be up for giving these substates explicit names? Eg `ProposerTimestampRoundedToPreviousMinute` / `LatestProposerTimestamp` or similar? :)

They can either be type aliases (with `type`) or explicit singleton structs, possibly with `#[sbor(transparent)]` if we want to just store them as `i64` for performance.

dhedey (approved):
Really nice stuff - happy to discuss the combining-set-and-round-update separately / as part of the epoch update change discussion.

radix-engine-interface/src/types/node_layout.rs
dhedey: Maybe best to keep them separate for abstraction reasons 🤷

PR #1056, opened 2023-05-24
tweak: Active Validator set is ordered by stake desc

dhedey:
## Summary
In the node, we were in some places ordering the active validator set by stake DESC and in some places using the engine ordering, this made the Core API confusing when we exposed the ValidatorIndex concept in the round change transactions.

So instead, this PR pushes that ordering up to the ActiveValidatorSet substate, so everything can be in harmony. (Note - if two validators have equal stake, they're ordered by their DbSortKey in the SortedIndex).

Also whilst I was at it, changed `next_epoch` method on the receipt to actually return the clearer `EpochChangeEvent`.

## Testing
Updated / expanded the consensus manager tests in `radix-engine-tests`.

### Node

Will need to update some API docs, and possibly remove some explicit ordering. Should be a small change :)

### DIFF
--- a/radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
+++ b/radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
@@ -46,7 +46,8 @@ pub struct Validator {
 
 #[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
 pub struct CurrentValidatorSetSubstate {
-    pub validator_set: BTreeMap<ComponentAddress, Validator>,
+    /// The current validator set in the epoch, ordered by stake descending.
+    pub validator_set: IndexMap<ComponentAddress, Validator>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
@@ -175,7 +176,7 @@ impl ConsensusManagerBlueprint {
                 round: 0,
             };
             let current_validator_set = CurrentValidatorSetSubstate {
-                validator_set: BTreeMap::new(),
+                validator_set: index_map_new(),
             };
             let current_proposal_statistic = CurrentProposalStatisticSubstate {
                 validator_statistics: Vec::new(),
@@ -503,7 +504,7 @@ impl ConsensusManagerBlueprint {
     where
         Y: ClientApi<RuntimeError>,
     {
-        // read previous validator set
+        // Read previous validator set
         let validator_set_handle = api.actor_lock_field(
             OBJECT_HANDLE_SELF,
             ConsensusManagerField::CurrentValidatorSet.into(),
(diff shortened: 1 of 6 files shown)


radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
```diff
@@ -532,19 +533,30 @@ impl ConsensusManagerBlueprint {
             api,
         )?;
 
-        // select next validator set
+        // Select next validator set
+        // NOTE - because the stake index is by u16 buckets, it's possible that there are multiple validators at the cut off point
+        // that fall into the same bucket, and it's possibly we arbitrarily (and incorrectly) choose a validator with lower stake,
+        // because they happen to have a higher DbSortKey in the index.
+        // This might cause us not to exactly select the top 100 validators. We accept this behaviour as a reasonable trade-off.
```
jakrawcz-rdx, line 540: wuuuuut only now I realized that the current impl in the Engine does not apply the solution that I discussed with Josh:
image
 
:(

I looked at the call stack down the "list sorted" op, and I acknowledge it's not easy to return an iterator from there (on which the business logic here could "iterate until the stake bin changes again after 100"). But it seems possible, if we decided it's worth pursuing.

(I also acknowledge that this solution has a potential of reading unbounded number of entries from DB; like e.g. all of them, if all validators have close-enough stakes)

Do you think "it's worth it"?
- if yes: please transform this comment into a `TODO` and I can give it a try.
- if no: please surface the information from this comment up, to our public docs (I totally expect collisions within the "stake bins" we have, and being chosen into a validator set [or not] is probably serious enough)

radix-engine/src/blueprints/consensus_manager/events/consensus_manager.rs
```diff
@@ -10,6 +10,6 @@ pub struct RoundChangeEvent {
 pub struct EpochChangeEvent {
     /// The *new* epoch's number.
     pub epoch: u64,
-    /// The *new* epoch's validator set.
-    pub validators: BTreeMap<ComponentAddress, Validator>,
+    /// The *new* epoch's validator set, ordered by stake descending.
+    pub validators: IndexMap<ComponentAddress, Validator>,
```
jakrawcz-rdx, line 14: instead of repeated `rustdoc + IndexMap`, could we have a single `ActiveValidatorSet` struct, re-used in the `EpochChangeEvent` and `CurrentValidatorSetSubstate`?

(it would also be a _perfect_ place for other repeated code like `.get_by_index(ValidatorIndex)` and `.get_by_address(ComponentAddress)` and `.total_active_stake()` etc)

radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
```diff
@@ -46,8 +46,44 @@ pub struct Validator {
...
+impl ActiveValidatorSet {
+    pub fn get_by_index(&self, index: ValidatorIndex) -> Option<(&ComponentAddress, &Validator)> {
+        self.validators_by_stake_desc.get_index(index as usize)
+    }
+
+    pub fn get_by_address(&self, address: &ComponentAddress) -> Option<&Validator> {
+        self.validators_by_stake_desc.get(address)
+    }
+
+    pub fn get_by_public_key(
```
jakrawcz-rdx, line 68: (minor) I imagine we want to use it only for tests (= worth annotating about poor runtime performance)

jakrawcz-rdx (approved):
The `ActiveValidatorSet` struct is absolutely sweet!

PR #1061, opened 2023-05-24
Advancing epoch on a "target wallclock duration + sanity round count range"

jakrawcz-rdx:
~~⚠️ This PR is based on https://github.com/radixdlt/radixdlt-scrypto/pull/1060.~~

This PR follows the https://radixdlt.atlassian.net/wiki/spaces/S/pages/2849865854/REP-41+-+Proposer+Time+and+the+Babylon+Engine#Section-D---Time-usage-in-the-protocol (regarding the epoch duration only).


radix-engine-interface/src/blueprints/consensus_manager/invocations.rs
```diff
@@ -70,6 +70,41 @@ impl ConsensusManagerInitialConfiguration {
...
 
+#[derive(Debug, Clone, PartialEq, Eq, Default, ScryptoSbor, ManifestSbor)]
+pub struct EpochChangeCondition {
+    /// A lower bound (inclusive) on number of rounds that is assumed to happen during the
+    /// [`duration_millis`].
+    /// If an actual number of rounds after [`duration_millis`] is less than this value, the epoch
+    /// change will wait until this value is reached.
+    pub min_round_count: u64,
+
+    /// An upper bound (exclusive) on number of rounds that is assumed to happen during the
```
dhedey, line 81: I'm slightly surprised that the upper bound is exclusive - as it means eg if something goes wrong* we get eg 9999 rounds rather than 10000 rounds but I don't think it matters too much :D.

* like validators slowing down time, which would be a surprising thing to happen.

dhedey: Actually, looks like this is inclusive in the logic? In the sense that once max round is hit, an epoch change occurs.

So perhaps the comment needs updating?

radix-engine-interface/src/blueprints/consensus_manager/invocations.rs
```diff
@@ -70,6 +70,41 @@ impl ConsensusManagerInitialConfiguration {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Default, ScryptoSbor, ManifestSbor)]
+pub struct EpochChangeCondition {
+    /// A lower bound (inclusive) on number of rounds that is assumed to happen during the
```
dhedey, line 75: Thanks again for such clear rust docs. Brilliant.

radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
```diff
@@ -446,20 +440,26 @@ impl ConsensusManagerBlueprint {
...
-        if round >= config.rounds_per_epoch {
-            let next_epoch = consensus_manager.epoch + 1;
-            Self::epoch_change(next_epoch, &config, api)?;
-            consensus_manager.epoch = next_epoch;
-            consensus_manager.round = 0;
+        let config = &config_substate.config;
+        let epoch_duration_millis = proposer_timestamp_milli - manager_substate.epoch_start_milli;
+        if config
+            .epoch_change_condition
+            .is_met(epoch_duration_millis, round)
```
talekhinezh, line 447: Nice!

radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
```diff
@@ -446,20 +440,26 @@ impl ConsensusManagerBlueprint {
...
+        {
+            let next_epoch = manager_substate.epoch + 1;
+            Self::epoch_change(next_epoch, config, api)?;
+            manager_substate.epoch = next_epoch;
+            manager_substate.epoch_start_milli = proposer_timestamp_milli;
```
dhedey, line 452: So this model will see some (slow) epoch drift over time, because they'll be a delay between when the 5 minutes is up and when a round change notices.

I wonder if a tiny bit of added complexity might be worth it to avoid this problem: we create separate fields `actual_epoch_start_milli` and `effective_epoch_start_milli` - and use `effective_epoch_start_milli` to measure `epoch_duration_millis`.

Then, when the epoch change condition is met, if `epoch_duration_millis` is sufficiently close to the epoch length (say, within 10%), then we set `manager_substate.effective_epoch_start_milli +=  standard_epoch_length` (and otherwise set it to `proposer_timestamp_milli` as currently).

This avoids the drift problem, as a slightly slow-to-change epoch will be offset in the next epoch :).

The `actual_epoch_start_milli` would just be as a sanity check and to let us monitor the drift per-epoch.

talekhinezh (approved):

dhedey (approved):
Really nice work!

Let me know what you think about the tweak to avoid drift (and make it easier to predict protocol updates accurately)

radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
jakrawcz-rdx: I see the guaranteed drift.

However, is there an end-user value in mitigating it? To me, "not so great":
- for short-term epoch number estimations, that drift won't affect things that much.
- for long-term estimations (like, "what epoch number will we have on 2026.06.06 at 13:37?") the probability of `>10% diff => set it to proposer_timestamp_milli` is totally not negligible - which makes the estimation as unreliable as with a drift.

Let's not add this complexity just yet.

radix-engine-interface/src/blueprints/consensus_manager/invocations.rs
jakrawcz-rdx: Hah, interesting, simply both words "inclusive" and "exclusive" can be used to describe this number :)

In my interpretation, this is an exclusive upper bound of a `range of rounds that we consider normal and under this condition, the timestamp drives the epoch`.
E.g. range `[3, 7)` means that "3, 4, 5 or 6 rounds are totally normal and if this is not exceeded, then timestamp drives the epoch". This implies: "if 7 rounds happen, then the timestamp no longer drives the epoch, but we cut the epoch right away". So 7 is not normal. It is not inside the "expected range of rounds". So it is an exclusive upper bound.

In fact, I wanted to use `expected_round_count: Range<Round>` there, but `Range<>` is not `Sbor` 🤷 

I see that under your definition, we have a "range of round counts that CAN happen in an epoch". Then we could have `[3, 7]` (inclusive) because `7` can be a number of rounds in an epoch (as a special case rule, where we disregard timestamp because of exceeding the normal round limit).

I now rephrased the docs a bit to make it all clear 👍

PR #1064, opened 2023-05-25
Tweak/manifest maps and semicolon change

dhedey:
## Summary

Captures two key changes:
* Maps are now written as `Map<X, Y>(x1 => y1, x2 => y2)`
* Canonical decompilation of instructions with arguments puts the `;` on a new line, so it's harder to miss, and less likely to cause issues when extending the argument list.

Also:
* Improves the manifest compiler Span to capture line number and offset, to aid with debugging compilation failures. Compilation may be slightly slower, but compilation is only performed off line, not on-ledger, so it's not a big issue.
* Fixed an ignored test in the e2e.rs, and improved the error messages on failure.

### For dApp Developers

Maps in the manifest are now written like:
```
Map<X, Y>(x1 => y1, x2 => y2)
```

### DIFF
--- a/transaction/src/manifest/lexer.rs
+++ b/transaction/src/manifest/lexer.rs
@@ -2,12 +2,22 @@ use sbor::rust::str::FromStr;
 
 /// The span of tokens. The `start` and `end` are Unicode code points / UTF-32 - as opposed to a
 /// byte-based / UTF-8 index.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Span {
     /// The start of the span, exclusive
-    pub start: usize,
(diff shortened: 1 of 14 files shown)


transaction/src/manifest/lexer.rs
```diff
@@ -2,12 +2,22 @@ use sbor::rust::str::FromStr;
...
     /// The start of the span, exclusive
-    pub start: usize,
+    pub start: Position,
     /// The end of the span, inclusive
-    pub end: usize,
+    pub end: Position,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Position {
```
iamyulong, line 14: Finally some proper implementation of `Position`.

iamyulong (approved):
LGTM

PR #1065, opened 2023-05-25
Refactor: Type-safe Epoch and Round throughout the codebase

jakrawcz-rdx:
~~⚠️ This is built on top of https://github.com/radixdlt/radixdlt-scrypto/pull/1061~~
Note: it does not mean it has to merge into `release/birch` (we can wait for the dep to merge to `release/birch`, and then for `release/birch` to merge into `develop`)

This is a 100% pure refactoring, no behavior change.


radix-engine-common/src/types/consensus.rs
```diff
@@ -0,0 +1,89 @@
+use sbor::Sbor;
+
+/// A type-safe consensus epoch number.
+/// Assuming one epoch per minute (i.e. much faster progression than designed), this gives us time
+/// until A.D. ~5700.
```
talekhinezh, line 5: Feels awfully close, and sounds like the same argument peeps had with `IPv4` and `32-bit` architecture. Is there a reason we don't just go with `u64`? I mightve missed this discussion.

jakrawcz-rdx: Hah, that's true... although those people didn't control the growth rate of the number of devices/networks connected to the internet globally.

Even with 1 epoch per second we have ~60 years. If our design changes to even more extreme rates (or if someone hacks the epoch progression) then we probably have more things to fix (and data migrations to perform) than `u32 -> u64` anyways.

That said: I don't push hard for this change myself. But more performance-aware people (@dhedey) argued that "it would save us a non-trivial amount of bytes in transactions" (see https://rdxworks.slack.com/archives/C01HK4QFXNY/p1684235019062509).

radix-engine-common/src/types/consensus.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /// Creates an epoch immediately preceding this one.
+    pub fn previous(&self) -> Self {
+        self.relative(-1)
+    }
+
+    /// Creates an epoch of a number relative to this one.
+    fn relative(&self, epoch_count: i64) -> Self {
+        Self(u32::try_from(self.0 as i64 + epoch_count).unwrap())
```
dhedey, line 52: A few things:
* It feels scary this panics if it overflows or underflows
* Why an i64? Rather than, say, an i32?
* Can we make this return an Option<Self> instead?

dhedey (approved):
Really nice, thanks.

I'd suggest we avoid panics on the Epoch API, or at least document clearly when they can occur.

dhedey (approved):
Really nice, thanks.

I'd suggest we avoid panics on the Epoch API, or at least document clearly when they can occur.

radix-engine-common/src/types/consensus.rs
dhedey: I guess we could keep it as u64 and only use u32 in the V1 transaction headers. I don't feel too strongly. Happy to be overruled if anyone does - everyone seemed pro the change on slack 👍 

But yeah - now that it's a single type, it's super easy to change I guess :)

radix-engine-common/src/types/consensus.rs
jakrawcz-rdx: > It feels scary this panics if it overflows or underflows. Can we make this return an Option instead?

AFAIU, the previous production behavior (i.e. `next_epoch = current_epoch + 1`) was not panicking, but silently wrapping back to 0. Is it a better behavior?

I mean, the code must do _something_ when it reaches those values. Returning an `Option` only changes _which_ code must do that.

My take is: we don't expect billions of epochs to be reached during our lifetimes. So if it happens, it is either because of a bug or a hack, and I'd appreciate if the system failed (with some obvious indication of the reason: a message or stack trace).

Let me explain this in rustdocs.

> Why an i64? Rather than, say, an i32?

This has a good reason actually:
A difference between two `u32`s can be in range `[-u32::MAX, u32::MAX]`.
The `i32` cannot represent that range (observe that `i32::MAX = u32::MAX / 2`), so `i64` is needed. (technically, `i33` could be fine, if it existed)

radix-engine-common/src/types/consensus.rs
jakrawcz-rdx: I've now migrated to `u64` (according to https://rdxworks.slack.com/archives/C01HK4QFXNY/p1685124666394689?thread_ts=1684235019.062509&cid=C01HK4QFXNY)

PR #1066, opened 2023-05-26
SCRY-438/Database read metrics + Fee table update

mstrug-rdx:
## Summary
Added dedicated tooling for RocksDB read overhead testing. 
Implemented `FeeTable` storage cost part `SubstateReadFromDb` costing value calculation.

## Details
More info [here](https://radixdlt.atlassian.net/wiki/spaces/~633a994d61dbef2805c10cb4/pages/3066298831/Storage+costing) (proposed algorithm and metrics graphs).


radix-engine-utils/Cargo.toml
```diff
@@ -4,12 +4,19 @@ version = "0.9.0"
 edition = "2021"
```
iamyulong, line 4: Minor: I'd suggest we rename `radix-engine-untils` to `radix-engine-profiling`, since a few people have been confused by this. :)

radix-engine-utils/src/rocks_db_metrics/mod.rs
```diff
@@ -0,0 +1,544 @@
...
+    pub read_metrics: RefCell<BTreeMap<usize, Vec<Duration>>>,
+}
+
+impl SubstateStoreWithMetrics<RocksdbSubstateStore> {
+    pub fn new_rocksdb(path: PathBuf) -> Self {
+        let mut factory_opts = BlockBasedOptions::default();
+        factory_opts.disable_cache();
+
+        let mut opt = Options::default();
+        opt.set_disable_auto_compactions(true);
```
iamyulong, line 26: Out of curiosity, why do we disable `auto_compactions`?

radix-engine-utils/src/rocks_db_metrics/mod.rs
```diff
@@ -0,0 +1,544 @@
...
+                    .unwrap()
+                    .push(duration);
+            } else {
+                self.read_metrics
+                    .borrow_mut()
+                    .insert(value.len(), vec![duration]);
+            }
+            Some(value)
+        } else {
+            None
```
iamyulong, line 73: I think we should record both, no? My understanding is that RocksDB will need to check the level files before issuing a `None`.

radix-engine-utils/src/rocks_db_metrics/mod.rs
```diff
@@ -0,0 +1,544 @@
...
+        export_graph_and_print_summary_for_two_series(
+            "RocksDB - InMemoryDB random reads",
+            &rocksdb_output_data,
+            &inmem_output_data,
+            "/tmp/scrypto_diff_1.png",
+        )
+        .unwrap();
+    }
+
+    fn drop_edge_values<S: SubstateDatabase + CommittableSubstateDatabase>(
```
iamyulong, line 204: Minor: `drop_highest_and_lowest` for explicitness?

radix-engine-utils/src/rocks_db_metrics/mod.rs
```diff
@@ -0,0 +1,544 @@
...
+
+                let mut partition = PartitionUpdates::new();
+                partition.insert(sort_key.clone(), value);
+
+                let partition_key = DbPartitionKey(p_key_cnt.to_be_bytes().to_vec());
+                p_key_cnt += 1;
+
+                data_index_vector.push((partition_key.clone(), sort_key, size));
+
+                input_data.insert(partition_key, partition);
```
iamyulong, line 260: We may use `to_db_partition_key(node_id: &NodeId, partition_num: PartitionNumber)` and `to_db_sort_key(key: &SubstateKey)` to generate data, as this is how RE uses substate store.

Also, we will need to re-run the tests by using the interface Jakub will provide, as the default store doesn't use merkle tree (which has huge overhead).

radix-engine-utils/src/rocks_db_metrics/mod.rs
```diff
@@ -0,0 +1,544 @@
...
+        );
+        println!(
+            "Linear approx.:  f(size) = {} * size + {}\n",
+            lin_slope, lin_intercept
+        );
+
+        Ok(())
+    }
+
+    pub fn export_graph_and_print_summary_for_two_series(
```
iamyulong, line 423: Amazing that you've automated linear regression and even graphing!

radix-engine-utils/src/rocks_db_metrics/mod.rs
mstrug-rdx: I just wanted to ensure that database will not be automatically compacted to replicate more real usage scenario, where we won't be using database in compacted state all the time.

radix-engine-utils/src/rocks_db_metrics/mod.rs
mstrug-rdx: Just using some dedicated crates.

radix-engine-utils/src/rocks_db_metrics/mod.rs
mstrug-rdx: I refactored db preparation code to use `to_db_partition_key()` and `to_db_sort_key()`, but this requires moving `db_key_mapper.rs` file from `radix-engine` project because of circular dependency (`radix-engine` uses `radix-engine-profiling`). I moved it to `radix-engine-store-interface` project.

iamyulong (approved):

PR #1072, opened 2023-05-30
Feature: Vault Freeze

talekhinezh:
## Summary
Add Vault Freeze/Unfreeze capability

## Details
* Generify direct vault manifest instructions
* Add Freeze/Unfreeze roles to ResourceManager
* Add AccessRules to Vaults (special cased for now)
* Add VAULT_FREEZE and VAULT_UNFREEZE manifest aliases

## Testing
* Add ResourceManager Freeze/Unfreeze role testing
* Add Vault freeze/unfreeze method testing

### For dApp Developers
You can now specify whether a resource's vaults are freezeable or not.

### For Internal Integrators
`RECALL_RESOURCE` instruction has been generified to be a generic direct vault method.


radix-engine-tests/tests/metering.rs
```diff
@@ -35,6 +35,7 @@ fn execute_with_time_logging(
 }
 
 #[test]
+#[ignore]
```
iamyulong, line 38: Accidental change?  You might find `cargo nextest run --no-fail-fast` useful. :)

transaction/src/manifest/decompiler.rs
```diff
@@ -514,8 +515,41 @@ pub fn decompile_instruction<F: fmt::Write>(
...
+        InstructionV1::CallDirectVaultMethod {
+            vault_id,
+            method_name,
+            args,
+        } => {
+            let mut fields = Vec::new();
+            let name = match method_name.as_str() {
+                VAULT_RECALL_IDENT => {
+                    fields.push(to_manifest_value(vault_id));
+                    "RECALL_RESOURCE"
```
iamyulong, line 527: How about `RECALL_VAULT`?

radix-engine-interface/src/api/object_api.rs
```diff
@@ -101,6 +101,13 @@ pub trait ClientObjectApi<E> {
     /// Get info regarding a visible object
     fn get_object_info(&mut self, node_id: &NodeId) -> Result<ObjectInfo, E>;
 
+    // TODO: Combine this with globalization process and/or find the right abstraction
```
iamyulong, line 104: I understand you have a `TODO` here, but I'm a bit concerned with the approach here, as it breaks the "access rules for global objects only" invariant.

It seems to me that we either need something like `auth => require(!vault.is_locked) && [other_constraints]` or we can do `if self.is_locked` check within withdraw methods.

iamyulong (approved):
Left some comments, but approving as it seems you wanted to address them in separate PR.

radix-engine-interface/src/api/object_api.rs
talekhinezh: I do understand this breaks that invariant but I think special casing for vaults here is the correct approach. We can discuss more as we do architecture refactors.

PR #1082, opened 2023-06-01
Bugfix: Fix non fungible vault summary

talekhinezh:
## Summary
Remove erroneous panic! from vault state_update_summary logic

## Details
The `panic!` was originally added as it was assumed that the vault index would use set logic. The current implementation uses more of a key-value map logic though which is why the panic! was being triggered. The fix is to just remove the panic. This should be re-added in the future if NonFungible vault logic would use better set-type logic.

## Testing
Added `add_and_remove_of_non_fungible_should_succeed` test which is the minimal case to cause the issue.


dhedey (approved):
Merging as already approved in #1086

PR #1085, opened 2023-06-05
Enable support for  WASM sign-ext opcodes

lrubasze:
## Summary

Enable support for `sign-extension` WASM opcodes.
Enable rust `stable` (unpin `1.69.0`).
Use `wasmparser` to validate WASM code.

## Details
Replace `wasmi_validation` with `wasmparser` since it didn't support `sign-extension` opcodes.

## Tests
Added WASM 'sign-extension' and 'mutable-global' features tests


radix-engine-tests/tests/wasm_non_mvp.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    assert!(WasmModule::init(&code).unwrap().contains_sign_ext_ops())
+}
+
+// Below tests verify WASM "mutable-global" feature, which allows importing/exporting mutable globals.
+// more details:
+// - https://github.com/WebAssembly/mutable-global/blob/master/proposals/mutable-global/Overview.md
+
+// NOTE!
+//  We test only WASM code, because Rust currently does not use the WASM "global" construct for globals
```
lrubasze, line 75: I spent a significant amount of time trying to test this feature in rust. But unfortunately Rust does not support it ATM. And yes, referenced issues are pretty old, but haven't found anything more up to date.

Taking above into account I am not really sure if testing it for WASM only makes sense. But we can always disable or remove it.

radix-engine/src/vm/wasm/wasmi.rs
```diff
@@ -1305,3 +1309,49 @@ impl WasmEngine for WasmiEngine {
         instance
     }
 }
+
+#[cfg(feature = "radix_engine_tests")]
+pub fn run_module_with_mutable_global(
```
lrubasze, line 1314: I know that this function makes no sense without `test_wasm_non_mvp_mutable_globals_execute_code()` from `radix-engine-tests/tests/wasm_non_mvp.rs`
But decided to put it here since it is easier to find some references when using `wasmi` module.

lrubasze: Moved the whole `test_wasm_non_mvp_mutable_globals_execute_code()` test to the `wasmi.rs`

radix-engine/src/vm/wasm/wasmi.rs
```diff
@@ -1305,3 +1309,132 @@ impl WasmEngine for WasmiEngine {
         instance
     }
 }
+
+// Below tests verify WASM "mutable-global" feature, which allows importing/exporting mutable globals.
```
lrubasze, line 1313: I spent a significant amount of time trying to test this feature in rust. But unfortunately Rust does not support it ATM. And yes, referenced issues are pretty old, but haven't found anything more up to date.

Taking above into account I am not really sure if testing it for WASM only makes sense. But we can always disable or remove it.

iamyulong (approved):

PR #1091, opened 2023-06-06
Use MetadataValue in genesis and actually set validator metadata

LukasGasior1:


radix-engine/src/system/bootstrap.rs
```diff
@@ -60,7 +60,10 @@ impl From<EcdsaSecp256k1PublicKey> for GenesisValidator {
             key,
             accept_delegated_stake: true,
             is_registered: true,
-            metadata: vec![],
+            metadata: vec![(
```
dhedey, line 63: This feels a little weird to include here.

It feels like perhaps this should just be done in particular test/s, rather than sticking this in the general "from" method?

LukasGasior1: well, this method is only really meant for tests

PR #1093, opened 2023-06-06
Feature: Instruction Schema Validation of well known blueprints

0xOmarA:


radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+) -> Result<Option<(LocalTypeIndex, &'s Schema<ScryptoCustomSchema>)>, ValidationError> {
+    let entity_type =
+        if let Some(entity_type) = invocation.global_address().as_node_id().entity_type() {
+            entity_type
+        } else {
+            return Err(ValidationError::InvalidAddress(
+                instruction_index,
+                invocation.global_address(),
+            ));
+        };
```
iamyulong, line 194: Minor: just to note that all addresses are entity-type-checked.

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+
+    let blueprint_schema = match invocation {
+        Invocation::Function(package_address @ PACKAGE_PACKAGE, ref blueprint, _) => {
```
iamyulong, line 197: `@` looks fancy!

transaction/examples/account/new.rtm
```diff
@@ -1,3 +1,4 @@
 CREATE_ACCOUNT_ADVANCED
-    Map<String, Tuple>();                  # Authority Rules
+    Enum<0u8>();
```
dhedey, line 2: In all of these `.rtm` files, can we instead use the string based values from `enums.rs` so people know what the type of this is?

transaction/src/manifest/e2e.rs
```diff
@@ -240,25 +240,25 @@ CALL_FUNCTION
             apply_address_replacements(
                 r##"
 CALL_METHOD
-    Address("${component_address}")
+    Address("component_sim1cqvgx33089ukm2pl97pv4max0x40ruvfy4lt60yvya744cvemygpmu")
```
dhedey, line 243: Why have we moved away from using `component_address` / `account_address` / `package_address` here? :)

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+            .map_err(|error| ValidationError::SchemaValidationError(index, format!("{:?}", error)))
+        } else {
+            Ok(())
+        }?;
+    }
+
+    Ok(())
+}
+
+macro_rules! define_lazy_package_definitions {
```
dhedey, line 108: Out of interest, why don't we use lazy static here?

It would be helpful to have this statically defined in the node :)

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+        match self {
+            Self::Method(..) => true,
+            Self::Function(..) => false,
+        }
+    }
+}
+
+// `usize` here is the index of the instruction has has an error.
+#[derive(Clone, Debug)]
+pub enum ValidationError {
```
dhedey, line 453: This is exposed as `radix_engine::utils::ValidationError` which doesn't feel like quite the right description.

Perhaps `ManifestSchemaValidationError`? Or else put this in `radix_engine::utils::schema::ValidationError` instead?

dhedey (approved):
Super nice!

Big thanks for fixing up all the manifests - it's a shame how much they managed to diverge when we didn't have these tests in place!

So happy to finally have all this come together! And I'm actually happily surprised how simple the native blueprint call validator code looks.

Have shared a few minor suggestions for some tweaks.

Relatedly, some minor thoughts on extensions - what are your thoughts on:
* We can re-use the blueprint lookup for formatting with RustLike syntax for a nicer instruction display (to eg show users what their access rules are)
* We can provide some kind of support for listing the non-native schemas we could use alongside the resolution result, and/or providing these non-native schemas directly.

radix-engine/src/utils/native_blueprint_call_validator.rs
0xOmarA: Can't say no to using fancy rust features :)

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+    }
+}
+
+fn get_blueprint_schema<'p>(
+    package_definition: &'p PackageDefinition,
+    package_address: PackageAddress,
+    blueprint: &str,
+    instruction_index: usize,
+) -> Result<&'p BlueprintSchema, ValidationError> {
+    package_definition.schema.blueprints.get(blueprint).map_or(
```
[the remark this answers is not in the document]
0xOmarA: Nice, did not know about that! Will start using that.

radix-engine/src/utils/native_blueprint_call_validator.rs
0xOmarA: Gotcha, so you're saying that unwrapping here is safe to do?

radix-engine/src/utils/native_blueprint_call_validator.rs
0xOmarA: It can definitely use lazy static here. My thought when I did this was that we would only be using it here, thus having the schemas as globals didn't make a lot of sesne. However, as you mentioned in your other comment the node wants to use this, and so does the toolkit. So, I will use lazy static here.

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+                        self.$field = Some($value);
+                        self.$field()
+                    }
+                }
+            )*
+        }
+    };
+}
+
+define_lazy_package_definitions! {
```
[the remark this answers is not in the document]
0xOmarA: Yup, will convert this to be lazy static and will make sure that they're public so the node and toolkit can use them.

radix-engine/src/utils/native_blueprint_call_validator.rs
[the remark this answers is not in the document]
0xOmarA: Sounds good! I've went ahead with names similar to the ones you have in your schema implementation. 

* `LocatedInstructionSchemaValidationError` for the error that includes the `cause` and the `instruction_index`.
* `InstructionSchemaValidationError` for the error enum

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+
+impl Invocation {
+    fn method(&self) -> &str {
+        match self {
+            Self::Method(_, _, method) => method,
+            Self::Function(_, _, method) => method,
+        }
+    }
+
+    fn global_address(&self) -> GlobalAddress {
```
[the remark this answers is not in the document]
0xOmarA: I've replaced that with an `entity_type` method that returns the `EntityType` of the object being invoked. How does that sound?

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+                .blueprints
+                .get(FUNGIBLE_VAULT_BLUEPRINT),
+            EntityType::InternalNonFungibleVault => package_definitions
+                .resource_package()
+                .schema
+                .blueprints
+                .get(NON_FUNGIBLE_VAULT_BLUEPRINT),
+
+            EntityType::GlobalGenericComponent
+            | EntityType::InternalGenericComponent
```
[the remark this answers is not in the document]
0xOmarA: I mainly have support for it in anticipation for Yulong's item from the "To Discuss" document that recommends that we add something along the lines of `Expression("LAST_OWNED")`.

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,460 @@
...
+                args,
+            } => (
+                Invocation::Method(
+                    *address,
+                    ObjectModuleId::AccessRules,
+                    method_name.to_owned(),
+                ),
+                args,
+            ),
+            _ => continue,
```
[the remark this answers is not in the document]
0xOmarA: Yup, I have added that and have added support for direct access methods.

radix-engine-tests/tests/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,243 @@
...
+
+fn account1() -> ComponentAddress {
+    ComponentAddress::virtual_account_from_public_key(&private_key1().public_key())
+}
+
+fn account2() -> ComponentAddress {
+    ComponentAddress::virtual_account_from_public_key(&private_key2().public_key())
+}
+
+fn apply_replacements(string: &str) -> String {
```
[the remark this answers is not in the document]
0xOmarA: Sounds good, will use the ones from `e2e` here to avoid duplication.

transaction/examples/call/call_method.rtm
```diff
@@ -6,15 +6,16 @@ CALL_METHOD
 
 CALL_ROYALTY_METHOD
     Address("${component_address}")
-    "some_method_1"
-    Decimal("1");
+    "set_royalty"
+    "my_method"
+    Enum<0u8>();
```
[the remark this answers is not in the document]
0xOmarA: Yup, sounds good.

transaction/examples/access_rule/access_rule.rtm
```diff
@@ -1,6 +1,5 @@
 UPDATE_ROLE
     Address("${resource_address}")
-    Enum<0u8>()
-    "auth"
-    Enum<1u8>(Enum<AccessRule::AllowAll>())
-    Enum<0u8>();
+    "hello"
```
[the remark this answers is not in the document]
0xOmarA: Sure thing, will add that.

transaction/examples/account/new.rtm
0xOmarA: Sounds good, will update that.

transaction/src/manifest/e2e.rs
0xOmarA: My bad, switching back.

radix-engine-tests/tests/native_blueprint_call_validator.rs
```diff
@@ -0,0 +1,243 @@
...
+    let manifest = ManifestBuilder::new()
+        .withdraw_from_account(account1(), RADIX_TOKEN, dec!("10"))
+        .try_deposit_batch_or_abort(account2())
+        .build();
+
+    // Act
+    let validation_result = validate_call_arguments_to_native_components(&manifest.instructions);
+
+    // Assert
+    assert!(validation_result.is_ok())
```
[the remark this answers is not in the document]
0xOmarA: Sounds good, We now use `.except(message)` here.

0xOmarA:
Thanks @iamyulong and @dhedey for the review! I've addressed and comments and updated the PR.

If we later on decide that we wish to add this to the `NotarizedTransactionValidator` it should be a very simple addition to add. For now, this will be used in the simulator and the toolkit.

> We can re-use the blueprint lookup for formatting with RustLike syntax for a nicer instruction display (to eg show users what their access rules are)

Certainly! I think we will first need to do your second point of allowing the caller to provide the schema of non-native components and packages. With that, we can format all instructions in a nice rust-like way.

> We can provide some kind of support for listing the non-native schemas we could use alongside the resolution result, and/or providing these non-native schemas directly.

This is definitely doable, in fact I imagine that it would be easy to do. The validation function (whatever the name we pick ends up being) could take an additional argument which is a `BTreeMap<NodeId, &'p PackageDefinition>` we can then use that package definition when we're trying to lookup the associated schema depending on whether it's a method or function call.

PR #1116, opened 2023-06-15
Add fee sanity checks and fix TestRunner

iamyulong:
## Summary
- Added sanity check in fee distribution
- Fixed execution configs in TestRunner


radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -58,7 +54,7 @@ pub struct ExecutionConfig {
     pub max_invoke_input_size: usize,
 }
 
-impl Default for ExecutionConfig {
+impl ExecutionConfig {
     fn default() -> Self {
```
talekhinezh, line 58: Can we add comment on why we don't use a trait for this

talekhinezh (approved):

PR #1119, opened 2023-06-16
Add transaction replay protection in engine

iamyulong:
## Summary

This PR implements intent-hash based replay protection in Radix Engine.

## Testing

New tests added

### For Internal Integrators

The following changes might have an integration impact:
- Added `TransactionTracker` node and substates;
- Updated `TransactionReceipt`, especially the `deleted_partitions` field;
- Updated `TransactionExecutor` interface;
- Fixed validator entity type id.


radix-engine/src/blueprints/intent_hash_store/package.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub start_partition: u8,
+
+    // parameters
+    pub partition_range_start: u8,
+    pub partition_range_end: u8,
+    pub epochs_per_partition: u64,
+}
+
+impl IntentHashStoreSubstate {
+    pub fn partition_of(&self, epoch: u64) -> Option<u8> {
```
dhedey, line 164: Can we add a rust doc explaining what this does?

I'd prefer call this `partition_for_expiry_epoch`

radix-engine/src/blueprints/intent_hash_store/package.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+#[derive(Debug, Clone, ScryptoSbor)]
+pub struct IntentHashStoreSubstate {
+    pub start_epoch: u64,
+    pub start_partition: u8,
+
+    // parameters
+    pub partition_range_start: u8,
+    pub partition_range_end: u8,
```
dhedey, line 159: It's surprising to me that an end index is inclusive, so I'd suggest  we rename this to include `_inclusive`?

radix-engine/src/blueprints/intent_hash_store/package.rs
```diff
@@ -0,0 +1,280 @@
...
+            partition_number -= num_partitions as u64;
+        }
+
+        assert!(partition_number >= self.partition_range_start as u64);
+        assert!(partition_number <= self.partition_range_end as u64);
+
+        Some(partition_number as u8)
+    }
+
+    pub fn advance(&mut self) -> u8 {
```
dhedey, line 186: This API is a bit confusing I think. To use it correctly leaks a lot of implementation details into (eg) the transaction executor.

I'd rather use a `handle_new_epoch(epoch: Epoch) -> EpochChangeHandling` which decides if it wants to advance this? With:
```
pub enum EpochChangeOutcome {
    None,
    DeleteFullyExpiredPartition(u8),
}
```

Or if we'd rather keep this, we should add a Rust doc explaining how to use this safely.

radix-engine/src/blueprints/intent_hash_store/package.rs
```diff
@@ -0,0 +1,280 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn calculate_coverage() {
+        let covered_days = EPOCHS_PER_PARTITION as f64
+            * (PARTITION_RANGE_END as f64 - (PARTITION_RANGE_START as f64 - 1.0) - 1.0)
```
dhedey, line 244: Why is this not (PARTITION_RANGE_END + 1 - PARTITION_RANGE_START)?

radix-engine/src/blueprints/intent_hash_store/package.rs
```diff
@@ -0,0 +1,280 @@
...
+            * 5.0
+            / 60.0
+            / 24.0;
+
+        assert_eq!(covered_days.floor() as usize, 65);
```
dhedey, line 249: Can we add an assertion that this is greater than the default epoch length from the transaction validator?

radix-engine/src/blueprints/intent_hash_store/package.rs
```diff
@@ -0,0 +1,280 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn calculate_coverage() {
+        let covered_days = EPOCHS_PER_PARTITION as f64
+            * (PARTITION_RANGE_END as f64 - (PARTITION_RANGE_START as f64 - 1.0) - 1.0)
+            * 5.0
```
dhedey, line 245: Can we maybe add a comment that this is 5 minutes / epoch?

radix-engine/src/blueprints/intent_hash_store/package.rs
dhedey: Maybe it would be clearer if it was a custom enum rather than option? EG:
```
pub enum ResolvedTransactionExpiryEpochPartition {
    Partition(u8),
    EpochNotInCurrentTrackableRange(Epoch),
}
```

dhedey, line 164: `epoch` should be of type `Epoch`, not `u64`.

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -243,97 +508,316 @@ where
...
+                );
+                track.release_lock(handle);
+            }
+        }
+
+        loop {
+            // TODO: how do we align this with TransactionValidator?
+            if substate
+                .partition_of(next_epoch.after(DEFAULT_MAX_EPOCH_RANGE).number())
+                .is_none()
```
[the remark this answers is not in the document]
dhedey: Ooo I think you're trying to keep a hold of the state for as long as possible?
Hmm I'd personally rather we keep things as simple as possible, and just delete the state as soon as it expires.

radix-engine-tests/tests/intent_hash_store.rs
```diff
@@ -0,0 +1,105 @@
...
+    receipt.expect_specific_rejection(|e| match e {
+        RejectionError::IntentHashCommitted => true,
+        _ => false,
+    });
+
+    // 3. Update the epoch
+    let new_epoch = current_epoch.after(EPOCHS_PER_PARTITION * 190);
+    test_runner.set_current_epoch(new_epoch);
+
+    // 4. Run another transaction
```
dhedey, line 52: I'm not sure this part of the test is too useful - I'd hope building a new transaction would not be expired :) - (but perhaps changing the nonce might be a better test)?

I think we should instead ensure that the replay prevention is "complete", ie that we:
* Wait to the `end_epoch_exclusive - 1`, submit the same transaction and check that we still get `RejectionError::IntentHashCommitted` 
* And then wait another epoch to `end_epoch_exclusive`, and then submit the same transaction again here, and ensure we get a different rejection (transaction expired)

This then ensures the cover is complete.

dhedey: That way the test is asserting that the transaction can't be committed again.

dhedey (changes requested):
Nice stuff! A few comments, some that I think need fixing before merge

radix-engine-common/src/native_addresses.rs
```diff
@@ -186,6 +186,12 @@ pub const FAUCET_PACKAGE: PackageAddress = PackageAddress::new_or_panic([
 /// The name of the faucet blueprint under the `FAUCET_PACKAGE`.
 pub const FAUCET_BLUEPRINT: &str = "Faucet";
 
+pub const INTENT_HASH_STORE_PACKAGE: PackageAddress = PackageAddress::new_or_panic([
```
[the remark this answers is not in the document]
iamyulong: Will use `TRANSACTION_TRACKER`.

radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -79,12 +80,20 @@ impl TransactionResult {
...
     pub outcome: TransactionOutcome,
     pub fee_summary: FeeSummary,
     pub fee_payments: IndexMap<NodeId, Decimal>,
     pub application_events: Vec<(EventTypeIdentifier, Vec<u8>)>,
     pub application_logs: Vec<(Level, String)>,
+    /// Metrics gathered during transaction execution.
+    pub execution_metrics: ExecutionMetrics,
+    /// Optional, only when `EnabledModule::ExecutionTrace` is ON.
+    /// Mainly for transaction preview.
+    pub execution_trace: TransactionExecutionTrace,
```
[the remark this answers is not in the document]
iamyulong: Yep, we can definitely improve this a bit more post-launch for rejections.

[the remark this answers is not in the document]
iamyulong: Yep, we can definitely improve this a bit more post-launch for rejections.

transaction/src/model/v1/preview_transaction.rs
```diff
@@ -54,24 +51,21 @@ impl ValidatedPreviewIntent {
             &intent.instructions.references,
             &intent.blobs.blobs_by_hash,
             ExecutionContext {
-                transaction_hash: intent_hash.into_hash(),
+                intent_hash: intent_hash.into_hash(),
```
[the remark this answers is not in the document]
iamyulong: Right. Given how intent hash is coupled with our transaction model, I was leaning towards no flexibility for now (for a simpler  system).

transaction/src/model/v1/preview_transaction.rs
```diff
@@ -54,24 +51,21 @@ impl ValidatedPreviewIntent {
...
+                epoch_range: if flags.skip_epoch_check {
+                    None
```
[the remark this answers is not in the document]
iamyulong: > It None is provided, I guess we can't actually read from / write to the hash store... which means the preview fee doesn't include this, meaning it's not accurate?

In the current iteration, both pre-transaction intent hash check and pos-transaction intent hash store update are `FREE`. 

It applies to all transactions, so I view it as accurate even from preview's standpoint.

radix-engine-common/src/native_addresses.rs
```diff
@@ -343,6 +355,11 @@ mod tests {
             EntityType::GlobalPackage,
             "package_rdx1pkgxxxxxxxxxfaucetxxxxxxxxx000034355863xxxxxxxxxfaucet",
         );
+        check_address(
+            INTENT_HASH_STORE_PACKAGE.as_ref(),
+            EntityType::GlobalPackage,
+            "package_rdx1p4ghn2tevauz68zu0xmc0apv5la8zqevtztzrle3r93wwzmwak7rzg",
```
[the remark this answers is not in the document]
iamyulong: Will do

radix-engine/src/errors.rs
```diff
@@ -52,6 +52,9 @@ pub enum RejectionError {
         valid_until: Epoch,
         current_epoch: Epoch,
     },
+    IntentHashExpiryEpochOutOfRange,
```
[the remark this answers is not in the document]
iamyulong: Yes, the expiry epoch is way into the future.

I added this error, thinking that we can potentially decouple the `epoch_range` from the epoch number used for transaction status tracking.

But for now, `end_epoch == expiry_epoch`, so I'm happy to make it panic.

radix-engine/src/errors.rs
```diff
@@ -52,6 +52,9 @@ pub enum RejectionError {
...
+    IntentHashCommitted,
```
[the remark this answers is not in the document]
iamyulong: Will rename

radix-engine/src/errors.rs
```diff
@@ -52,6 +52,9 @@ pub enum RejectionError {
...
+    IntentHashCancelled,
```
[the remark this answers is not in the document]
iamyulong: Will rename

transaction/src/model/executable.rs
```diff
@@ -1,20 +1,25 @@
...
+pub struct EpochRange {
+    pub start_epoch_inclusive: Epoch,
+    pub end_epoch_exclusive: Epoch,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
 pub struct ExecutionContext {
-    pub transaction_hash: Hash,
+    pub intent_hash: Hash,
+    pub epoch_range: Option<EpochRange>,
```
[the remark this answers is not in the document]
iamyulong: How about I change the interface to `intent_hash: (Hash, Epoch)`?

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -177,33 +181,295 @@ where
...
+                LockFlags::read_only(),
+            )
+            .unwrap()
+            .0;
+        let substate: IntentHashStoreSubstate = track.read_substate(handle).0.as_typed().unwrap();
+        track.release_lock(handle);
+
+        let partition_number = substate
+            .partition_of(expiry_epoch.number())
+            .ok_or(RejectionError::IntentHashExpiryEpochOutOfRange)?;
```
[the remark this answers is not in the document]
iamyulong: Will act based on resolution to comments above.

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -243,97 +508,316 @@ where
...
+                    .0;
+                track.update_substate(
+                    handle,
+                    IndexedScryptoValue::from_typed(&KeyValueEntrySubstate {
+                        value: Some(if is_success {
+                            IntentHashStatus::CommittedSuccess
+                        } else {
+                            IntentHashStatus::CommittedFailure
+                        }),
+                        // TODO: maybe make it immutable, but how does this affect partition deletion?
```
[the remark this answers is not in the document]
iamyulong: Yeah, not going to worry about it for now lol

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -243,97 +508,316 @@ where
...
+                    }),
+                );
+                track.release_lock(handle);
+            }
+        }
+
+        loop {
+            // TODO: how do we align this with TransactionValidator?
+            if substate
+                .partition_of(next_epoch.after(DEFAULT_MAX_EPOCH_RANGE).number())
```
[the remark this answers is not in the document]
iamyulong: > partitions are stored by expiry epoch

Yes, and because of this, I think it should be `next_epoch + max_support_epoch_range_in_header`?

radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -79,12 +80,20 @@ impl TransactionResult {
 #[derive(Debug, Clone, ScryptoSbor)]
 pub struct CommitResult {
     pub state_updates: StateUpdates,
+    /// Unstable, for intent hash intent only; Must be applied after committing the state updates above.
+    /// TODO: if time allows, consider merging into tracked nodes.
+    pub partition_deletions: IndexSet<DbPartitionKey>,
```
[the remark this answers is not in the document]
iamyulong: Was worried about impact on integration interfaces. Guess merged struct will help force integrator to update lol.

Will merge it into `StateUpdates`.

radix-engine-common/src/native_addresses.rs
```diff
@@ -362,6 +379,11 @@ mod tests {
             EntityType::GlobalGenericComponent,
             "component_rdx1cptxxxxxxxxxgenssxxxxxxxxxx000977302539xxxxxxxxxgenssx",
         );
+        check_address(
+            INTENT_HASH_STORE.as_ref(),
+            EntityType::GlobalGenericComponent,
+            "component_rdx1crljck99j52j4dc230ecvj762s89wfqfyg0amaefnst6j0g7tgfnm6",
```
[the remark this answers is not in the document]
iamyulong: Will do.

radix-engine/src/track/track.rs
```diff
@@ -951,13 +981,16 @@ impl<'s, S: SubstateDatabase, M: DatabaseKeyMapper> SubstateStore for Track<'s,
...
         }
 
-        let substate = tracked
-            .get_runtime_substate_mut()
-            .ok_or(AcquireLockError::NotFound(
-                *node_id,
-                partition_num,
-                substate_key.clone(),
-            ))?;
+        let substate = match tracked.get_runtime_substate_mut() {
```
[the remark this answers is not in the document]
iamyulong: I like `ok_or`. The problem is it's not breakpoint friendly, which made it harder to debug issues like the one in `Track`.

radix-engine/src/blueprints/intent_hash_store/package.rs
```diff
@@ -0,0 +1,280 @@
...
+                    events: BlueprintEventSchemaInit::default(),
+                    functions: BlueprintFunctionsSchemaInit {
+                        virtual_lazy_load_functions: btreemap!(),
+                        functions,
+                    },
+                },
+
+                royalty_config: RoyaltyConfig::default(),
+                auth_config: AuthConfig {
+                    function_auth: btreemap!(
```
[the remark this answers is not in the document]
iamyulong: I thought it would cause no harm by allowing Scrypto to call the constructor.

But maybe better to be more conservative.

Will add

radix-engine/src/blueprints/intent_hash_store/package.rs
iamyulong: Happy to rename, but don't feel necessity to add further abstraction.

Will just add some comments to the interface.

radix-engine/src/blueprints/intent_hash_store/package.rs
iamyulong: Will rename.

radix-engine/src/blueprints/intent_hash_store/package.rs
iamyulong: Because there is one partition that's updated actively and does not contain the full `EPOCHS_PER_PARTITION` number of hashes.

radix-engine/src/blueprints/intent_hash_store/package.rs
iamyulong: Yes, will add!

radix-engine/src/blueprints/intent_hash_store/package.rs
iamyulong: Yes, will add!

radix-engine/src/blueprints/intent_hash_store/package.rs
iamyulong: Found it's much easier to use `u64` for math.

radix-engine/src/transaction/transaction_executor.rs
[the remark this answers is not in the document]
iamyulong: > I think you're trying to keep a hold of the state for as long as possible?

Partially correct. The main invariant is that the tracked intent hash range is greater than the max header epoch range.

>  just delete the state as soon as it expires.

This isn't as simple as it sounds. It would require scanning of a partition and delete individuals manually.

radix-engine/src/blueprints/intent_hash_store/package.rs
iamyulong: Haha, will just add comments. 

The motivation here is primarily to follow [ring buffer design](https://en.wikipedia.org/wiki/Circular_buffer), and not coupled with epoch update handling.

Transaction executor is already full of logic lol.

radix-engine-tests/tests/intent_hash_store.rs
iamyulong: Will see if I can add a test for the described scenario.

transaction/src/model/v1/preview_transaction.rs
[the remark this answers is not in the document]
dhedey: Oh okay, makes sense! So the cost needs to be factored into the base cost of a transaction?

radix-engine-common/src/native_addresses.rs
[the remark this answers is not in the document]
dhedey: Have updated the vanity addresses REP.

The component address is now"
```
transactiontracker_rdx1stxxxxxxxxxxtxtrakxxxxxxxxx006844685494xxxxxxxxxtxtrak
[130, 204, 99, 24, 198, 49, 140, 101, 153, 99, 237, 140, 99, 24, 198, 49, 140, 247, 232, 245, 174, 143, 74, 150, 166, 49, 140, 99, 24, 198]
```

radix-engine/src/transaction/transaction_executor.rs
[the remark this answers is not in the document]
dhedey: The transaction tracker stores hashes by their expiry epoch exclusive, right? 

So if, when we advance an epoch to epoch N, and this causes the transaction tracker to advance the minimum partition it tracks from P to P + 1, then P must have handled expiry epochs M to N inclusive (for some M < N), and therefore must now be fully expired (ie all the epochs that map into it are in the past) - so that partition can be deleted.

radix-engine/src/blueprints/intent_hash_store/package.rs
dhedey: It should also take an `Epoch` rather than a u64 :)

radix-engine/src/blueprints/intent_hash_store/package.rs
dhedey: Hmm but I'd rather the deletion logic was better defined by the protocol and didn't depend on the transaction validator logic, or require loops in the clear up process.

transaction/src/model/executable.rs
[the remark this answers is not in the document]
dhedey: Sounds good.

Perhaps we could be even more explicit, eg:
``` intent_hash: TransactionIntentHash```
and:
```
pub enum TransactionIntentHash {
    /// This hash will be checked for cancellation/replay
    User { intent_hash: Hash, expiry_epoch_exclusive: Epoch }
    /// Doesn't do any transaction tracking / replay prevention - the node is expected to ensure
    /// each system transaction has a unique hash
    System { intent_hash: Hash }
}
```

File: radix-engine/src/transaction/transaction_executor.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (dhedey) [reply radix-engine/src/transaction/transaction_executor.rs:787]: I'd personally rather the deletion logic cleans up sensibly/deterministically so we store minimal additional state, and can explain it more easily.

**Comment by iamyulong** [comment]
TODOs:
- Make `IntentHashExpiryEpochOutOfRange` panic
- Add enum
  ```
  pub enum TransactionIntentHash {
    /// This hash will be checked for cancellation/replay
    User { intent_hash: Hash, expiry_epoch_exclusive: Epoch }
    /// Doesn't do any transaction tracking / replay prevention - the node is expected to ensure
    /// each system transaction has a unique hash
    System { intent_hash: Hash }
  }
  ```
- Add function auth
- Add more test scenarios
- Merge `state_updates` and `deleted_partitions`

File: radix-engine-common/src/types/entity_type.rs [hunk 25]
```diff
@@ -24,7 +24,7 @@ pub enum EntityType {
     GlobalConsensusManager = 0b10000110, //-------------- 10000 => s, 110xx => ce6m [se vanity prefix]
 
     /// A global validator entity (130 in decimal). Gives Bech32 prefix: `s` followed by one of `v`, `d`, `w` or `0`.
-    GlobalValidator = 0b10000010, //----------------- 10000 => s, 110xx => vdw0
+    GlobalValidator = 0b10000011, //--------------------- 10000 => s, 110xx => vdw0
```
lrubasze, line 27: Comment no longer valid
_131 in decimal_
_011xx => vdw0_

iamyulong: Nice spot. will update!

radix-engine-common/src/types/entity_type.rs
```diff
@@ -42,13 +42,16 @@ pub enum EntityType {
...
 
     /// A global native pool entity (197 in decimal). Gives Bech32 prefix: `c` followed by one of `5`, `4`, `k` or `h`.
-    GlobalTwoResourcePool = 0b11000101, //---------- 11000 => c, 101xx => 54kh (101 = pool)
+    GlobalTwoResourcePool = 0b11000101, //----------- 11000 => c, 101xx => 54kh (101 = pool)
 
     /// A global native pool entity (197 in decimal). Gives Bech32 prefix: `c` followed by one of `c`, `e`, `6` or `m`.
-    GlobalMultiResourcePool = 0b11000110, //---------- 11000 => c, 110xx => ce6m (101 = pool)
+    GlobalMultiResourcePool = 0b11000110, //--------- 11000 => c, 110xx => ce6m (101 = pool)
+
+    /// A global transaction tracker (130 in decimal). Gives Bech32 prefix: `s` followed by one of `g`, `f`, `2` or `t`.
```
dhedey, line 53: Very minor, but in terms of grouping, this should be under `System Components` :)

radix-engine-queries/src/typed_substate_layout.rs
```diff
@@ -123,6 +124,7 @@ pub enum TypedMainModuleSubstateKey {
     OneResourcePoolField(OneResourcePoolField),
     TwoResourcePoolField(TwoResourcePoolField),
     MultiResourcePoolField(MultiResourcePoolField),
+    TransactionTrackerField(TransactionTrackerField),
```
dhedey, line 127: We should also include the TransactionTrackerTrackedTransactionEntry here, otherwise we'll get errors when we try to ingest any user transactions.

radix-engine-tests/tests/metering.rs
```diff
@@ -200,11 +200,11 @@ fn test_radiswap() {
         + 11988 /* DropLock */
         + 3255 /* DropNode */
         + 3130196 /* Invoke */
-        + 2279411 /* LockSubstate */
+        + 2319706 /* LockSubstate */
         + 18200 /* ReadSubstate */
         + 125000 /* RunNative */
         + 20000 /* RunSystem */
-        + 602635 /* RunWasm */
+        + 613895 /* RunWasm */
```
dhedey, line 207: It's impressive how much this can differ depending on where the WASM was last compiled

radix-engine-tests/tests/intent_hash_store.rs
```diff
@@ -0,0 +1,105 @@
...
+    NotarizedTransactionV1, TransactionHeaderV1, TransactionPayload,
+    ValidatedNotarizedTransactionV1,
+};
+use transaction::signing::secp256k1::Secp256k1PrivateKey;
+use transaction::validation::{
+    NotarizedTransactionValidator, TransactionValidator, ValidationConfig,
+};
+
+#[test]
+fn test_transaction_replay_protection() {
```
dhedey, line 19: This file could probably do with a rename :)

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -726,38 +720,43 @@ where
...
 
         // Update the status of the intent hash
-        if let Some(expiry_epoch) = expiry_epoch {
-            if let Some(partition_number) = substate.partition_of(expiry_epoch.number()) {
+        if let TransactionIntentHash::ToCheck {
+            expiry_epoch,
+            intent_hash,
+        } = intent_hash
+        {
+            if let Some(partition_number) =
```
dhedey, line 748: If this isn't true, (eg due to epoch jumping forward, or going backward) we should panic IMO - because it's an invariant break? We should protect against replay not being tracked.

(And for this, I quite like the `let .. else .. ` construct)

dhedey (approved):
Really nice. I've noticed a couple of other small issues (including two which I think we must fix - in the executor and in the typed substate queries).

But happy for you to merge once they're done

PR #1121, opened 2023-06-16
Feature/metadata and royalty roles

talekhinezh:
## Summary
Move role definition to per module rather than per instance.

## Details
* Add discrete `OwnerRole` substate to AccessRules
* Update AccessRules to store rules by ObjectModuleId
* Unified `globalize()` and `globalize_with_address()` api to just `globalize()`
* Defined Metadata setter and Royalty owner roles

### For dApp Developers
Roles are now assigned by module rather than per instance

### For Internal Integrators
Access Rule Substates have been refactored to account for the new model.


iamyulong (approved):
LGTM

PR #1125, opened 2023-06-18
feat: Implemented transaction messages

dhedey:
## Summary
Implements plaintext and encrypted message data models inside user transactions, as per REP-70.

## Details
Ideally we'd also provide an example implementation and test vectors for transaction messages, but these can come later.

## Testing
Added transaction validation tests.

### For Internal Integrators

The encoding of Transaction attachments has changed, which will have some impact on transaction encoding (for eg Ledger implementation).

* The toolkit will need to allow for the user to provide the message data model for transaction building
* The ledger:
  * Will need to support the new attachments format for decoding (I think it might get this for free)
  * May wish to display the message if it is string and plaintext


transaction/src/model/v1/attachments.rs
```diff
@@ -1,7 +1,177 @@
...
+///   known as ASN1 X9.63 variant of ECDH. Be careful - libsecp256k1 uses another non-standard variant.
+/// - We persist 128-bit symmetric keys because we wish to save on payload size, and:
+///   * 128-bit AES is considered secure enough for most use cases (EG bitcoin hash rate is only 2^93 / year)
+///   * It's being used with a transient key - so a hypothetical successful attack would only decrypt one message
+#[derive(Debug, Clone, PartialEq, Eq, ManifestSbor)]
+pub struct EncryptedMessageV1 {
+    pub encrypted: AesGcmPayload,
+    // Note we use a collection here rather than a struct to be forward-compatible to adding more curve types.
+    // The engine should validate each DecryptorsByCurve matches the CurveType.
+    pub decryptors_by_curve: BTreeMap<CurveType, DecryptorsByCurve>,
```
iamyulong, line 82: How about?
```rust
decryptors: Vec<Decryptors>,

pub enum Decryptors {
    dh_ephemeral_public_key: PublicKey, // This is an enum
    decryptors: BTreeMap<PublicKeyFingerprint, AesWrapped128BitKey>,
}
```

`BTreeMap` breaks round-trip property.

transaction/src/model/v1/attachments.rs
```diff
@@ -1,7 +1,177 @@
 use super::*;
 use crate::prelude::*;
 
+#[derive(Debug, Clone, Eq, PartialEq, ManifestSbor, Default)]
+#[sbor(transparent)]
+pub struct AttachmentsV1 {
+    pub message: MessageV1,
+}
```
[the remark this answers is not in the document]
dhedey: Based on discussion, will remove the `AttachmentsV1` wrapper and just make it `MessageV1`

transaction/src/model/v1/attachments.rs
dhedey: Will change to `IndexMap` to get the round-trip property back.

iamyulong (approved):
Discussed
- Will merge `AttachmentV1` and `MessageV1`
- Will change `decryptors_by_curve` to use `IndexMap`
- Will not update MIME type for now

PR #1126, opened 2023-06-19
RUID

iamyulong:
## Summary

This PR introduces Radix Unique IDentifier (RUID), a unique 32-byte identifier derived from transaction intent hash. 

Format:
```
{7b003d8e0b2c9e3a-516cf99882de64a1-f1cd6742ce3299e0-357f54f0333d25d0}
```

## Details

- Removed `Runtime::generate_uuid` from Scrypto
- Added 32-byte RUID implementation
- Updated `NonFungibleLocalId` to use RUID


radix-engine-common/src/crypto/public_key_ed25519.rs
```diff
@@ -47,7 +47,7 @@ impl TryFrom<&[u8]> for Ed25519PublicKey {
 #[cfg_attr(feature = "radix_engine_fuzzing", derive(Arbitrary))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Sbor)]
 #[sbor(transparent)]
-pub struct Ed25519PublicKeyHash(pub [u8; NodeId::UUID_LENGTH]);
+pub struct Ed25519PublicKeyHash(pub [u8; NodeId::RANDOM_LENGTH]);
```
talekhinezh, line 50: Can we use a different name than `RANDOM_LENGTH`? Sounds like we're choosing a random length.

talekhinezh (approved):

PR #1136, opened 2023-06-21
Bugfix/access to tx processor

talekhinezh:
## Summary
Disallow access to Transaction Processor `run` function

## Details
* Added `RootOnly` function auth
* Only allow root call frame to use initial virtual badges
* Fix multiple `extern_blueprint!` macro use in Scrypto

### For Internal Integrators
Updated Function Auth to include `RootOnly` option


radix-engine-tests/tests/blueprints/tx_processor_access/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+    None,
+}
+
+#[blueprint]
+mod execute_manifest {
+    extern_blueprint!(
+        "package_rdx1pkgxxxxxxxxxtxnpxrxxxxxxxxx002962227406xxxxxxxxxtxnpxr",
+        TransactionProcessor {
+            fn run(
+                transaction_hash: Hash,
```
iamyulong, line 15: I thought you intended to remove this, no?

iamyulong (commented):
Can we update the PR description?

iamyulong (approved):
Approving, assuming the comments will be addressed.

PR #1137, opened 2023-06-21
Test kernel lock_substate

iamyulong:
## Summary

I've checked `kernel.rs` and the hack logic (replacing "any" module id with "main") has been removed.

Added a test to verify the kernel behaviour when package substate is not found.


radix-engine/src/kernel/kernel.rs
```diff
@@ -37,6 +37,17 @@ pub struct KernelBoot<'g, V: SystemCallbackObject, S: SubstateStore> {
 }
 
 impl<'g, 'h, V: SystemCallbackObject, S: SubstateStore> KernelBoot<'g, V, S> {
+    pub fn start_test_kernel(&mut self) -> Kernel<SystemConfig<V>, S> {
```
talekhinezh, line 40: Can we add a comment on what this is for, since this should only be used by testing

talekhinezh (approved):

PR #1139, opened 2023-06-21
Add fee locks to execution trace

0xOmarA:
## Summary

This PR adds the amount of fee locks and contingent fee locks to the transaction execution trace. This is needed by the RET and wallet for accurate fee calculation.


iamyulong (approved):
LGTM

PR #1141, opened 2023-06-21
Feature: Transaction Hashes Bech32m

0xOmarA:
## Summary

This PR implements another set of Bech32m encoders and decoders for the transaction hash types `IntentHash`, `SignedIntentHash`, and `NotarizedTransactionHash` that allows these types to be represented as Bech32m encoded strings.

With this, the PR renamed the existing `Bech32Encoder` and `Bech32Decoder` classes to be prefixed with `Address` to make it clear that they're used for addresses. 

A big chunk of this PR is just renames, all that you need to review is the `transaction/src/models/hash` module


radix-engine-common/src/address/hrpset.rs
```diff
@@ -5,20 +5,26 @@ use sbor::rust::prelude::*;
...
+    pub transaction_tracker: String,
+    pub internal_vault: String,
+    pub internal_account: String,
+    pub internal_component: String,
+    pub internal_key_value_store: String,
+
+    /* Transaction Parts */
+    pub transaction_intent: String,
+    pub signed_transaction_intent: String,
+    pub notarized_transaction: String,
```
dhedey, line 27: So we also have quite a few different transaction hashes in the node (ledger, round update, genesis).

So I wonder if (to be extensible) we can instead put these (network-independent) prefixes on the `HashHasHrp`? (maybe call it `TransactionHash`)? And then combine them with the network suffix inside the TransactionBech32Encoder/Decoder?

transaction/src/model/versioned.rs
```diff
@@ -161,7 +161,7 @@ mod tests {
         let intent_hash = prepared_intent.intent_hash();
 
         assert_eq!(
-            intent_hash.to_string(),
+            intent_hash.to_string(None),
```
dhedey, line 164: Might be interesting to also output the transaction ids eg `intent_hash.to_string(NetworkDefinition::simulator())` here :) - as this file is basically a demonstration of the various transactions.

dhedey (approved):
Nice 👍

radix-engine-common/src/address/hrpset.rs
0xOmarA: Why not extend the HRP set to include the other transaction types such as ledger transactions, round update, genesis, etc... and then implement the `HashHasHrp` trait in the node?

transaction/src/model/versioned.rs
0xOmarA: Sounds good, will change that.

PR #1144, opened 2023-06-22
Tweak/finalize epoch constants

dhedey:
## Summary
Three parts:
* Fix `DEFAULT_MAX_EPOCH_RANGE` for a transaction to be ~1 month
* Tweak the consensus manager epoch change logic to not be prone to systematic drift - this means we can avoid handling / second-guessing this in our constants on the node side.
* Ensure the proposer timestamp is non-decreasing. This formalizes an invariant that an honest validator set is required to maintain.

### DIFF
--- a/radix-engine-interface/src/blueprints/consensus_manager/invocations.rs
+++ b/radix-engine-interface/src/blueprints/consensus_manager/invocations.rs
@@ -122,6 +122,10 @@ impl EpochChangeCondition {
         round: Round,
     ) -> EpochChangeOutcome {
         let epoch_duration_millis =
+            // The application invariants in `check_non_decreasing_and_update_timestamps`
+            // ensures that current_time > effective_start, and genesis should ensure
+            // effective_start > 0.
+            // This is just a sanity-check to avoid overflow if something invaraint fails.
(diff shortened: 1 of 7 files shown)


radix-engine-interface/src/blueprints/consensus_manager/invocations.rs
```diff
@@ -105,16 +105,67 @@ pub struct EpochChangeCondition {
...
+}
+
 impl EpochChangeCondition {
     /// Determines whether this condition is met by the given actual state.
     /// See the condition's field definitions for exact rules.
-    pub fn is_met(&self, duration_millis: i64, round: Round) -> bool {
+    pub fn should_epoch_change(
+        &self,
+        effective_start: i64,
+        current_time: i64,
```
iamyulong, line 121: What's the behaviour if a proposer sets `current_time` to a negative value?

iamyulong (approved):

radix-engine-interface/src/blueprints/consensus_manager/invocations.rs
dhedey: Have added a new assertion and test case to ensure that the current time is non-decreasing.

PR #1166, opened 2023-06-24
Feature: Metadata Limits

talekhinezh:
## Summary
Add Metadata Key and value limits (Currently set to 100 string length size for key and 512 for sbor value size)


radix-engine-constants/src/lib.rs
```diff
@@ -114,3 +114,7 @@ pub const DEFAULT_MAX_LOG_SIZE: usize = 64 * 1024;
 pub const DEFAULT_MAX_PANIC_MESSAGE_SIZE: usize = 64 * 1024;
 pub const DEFAULT_MAX_NUMBER_OF_EVENTS: usize = 256;
 pub const DEFAULT_MAX_NUMBER_OF_LOGS: usize = 256;
+
+/* Metadata */
+pub const DEFAULT_MAX_METADATA_KEY_STRING_LEN: usize = 100;
+pub const DEFAULT_MAX_METADATA_VALUE_SBOR_LEN: usize = 512;
```
iamyulong, line 120: Minor: we may need to adjust this number later.

iamyulong (approved):

PR #1167, opened 2023-06-24
Release/damson

dhedey:
## Summary
Various fixes I had to put into the the `release/damson` branch so it could be merged into the node.

## Details

I now also run all transaction scenarios through the typed substate checker - which should make this more fool-proof if we actually add scenarios as we go (which should be really easy to do - I encourage you all to do it!)


dhedey:
Closing in favour of #1173

PR #1191, opened 2023-06-28
Another batch of costing adjustment

iamyulong:
## Summary

- Added substate expansion cost (to limit state bloat; different from execution, tipping and royalty costs)
- Removed substate read/write limiting (not effective, as they’re not database metrics)
- Added `max_number_of_substates_in_track` limit
- Refactored database read tracing in Track
- Moved database write costing to the commit phase
- Separated `total_tipping_cost_xrd` from execution cost in transaction receipt
- Adjusted cost unit price to increase fees while maintaining constraints


radix-engine-tests/benches/radiswap.rs
```diff
@@ -36,7 +36,7 @@ fn bench_radiswap(c: &mut Criterion) {
     let component_address: ComponentAddress = test_runner
         .execute_manifest(
             ManifestBuilder::new()
-                .lock_fee(account2, 50u32.into())
+                .lock_fee(account2, 500u32.into())
```
mstrug-rdx, line 39: Maybe move this value to a constant, so future changes will be easier.

iamyulong: That's a good idea. If I have to change it one more time, will make it a constant :)

mstrug-rdx (approved):

PR #1194, opened 2023-06-29
Bugfix: Fix genesis staking

talekhinezh:
## Summary
Fix genesis staking when a validator is set to not accept delegated stake.

## Details
* Exposed a method to retrieve the current delegated stake flag on a validator
* Use the delegated stake flag method to save, turn on, restore delegated stake in genesis


iamyulong (approved):
Looks good.

Note that auth is turned off for genesis transaction. This is why we can change the state of a validator in GenesisHelper.

PR #1195, opened 2023-06-29
Feature/tweaks for scenario running in node

dhedey:
## Summary

The change is to move the faucet creation into the bootstrap transaction - this enables us to inject scenarios requiring the faucet into the genesis epoch, after the genesis data chunks, before we finalize with the wrap-up transaction.


LukasGasior1 (approved):
LGTM

PR #1201, opened 2023-06-29
Make prices unrealistic before ready

iamyulong:
## Summary

Per team discussion, this PR makes prices unrealistic before they're finalised.

- `COST_UNIT_PRICE_IN_XRD`: The price of a single execution cost unit, in XRD
- `STATE_EXPANSION_PRICE_IN_XRD`: The price of adding one byte to substate store, in XRD
- `USD_PRICE_IN_XRD`: The price of USD, in XRD
- `MAX_PER_FUNCTION_ROYALTY_IN_XRD`: The max amount of a function/method, in XRD

Plus some renamings.


radix-engine-tests/benches/costing.rs
```diff
@@ -74,22 +74,24 @@ fn bench_spin_loop(c: &mut Criterion) {
...
+            ));
             let mut instance = wasm_engine.instantiate(Hash([0u8; 32]), &instrumented_code);
             instance
                 .invoke_export("Test_f", vec![Buffer(0)], &mut runtime)
                 .unwrap();
         })
     });
 
-    println!("Gas consumed: {}", gas_consumed);
+    println!("Gas consumed: {}", wasm_execution_units_consumed);
```
dhedey, line 94: Very minor - perhaps this warrants a rename?

radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -138,27 +138,27 @@ impl FeeTable {
...
     }
 
     #[inline]
     pub fn run_wasm_code_cost(
         &self,
         _package_address: &PackageAddress,
         _export_name: &str,
-        gas: u32,
+        wasm_execution_units: u32,
     ) -> u32 {
```
dhedey, line 157: Not a suggestion for now, but perhaps in future, we might want a CostUnit new type, to save confusion :)

radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
```diff
@@ -371,10 +374,10 @@ where
             .consume_cost_units(ClientCostingEntry::RunWasmCode {
```
dhedey, line 374: This code took me longer than I care to admit to understand and validate. Could we update the comment to explain that we're taking:
```
/// If we need to request more from the fee reserve, we round `n` up to the nearest `1_000_000` and request
/// that much additional gas
let amount_to_request = ...
```

radix-engine/src/vm/wasm_runtime/no_op_runtime.rs
```diff
@@ -182,8 +185,11 @@ impl<'a> WasmRuntime for NoOpWasmRuntime<'a> {
         Err(InvokeError::SelfError(WasmRuntimeError::NotImplemented))
     }
 
-    fn consume_gas(&mut self, n: u32) -> Result<(), InvokeError<WasmRuntimeError>> {
-        self.gas_consumed.add_assign(n);
+    fn consume_wasm_execution_units(
+        &mut self,
+        n: u32,
+    ) -> Result<(), InvokeError<WasmRuntimeError>> {
+        self.wasm_execution_units_consumed.add_assign(n);
```
dhedey, line 192: We just saw a panic here in tests - perhaps we should catch / handle this as a fee reserve overflow?

dhedey (approved):

PR #1203, opened 2023-06-30
Add metadata scenarios

iamyulong:
## Summary

Add metadata transaction scenarios.


dhedey (approved):
Looks great!

But another quick thought:
* Could we add a "locked_on_create" metadata entry which starts locked?
  * We could also have a transaction which tries to update the locked metadata and fails - and have the scenario that asserts it fails.
* And "locked_later" metadata entry which is locked in a follow-up transaction?

This can be used by Gateway + Dashboard to check that the locking status also gets pulled through correctly 👍

PR #1204, opened 2023-06-30
Add fungible resource transaction scenarios

iamyulong:
## Summary

Adds fungible resource transaction scenarios, covering basic minting, burning, transferring, freezing, unfreezing and recalling.


transaction-scenarios/generated-examples/fungible_resource/1--fungible-max-div-create.rtm
```diff
@@ -0,0 +1,51 @@
+CALL_METHOD
```
dhedey, line 1: The ordering of these has a wonky order, maybe we should left-pad the numbers in the file names, eg `001--`?

transaction-scenarios/src/scenarios/fungible_resource.rs
```diff
@@ -0,0 +1,435 @@
...
+use radix_engine_interface::*;
+
+pub struct FungibleResourceScenarioConfig {
+    pub user_account_1: VirtualAccount,
+    pub user_account_2: VirtualAccount,
+}
+
+#[derive(Default)]
+pub struct FungibleResourceScenarioState {
+    pub max_divisibility_fungible_resource: Option<ResourceAddress>,
```
dhedey, line 14: The `State` wrapper was designed to be slightly more intuitive to get/set, but using `Option` is totally fine too

transaction-scenarios/src/scenarios/fungible_resource.rs
```diff
@@ -0,0 +1,435 @@
...
+                        vec![],
+                    )
+                },
+                |core, config, state, result| {
+                    state.max_divisibility_fungible_resource = Some(result.new_resource_addresses()[0]);
+                    state.vault1 = Some(result.new_vault_addresses()[0]);
+                    Ok(())
+                },
+            )
+            .successful_transaction_with_result_handler(
```
dhedey, line 80: You can just use `successful_transaction` to avoid the empty result handler, but I'm happy if you want to keep this for symmetry :)

transaction-scenarios/src/scenarios/fungible_resource.rs
```diff
@@ -0,0 +1,435 @@
...
+                    )
+                },
+                |core, config, state, result| {
+                    Ok(())
+                },
+            )
+            .successful_transaction_with_result_handler(
+                |core, config, state| {
+                    core.next_transaction_with_faucet_lock_fee(
+                        "fungible-max-div-freeze-deposit",
```
dhedey, line 166: freeze-burn *

transaction-scenarios/src/scenarios/fungible_resource.rs
```diff
@@ -0,0 +1,435 @@
...
+                    )
+                },
+                |core, config, state, result| {
+                    Ok(())
+                },
+            )
+            .successful_transaction_with_result_handler(
+                |core, config, state| {
+                    core.next_transaction_with_faucet_lock_fee(
+                        "fungible-max-div-unfreeze-deposit",
```
dhedey, line 219: `unfreeze-burn`*

transaction-scenarios/src/scenarios/fungible_resource.rs
```diff
@@ -0,0 +1,435 @@
...
+                    )
+                },
+                |core, config, state, result| {
+                    Ok(())
+                },
+            )
+            .successful_transaction_with_result_handler(
+                |core, config, state| {
+                    core.next_transaction_with_faucet_lock_fee(
+                        "fungible-max-div-recall-unfreezed-vault",
```
dhedey, line 231: unfrozen*

transaction-scenarios/src/scenarios/fungible_resource.rs
```diff
@@ -0,0 +1,435 @@
...
+                Ok(ScenarioOutput {
+                    interesting_addresses: DescribedAddresses::new()
+                    .add("user_account_1", config.user_account_1.address.clone())
+                    .add("user_account_2", config.user_account_2.address.clone())
+                    .add(
+                        "max_divisibility_fungible_resource",
+                        state.max_divisibility_fungible_resource.unwrap(),
+                    )
+                    .add(
+                        "min_divisibility_fungible_resource",
```
dhedey, line 429: Would be quite cool if we leave a vault frozen and we include the vault address here - for the dashboard team -- I guess we need some kind of `AnyAddress` model for that though.

dhedey (approved):
A couple of fixes, and a few optional suggestions. Merge when you're happy.

transaction-scenarios/src/scenarios/fungible_resource.rs
dhedey: TBH we could probably just have these Addresses String instead of an Address object, which would allow us to include EG vault addresses and eg NonFungibleGlobalIds :) - or maybe we could create an Enum:
```
pub enum DescribedAddress {
    Global(GlobalAddress),
    Internal(InternalAddress),
    NonFungible(NonFungibleGlobalId),
}
```

transaction-scenarios/generated-examples/fungible_resource/1--fungible-max-div-create.rtm
iamyulong: Good idea!

PR #1208, opened 2023-06-30
Drop parity-wasm dependency

lrubasze:
## Summary

This PR is about switching to forked `wasm-instrument` crate (see [here](https://github.com/radixdlt/wasm-instrument/pull/1)), which drops a dependency on no longer maintained `parity-wasm` crate and switches to `wasmparser`

## Details

- WASM deserialization is faster than with `parity-wasm`
- WASM instrumentation is slower (`costing::bench_prepare_wasm +30%`, `costing::validate_wasm +107%`). This is because of the wasmparser and wasm-encoder design, which require reencoding whole section when applying some change)
- forked `wasm-instrument` removes custom section (which is used to keep debug information, not needed for our use cases). It results in a slightly smaller code, which is good.


radix-engine/src/vm/wasm/prepare.rs
```diff
@@ -930,32 +789,35 @@ impl WasmModule {
     }
 
     pub fn enforce_global_limit(self, max_number_of_globals: u32) -> Result<Self, PrepareError> {
-        if let Some(section) = self.module.global_section() {
-            if section.entries().len() > max_number_of_globals as usize {
-                return Err(PrepareError::TooManyGlobals);
-            }
+        if self.module.num_local_globals() > max_number_of_globals {
```
iamyulong, line 792: Ditto

radix-engine/src/vm/wasm/prepare.rs
```diff
@@ -930,32 +789,35 @@ impl WasmModule {
...
         Ok(self)
     }
 
     pub fn enforce_export_constraints<'a, I: Iterator<Item = &'a BlueprintDefinitionInit>>(
         self,
         blueprints: I,
     ) -> Result<Self, PrepareError> {
+        if self.module.exports_count == 0 {
+            return Err(PrepareError::NoExportSection);
+        }
```
iamyulong, line 805: Why do we need to enforce this?

radix-engine/src/vm/wasm/wasm_validator_config.rs
```diff
@@ -1,21 +1,23 @@
...
     max_stack_size: u32,
+    call_per_local_cost: u32,
 }
 
 impl WasmValidatorConfigV1 {
     pub fn new() -> Self {
         Self {
             weights: InstructionWeights::default(),
             max_stack_size: 1024,
+            call_per_local_cost: 1,
```
iamyulong, line 20: What is this?

radix-engine/src/vm/wasm/wasm_validator_config.rs
```diff
@@ -29,57 +31,57 @@ impl WasmValidatorConfigV1 {
...
+            I32Store { .. } => Some(self.weights.i64store),
+            I64Store { .. } => Some(self.weights.i64store),
+            F32Store { .. } => None,
+            F64Store { .. } => None,
+            I32Store8 { .. } => Some(self.weights.i64store),
+            I32Store16 { .. } => Some(self.weights.i64store),
+            I64Store8 { .. } => Some(self.weights.i64store),
+            I64Store16 { .. } => Some(self.weights.i64store),
+            I64Store32 { .. } => Some(self.weights.i64store),
+            //            CurrentMemory { .. } => Some(self.weights.memory_current),
```
iamyulong, line 79: ?

radix-engine/src/vm/wasm/wasm_validator_config.rs
```diff
@@ -179,31 +181,37 @@ impl Rules for WasmValidatorConfigV1 {
...
             F32ReinterpretI32 => None,
             F64ReinterpretI64 => None,
-            SignExt(_) => Some(self.weights.i64extendsi32),
+            // sign-extension instructions
+            I32Extend8S => Some(self.weights.i64extendsi32),
+            I32Extend16S => Some(self.weights.i64extendsi32),
+            I64Extend8S => Some(self.weights.i64extendsi32),
+            I64Extend16S => Some(self.weights.i64extendsi32),
+            I64Extend32S => Some(self.weights.i64extendsi32),
+            _ => todo!("instruction {:?} not covered", instruction),
```
iamyulong, line 214: Can we list all the instructions that's not covered and promote the `TODO` to `FIXME`?

radix-engine/src/vm/wasm/wasmer.rs
```diff
@@ -576,11 +576,13 @@ impl WasmerModule {
 
         fn consume_wasm_execution_units(
             env: &WasmerInstanceEnv,
-            n: u32,
+            n: u64,
```
iamyulong, line 579: As `wasm-instrument` supports u64, let's bump our stack to use `u64`.

radix-engine/src/vm/wasm/errors.rs
```diff
@@ -67,6 +69,8 @@ pub enum InvalidMemory {
     MemorySizeLimitExceeded,
     /// The wasm module does not have the `memory` export.
     MemoryNotExported,
+    /// The wasm module misses max memory definition.
+    MemoryMaxMissing,
```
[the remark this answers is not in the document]
lrubasze: Right. Initially I thought it won't be possible in new implementation, so I added this error. Afterwards I realised it is possible I simply forgot about this.

radix-engine/src/vm/wasm/wasmer.rs
```diff
@@ -576,11 +576,13 @@ impl WasmerModule {
...
         ) -> Result<(), RuntimeError> {
             let (_instance, runtime) = grab_runtime!(env);
+            // TODO: wasm-instrument uses u64 for cost units. We need to decide if we want to move from u32
+            // to u64 as well.
             runtime
-                .consume_wasm_execution_units(n)
+                .consume_wasm_execution_units(n as u32)
```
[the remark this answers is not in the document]
lrubasze: yes, it could. But I treat it as interim approach (see comment above)

radix-engine/src/vm/wasm/wasm_validator_config.rs
lrubasze: Sure. But be aware that there are 541 variants at the moment. We support ~177 right now. So there will be plenty of TODOs 😄

radix-engine/src/vm/wasm/prepare.rs
```diff
@@ -1,170 +1,43 @@
...
-                            {
-                                return Err(PrepareError::FloatingPointNotAllowed);
-                            }
-                        }
-                    }
-                }
+        module.validate(features).map_err(|err| {
+            let err = err.to_string();
+            if err.starts_with("floating-point support is disabled")
+                | err.starts_with("floating-point instruction disallowed")
```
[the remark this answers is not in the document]
lrubasze: Yeah, it looks bad. The thing is that `wasmparser` specifies a [BinaryReaderError struct](https://docs.rs/wasmparser/latest/wasmparser/struct.BinaryReaderError.html#), which is wrapped by [anyhow::Error](https://docs.rs/anyhow/latest/anyhow/struct.Error.html). 
BinaryReaderError is not simply matchable.
How about such approach?
- get rid of `PrepareError::FloatingPointNotAllowed`
- simply wrap error with `PrepareError::ValidationError(err.to_string())` or `PrepareError::WasmParserError(err.to_string())`

radix-engine/src/vm/wasm/prepare.rs
lrubasze: Trying to keep the same approach as so [far](https://github.com/radixdlt/radixdlt-scrypto/pull/1208/files#diff-f8451cacac25f49e552023e9aaa0d43c5a7b82b5ec528059d2c5f5428af94f81L948).

radix-engine/src/vm/wasm/prepare.rs
```diff
@@ -823,71 +673,73 @@ impl WasmModule {
...
+            .module
+            .memory_section()
+            .map_err(|err| PrepareError::WasmParserError(err.to_string()))?;
 
         // Check if there is only one memory definition
-        let memory = match memory_section.entries().len() {
+        let mut memory = match memory_section.len() {
             0 => Err(PrepareError::InvalidMemory(
-                InvalidMemory::NoMemoryDefinition,
+                InvalidMemory::MissingMemorySection,
```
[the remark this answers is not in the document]
lrubasze: Current memory_section() might return empty vector if section is empty or does not exist, which is misleading. 
Will fix it to return Option. 
None - section does not exist
vector - section items (might be empty)

radix-engine/src/vm/wasm/prepare.rs
```diff
@@ -918,10 +779,8 @@ impl WasmModule {
         self,
         max_number_of_functions: u32,
     ) -> Result<Self, PrepareError> {
-        if let Some(section) = self.module.function_section() {
-            if section.entries().len() > max_number_of_functions as usize {
-                return Err(PrepareError::TooManyGlobals);
-            }
+        if self.module.num_local_functions() > max_number_of_functions {
```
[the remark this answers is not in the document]
lrubasze: self.module.function_section() used previously contains vector of function types specified in `funcs` component (see https://webassembly.github.io/spec/core/binary/modules.html#function-section).
I understand it as locally defined functions.

radix-engine/src/vm/wasm/prepare.rs
lrubasze: see explanation above

iamyulong (approved):
LGTM. 

The only remaining TODOs are
1. To decide the right value for `call_per_local_cost`
2. To update cost to `u64`

lrubasze:
Thanks Yulong.
TODOs will be added in separate PRs

PR #1210, opened 2023-06-30
Feature/SCRY-425 native blueprint costing

mstrug-rdx:
## Summary
Added tests to cover missing calls of all native blueprints exported functions.

## Details
Added two functions: `publish_native` and `publish_wasm_advanced` as size dependent (graphs: https://docs.google.com/spreadsheets/d/1s6Te1sOpy2E9d10G_jSqcy4FyTZN9zvsS_dE5m7tCH8/edit?usp=sharing).
Updated profiling scripts.


radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -137,15 +151,30 @@ impl FeeTable {
...
+                        "Native function not found: {:?}::{}. ",
+                        package_address, export_name
+                    ))
+            });
 
-        cpu_instructions / 10
+        // Reference EC2 instance c5.4xlarge has CPU clock 3.4 GHz which means in 1 µs it executes 3400 instructions
+        // (1 core, single-threaded operation, skipping CPU cache influence).
+        // Basing on above assumptions return native function execution time in µs:
+        cpu_instructions / 3400
```
iamyulong, line 177: We need to multiply this number by 100 because 1 us = 100 cost units.

iamyulong (approved):
LGTM.

I noticed that you've added `ASSERT_WORKTOP_CONTAINS_ANY` which means we will have to update the compiler and add e2e test to cover it. 

Although, it can be done in another PR.

PR #1211, opened 2023-07-01
feature: ManifestBuilder overhaul (buckets without callbacks)

dhedey:
## Summary

Update `ManifestBuilder` - see the `transfer_xrd` or `radiswap` scenarios for how it works.

This builder has the following benefits:

* It gets rid of those pesky callbacks, making manifest building easier to write and read! This has been asked for time and again by our community. Instead, it exposes a "Namer" which you can use for creating new proofs, and using the proofs, in the relevant placeholders.
* It conserves the naming of buckets/proofs etc which can be passed to the decompiler - see eg the named buckets in the decompiled radiswap manifests
* It supports more methods such as "deposit" and "burn_resource", which I guess were missing from the old builder because they were a pain to use

### For dApp Developers

Use the new builder like this:
```rust
use transaction::prelude::*;

    .lock_fee_from_faucet()
    .withdraw_from_account(from_account_address, XRD, 1)
    .take_from_worktop(XRD, 1, "xrd_bucket")
    .try_deposit_or_abort(to_account_address, "xrd_bucket")
    .build();
```

Or for more complicated cases, where you need to embed a bucket/proof into a model, you can use a name `lookup` like this:
```rust
use transaction::prelude::*;

    .lock_fee_from_faucet()
    .withdraw_from_account(from_account_address, XRD, 1)
    .take_from_worktop(XRD, 1, "xrd_bucket")
    .call_function_with_name_lookup(
        package_address,
        "SomeBlueprint",
        "some_function",
        |lookup| (
            "argument1",
            lookup.bucket("xrd_bucket"),
            dec!("13.33")
        ),
    )
    .build();
```

Note:
* The builder now has methods which are much more permissive in terms of arguments they accept - this means you can remove `.into()`, `.to_string()` calls etc. If using `to_manifest_value_and_unwrap!(&TypedManifestArguments { .. })` then you can drop the `to_manifest_value_and_unwrap!(&..)` wrapper.
* The `test_runner` isn't yet quite so permissive - such tweaks will come there soon.


transaction/src/builder/manifest_builder_v2.rs
```diff
@@ -0,0 +1,1394 @@
...
+use radix_engine_interface::blueprints::access_controller::*;
+use radix_engine_interface::blueprints::account::*;
+use radix_engine_interface::blueprints::consensus_manager::*;
+use radix_engine_interface::blueprints::identity::*;
+use radix_engine_interface::blueprints::package::*;
+use radix_engine_interface::blueprints::resource::ResourceAction::{Burn, Mint};
+use radix_engine_interface::blueprints::resource::*;
+
+/// Utility for building transaction manifest.
+pub struct ManifestBuilderV2 {
```
0xOmarA, line 16: It seems like we now have a lot of duplicate logic between the ManifestBuilderV1 and the ManifestBuilderV2. I think that we can improve the situation there down to something that allows us to maintain both easily. Consider defining a `ManifestBuilder` trait that implements all of the shared logic between V1 and V2 of the manifest builder, something along the lines of:

```rs
pub trait ManifestBuilder {
    pub fn add_instruction(&mut self, instruction: InstructionV1) -> &mut Self;

    pub fn add_blob(&mut self, blob: Vec<u8>) -> &mut Self;

    pub fn assert_worktop_contains(&mut self, resource_address: ResourceAddress, amount: Decimal) -> &mut Self {
        self.add_instruction(InstructionV1::AssertWorktopContains {
            resource_address,
            amount
        })
    }

    /* Other Shared Methods */
}
```

`ManifestBuilderV1` and `ManifestBuilderV2` could then implement this trait and implement `add_instruction` and `add_blob`. Other methods that differ between the two manifest builders could then be implemented in an `impl` block for that specific builder.

dhedey: So chatting with Yulong, I've now replaced ManifestBuilder with this new version.

The ManifestBuilder as-is isn't Toolkit/FFI friendly - it:
* Makes heavy use of traits to make it super-nice to use from Rust, but I doubt translate well over FFI
* Panics if thing aren't correct
* Uses lots of Rust-specific types

I imagine you'll want to instead have a different ManifestBuilder paradigm for the FFI  (much more string-based and less clever). But happy to discuss

iamyulong (approved):
Thanks for pulling this together! The new interface looks much better.

transaction/src/builder/manifest_namer.rs
```diff
@@ -0,0 +1,733 @@
...
+
+    /// This is intended for registering a bucket name to an allocated identifier, as part of processing a manifest
+    /// instruction which creates a bucket.
+    pub fn register_bucket(&mut self, new: NamedManifestBucket) {
+        if self.namer_id != new.namer_id {
+            panic!("NewManifestBucket cannot be registered against a different ManifestNamer")
+        }
+        let new_bucket = self.id_allocator.new_bucket_id();
+        match self.named_buckets.get_mut(&new.name) {
+            Some(allocated @ ManifestObjectState::Unregistered) => {
```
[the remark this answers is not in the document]
dhedey: That triggers the code path `Some(_) => unreachable!("NewManifestBucket was somehow registered twice")`

But indeed - it's unreachable, because to get here, you need to have created two `NamedManifestBucket`s with `new_bucket` (because they're not clone) - and we get the user-facing error message there about a duplicate bucket name.

PR #1212, opened 2023-07-01
Feature: Update Validator Create Costing

talekhinezh:
## Summary
Price validator creation in USD and return back bucket of change on creation


radix-engine/src/system/bootstrap.rs
```diff
@@ -50,7 +50,9 @@ use transaction::validation::ManifestIdAllocator;
 
 lazy_static! {
     pub static ref DEFAULT_TESTING_FAUCET_SUPPLY: Decimal = dec!("100000000000000000");
-    pub static ref DEFAULT_VALIDATOR_XRD_COST: Decimal = dec!("1000");
+    pub static ref DEFAULT_VALIDATOR_USD_COST: Decimal = dec!("100");
+    pub static ref DEFAULT_VALIDATOR_XRD_COST: Decimal =
+        *DEFAULT_VALIDATOR_USD_COST * Decimal::try_from(DEFAULT_USD_PRICE_IN_XRD).unwrap();
```
iamyulong, line 55: Should we remove the `DEFAULT_VALIDATOR_XRD_COST` given that `usd_price` is a protocol defined variable?

iamyulong (approved):

radix-engine/src/system/bootstrap.rs
talekhinezh: I think it should be fine to keep as `DEFAULT_VALIDATOR_XRD_COST` is derived from the usd price and is useful for testing.

PR #1216, opened 2023-07-03
Feature: Dont expose OwnerRole::UpdatableByObject enum to scrypto

talekhinezh:
## Summary
Dont expose OwnerRole::UpdatableByObject enum to scrypto

## Defails
* Reordered Publish Package argument ordering in the manifest to be more consistent with Resource Creation functions

## For dApp Developers
Publish Package manifest instruction has reordered arguments:
```
PUBLISH_PACKAGE_ADVANCED
    Enum<AccessRule::AllowAll>() # Owner AccessRule
    Tuple(                       # Package Definition
        Map<String, Tuple>()
    )
    Blob("${code_blob_hash}")    # Package Code
    Map<String, Tuple>()         # Metadata
    None;                        # Address Reservation
```


dhedey (changes requested):
As a point of order, I'm going to mark request changes on all PRs which have a breaking change which haven't:
* Got the correct tags - in this case, it needs `native` and `manifest`
* Got clear documentation on how integrators should respond to the change - in this case, example manifest orders for the change
* Doesn't have well commented example manifests in `e2e.rs`, using the correct enum tags (eg None/Some), and `enums.rs` variant names.

iamyulong (approved):

PR #1221, opened 2023-07-04
Feature/take advanced

iamyulong:
## Summary

- Renamed `RoundingMode` variants
- Added `RoundingMode::MidpointNearestEven`
- Added `take_advanced(amount: Decimal, strategy: WithdrawStrategy)` to buckets and vaults
- Added `amount_for_withdrawal(amount: Decimal, strategy: WithdrawStrategy)` to resource managers
- Updated `protected_withdraw` to accept a `WithdrawStrategy` for pools
- Various clean-ups to the bucket and vault native blueprints

## Testing

Tests added for the new features.

### For dApp Developers

You can now call `.take_advanced()` method to withdraw resource of any divisiblitty. This is more cost-efficient than rounding and withdrawing in application space.


radix-engine-common/src/math/rounding_mode.rs
```diff
@@ -1,16 +1,23 @@
...
+    ToZero,
+    /// The number is always rounded away from zero, e.g. `3.1 -> 4`, `-3.1 -> -4`.
     AwayFromZero,
-    /// Rounds to the nearest and when a number is halfway between two others, it's rounded towards zero, e.g. `3.5 -> 3`, `-3.5 -> -3`.
-    TowardsNearestAndHalfTowardsZero,
-    /// Rounds to the nearest and when a number is halfway between two others, it's rounded away zero, e.g. `3.5 -> 4`, `-3.5 -> -4`.
-    TowardsNearestAndHalfAwayFromZero,
+
+    /// The number is rounded to the nearest, and when it is halfway between two others, it's rounded toward zero, e.g. `3.5 -> 3`, `-3.5 -> -3`.
+    MidpointTowardZero,
```
dhedey, line 18: I'd suggest the following names are slightly clearer:
```
* ToNearestMidpointTowardZero
* ToNearestMidpointArrayFromZero
* ToNearestMidpointToEven
```

But also happy with the current names if you prefer :)

dhedey: Although actually being consistent with https://docs.rs/rust_decimal/latest/rust_decimal/enum.RoundingStrategy.html is okay, although I don't think necessarily we need to fully copy it - it looks like they have a lot of legacy names they want to ditch so may have kept some names around for backwards compatibility

radix-engine-common/src/math/decimal.rs
```diff
@@ -98,20 +98,20 @@ impl Decimal {
...
+        self.round(0, RoundingMode::ToNegativeInfinity)
     }
 
     /// Returns the smallest integer that is equal to or greater than this number.
     pub fn ceiling(&self) -> Self {
-        self.round(0, RoundingMode::TowardsPositiveInfinity)
+        self.round(0, RoundingMode::ToPositiveInfinity)
     }
 
     pub fn round(&self, decimal_places: u32, mode: RoundingMode) -> Self {
```
dhedey, line 109: I wonder if we want `round(&self, decimal_places: impl Into<i32>, mode: RoundingMode)`, and allow negative decimal places?

dhedey: Separately, this method asserts on decimal_places - could we add this to the Rust doc as a clear precondition for calling this?

radix-engine-common/src/math/decimal.rs
```diff
@@ -181,6 +181,35 @@ impl Decimal {
                     }
                 }
             }
+            RoundingMode::MidpointNearestEven => {
+                if self.0 % divisor == BnumI256::zero() {
+                    self.clone()
+                } else {
```
dhedey, line 187: This all feels like quite an expensive calculation, but I can't see any obvious speed-ups (although I imagine there would be some if we wanted to do a separate optimisation pass in a future workstream).

I'm glad that we're outsourcing this to engine where we can, because I imagine this would be quite gas intensive.

radix-engine-common/src/math/decimal.rs
```diff
@@ -1052,6 +1081,38 @@ mod tests {
         assert_eq!(num.round(18, mode).to_string(), "-2.555555555555555555");
     }
 
+    #[test]
+    fn test_rounding_midpoint_nearest_even_zero_decimal() {
```
dhedey, line 1085: Nice tests

radix-engine-interface/src/blueprints/resource/resource.rs
```diff
@@ -142,8 +128,9 @@ impl LiquidNonFungibleResource {
     }
 
     pub fn take_all(&mut self) -> LiquidNonFungibleResource {
-        self.take_by_amount(self.amount())
-            .expect("Take all from `Resource` should not fail")
+        LiquidNonFungibleResource {
+            ids: core::mem::replace(&mut self.ids, btreeset!()),
```
dhedey, line 132: :') cute.

radix-engine/src/blueprints/resource/fungible/fungible_bucket.rs
```diff
@@ -10,41 +10,50 @@ use radix_engine_interface::blueprints::resource::*;
...
+        amount: Decimal,
+        withdraw_strategy: WithdrawStrategy,
+        api: &mut Y,
+    ) -> Result<Bucket, RuntimeError>
     where
         Y: KernelNodeApi + ClientApi<RuntimeError>,
     {
+        // Apply withdraw strategy
+        let divisibility = Self::get_divisibility(api)?;
+        let amount = amount.for_withdrawal(divisibility, withdraw_strategy);
```
dhedey, line 47: I imagine given how often this code path will be hit with `WithdrawStrategy::Exact` (eg from the `take` call), it's worth us skipping the divisibility read in this case?

Probably simplest is to inline `for_withdrawal` for performance, and only doing the divisibility read where it's required.

dhedey: Although actually I guess we still need to do the `check_fungible_amount` call anyway with the divisibility, so no biggy!

radix-engine/src/blueprints/resource/non_fungible/non_fungible_bucket.rs
```diff
@@ -7,199 +7,187 @@ use radix_engine_interface::api::field_lock_api::LockFlags;
...
-        Ok(amount)
+        Self::take_advanced(amount, WithdrawStrategy::Exact, api)
     }
 
-    pub fn locked_amount<Y>(api: &mut Y) -> Result<Decimal, RuntimeError>
+    pub fn take_advanced<Y>(
+        amount: &Decimal,
+        withdraw_strategy: WithdrawStrategy,
+        api: &mut Y,
+    ) -> Result<Bucket, RuntimeError>
```
dhedey, line 24: Much nicer interface / encapsulation returning a bucket here 👍

radix-engine/src/blueprints/resource/non_fungible/non_fungible_vault.rs
```diff
@@ -251,33 +275,78 @@ impl NonFungibleVaultBlueprint {
...
+        Self::internal_take_non_fungibles(&delta, api)?;
+
+        // Increase lock count
+        for id in ids {
+            locked.ids.entry(id.clone()).or_default().add_assign(1);
+        }
+
+        api.field_lock_write_typed(handle, &locked)?;
+
+        // Issue proof
```
dhedey, line 310: Very minor, but this comment isn't accurate anymore after the refactor

dhedey (approved):
Really nice! I like the improvements of the interfaces / abstractions in the native blueprints too.

radix-engine-common/src/math/rounding_mode.rs
iamyulong: Will change to the above.

radix-engine-common/src/math/decimal.rs
iamyulong: I will change to `i32` and leaving support of negative decimal places to future.

radix-engine/src/blueprints/resource/fungible/fungible_bucket.rs
iamyulong: Yeah, we need that for amount checking anyway.

radix-engine-common/src/math/decimal.rs
iamyulong: In the fixed version of implementation, `remainder` will be checked against `mid_point`, so it's not a waste, assuming the `eq` check is relatively cheap. We can benchmark more later on for opitmisation.

PR #1223, opened 2023-07-04
Feature: Remove lockable access rules from all interfaces

talekhinezh:
## Summary
Removed lock flag from `RolesInit` effectively removing the lock flag for roles in all interfaces.

## Details
* Removed AccessRules methods: `SET_AND_LOCK_OWNER_ROLE`, `LOCK_ROLE`, `SET_AND_LOCK_ROLE`
* Refactored `RolesInit` and associated macros to remove lock flag

### For dApp Developers
* Removed `.set_and_lock_owner_role()`, `.lock_role()`, `.set_and_lock_role()` methods from global component interface

### For Internal Integrators
Resource Manager Creation in the Manifest has been simplified to not require a lock flag. An example manifest now looks like:
```python
CREATE_FUNGIBLE_RESOURCE
    # Owner role - This gets metadata permissions, and is the default for other permissions
    # Can set as Enum<OwnerRole::Fixed>(access_rule)  or Enum<OwnerRole::Updatable>(access_rule)
    Enum<OwnerRole::None>() 
    true             # Whether the engine should track supply (avoid for massively parallelizable tokens)
    18u8             # Divisibility (between 0u8 and 18u8)
    Tuple(
       Some(         # Mint Roles (if None: defaults to DenyAll, DenyAll)
            Tuple(
                Some(Enum<AccessRule::AllowAll>()),  # Minter (if None: defaults to Owner)
                Some(Enum<AccessRule::DenyAll>())    # Minter Updater (if None: defaults to Owner)
            )
        ),
        None,        # Burn Roles (if None: defaults to DenyAll, DenyAll)
        None,        # Freeze Roles (if None: defaults to DenyAll, DenyAll)
        None,        # Recall Roles (if None: defaults to DenyAll, DenyAll)
        None,        # Withdraw Roles (if None: defaults to AllowAll, DenyAll)
        None         # Deposit Roles (if None: defaults to AllowAll, DenyAll)
    )
    Tuple(
        Map<String, Tuple>(),
        Map<String, Enum>()
    )
    None             # No Address Reservation
;
```


transaction/examples/access_rule/access_rule.rtm
```diff
@@ -5,22 +5,7 @@ SET_OWNER_ROLE
 LOCK_OWNER_ROLE
```
iamyulong, line 5: We still have lock owner role?

iamyulong (approved):

transaction/examples/resources/creation/fungible/no_initial_supply.rtm
```diff
@@ -16,52 +16,36 @@ CALL_METHOD
...
-                Enum<Option::Some>(
-                    Enum<Metadata::String>("A very innovative and important resource")    # Resource Description
-                ),
-                true                                                                      # Locked Resource Description
+                Some(Enum<Metadata::String>("MyResource")),    # Resource Name
+                true                                                         # Locked
             )
         ),
-        Map<String, Tuple>()
+        Map<String, Enum>(                                                   # Metadata roles
```
dhedey, line 44: Not something we need to change - just an observation - that this is inconsistent with the tuple structure above. I guess it makes sense, because they have different defaulting strategies.

transaction/src/manifest/ast.rs
```diff
@@ -227,22 +227,10 @@ pub enum Instruction {
         address: Value,
         args: Vec<Value>,
     },
-    SetAndLockOwnerRole {
```
dhedey, line 230: I see we've removed SetAndLockOwnerRole but are keeping LockOwnerRole - I think this is fine tbh, you can do a Set and then the new Owner can choose to lock if they wish. Perhaps that's better even.

dhedey (approved):
Looks good! I guess we're keeping `LOCK_OWNER_ROLE` because it possibly does make sense for the owner role to be able to change itself to be `Fixed`.

transaction/examples/access_rule/access_rule.rtm
talekhinezh: Yup, as owner role is a little special and should be able to go from `OwnerRole::Updatable` to `OwnerRole::Fixed`.

transaction/examples/resources/creation/fungible/no_initial_supply.rtm
talekhinezh: Yup, exactly

transaction/src/manifest/ast.rs
talekhinezh: Yup, exactly

transaction/examples/resources/creation/non_fungible/no_initial_supply.rtm
```diff
@@ -15,48 +15,38 @@ CALL_METHOD
...
-                    Enum<Metadata::String>("MyResource")                               # Resource Name
-                ),
-                true                                                                   # Locked Resource Name
-            ),
-            "description" => Tuple(
-                Enum<Option::Some>(
-                    Enum<Metadata::String>("A very innovative and important resource")    # Resource Description
-                ),
-                false
+                Some(Enum<Metadata::String>("MyResource")),    # Resource Name
```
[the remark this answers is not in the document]
talekhinezh: np!

PR #1226, opened 2023-07-05
Feature: update proof creation interface

iamyulong:
## Summary

- Removed `CREATE_PROOF` (instructions & APIs) from `Account`, `AuthZone`, `Bucket` and `Vault`
- Removed `create_proof_of_amount` from non-fungible `Bucket` and `Vault`
- Removed `.authorize()` from `Bucket` and `Vault`
- Added `.authorize_with_all(fn)` for any bucket
- Added `.authorize_with_amount(amount, fn)` for fungible buckets and proofs
- Added `.authorize_with_non_fungibles(ids, fn)` for non-fungible buckets and proofs
- Changed `create_proof` of `AccessController` to create proof of all

### For dApp Developers

With the removal of `create_proof`, you can use `create_proof_of_amount`, `create_proof_of_non_fungibles` or `create_proof_of_all` depending on resource fungibility.

For manifest, 
- Use `CREATE_FROM_BUCKET_OF_AMOUNT`,  `CREATE_FROM_BUCKET_OF_NON_FUNGIBLES`, and `CREATE_FROM_BUCKET_OF_ALL` for creating proof from buckets
- Use `CREATE_FROM_AUTH_ZONE_OF_AMOUNT`,  `CREATE_FROM_AUTH_ZONE_OF_NON_FUNGIBLES`, and `CREATE_FROM_AUTH_ZONE_OF_ALL` for composing proof from auth zone.

To get a proof of Validator / Account / Identity Owner Badges:
```rust
CALL_METHOD Address("ACCOUNT_ADDRESS")
    "create_proof_of_non_fungibles"
    Address("OWNER_BADGE_RESOURCE_ADDRESS")
    Array<NonFungibleLocalId>(
        NonFungibleLocalId("[<hex encoded raw bytes of address>]"),
    )
;
```

### For Internal Integrators

Wallets, dashboard and toolkit might be heavily impacted by the account interface change.


transaction/src/builder/manifest_builder.rs
```diff
@@ -303,13 +301,13 @@ impl ManifestBuilder {
     pub fn take_non_fungibles_from_worktop(
         self,
         resource_address: impl ResolvableResourceAddress,
-        ids: BTreeSet<NonFungibleLocalId>,
+        ids: &BTreeSet<NonFungibleLocalId>,
```
dhedey, line 304: I purposefully changed it to remove & to make the interface nicer / so that we could change to an IntoIterator model :) - the caller can just call clone if they need to (very rarely)

dhedey (approved):
Really nice!

Although minor - would rather keep without & (or use IntoIterator) on the ManifestBuilder - perhaps we can merge as-is and then fix in a follow up PR?

transaction/src/builder/manifest_builder.rs
iamyulong: I wanted to use `AsRef<BTreeSet<NonFungibleLocalId>>` so both reference and owned can be used. We can improve this later.

dhedey: Hmm but then it requires a btreeset, rather than, say, allowing an array. I think I'd prefer `IntoIterator` and then doing a count on the length of each to verify uniqueness :).

iamyulong: Ah, I think my main concern was about passing ownership of the non-fungible local ids.

How about `AsRef<[NonFungibleLocalId]>`?

dhedey: I'm not sure I follow the concern over ownership? `BTreeSet<NonFungibleLocalId>` doesn't implement `AsRef<[NonFungibleLocalId]>` I believe

PR #1227, opened 2023-07-05
Architecture: Locked Fields

talekhinezh:
## Summary
Add ability to lock fields similarly to how key value entries are lockable.

## Details
* Field substates are now wrapped with additional mutability info
* Updated `new_object` API to include mutability of fields
* Fixed object APIs to use `open`/`close` semantics rather than `lock`

### For Internal Integrators
All Field substates are now wrapped in a similar manner key value entries are wrapped


radix-engine-interface/src/api/field_lock_api.rs
```diff
@@ -31,24 +31,26 @@ pub type FieldLockHandle = u32;
 
 /// A high level api to read/write fields
 pub trait ClientFieldLockApi<E: Debug> {
```
iamyulong, line 33: Rename to `ClientFieldApi`?

radix-engine-interface/src/api/key_value_entry_api.rs
```diff
@@ -34,7 +34,7 @@ pub trait ClientKeyValueEntryApi<E> {
 
     fn key_value_entry_remove(&mut self, handle: KeyValueEntryHandle) -> Result<Vec<u8>, E>;
 
-    fn key_value_entry_freeze(&mut self, handle: KeyValueEntryHandle) -> Result<(), E>;
+    fn key_value_entry_lock(&mut self, handle: KeyValueEntryHandle) -> Result<(), E>;
 
     fn key_value_entry_release(&mut self, handle: KeyValueEntryHandle) -> Result<(), E>;
```
iamyulong, line 39: `key_value_entry_release` to `key_value_entry_close`?

iamyulong (approved):
Nice!

PR #1229, opened 2023-07-06
Fix: Separate pool manager and pool owner

0xOmarA:
## Summary

This PR separates the pool manager and the pool owner such that the pool owner is not necessarily also the pool manager. This is because the pool manager badge is most likely going to be held by some component while the pool owner badge (which has the ability to update the resource and pool metadata) would likely be stored in an account to allow for the metadata to be updated later on.

## Details

The `instantiate` function of all three pool blueprints has been modified to allow the instantiator to pass in an optional `OwnerRole` to it. 
* The pool manager role is given the ability to call the `protected_withdraw` and `protected_deposit` methods on the pool.
* The pool owner is given all of the abilities given to it by default when nothing is overridden, including the ability to update and lock metadata.

## Testing

Tests for this already exited. I've updated them slightly to reflect the new interfaces.

### For dApp Developers

The `instantiate` function on all three pool blueprints now takes an additional `OwnerRole` argument. If you're using pool in Scrypto then you will get a compile-time error with this update since the pool stubs have been updated to reflect this.


radix-engine-interface/src/blueprints/pool/multi_resource_pool/invocations.rs
```diff
@@ -10,11 +10,13 @@ define_invocation! {
     function_name: instantiate,
     input: struct {
         resource_addresses: BTreeSet<ResourceAddress>,
+        owner_rule: OwnerRole,
```
dhedey, line 13: The OwnerRole should come first for consistency with other calls which use an OwnerRole

radix-engine-interface/src/blueprints/pool/one_resource_pool/invocations.rs
```diff
@@ -10,11 +10,13 @@ define_invocation! {
     function_name: instantiate,
     input: struct {
         resource_address: ResourceAddress,
+        owner_rule: OwnerRole,
```
dhedey, line 13: The OwnerRole should come first for consistency with other calls which use an OwnerRole

radix-engine-interface/src/blueprints/pool/two_resource_pool/invocations.rs
```diff
@@ -10,11 +10,13 @@ define_invocation! {
     function_name: instantiate,
     input: struct {
         resource_addresses: (ResourceAddress, ResourceAddress),
+        owner_rule: OwnerRole,
```
dhedey, line 13: The OwnerRole should come first for consistency with other calls which use an OwnerRole

dhedey (approved):

PR #1235, opened 2023-07-06
Architecture: FIXME: Check calls to module methods

talekhinezh:
## Summary
Add checks to verify if an object has a given module on method calls

## Details
* Expand `ObjectInfo` to include versions for every module. Secondarily, this map acts as a way of describing whether the given object has the module or not.
* Remove requirement that `ComponentRoyalty` must be specified by all components.

### For Internal Integrators
`ObjectInfo` structure which is part of the `TypeInfoSubstate` has been updated to specify which modules a particular object has. In particular, royalty modules are now optional.

```
pub struct ObjectInfo {
    pub global: bool,
    pub main_blueprint_id: BlueprintId,
    pub module_versions: BTreeMap<ObjectModuleId, BlueprintVersion>,

    // Blueprint arguments
    pub blueprint_info: ObjectBlueprintInfo,
    pub features: BTreeSet<String>,
    pub instance_schema: Option<InstanceSchema>,
}
```


radix-engine/src/system/system.rs
```diff
@@ -1215,6 +1143,42 @@ where
 
         Ok(enabled)
     }
+
+    fn get_module_object_info(
+        &mut self,
+        node_id: &NodeId,
+        object_module_id: ObjectModuleId,
+    ) -> Result<ObjectInfo, RuntimeError> {
```
iamyulong, line 1151: The abstraction doesn't feel right - "module of a global object is not an object".

radix-engine/src/system/system.rs
```diff
@@ -1215,6 +1143,42 @@ where
...
+                    } else {
+                        return Err(RuntimeError::SystemError(
+                            SystemError::ObjectModuleDoesNotExist(object_module_id),
+                        ));
+                    };
+
+                // Modules outside of Main do not have their own modules
+                ObjectInfo {
+                    global: true,
+                    main_blueprint_id: object_module_id.static_blueprint().unwrap(),
```
iamyulong, line 1168: Following the above comment, the `main_blueprint_id` renaming doesn't look necessary.

In fact, it's actually confusing, because when the callee is an `Owned<Metadata>`, the `main_blueprint_id` is `metadata`.

```rust
MethodActor {
    node_id: NodeId,
    module_id: ObjectModuleId,
    ident: String,
    object_info: ObjectInfo {
         global: bool,
         blueprint_id: BlueprintId,
         modules: BTreeMap<ObjectModuleId, BlueprintVersion>  // module blueprint id statically inferred
    }
}
```

PR #1238, opened 2023-07-10
Refactor actor and on_virtualize invocation

iamyulong:
## Summary

- Refactored `on_virtualize` invocation
- Addresses various `Actor` FIXMEs


radix-engine-interface/src/types/invocation.rs
```diff
@@ -24,45 +24,10 @@ impl FunctionIdentifier {
     }
 }
 
-#[derive(Clone, Eq, PartialEq, ScryptoSbor)]
-pub enum FnIdent {
```
talekhinezh, line 28: Nice cleanup!

radix-engine/src/blueprints/account/blueprint.rs
```diff
@@ -89,32 +89,36 @@ impl AccountBlueprint {
...
-    pub fn create_virtual_ed25519<Y>(
-        input: VirtualLazyLoadInput,
-        api: &mut Y,
-    ) -> Result<VirtualLazyLoadOutput, RuntimeError>
-    where
-        Y: ClientApi<RuntimeError>,
-    {
-        let public_key_hash = PublicKeyHash::Ed25519(Ed25519PublicKeyHash(input.id));
-        Self::create_virtual(public_key_hash, api)
+        match input.node_id.entity_type() {
```
talekhinezh, line 99: I much prefer the application layer being unaware of entity types (should be a system layer invariant). Is there a reason why this was moved into this layer?

radix-engine/src/blueprints/package/package.rs
```diff
@@ -1145,21 +1145,19 @@ impl PackageNativePackage {
...
                                 },
-                            )
-                        })
-                        .collect(),
+                            );
+                        }
+                        hook_exports
+                    },
                 };
                 definitions.insert(blueprint.clone(), definition);
```
talekhinezh, line 1162: Not sure if it's here but we should also be preventing Scrypto/WASM packages from registering hooks

radix-engine/src/kernel/actor.rs
```diff
@@ -95,54 +109,33 @@ impl Actor {
...
             }) => object_info.global,
             Actor::Function { .. } => true,
-            Actor::VirtualLazyLoad { .. } => false,
+            Actor::BlueprintHook { .. } => true,
             Actor::Root { .. } => false,
         }
     }
 
-    pub fn fn_identifier(&self) -> FnIdentifier {
+    pub fn fn_identifier(&self) -> Option<FnIdentifier> {
```
talekhinezh, line 117: It feels to me cleaner to mock the Root with a fake `FnIdentifier`/`BlueprintId`/`PackageAddress`/etc. rather than forcing the rest of the system to deal with the never used case of weird things happening at the root call frame. Thoughts?

talekhinezh (commented):
Looks good overall, nice cleanups! Just two comments that I think we should resolve.

radix-engine/src/blueprints/account/blueprint.rs
iamyulong: This is because there is only one blueprint (which implies one `on_virtualize` hook) for account.

I don't imagine we'd like `Ed25519Account` and `Secp256k1Account` lol.

radix-engine/src/blueprints/package/package.rs
iamyulong: The same way as we prevents users from using collections.

radix-engine/src/kernel/actor.rs
iamyulong: Unless there is substantial value in maintaining invariant, `FnIdentifier` is always available,  I'm leaning towards keeping the implementation simple and close to its true behaviour. 

This way we don't have to mock/reason about anything, and only need to follow compiler's guide and handle the extra case at consumers' side.

radix-engine-interface/src/api/system_modules/virtualization.rs
```diff
@@ -6,8 +6,9 @@ use radix_engine_common::types::NodeId;
 use sbor::rust::collections::BTreeMap;
 
 #[derive(Debug, Clone, Eq, PartialEq, ScryptoSbor, ManifestSbor)]
-pub struct VirtualLazyLoadInput {
-    pub id: [u8; NodeId::RID_LENGTH],
+pub struct OnVirtualizeInput {
```
talekhinezh, line 9: Nice!

radix-engine-interface/src/blueprints/account/invocations.rs
```diff
@@ -36,9 +36,6 @@ pub enum AccountDefaultDepositRule {
 
 pub const ACCOUNT_BLUEPRINT: &str = "Account";
 
-pub const ACCOUNT_CREATE_VIRTUAL_SECP256K1_ID: u8 = 0u8;
-pub const ACCOUNT_CREATE_VIRTUAL_ED25519_ID: u8 = 1u8;
```
talekhinezh, line 40: Seems like these are relevant again?

radix-engine-interface/src/blueprints/identity/invocations.rs
```diff
@@ -7,9 +7,6 @@ use sbor::rust::fmt::Debug;
 
 pub const IDENTITY_BLUEPRINT: &str = "Identity";
 
-pub const IDENTITY_CREATE_VIRTUAL_SECP256K1_ID: u8 = 0u8;
-pub const IDENTITY_CREATE_VIRTUAL_ED25519_ID: u8 = 1u8;
```
talekhinezh, line 11: Seems like these are relevant again?

talekhinezh (approved):

radix-engine-interface/src/blueprints/account/invocations.rs
iamyulong: YES, will use it again.

PR #1240, opened 2023-07-10
Partition-Tier in JMT

jakrawcz-rdx:
This is the refactoring described at https://rdxworks.slack.com/archives/C01HK4QFXNY/p1688482720264219


jakrawcz-rdx:
Performance charts (from @mstrug-rdx):

Current `develop`:
image

This PR:
image

My observations:
- the change captured by "linear approx" formula suggests that handling the extra JMT tier costs us +0.5ms per commit
- the workloads simulated by this perf test do not benefit from cache locality at all
- I kinda like how the JMT's `NodePayload` now fits the Engine's "Nodes -> Partitions -> Substates" structure
- The existence of a very natural "RE Node hash" is also undeniable, although we don't yet have any use-cases for it

My conclusions: 
- 🤷‍♂️

jakrawcz-rdx:
The same experiment as above, but running on a dev MAC (not Michał's EC2):

Develop:
develop

PR:
plain_branch

Basically: the same effect, just the constant added commit delay is ~0.24ms, not ~0.5ms.

jakrawcz-rdx:
The [latest commit](https://github.com/radixdlt/radixdlt-scrypto/pull/1240/commits/90f835deff1382d05db38e2775abb9826fd273e0) changes the approach a bit, and here are its numbers:

image

(i.e. same as `develop` on this artificial workload)

jakrawcz-rdx:
Ok, so here may be a bit of explanation why the synthetic test sees no improvement, while a more real `` does:

I inspected actual nature of Node->Partition->Substate update set from a commit.

Here is what the synthetic test loops through:
```
commit touches 1 Nodes:
- Node ..67c1 touches 1 Partitions (currently having 100):
  - Partition 98 touches 98 Substates
commit touches 1 Nodes:
- Node ..67c1 touches 1 Partitions (currently having 100):
  - Partition 99 touches 99 Substates
commit touches 1 Nodes:
- Node ..67c1 touches 1 Partitions (currently having 100):
  - Partition 100 touches 100 Substates
Round 100/100  commit touches 1 Nodes:
- Node ..67c1 touches 1 Partitions (currently having 100):
  - Partition 1 touches 1 Substates
commit touches 1 Nodes:
- Node ..67c1 touches 1 Partitions (currently having 100):
  - Partition 2 touches 2 Substates
```
(a Node has very many Partitions, and updates 1 of them, writing ~50 substates)

Meanwhile, the `cargo bench --bench radiswap --features rocksdb` loops through:
```
commit touches 9 Nodes:
- Node ..18c6 touches 1 Partitions (currently having 6):
  - Partition 64 touches 1 Substates
- Node ..18c6 touches 2 Partitions (currently having 5):
  - Partition 64 touches 1 Substates
  - Partition 65 touches 1 Substates
- Node ..0c80 touches 1 Partitions (currently having 4):
  - Partition 64 touches 1 Substates
- Node ..bec8 touches 1 Partitions (currently having 2):
  - Partition 64 touches 1 Substates
- Node ..e50b touches 1 Partitions (currently having 2):
  - Partition 64 touches 1 Substates
- Node ..fcb2 touches 1 Partitions (currently having 2):
  - Partition 64 touches 1 Substates
- Node ..e697 touches 1 Partitions (currently having 2):
  - Partition 64 touches 1 Substates
- Node ..7bee touches 1 Partitions (currently having 2):
  - Partition 64 touches 1 Substates
- Node ..ff1e touches 1 Partitions (currently having 2):
  - Partition 64 touches 1 Substates
```
(a commit touches 9 Nodes, but still only 1-2 Partitions from each - however, the Partition counts are low)

From this numbers it is obvious why migrating the `Node-tier Leaf` impl from `Map<PartitionNum, Substate-tier Root>` to `Map<PartitionNum, Substate-tier root Hash>` (in https://github.com/radixdlt/radixdlt-scrypto/pull/1240/commits/90f835deff1382d05db38e2775abb9826fd273e0) has fixed the performance of the synthetic test (see https://github.com/radixdlt/radixdlt-scrypto/pull/1240#issuecomment-1630678459).

**Now the hard part:** why introducing a Partition-Tier improves the `radiswap` performance by >2x?
(actually the improvement against `develop` grows with the DB size)

It would be easy to explain if the `radiswap` had more cases like `Node X touches 2 Partitions` (i.e. it would mean far less updates to the Node-tier of JMT). But that's not so true (just 1 such case out of 9)...
Any other ideas? Or obvious omissions in the impl of this performance improvement?

jakrawcz-rdx:
(maybe I'll actually try to measure the number of hash computations / DB reads / DB writes within JMT update logic...)

jakrawcz-rdx:
I have some numbers for "JMT node DB read/write/delete + Blake hash computations during JMT update logic":
(I log cumulative numbers after each transaction)

`radiswap`:

Develop:
```
version Some(1336):
subtier: InvMetrics { jmt_db_reads: 153698, jmt_db_writes: 41761, jmt_db_stales: 18385, blakes: 891432 }
partitier: InvMetrics { jmt_db_reads: 70252, jmt_db_writes: 92436, jmt_db_stales: 70252, blakes: 645281 }
version Some(1337):
subtier: InvMetrics { jmt_db_reads: 153798, jmt_db_writes: 41789, jmt_db_stales: 18411, blakes: 891892 }
partitier: InvMetrics { jmt_db_reads: 70296, jmt_db_writes: 92480, jmt_db_stales: 70296, blakes: 645630 }
version Some(1338):
subtier: InvMetrics { jmt_db_reads: 153898, jmt_db_writes: 41817, jmt_db_stales: 18437, blakes: 892356 }
partitier: InvMetrics { jmt_db_reads: 70340, jmt_db_writes: 92524, jmt_db_stales: 70340, blakes: 645979 }
```

This PR:
```
version Some(1336):
subtier: InvMetrics { jmt_db_reads: 74054, jmt_db_writes: 41761, jmt_db_stales: 18385, blakes: 385839 }
nodetier: InvMetrics { jmt_db_reads: 30866, jmt_db_writes: 35019, jmt_db_stales: 30866, blakes: 237491 }
version Some(1337):
subtier: InvMetrics { jmt_db_reads: 74127, jmt_db_writes: 41789, jmt_db_stales: 18411, blakes: 386211 }
nodetier: InvMetrics { jmt_db_reads: 30895, jmt_db_writes: 35048, jmt_db_stales: 30895, blakes: 237724 }
version Some(1338):
subtier: InvMetrics { jmt_db_reads: 74200, jmt_db_writes: 41817, jmt_db_stales: 18437, blakes: 386587 }
nodetier: InvMetrics { jmt_db_reads: 30924, jmt_db_writes: 35077, jmt_db_stales: 30924, blakes: 237957 }
```

(note: `jmt_db_reads` is an actual number of "fetch a JMT node" roundtrips to DB; `jmt_db_writes` and `jmt_db_stales` are the number of JMT nodes which are [created | recorded as stale] in a single batch roundtrip per transaction)

Some observations: (inclusing my sudden realizations **in bold**)
- on This PR, there are considerably less Node-tier operations than Develop's (Node+Partition)-tier operations (like, 2-3x less), and the explanation is simple:
  - a number of Partitions in radiswap-related Nodes is 2-6 (see https://github.com/radixdlt/radixdlt-scrypto/pull/1240#issuecomment-1631362901) - which means that a Node-tier JMT is at least 2x smaller than a Node+Partition)-tier JMT.
  - while it is true that JMT operation latency grows with a `log(leaf count)`, **we still have to spend time creating those JMT Nodes - and there is >2x more of them** 🤷 
- also, on This PR, there are considerably less Substate-tier JMT DB read operations!
  - you may ask "why? the number of Substates is the same in both case!" (and the JMT DB write | stale counters confirm this very exactly)
  - and I will respond: "look, when you want to insert a Substate into _the right_ lower-tier JMT, you need to obtain the root of that particular lower-tier JMT - and it is not enough to know the upper-tier JMT prefix: you also need its version. So **you need to actually traverse the upper-tier JMT down, to locate the upper-tier Leaf == lower-tier Root** (it happens [here](https://github.com/radixdlt/radixdlt-scrypto/blob/40493299dbdb787e309d76e5fb5ca04f478d62e8/radix-engine-stores/src/hash_tree/mod.rs#L105))"
  - ...and since we already established that there is >2x more of upper-tier JMT Leafs on Develop, it is also guaranteed that we spend time locating each of them.

Above, I kinda convinced myself that ~2x improvement is legit (if we assume that "update JMT" is a significant contributor in the "execute logic + update flat Substate store + update JMT" set-up used by the `radiswap` benchmark). Maybe some other effects (e.g. cache-locality, smaller overall DB size) also boost it.
Let me know if I convinced anyone else :p

radix-engine-stores/src/hash_tree/mod.rs
```diff
@@ -191,16 +210,25 @@ fn put_changes<S: TreeStore<P>, P: Payload>(
...
-        new_payload: change.changed.map(|root| (root.hash, root.node)),
+        key: LeafKey::new(&change.id),
+        new_payload: change
+            .changed
+            .map(|payload| (calculate_node_hash(&payload), payload)),
     }
 }
 
-fn to_substate_change(change: IdChange<DbSortKey, Hash>) -> LeafChange<()> {
+fn calculate_node_hash(node_payload: &NodePayload) -> Hash {
```
jakrawcz-rdx, line 222: this is an interesting place:

When any of the Substates of a Node changes, its Partition root hash changes (it happens via regular JMT update), and then its Node root hash also must change (I made it happen myself, and here is the logic that calculates the new hash).

The calculation you see here only does 1x `hash()`, but I now wonder if it rather should do some funny JMT-like (or at least Merkle-like?) hierarchical hashing, in order to allow for smaller proofs...
Any thoughts welcome.

dhedey: Yes, I feel like this is potentially an overhead. Would it make sense to just make this middle tier a JMT?

jakrawcz-rdx: In terms of "compute merkle hash" (and "generate proof") operation, the 2-level JMT is same as 8-level Merkle tree (i.e. outputs 8 uncle-hashes). Does the average ReNode have >8 partitions?

Anyways: storing this embedded `Map<PartitionNumber, SubstateRootHash>` encoded as a set of JMT nodes would really be a pity (because there is a more compact and easier to understand encoding - the `Map<>`).

However, we can totally interpret the `Map` as a Merkle tree (a.k.a. "do some funny JMT-like (or at least Merkle-like?) hierarchical hashing`). Maybe I can even use the JMT's original methods (i.e. [this](https://github.com/radixdlt/radixdlt-scrypto/blob/99a674e5151565a7bb1e3cacefc510a405e52c2b/radix-engine-stores/src/hash_tree/types.rs#L821-L825))...

dhedey: For A1 - bear in mind every user transaction is updating a node with like 160 partitions (the TransactionTracker) - so we need to optimise for this case too.

Personally, I'd opt for a permanent JMT for Q1:
=> The root node of the partition tree can be stored in the leaf of the node tree.
=> We can re-use code / implementation for our substate tree

jakrawcz-rdx: Indeed a permanent true JMT has the "consistency" advantages... I guess the only downsides I see are:
- _(performance, minor)_ guaranteed +2 roundtrips to DB (per partition touched in the commit)
- _(conceptual)_ using a JMT for a map of bounded [and TBH absolutely tiny] cardinality seems like an overkill

Let me give it a try - if it is simple/elegant enough from implementation PoV, then I can live with it 👍

dhedey: Well most updates will be in the user partition, which for scrypto components will be the only partition in that nibble, so it will only be +1 round trip to the DB. But for native components, you're right, this will be +2 round trips because most of these use multiple partitions under the same first nibble. 

We are also decreasing the average number of distinct partitions in the upper tree by about 3x, which is logarithmically -1/2 of a layer in the top tree :D.

But yes, it will have some impact... I wonder what size of map is required before this is cheaper - I thought it was surprisingly expensive? But maybe I'm misremembering or we were capturing the other slowdowns?

jakrawcz-rdx: > I wonder what size of map is required before this is cheaper - I thought it was surprisingly expensive? But maybe I'm misremembering or we were capturing the other slowdowns?

I remember our impact attribution was wrong all the way. I have 50% of the "true JMT" impl ready, so we'll be soon able to run some true comparison benchmarks.

jakrawcz-rdx:
After the latest commit:

The permanent, true Partition-Tier JMT: `transaction::radiswap   time:   [4.8640 ms 4.9074 ms 4.9604 ms]`
Before that latest commit: `transaction::radiswap   time:   [5.1237 ms 5.1552 ms 5.1879 ms]`

radix-engine-stores/src/hash_tree/mod.rs
jakrawcz-rdx: Finally done ✅

jakrawcz-rdx:
...and regarding the benchmarks, just for completion:

`develop`: `transaction::radiswap   time:   [5.3076 ms 5.3750 ms 5.4604 ms]`

So:
- splitting the ReNode vs Partition in the DB mapper brings `5.37ms -> 5.15ms` = ~4% faster
- doing the above + using a true Partition-Tier JMT brings `5.37ms -> 4.90ms` = ~8% faster

talekhinezh (approved):

PR #1243, opened 2023-07-11
Feature: object drop hook

iamyulong:
## Summary

Adds `BlueprintHook::OnDrop` and refactors `Proof` blueprints to use this feature.

### For Internal Integrators
Slightly renamed fields in `TypeInfoSubstate` - no structural change.


radix-engine-interface/src/types/blueprint.rs
```diff
@@ -11,26 +11,24 @@ use scrypto_schema::{InstanceSchema, KeyValueStoreSchema};
...
 pub struct ObjectInfo {
     pub global: bool,
-
     pub blueprint_id: BlueprintId,
-    pub version: BlueprintVersion,
-
-    // Blueprint arguments
-    pub blueprint_info: ObjectBlueprintInfo,
+    pub blueprint_version: BlueprintVersion,
+    pub outer_object: OuterObjectInfo,
```
talekhinezh, line 31: Can you revert these renames? Some of my upcoming PRs refactor `ObjectInfo` a bit.

radix-engine/src/errors.rs
```diff
@@ -150,6 +149,8 @@ pub enum KernelError {
 
     // Invoke
     InvalidInvokeAccess,
+
+    DanglingNodes(Vec<NodeId>),
```
talekhinezh, line 153: I quite like the old name better or `OrphanedNodes`

radix-engine/src/kernel/actor.rs
```diff
@@ -9,23 +10,26 @@ pub struct InstanceContext {
...
-    pub module_object_info: ObjectInfo,
+    pub is_direct_access: bool,
+    pub object_info: ObjectInfo,
+}
 
+#[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
+pub struct MethodActor {
     pub ident: String,
+    pub receiver_info: RuntimeReceiverInfo,
+    pub global_address: Option<GlobalAddress>,
```
talekhinezh, line 25: I've also refactored `MethodActor` a bit in https://github.com/radixdlt/radixdlt-scrypto/pull/1246/files#diff-a1773c84b7bd8f3422fd46bd349a608119e9beb86872c8e1f70c1a04ef4d3ecdL14

Could we revert these cleanups?

radix-engine/src/system/system_callback.rs
```diff
@@ -82,6 +86,65 @@ pub struct SystemConfig<C: SystemCallbackObject> {
     pub modules: SystemModuleMixer,
 }
 
+pub fn load_blueprint_definition<Y, C>(
```
talekhinezh, line 89: Feels to me that this should still be in `system.rs` as it's part of the "service" the system provides. Is there a reason to move it here?

radix-engine/src/kernel/actor.rs
iamyulong: Briefly looked at the other PR. I'm not sure I agree with that refactor.

I'm happy to revert any specific change here, if it doesn't look right to you.

radix-engine/src/system/system_callback.rs
iamyulong: It was to make it usable by both `system.rs` and `system_call.rs`, as it's now purely a utility function.

Happy to move to `SystemService` (seems like another centralised place for such functionalities).

iamyulong: Moved to `SystemService`.

All I need to do is to construct a `SystemService::new(api)`. No idea why I didn't do that initially.

radix-engine/src/errors.rs
iamyulong: Yes, will change.

radix-engine/src/kernel/actor.rs
talekhinezh: Yup, I think just removing `RuntimeReceiverInfo` and having all info flat in `MethodActor` for now would make it easiest. For example, I consider `object_info` as not required but is essentially just a cache of this info for `MethodActor`.

talekhinezh (approved):

radix-engine/src/kernel/actor.rs
iamyulong: Yes, removed it.

radix-engine-interface/src/types/blueprint.rs
iamyulong: Reverted `blueprint_version`, but kept `outer_object` as I feel it's more accurate.

PR #1248, opened 2023-07-12
Fix: Character Escaping in Decompiler

0xOmarA:
## Summary

This PR adds character escaping for the manifest decompiler.


file.tx
iamyulong, line 1: Accidental commit?

iamyulong (approved):
LGTM

transaction/src/data/formatter.rs
```diff
@@ -565,3 +565,21 @@ impl<'a> ContextualDisplay<ManifestDecompilationDisplayContext<'a>> for Manifest
         format_custom_value(f, self, context, false, 0)
     }
 }
+
+const ESCAPE_SEQUENCES: [(&str, &str); 7] = [
```
dhedey, line 569: Just a couple of thoughts, if you want to fix them:
* I'd have expected a comment here with a reference back to the function in the compiler, explaining where these escape sequences come from
* It looks like the only required escape sequence is " => \" - the others are arbitrary choices.  And we seem to have opted to do the easy ones and avoid eg escaping unicode. Choosing what to escape in unicode is a hard problem, but at the very least, we should probably escape control characters. This would avoid things like RTL overrides in our outputs.
* The below escape isn't very performant - we're repeating iteration 7 times.

I'd suggest as a non-urgent task, changing this to instead iterate over &str.chars(), check if it needs to be replaced, and and output itself, or its replacement, directly to the formatter `f`.

EG logic could be - first use `match` on the character for these checks. If it would output as is, then check for `.is_control()` and then output as the unicode escape, eg see: https://doc.rust-lang.org/std/primitive.char.html

0xOmarA: Regarding adding comments to the implementation to reference back to the bit in the lexer makes sense, we should do that.

Also, the suggested implementation would improve the performance and should reduce the number of iterations we do over the string.

I'd be happy if we add a non-urgent ticket for this.

dhedey: Will stick something on the work tracker 👍

PR #1250, opened 2023-07-12
Restrict globalize and drop to package only

iamyulong:
## Summary

Restricts `globalize` and `drop` to package only.


radix-engine-tests/tests/core.rs
```diff
@@ -34,3 +37,51 @@ fn test_call() {
     );
     receipt.expect_commit_success();
 }
+
+#[test]
```
talekhinezh, line 41: Can we add tests specifically for trying to globalize:
* Bucket/Vault/Proof
* Metadata/Royalty/RoleAssignment

talekhinezh (approved):

radix-engine-tests/tests/core.rs
talekhinezh: This would be more of product "behavior" tests rather than system tests

iamyulong: Yes, I will add such tests in https://github.com/radixdlt/radixdlt-scrypto/pull/1268, because I added `is_transient` flag there for all the listed blueprints.

PR #1255, opened 2023-07-12
Allowing for configurable Database of TestRunner (deduplicating BasicRocksdbTestRunner)

jakrawcz-rdx:
When working on https://github.com/radixdlt/radixdlt-scrypto/pull/1240 I wanted to repeatedly run benchmarks on memoryDb vs rocksDb.
This was problematic, because `BasicRocksdbTestRunner` was very separate from `TestRunner`. But in fact they only differed by their `substate_db` dependency, so I made it configurable at `TestRunner` and then deleted `BasicRocksdbTestRunner`.


scrypto-unit/src/test_runner.rs
```diff
@@ -235,24 +235,38 @@ impl CustomGenesis {
...
     custom_genesis: Option<CustomGenesis>,
+    custom_extension: E,
+    custom_database: D,
     trace: bool,
     state_hashing: bool,
 }
 
-impl TestRunnerBuilder {
-    pub fn new() -> TestRunnerBuilder {
+impl TestRunnerBuilder<NoExtension, InMemorySubstateDatabase> {
```
talekhinezh, line 254: Cleaner may be `impl DefaultTestRunner {`

scrypto-unit/src/test_runner.rs
```diff
@@ -235,24 +235,38 @@ impl CustomGenesis {
     }
 }
 
-pub struct TestRunnerBuilder {
+pub trait Database:
```
talekhinezh, line 238: Perhaps change name to `TestDatabase` as it's the fact that it's a test that we need all of the listed traits.

talekhinezh (approved):

scrypto-unit/src/test_runner.rs
jakrawcz-rdx: I had the same thought!
...but this is a `TestRunnerBuilder`, not `TestRunner` :(

scrypto-unit/src/test_runner.rs
jakrawcz-rdx: I always welcome a good naming suggestion 👍

scrypto-unit/src/test_runner.rs
talekhinezh: ah!

jakrawcz-rdx:
@talekhinezh is the `Cancelled after 23m` normal for these ☝️ 2 checks?
(or please click merge if safe)

talekhinezh:
> @talekhinezh is the `Cancelled after 23m` normal for these ☝️ 2 checks? (or please click merge if safe)

Hrm, let me try and rerun. Not normal but sometimes it happens.

jakrawcz-rdx:
I retried it again in the morning and consistently these 2 (`Radix Engine tests (wasmer), Radix Engine tests (k8s-linux-runner)`) are getting cancelled, after 19-24 minutes.
Other current PRs against develop (e.g. https://github.com/radixdlt/radixdlt-scrypto/pull/1256) don't have this problem.

I cannot find any logs (or information on "why cancelled") on CI. Maybe I'm clicking wrong things. How to find out more?

Is `cancelled == timeout`? Can my change even affect the performance?
Or is it about something else?

iamyulong:
Thanks for cleaning this up! 

Wanted to do the same but needed to get something in quick,  so added `BasicRocksdbTestRunner` :-D

jakrawcz-rdx:
Let's see if updating from `develop` helps the CI.

Even if not, I will hit "Merge", backed by this 🟢 :

image

jakrawcz-rdx:
yup, same 2 cancellations, but I am merging anyway, as promised.

PR #1257, opened 2023-07-13
Demonstrating surprising performance of "not hashing partition number" alone

jakrawcz-rdx:
I decided to test this after hearing that my wonderful and flawless explanation (https://github.com/radixdlt/radixdlt-scrypto/pull/1240#issuecomment-1632268424) is not convincing enough to _some_ people.

These people must have been right, because this PR **only** changes the `SpreadPrefixKeyMapper` impl a little bit (i.e. includes the partition num outside the hashed part, not inside it). I still cannot explain why the speed-up is so dramatic, but now everyone can repro this on this minimal, `develop`-based change:

_(I introduced feature flags to test this easily)_

The baseline (equivalent to plain `develop`):
```
$ cargo bench --bench radiswap --features rocksdb,parnum_hashed
...
Radiswap::run           time:   [7.1027 ms 7.1604 ms 7.2341 ms]
```

The changed DB key mapper impl:
```
$ cargo bench --bench radiswap --features rocksdb,parnum_not_hashed
...
Radiswap::run           time:   [3.2403 ms 3.2532 ms 3.2665 ms]
```

That's a >2x faster `radiswap`. Even without any changes to JMT internal structure. Any ideas where it comes from?


jakrawcz-rdx:
Ahhh forgot: 

⚠️ **DO NOT MERGE, THIS IS ONLY AN EXPERIMENT** ⚠️ 

(the CI won't pass due to new feature flags not being handled)

PR #1268, opened 2023-07-17
Feature: transient blueprint

iamyulong:
## Summary

Adds flag `is_transient` to `BlueprintDefinition`. 

This is to mark whether any component of a blueprint can be persisted or not

When set to `true`, objects of the blueprint will be prohibited from being persisted.

### For Internal Integrators

Updated `BlueprintDefinition` and `BlueprintDefinitionInit` to include the new flag.

Package publishing manifest now includes an additional parameter.


radix-engine/src/kernel/call_frame.rs
```diff
@@ -376,33 +372,6 @@ impl<L: Clone> CallFrame<L> {
         &self.actor
     }
 
-    // TODO: Remove
-    fn get_type_info<S: SubstateStore>(
```
talekhinezh, line 380: YAY

radix-engine/src/kernel/call_frame.rs
```diff
@@ -511,12 +480,18 @@ impl<L: Clone> CallFrame<L> {
         Ok((lock_handle, substate_value.len(), store_access))
     }
 
-    pub fn close_substate<S: SubstateStore>(
+    pub fn close_substate<S: SubstateStore, F, M>(
         &mut self,
         heap: &mut Heap,
         store: &mut S,
+        callback: &mut M,
```
talekhinezh, line 487: Can we remove callback object from this layer? I think the kernel itself should be able to add callback object on it's own.

radix-engine/src/kernel/call_frame.rs
```diff
@@ -715,14 +690,20 @@ impl<L: Clone> CallFrame<L> {
...
-    pub fn create_node<'f, S: SubstateStore>(
+    pub fn create_node<'f, S: SubstateStore, F, M>(
         &mut self,
         node_id: NodeId,
         node_substates: NodeSubstates,
         heap: &mut Heap,
         store: &'f mut S,
-        push_to_store: bool,
-    ) -> Result<StoreAccessInfo, CreateNodeError> {
+        callback: &mut M,
```
talekhinezh, line 699: Ditto here

radix-engine/src/kernel/call_frame.rs
```diff
@@ -824,15 +805,21 @@ impl<L: Clone> CallFrame<L> {
...
+    pub fn move_module<'f, S: SubstateStore, F, M>(
         &mut self,
         src_node_id: &NodeId,
         src_partition_number: PartitionNumber,
         dest_node_id: &NodeId,
         dest_partition_number: PartitionNumber,
         heap: &'f mut Heap,
         store: &'f mut S,
-    ) -> Result<StoreAccessInfo, MoveModuleError> {
+        callback: &mut M,
```
talekhinezh, line 816: Ditto here

radix-engine/src/kernel/call_frame.rs
```diff
@@ -1069,15 +1056,21 @@ impl<L: Clone> CallFrame<L> {
         Ok((substates, store_access))
     }
 
-    pub fn drop_all_locks<S: SubstateStore>(
+    pub fn drop_all_locks<S: SubstateStore, F, M>(
         &mut self,
         heap: &mut Heap,
         store: &mut S,
-    ) -> Result<(), CloseSubstateError> {
+        callback: &mut M,
```
talekhinezh, line 1063: Ditto here

radix-engine/src/kernel/call_frame.rs
```diff
@@ -1102,37 +1095,19 @@ impl<L: Clone> CallFrame<L> {
         self.owned_root_nodes.keys().cloned().collect()
     }
 
-    pub fn move_node_to_store<S: SubstateStore>(
+    pub fn move_node_to_store<S: SubstateStore, F, M>(
         heap: &mut Heap,
         store: &mut S,
+        callback: &mut M,
```
talekhinezh, line 1101: Ditto

radix-engine/src/kernel/call_frame.rs
iamyulong: I kind of need it inside the closure.

But we don't need `M` to be `KernelCallbackObject`. Maybe I just remove the type bound?

Another option is have trait object which encapsulates both the `callback` and the `filter`.

talekhinezh: Yes, exactly, the call_frame doesn't need to know that there is even such a thing as filter.

See a similar pattern here: https://github.com/radixdlt/radixdlt-scrypto/pull/1270/files#diff-d92ad4919aac15aa05fcea4887fe2e880ecd22d05a6b77dfd51479662372a17fL261

talekhinezh: right, that's what I meant

talekhinezh (approved):

PR #1269, opened 2023-07-17
Costing of publish package

mstrug-rdx:
## Summary
Results of investigation of publishing package costing.

## Details
Enabled Fee table `publish_native` and `publish_wasm_advanced` coefficients.
Added `large_package` blueprint which just merges other two blueprints to provide one package with size very close to 1 MiB limit, used for transaction payload cost verification - reduced payload byte cost to 40 (from 50).
Added `bench_prepare_wasm` which publishes `radiswap` package, used for validation of CPU instructions accuracy.

## Testing
Added new test `publish_package_1mib`.


iamyulong (approved):
LGTM with minor comment.

radix-engine-tests/tests/metering.rs
```diff
@@ -695,3 +695,18 @@ struct TestNonFungibleData {
...
+/// SCRYPTO_SBOR_V1_PAYLOAD_PREFIX and the actor size.
+/// If this test fails with an error TransactionLimitsError::MaxInvokePayloadSizeExceeded,
+/// go to `blueprints/large_package/Cargo.toml` file and change `package2` reference package name and path
+/// some other blueprint, but verify transaction payload size if it is close enough to the limit (it can be
+/// done by checking TxPayloadCost from cost breakdown table divided by payload byte cost, which can be taken
+/// from fee_table.rs tx_payload_cost() function). List of blueprints and its size can be displayed using command
+/// `ls -lSk ./radix-engine-tests/tests/blueprints/target/wasm32-unknown-unknown/release/*.wasm`
+fn publish_package_1mib() {
+    let mut test_runner = TestRunnerBuilder::new().build();
+    test_runner.compile_and_publish("./tests/blueprints/large_package");
```
iamyulong, line 711: It would be nicer if we can make the wasm a static asset, to avoid flaky results caused by host environment.

PR #1272, opened 2023-07-18
Fix transaction preview issues

iamyulong:
## Summary

- Fixed preview transaction payload size
- Fixed an issue with previewing without locking fee
- Updated the hello world template

### DIFF
--- a/assets/template/tests/lib.rs
+++ b/assets/template/tests/lib.rs
@@ -32,7 +32,11 @@ fn test_hello() {
     // Test the `free_token` method.
     let manifest = ManifestBuilder::new()
         .call_method(component, "free_token", manifest_args!())
-        .deposit_batch(account)
+        .call_method(
+            account,
+            "deposit_batch",
+            manifest_args!(ManifestExpression::EntireWorktop),
+        )
         .build();
     let receipt = test_runner.execute_manifest_ignoring_fee(
         manifest,
--- a/radix-engine-tests/tests/preview.rs
+++ b/radix-engine-tests/tests/preview.rs
@@ -53,6 +53,32 @@ fn test_transaction_preview_cost_estimate() {
 }
 
 #[test]
+fn test_transaction_preview_without_locking_fee() {
+    // Arrange
+    let mut test_runner = TestRunner::builder().build();
+    let network = NetworkDefinition::simulator();
+    let manifest = ManifestBuilder::new()
+        //.lock_fee_from_faucet()
+        .clear_auth_zone()
+        .build();
+    let preview_flags = PreviewFlags {
+        use_free_credit: true,
+        assume_all_signature_proofs: false,
+        skip_epoch_check: false,
+    };
+    let (_, preview_intent) = prepare_matching_test_tx_and_preview_intent(
+        &mut test_runner,
+        &network,
+        manifest,
+        &preview_flags,
+    );
+
+    // Act
+    let preview_receipt = test_runner.preview(preview_intent, &network).unwrap();
+    println!("{:?}", preview_receipt);
+}
+
+#[test]
 fn test_assume_all_signature_proofs_flag_method_authorization() {
     // Arrange
     // Create an account component that requires a key auth for withdrawal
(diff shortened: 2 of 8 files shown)


radix-engine-tests/tests/preview.rs
```diff
@@ -27,25 +28,62 @@ fn test_transaction_preview_cost_estimate() {
...
-    preview_receipt.expect_commit_success();
-
-    let receipt = test_runner.execute_transaction(
+    let preview_receipt = test_runner.preview(preview_intent, &network).unwrap();
+    let preview_result = preview_receipt.expect_commit_success();
+    let actual_receipt = test_runner.execute_transaction(
         validate(&network, &notarized_transaction).get_executable(),
         FeeReserveConfig::default(),
-        ExecutionConfig::for_preview(),
+        ExecutionConfig::for_notarized_transaction()
```
dhedey, line 41: Can we explicitly have a `for_preview()` do this, and test with that configuration?

So we know we're using a mode which has been tested in the engine?

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -667,16 +671,34 @@ where
             // Record final payments
             *fee_payments.entry(vault_id).or_default() += amount;
         }
+        // Free credit is locked first and thus used last
+        if free_credit.is_positive() {
```
dhedey, line 675: I wonder if we should include `used_free_credit` on the fee_summary?

I guess your feeling is maybe not, as it will always be zero for actual transactions?

dhedey (approved):
Thanks Yulong. A few minor/medium markups. Happy for you to merge when you're happy.

assets/template/tests/lib.rs
```diff
@@ -32,7 +32,11 @@ fn test_hello() {
     // Test the `free_token` method.
     let manifest = ManifestBuilder::new()
         .call_method(component, "free_token", manifest_args!())
-        .deposit_batch(account)
+        .call_method(
```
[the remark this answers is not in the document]
iamyulong: Ah, I was just copying and pasting the `hello-world` example.

I saw the diff, but didn't bother to change, as it would be an example to demo `ENTIRE_WORKTOP`.

BTW we should probably rename `.deposit_batch` to`.deposit_batch_with_entire_worktop`.

radix-engine-tests/tests/preview.rs
iamyulong: We have `for_preview()` configuration.  

This test is about comparing the estimated results with **actual** results, which is why I picked `for_notarized_transaction()`.

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -174,31 +180,24 @@ where
 
     pub fn execute(
         &mut self,
-        transaction: &Executable,
+        executable: &Executable,
         fee_reserve_config: &FeeReserveConfig,
         execution_config: &ExecutionConfig,
     ) -> TransactionReceipt {
+        let fee_credit = executable.fee_payment().free_credit_in_xrd;
```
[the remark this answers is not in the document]
iamyulong: Nice spot. Will fix

radix-engine/src/transaction/transaction_executor.rs
iamyulong: I don't have a strong opinion, as `total_cost() - total_payments()` would give the free credit amount.

iamyulong: I will add a function and see how useful it is. :)

radix-engine-tests/tests/preview.rs
dhedey: Oooh yes, good point!

radix-engine/src/transaction/transaction_executor.rs
dhedey: I meant more as an actual field, but you're right, because it's calculated, it's probably not needed explicitly!

assets/template/tests/lib.rs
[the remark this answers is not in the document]
dhedey: Yes, I agree on the naming... It's on my list of tweaks I want to make in scrypto when I get a chance, but feel free to make it yourself :)

PR #1276, opened 2023-07-20
Refactor/consolidate protocol constants

iamyulong:
## Summary

Move all protocol defined constants from various places to `radix-engine-common/constants`, to make evaluation as a whole easier.

### For Internal Integrators

The constants are now defined under `radix-engine-common/constants`. 

I've also removed `DEFAULT_` prefix for some of them, to emphasise that it's THE configuration and for consistency.


radix-engine-common/src/constants/transaction_construction.rs
```diff
@@ -0,0 +1,2 @@
+/// The default tip percentage to use when building a transaction.
```
dhedey, line 1: I feel like in tests we may wish to default to this? But the actual transaction builder should default to 0?

iamyulong: Yes, let me change it to zero.

talekhinezh (approved):

radix-engine-common/src/constants/transaction_execution.rs
```diff
@@ -0,0 +1,74 @@
+/// The default system loan amount, used by transaction executor.
```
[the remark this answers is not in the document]
iamyulong: Yes!

PR #1281, opened 2023-07-21
Avoid recursion when persisting nodes

iamyulong:
## Summary

Limits the max depth allowed during recursive persistence. 

Without the limit, one can crash the engine with stack overflow, even in release mode.


radix-engine-common/src/constants/transaction_execution.rs
```diff
@@ -57,6 +57,9 @@ pub const MAX_METADATA_KEY_STRING_LEN: usize = 100;
 /// The max SBOR size of metadata value
 pub const MAX_METADATA_VALUE_SBOR_LEN: usize = 512;
 
+/// The max depth to support, per recursive move-to-store operation.
```
talekhinezh, line 60: To clarify, this doesn't put bounds on the actual max persisted depth right? If for example, I call move-to-store on a leaf node and then call move-to-store on the new leaf node.

talekhinezh: If so, can we rename to `MAX_RECURSIVE_ON_PERSIST_DEPTH` to make this clear?

talekhinezh:
Is a better solution to use Heap rather than Stack (ie no recursion) when doing move-to-store?

radix-engine/src/kernel/call_frame.rs
```diff
@@ -1089,34 +1089,39 @@ impl<L: Clone> CallFrame<L> {
...
-                return Err(PersistNodeError::NodeBorrowed(node_id, count));
-            }
-        };
-        for (_partition_number, module_substates) in &node_substates {
-            for (_substate_key, substate_value) in module_substates {
-                for reference in substate_value.references() {
-                    if !reference.is_global() {
-                        return Err(PersistNodeError::ContainsNonGlobalRef(*reference));
-                    }
+        let mut queue = LinkedList::new();
```
talekhinezh, line 1092: Perhaps `VecDeque` would be better?

iamyulong: It really depends on the workload, how dynamic the queue is. Hard to tell without serious benchmarks.

`LinkedList` gives "constant time" for each pop/push, but each requires a small malloc each time.

`VecDeque` (ring buffer) is fast in push/pop, if everything fits in the buffer. When resize is needed, it requires large malloc and memcpy, which can be slow.

talekhinezh: Fair enough, I'd just usually consider using array based lists/queues as a default (especially if size is somewhat bounded) and then using linkedlists as an optimization.

PR #1284, opened 2023-07-24
Architecture: Cleanup AuthZone

talekhinezh:
## Summary
Cleanup AuthZone

## Details
* Replace `assert_access_rule` system call with a `assert_access_rule` method
* Add AuthZone `NodeId` into `Actor`
* Prevent movement and dropping of AuthZone by creating it in previous frame and passing a reference to it
* Cleanup authorization logic without `ActingLocation`
* Remove `BeforePushFrame` and `AfterPopFrame` kernel callbacks, replace with in system wrappers

### For Internal Integrators
Added `assert_access_rule` to AuthZone blueprint


iamyulong (commented):
As discussed yesterday, the only required change is to make sure that auth module is fully gated through `SystemModuleMixer`.

iamyulong (approved):

PR #1289, opened 2023-07-25
tests: Add scenarios with example schemas, and metadata-standard nf data

dhedey:
## Summary
Theo needed to generate some example schemas, for their tests, and I suggested the easiest mechanism was to create some scenarios for them.

So we've created some scenarios/tests for a few of these cases.

I've merged into `rcnet-v2-phase-2` because I want it to be clear as examples for that release, although I'm not intending we include this in the scenarios / wipe genesis again as I don't want to disrupt upstream teams. I'll need to merge `release/rcnet-v2-phase-2` into `develop` after this branch is merged.

## For downstream teams

From the next release, the non-fungible resource scenario will have a `"integer_non_fungible_resource_with_metadata_standard_data"` resource address which has non-fungible data which aligns with the Metadata Standard.

### DIFF
--- a/radix-engine-interface/src/blueprints/resource/non_fungible/non_fungible_resource_manager.rs
+++ b/radix-engine-interface/src/blueprints/resource/non_fungible/non_fungible_resource_manager.rs
@@ -211,6 +211,14 @@ pub struct NonFungibleResourceManagerUpdateDataInput {
     pub data: ScryptoValue,
 }
 
+#[cfg_attr(feature = "radix_engine_fuzzing", derive(Arbitrary))]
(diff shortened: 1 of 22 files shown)


iamyulong (approved):

talekhinezh:
@dhedey Are we good to merge this?

dhedey:
@talekhinezh  - hmmm I think we probably shouldn't merge this to `phase-2` as it breaks genesis. Although I wanted this PR as a demonstration for other teams.

Let's close this PR and merge into `develop` instead before we take the fig cut.

PR #1292, opened 2023-07-27
Feature/Kernel API costing

mstrug-rdx:
## Summary
Added base cost to Kernel API basing on QEMU CPU instructions.

## Details
Costing of functions in `fee_table.rs` related to Kernel API has been established basing on CPU instructions generated on QEMU. Some functions are dependent on arguments, for others there is no clear relationship with arguments so median value from all calls has been used.
Updated QEMU scripts.


radix-engine/src/kernel/call_frame.rs
```diff
@@ -744,7 +744,7 @@ impl<C, L: Clone> CallFrame<C, L> {
         heap: &'f Heap,
         store: &'f mut S,
         lock_handle: LockHandle,
-    ) -> Result<&'f IndexedScryptoValue, ReadSubstateError> {
+    ) -> Result<(&'f IndexedScryptoValue, bool), ReadSubstateError> {
```
talekhinezh, line 747: Instead of returning `bool` can we pass the callback into this method instead, so that we don't leak weird abstractions in the interface.

mstrug-rdx: I added `CallFrameEventHandler` as an parameter through which information about reading from heap is passed back to the Kernel (to `KernelHandler`).

radix-engine/src/kernel/kernel.rs
```diff
@@ -766,12 +787,19 @@ where
...
+        let value = self
             .current_frame
-            .read_substate(&self.heap, self.store, lock_handle)
+            .read_substate(&self.heap, self.store, lock_handle, &mut handler)
             .map_err(CallFrameError::ReadSubstateError)
             .map_err(KernelError::CallFrameError)?;
 
+        let read_from_heap = handler.read_from_heap;
         let mut read_only = as_read_only!(self);
         M::on_read_substate(
```
talekhinezh, line 804: This is the event handler which should be passed into `read_substate`

radix-engine/src/kernel/kernel.rs
```diff
@@ -215,6 +215,27 @@ struct KernelHandler<
     callback: &'a mut M,
     prev_frame: Option<&'a CallFrame<M::CallFrameData, M::LockData>>,
     on_store_access: F,
+    read_from_heap: bool,
```
talekhinezh, line 218: Can we just combine this with `on_store_access` so that `StoreAccess` is repaced with a bigger enum. For example,
```
enum CallFrameEvent {
    StoreAccess(StoreAccess),
    HeapAccess,
}
```

mstrug-rdx: Yes, actually such a solution I had before, but after some heavy changes on develop around that enum I switched to boolean, but I will change that.

mstrug-rdx: I've extended `StoreAccess` enum with dedicated heap read variant, because it is more straightforward in my opinion than modification of `KernelHandler` (that will also require to add another event handler for heap store access).

radix-engine/src/track/interface.rs
```diff
@@ -207,6 +207,8 @@ pub enum StoreAccess {
     /// A new entry has been added to track
     /// System limits how many items that can be tracked.
     NewEntryInTrack,
+    /// Substate has been read from heap
+    ReadFromHeap,
```
talekhinezh, line 211: Hrm, this is the wrong place to add "Heap" related info as `Track` should not have any notion of a `Heap`

talekhinezh (approved):

radix-engine/src/kernel/kernel.rs
```diff
@@ -766,31 +798,24 @@ where
...
         lock_handle: LockHandle,
     ) -> Result<&IndexedScryptoValue, RuntimeError> {
-        let mut read_from_heap = false;
         let mut handler = KernelHandler {
             callback: self.callback,
             prev_frame: self.prev_frame_stack.last(),
-            on_store_access: |_, store_access| {
-                read_from_heap = matches!(store_access, StoreAccess::ReadFromHeap);
-                Ok(())
+            on_store_access: |_, _| {
```
mstrug-rdx, line 804: @talekhinezh maybe we should change that field to `Option<F>`, so here we can just pass `None` and in other places `Some(...)`?

mstrug-rdx: But I see there is a complication of using `None` with `Option<F>`.

PR #1295, opened 2023-07-30
Feature / Add BalancedDecimal with 38 decimal places using BnumI256

fpieper:
## Summary
This PR adds `BalancedDecimal` with 38 decimal places using `BnumI256` like `Decimal` for identical performance and cost. For financial applications you often need prices with many decimal places since having inverted prices based on arbitrary token addresses is common. The current `Decimal` type often doesn't have enough decimal places and the `PreciseDecimal` is more expensive and often overkill.

## Details
For example, if you want to represent a price of two tokens and if you determine the order via the token address the price will be either very large or very small (potentially).
If it is small you don't have much decimals after the price value to differentiate (and also account for accumulating errors). In contrast if the price is (by chance based on the address) very large you still have all the decimals to differentiate different prices.

Meaning what you often want is an imbalanced decimal type with more decimals places for the fractions than for the integer part. Uniswap for example is using Q64.96 (https://en.wikipedia.org/wiki/Q_(number_format ) to represent prices for that reason.

Right now, we have `Decimal` and `PreciseDecimal` in Scrypto. Both are not very suited for that use case:
`Decimal`  (~ Q196.60) is relatively cheap and fast but 18 decimal places / fractions is often not enough and results in inaccurate results (or even unable to properly represent certain values in the first place.
`PreciseDecimal` (~ Q299.213) has enough fractional precision but is inefficient due to large integer part (often unused) and has much higher fees compared to `Decimal`.

To avoid accumulating round-off errors you need high precision, but often don't need these many integer places.

My proposal is to add  decimal datatype which I called `BalancedDecimal` (for now) which would use the same `BnumI256` bit integer like `Decimal`, but has 38 decimal places instead of 18. Which would make it something like a Q128.128 (which also still fits into `BnumI384` for overflow - only exception is `cbrt()` which now uses `BnumI512`).

Instead of adding it to our blueprint, I made this PR to properly support `BalancedDecimal` as a Scrypto datatype in the whole stack including in the transaction manifest and gateway. `BalancedDecimal` currently implements something very close to ~ Q128.128, that's why I have named it `BalancedDecimal`. I don't have a strong opinion on the name, but wanted to provide a first reasonable effort.

Further down the road there could be even the possibility to implement an Q64.96, but this would require changes in the `BnumI` type which I didn't went into for now to keep the changes straightforward.

## Discussion & Questions
- Naming: are there better alternative for naming the datatype instead of `BalancedDecimal`? (I don't have a strong opinion)
- Is this constant a reasonable choice? `VALUE_KIND_BALANCED_DECIMAL: u8 = 0xd0;`
- I additionally implemented `floor_to_decimal` and `ceil_to_decimal` (flooring & ceiling to 18 decimals) for `BalancedDecimal`. This is very useful for token input/output calculations and rounding the result to token amounts afterwards. This could be added to `PreciseDecimal` too in another PR (we are using it with PreciseDecimal in our own blueprint right now).
- In general, this is meant as proposal. I don't have any strong opinions regarding specific names or conventions. Overall I tried to make as least changes as possible.

## Testing
I added a comprehensive test suite and added the new type including tests everywhere where I have found `PreciseDecimal`.

## Update Recommendations
`BalanceDecimal` is a new decimal datatype for Scrypto and would be needed to be added if not supported indirectly already via `radix-engine-common`. Especially in existing Rust `match` constructs the `BalanceDecimal` type would need to be added additionally. Should be mostly relevant for internal integrators.

Tags:
* scrypto-lib
* sbor
* manifest
* transaction

Thank you very much for reading and taking that into consideration. This would be very great to have for Babylon.


fpieper:
@iamyulong thank you, you are welcome 👍 .

Yes, sure. I will close it (seems like I can do that).

The new PR looks cool - I really appreciate that you considered my proposal and improved upon it 💯

PR #1299, opened 2023-08-01
Fix decimal and precise decimal arithmetics

lrubasze:
## Summary
Fix basic arithmetics for `Decimal` and `PreciseDecimal` types.

## Details

- Assure that basic arithmetic operations between: `Decimal` and `PreciseDecimal`, `Decimal` and primitive types, `PreciseDecimal` and primitive types are transitive. Which means they produce  expected type no matter the order of the operand types.
- Additionally results of such operations shall be equal, if operands are derived from the same value
Example 1:
    `Decimal(10)` * `PreciseDecimal(10)` = `PreciseDecimal(100)`
    `PreciseDecimal(10)` * `Decimal(10)` = `PreciseDecimal(100)`
Example 2:
    `Decimal(10)` * `10_u32` = `Decimal(100)`
    `10_u32` * `Decimal(10)` = `Decimal(100)`

This PR addresses the [issue reported by the community](https://discord.com/channels/417762285172555786/765994894749597697/1129167191057645709) 

## Testing
Respective tests were implemented

## Update Recommendations

All should work as before. Backward compatibility has been assured.

### DIFF
--- a/radix-engine-common/src/math/decimal.rs
+++ b/radix-engine-common/src/math/decimal.rs
@@ -432,8 +432,9 @@ macro_rules! impl_arith_ops {
         impl Add<Decimal> for $type {
             type Output = Decimal;
 
+            #[inline]
             fn add(self, other: Decimal) -> Self::Output {
-                Decimal::from(self) + other
+                other + self
             }
         }
 
@@ -448,8 +449,9 @@ macro_rules! impl_arith_ops {
         impl Mul<Decimal> for $type {
             type Output = Decimal;
 
+            #[inline]
             fn mul(self, other: Decimal) -> Self::Output {
-                Decimal::from(self) * other
+                other * self
             }
         }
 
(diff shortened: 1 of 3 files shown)


talekhinezh (approved):

radix-engine-common/src/math/decimal.rs
```diff
@@ -345,75 +345,166 @@ impl From<bool> for Decimal {
...
+
+            fn div(self, other: $type) -> Self::Output {
+                self / Decimal::from(other)
+            }
+        }
+
+        impl Add<Decimal> for $type {
+            type Output = Decimal;
+
+            fn add(self, other: Decimal) -> Self::Output {
```
dhedey, line 435: I believe these "flipped" methods would be smaller code size if they were all just `#[inline]` and the body was `other + self` - so it basically just reuses the other method, but with the arguments flipped.

lrubasze: In case of:
a) multiplication addition flipping has no impact
b) substraction we would need to negate the second argument: `-other + self`
c) division we would have to invert the argument: `(Decimal::ONE /other) * self` and I am afraid there could appear some rounding issues, meaning we could receive different results than with current version `Decimal::from(self) / other`

So I would suggest to apply suggested change for a) and b)

PR #1300, opened 2023-08-01
Feature: Add Substate Schema to Transaction Receipt

talekhinezh:
## Summary
Add Substate Schema Pointer for all substates specified in receipt

## Details
* Add `SortedKey` to return of sorted index scan
* Add scrypto schemas to all `BlueprintCollectionSchema` variants
* Add reference handling for substate keys
* Add `SubstateSchemaPointer` map for all substates in receipt

### For Internal Integrators

Simplified the sorted index schema for validators by replacing the `EpochRegisteredValidatorByStakeEntry` substate with `Validator`.

Old:
```rust
pub struct EpochRegisteredValidatorByStakeEntry {
    pub component_address: ComponentAddress,
    pub validator: Validator,
}

#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, ScryptoSbor)]
pub struct Validator {
    pub key: Secp256k1PublicKey,
    pub stake: Decimal,
}
```

New (Just `Validator`):
```rust
#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, ScryptoSbor)]
pub struct Validator {
    pub key: Secp256k1PublicKey,
    pub stake: Decimal,
}
```


dhedey (approved):

radix-engine-common/src/types/type_identifier.rs
```diff
@@ -0,0 +1,6 @@
+use crate::crypto::Hash;
+use crate::Sbor;
+use sbor::LocalTypeIndex;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Sbor)]
+pub struct TypeIdentifier(pub Hash, pub LocalTypeIndex);
```
dhedey, line 6: Minor - but would prefer this to have named fields so the hash can be named clearly; and a Rust doc explaining what this type is for.

PR #1301, opened 2023-08-02
Features/SCRY-452 Blueprint and Wasm export names validation

mstrug-rdx:
## Summary
All Wasm exported names and Blueprint name and its components should follow Rust Identifier specification

## Details
Added new functions: `enforce_export_names()` into `WasmModule` and `validate_names` into `package` module which uses `syn` crate `parse_str()` function to validate `Ident` conformance.

## Testing
Added new test which verifies if prohibited names are rejected with appropriate error.

## Update Recommendations
If anyone was using non 'Ident' conformant names, such Wasm/Blueprint will fail now.

### DIFF
--- a/radix-engine-tests/tests/package.rs
+++ b/radix-engine-tests/tests/package.rs
@@ -369,14 +369,14 @@ fn name_validation_blueprint() {
     let (code, mut definition) = Compile::compile("./tests/blueprints/publish_package");
 
     definition.blueprints = BTreeMap::from([(
-            String::from("wrong_bluepint_name_*"),
-            definition
-                .blueprints
-                .first_entry()
-                .unwrap()
-                .get()
-                .to_owned(),
-        )]);
+        String::from("wrong_bluepint_name_*"),
+        definition
+            .blueprints
+            .first_entry()
+            .unwrap()
+            .get()
+            .to_owned(),
+    )]);
 
     // Act
     let manifest = ManifestBuilder::new()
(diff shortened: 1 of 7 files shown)


radix-engine/src/blueprints/package/package.rs
```diff
@@ -412,6 +414,43 @@ fn validate_auth(definition: &PackageDefinition) -> Result<(), PackageError> {
...
+            for (name, _) in list.iter() {
+                condition(name)?;
+            }
+        }
+
+        if let FunctionAuth::AccessRules(list) = &bp_init.auth_config.function_auth {
+            for (name, _) in list.iter() {
+                condition(name)?;
+            }
+        }
```
talekhinezh, line 448: Almost all covered! I think still missing two:
* `BlueprintDefinitionInit -> feature_set`
* `BlueprintDefinitionInit -> method_auth -> StaticRoles -> roles -> Normal -> RoleKey -> key`

radix-engine-tests/tests/wasm_validation.rs
```diff
@@ -26,3 +27,51 @@ fn test_large_memory() {
         result
     );
 }
+
```
talekhinezh, line 30: Can we add atleast one test which captures the `PackageError::InvalidName` error

mstrug-rdx: Done

radix-engine-tests/tests/package.rs
```diff
@@ -361,3 +361,77 @@ fn should_not_be_able_to_publish_native_packages_in_scrypto() {
         )
     });
 }
+
+#[test]
+fn package_definition_names_validation() {
+    let mut test_runner = TestRunnerBuilder::new().build();
+
+    for i in 1..=3 {
```
talekhinezh, line 369: Last thing, can we separate these 3 into separate tests

PR #1307, opened 2023-08-03
Feature: Max substate key size

talekhinezh:
## Summary
Add Max substate key size

## Details
* Expanded events to include substate keys
* Extended Limits module to enforce a max substate key length (currently set to 1024)


radix-engine/src/system/system_modules/limits/module.rs
```diff
@@ -112,27 +133,27 @@ impl<V: SystemCallbackObject> SystemModule<SystemConfig<V>> for LimitsModule {
...
+                        limits.process_substate_key(key)?;
                     }
                 }
             }
             CreateNodeEvent::StoreAccess(store_access) => {
-                api.kernel_get_system()
-                    .modules
-                    .limits
-                    .process_store_access(store_access)?;
+                limits.process_store_access(store_access)?;
```
iamyulong, line 156: We only enforce this limit to substates in Store only?

iamyulong (approved):

PR #1308, opened 2023-08-03
Feature: AccessRule limits

talekhinezh:
## Summary
Add AccessRule complexity limits

## Details
* Added maximum depth check to AccessRules (currently set to 8)
* Added maximum number of nodes check to AccessRules (currently set to 64)


iamyulong:
Do we have a test case to validate that one can abuse this? I vaguely remember that I was trying to exploit this but failed because the system is protected by some other limit.

talekhinezh:
> Do we have a test case to validate that one can abuse this? I vaguely remember that I was trying to exploit this but failed because the system is protected by some other limit.

The `system_access_rule.rs` test tries to abuse these limits. I don't remember if these test passed or failed before the addition of limits though.

iamyulong:
As discussed, this is more to provide depth guarantee for downstream integrators, rather than a fix to a vulnerability.

iamyulong (approved):

PR #1316, opened 2023-08-07
Benchmarks to compare native and WASM performance

lrubasze:
## Summary

This is to compare performance of `native`, `wasmi`, `wasmer` for some operations such as:
- primitive types arithmetics
- Decimal and PreciseDecimal arithmetics

To run:
cargo bench -p radix-engine  --bench wasm_benchmarks


radix-engine/benches/wasm_benchmarks.rs
```diff
@@ -0,0 +1,311 @@
...
+                let wasmer_instance = get_wasmer_instance(&wasm_code[..]);
+                let wasmer_func = wasmer_instance.exports.get_function(func_name).unwrap();
+
+                for i in $range {
+                    // native
+                    group.bench_with_input(BenchmarkId::new("rust-native", i), &i, |b, i| {
+                        black_box(b.iter(|| {
+                            for _ in 0..*i {
+                                [< $t _ $ops >] ($x, $y, 0);
+                            }
```
iamyulong, line 119: Should we move the `black_box` to the statement inside the `for` loop, so compiler won't strip the code completely?

radix-engine/benches/wasm_benchmarks.rs
```diff
@@ -0,0 +1,311 @@
...
+                        }))
+
+                    });
+                    // wasmi call native
+                    group.bench_with_input(BenchmarkId::new("wasmi-call-native", i), &i, |b, i| {
+                        black_box(b.iter(|| {
+                            for _ in 0..*i {
+                                wasmi_call_host_func.call(store_call_host.as_context_mut(), ($x, $y, 0)).unwrap();
```
iamyulong, line 127: How does `Decimal`/`PreciseDecimal` cross the boundary?

radix-engine/wasm-benchmarks/src/precise_decimal.rs
```diff
@@ -0,0 +1,98 @@
+use radix_engine_common::math::PreciseDecimal;
+
+#[no_mangle]
+pub fn precise_decimal_add(x: i64, y: i64, _cnt: i64) -> i64 {
+    let x = PreciseDecimal::from(x);
+    let y = PreciseDecimal::from(y);
+    let z = x + y;
```
iamyulong, line 7: Is the `from` cheap, compared with `+`?

lrubasze: This is the reason I added `add_no_conversion` benchmark, where I benchmark only `+`.
No, it is not cheap at all. It is exactly the opposite.

Here is the quick comparison
```
                                            + and from      +               calculated from
precise_decimal_add/rust-native/1           137.44 ns       19.405 ns       59 ns
precise_decimal_add/wasmi/1                 60.212 µs       2.4815 µs       29 µs
precise_decimal_add/wasmer/1                2.4495 µs       290.85 ns       1 us
```

radix-engine/benches/wasm_benchmarks.rs
lrubasze: `local_call_host_func` accepts 3 x `i64`, which are further:
- converted to `Decimal`/`PreciseDecimal` (in case of eg. `decimal_mul`)
- passed for calculation (in case of eg `primitive_mul`)
- ignored (in case of eg. `decimal_mu_no_conversion`)

iamyulong (approved):

PR #1318, opened 2023-08-09
[ON HOLD] Feature: Full `declare_native_blueprint_state` macro

dhedey:
## PR status update

This full PR has been parked, and instead, will do it in bits (starting with just the core versioned models + schema stuff) - will create a new PR to capture that.

Replacement PR 2023: #1358
Extension PR 2024: #1744

## Summary

The goal of this PR is to introduce a more structured / type-safe mechanism for interfacing with our native substates... and to enable easily wrapping them with lazily-upgradable versioned enum wrappers in a type-safe way.

This PR has:
* (Small bit) Created a `versioned!` macro to make versioning easy
* (Bulk) Created a `declare_native_blueprint_state` macro to create typed native substate models which creates:
  * Data Models for Substate data and their keys
     * "Latest" version / versioned wrapper for content (reading layer 3)
     * (System) Payload (reading at layer 2)
     * For substates (reading at layer 1)
  * Data model for state init
     * Including for system layer
     * And output at kernel layer (for eg flashing) 
  * "Node layout" models including Partition enum
  * Typed models for Core API and their mapping
  * Creation of Schema
  * Feature enum and "FeatureInit"

## Generic "Versioned" data model

In terms of versioning conventions, I believe we are going to:
* Version substate contents for all native substates at the application layer
* Not version function arguments, because we can create v2 methods in future if we need to.

When implementing for X = Schema, I decided the following naming schema fitted best - keeping code looking as neat as possible:
* `VersionedX` is the versioned enum type. This will be used for substate contents, or other types which need versioning.
* `X` is a type alias created for "latest version of this type"
* `XV1` is the concrete type which should be created, which maps to the type alias.

Questions to be answered before proceeding:
* As part of this change, do we wish to create macros to define Native State types more easy / clear / type-safe?
* If so, I'll implement that macro... then I'll work native blueprint-by-blueprint with separate PRs _into this branch_

## Changes for integrators

I'll merge this into the node. This shouldn't have any impact outside of Scrypto and Node.


radix-engine/src/blueprints/package/package.rs
```diff
@@ -1493,16 +1154,16 @@ impl PackageRoyaltyNativeBlueprint {
         if royalty_charge.is_non_zero() {
             let handle = api.kernel_open_substate(
                 receiver,
-                MAIN_BASE_PARTITION,
-                &PackageField::Royalty.into(),
+                PackagePartition::Field.as_main_partition(),
```
talekhinezh, line 1157: Again, would rather not rely on outside system code to specify what the main_partition is.

radix-engine/src/blueprints/macros.rs
```diff
@@ -0,0 +1,1197 @@
...
+                            PartitionDescription::Physical(partition_num) => Err(()),
+                        }
+                    }
+                }
+
+                impl TryFrom<PartitionOffset> for [<$blueprint_ident Partition>] {
+                    type Error = ();
+
+                    fn try_from(offset: PartitionOffset) -> Result<Self, Self::Error> {
+                        // NOTE: This should really be a fixed mapping - but it's hard to do with declarative macros
```
talekhinezh, line 505: The problem is this mapping really shouldn't be done here.

radix-engine/src/blueprints/package/substates.rs
```diff
@@ -0,0 +1,176 @@
...
+declare_native_blueprint_state! {
+    blueprint_ident: Package,
+    blueprint_snake_case: package,
+    features: {
+        package_royalty: {
+            ident: PackageRoyalty,
+            description: "Enables the package royalty substate",
+        }
+    },
+    instance_schema_types: {},
```
talekhinezh, line 14: generic_arguments?

dhedey:
Closing in favour of:

Replacement PR 2023 - merged: https://github.com/radixdlt/radixdlt-scrypto/pull/1358
Extension PR 2024 - draft: https://github.com/radixdlt/radixdlt-scrypto/pull/1744

PR #1319, opened 2023-08-10
Refactor Transaction Receipt

iamyulong:
## Summary

Various refactors to transaction receipt:
- Grouped fields based on their characteristics
- Added placeholder for finalization cost in receipt and added basic fee reserve support
- Renamed costs to align with product requirement

### For dApp Developers and Internal Integrators

`FeeSummary` in `CommitResult` has been split into 
- `CostingParameters`
- `TransactionCostingParameters`
- `TransactionFeeSummary`
- `TransactionFeeDetails`
- `FeeSource` (Commit only)
- `FeeDestination`  (Commit only)

Conceptually, transaction fee now is comprised of :
- Network execution (tipped)
- Network finalization (tipped)
- Tip
- Network storage
- Royalties


radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -336,14 +358,14 @@ where
...
                             ),
                             scrypto_encode(&BurnFungibleResourceEvent {
-                                amount: fee_summary.to_burn_amount(),
+                                amount: fee_destination.to_burn,
                             })
                             .unwrap(),
                         ));
 
                         // Finalize execution trace
                         let execution_trace =
```
0xOmarA, line 367: It would be nice if `execution_trace` was an `Option<TransactionExecutionTrace>` which we only have if the execution trace module is enabled. This allows clients such as the toolkit to be able to easily check if the receipt they have contains execution trace or not instead of checking if the execution trace is empty or not.

radix-engine/src/transaction/state_update_summary.rs
```diff
@@ -22,8 +22,10 @@ pub struct StateUpdateSummary {
     pub new_components: IndexSet<ComponentAddress>,
     pub new_resources: IndexSet<ResourceAddress>,
     pub new_vaults: IndexSet<InternalAddress>,
+    /// TODO: remove
```
0xOmarA, line 25: Is there currently anything missing to remove this?

0xOmarA (approved):
Looks good from my side! Left a few minor comments.

radix-engine/src/transaction/transaction_receipt.rs
```diff
@@ -15,13 +15,95 @@ use radix_engine_interface::blueprints::transaction_processor::InstructionOutput
...
+#[derive(Clone, ScryptoSbor)]
+pub struct TransactionReceipt {
+    /// Costing parameters
+    pub costing_parameters: CostingParameters,
+    /// Transaction costing parameters
+    pub transaction_costing_parameters: TransactionCostingParameters,
+    /// Transaction fee summary
+    pub fee_summary: TransactionFeeSummary,
+    /// Transaction fee detail
+    pub fee_details: TransactionFeeDetails,
```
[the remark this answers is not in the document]
iamyulong: Yes, we will make it an option. I was following existing pattern but agree Option will be more explicit.

radix-engine/src/transaction/transaction_executor.rs
iamyulong: Yes!

radix-engine/src/transaction/state_update_summary.rs
iamyulong: Core API still depends on it. We will be able to remove them once Node-side balance change accounting is implemented.

PR #1321, opened 2023-08-11
Account Events + Events clean up

0xOmarA:
## Summary

- Added a check to the `blueprint` macro to reject attribute macros not supported by the blueprint macro. Previously, if you added an `#[event(...)]` attribute macro instead of `#[events(...)]` you would get no compile time error and you would only get an error at runtime that the events are not registered.
- Split up the events for fungible and non-fungible vaults since they pre-date their split.
- Removed the following events since they were not being used:
    - Role Assignment - `SetAndLockOwnerRoleEvent`
    - Role Assignment - `LockRoleEvent`
    - Role Assignment - `SetAndLockRoleEvent`
- Added events for the account blueprint. Full event list can be found in the events module of the account blueprint.
- Minor clean ups to `to_typed_native_event`.

## Update Recommendations

* Everything in this PR is new. No changes to existing integrations or implementations is needed.


radix-engine/src/blueprints/account/events.rs
```diff
@@ -0,0 +1,49 @@
+use crate::types::*;
+use radix_engine_common::math::Decimal;
+use radix_engine_common::{ScryptoEvent, ScryptoSbor};
+use radix_engine_interface::blueprints::account::*;
+use sbor::rust::prelude::*;
+
+#[derive(ScryptoSbor, ScryptoEvent, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum WithdrawEvent {
+    Amount(ResourceAddress, Decimal),
+    Ids(ResourceAddress, BTreeSet<NonFungibleLocalId>),
```
iamyulong, line 10: How about we rename `Amount`/`Ids` to `Fungible`/`NonFungibles` for all?

iamyulong (approved):
Nice clean-ups!

PR #1323, opened 2023-08-11
Feature/formalize finalization costs

iamyulong:
## Summary

This PR adds finalization costs to account for additional work on Node side. Currently, they're
- Base cost
- Commit states cost
- Commit events cost
- Commit logs cost

### For dApp Developers

Be reminded that finalization costs are applied after the execution of manifest instructions.

`cost_unit_price` and `cost_unit_limit` have been renamed to `execution_cost_unit_price` and `execution_cost_unit_limit` respectively.

Also, new APIs `finalization_cost_unit_price` and `finalization_cost_unit_limit` have been added.


radix-engine/src/track/interface.rs
```diff
@@ -170,12 +170,22 @@ pub trait SubstateStore {
 #[derive(Debug, Clone, Copy)]
 pub enum StoreAccess {
     /// Some substate was read from database.
-    ReadFromDb(usize),
+    ReadFromDb(NodeId, usize),
     /// Non-existent substate was read from database.
-    ReadFromDbNotFound,
+    ReadFromDbNotFound(NodeId),
```
talekhinezh, line 175: If we're adding `NodeId`, is there a reason not to include the full substate path? `NodeId::PartitionNum::SubstateKey`

talekhinezh (approved):

radix-engine/src/track/interface.rs
iamyulong: It's primarily used for costing breakdown and so far we only need node id. 

Can definitely add more in future.

PR #1324, opened 2023-08-11
Add protocol update readiness summary to EpochChange event

LukasGasior1:
## Summary
This PR adds a summary of protocol update readiness signals to the EpochChange event. For storage space optimization it only includes significant entries, that is protocol updates having at least 10% of new epoch validators (by stake) that have signalled the readiness.


radix-engine/src/blueprints/consensus_manager/validator.rs
```diff
@@ -465,6 +481,7 @@ impl ValidatorBlueprint {
             LockFlags::read_only(),
         )?;
         let validator: ValidatorSubstate = api.field_read_typed(handle)?;
+        api.field_close(handle)?;
```
LukasGasior1, line 484: not sure if this was missed or skipped on purpose?

LukasGasior1: some tests fail after adding this, so I guess this must have been on purpose :man_shrugging: , reverted

talekhinezh: The field needs to remain open in order to have access to the vaults within. The system will close the field automatically. Otherwise, if you'd like you can close it at the end of the method but not required.

talekhinezh (approved):
Nice!

PR #1331, opened 2023-08-14
Add Runtime::generate_ruid for Scrypto

iamyulong:
## Summary

Adds `Runtime::generate_ruid() -> [u8; 32]` for Scrypto.


scrypto/src/runtime/runtime.rs
```diff
@@ -75,6 +75,11 @@ impl Runtime {
         ScryptoEnv.get_transaction_hash().unwrap()
     }
 
+    /// Returns the transaction hash.
+    pub fn generate_ruid() -> [u8; 32] {
+        ScryptoEnv.generate_ruid().unwrap()
```
talekhinezh, line 80: lol did we never remove the actual scrypto env call?

talekhinezh (approved):

scrypto/src/runtime/runtime.rs
iamyulong: Yeah, possibly due to some native things depending on it.

PR #1343, opened 2023-08-16
Refactor track substate size limiting

iamyulong:
## Summary

This PR refactors tracked substate size limiting, to use the total bytes instead of total number.

Note that I've added heap-related substate events, but the emission part will be implemented in a following PR.

### For Internal Integrators

If you're customising the substate size limits, you will need to reconfigure it to use "bytes".


radix-engine/src/track/interface.rs
```diff
@@ -208,17 +208,41 @@ pub enum StoreAccess {
...
-    /// A new entry has been added to track
-    /// System limits how many items that can be tracked.
-    NewEntryInTrack(CanonicalSubstateKey, usize),
+
+    /// A substate in track has been updated
+    UpdateSubstateInTrack {
+        /// The canonical substate key
+        canonical_substate_key: CanonicalSubstateKey,
+        /// Previous size of the substate, or `None` if it's a new entry
+        old_size: Option<usize>,
```
talekhinezh, line 217: Interesting! Can you add a comment that `old_size` refers to the current size before the update rather than the size in the underlying store. That was a little confusing to me.

radix-engine/src/track/track.rs
```diff
@@ -903,18 +960,32 @@ impl<'s, S: SubstateDatabase, M: DatabaseKeyMapper + 'static> SubstateStore for
             new_updates
         };
 
+        let num_iterations = tracked_iter.num_iterations;
+        drop(tracked_iter);
```
talekhinezh, line 964: Minor: probably better to use standard use of `{}` to limit scoping and adding `tracked_iter.num_iterations;` into the above code block

talekhinezh (approved):

PR #1344, opened 2023-08-17
Architecture: Add System Database Checker

talekhinezh:
## Summary
Add System Database Checker to check that the database matches all system layer invariants. To enable this for testing run with feature `--features post_run_db_check`

## Details
* Add `SystemDatabaseChecker` which scans a `SubstateDB` and verifies that the data underlying does not break any system invariants
* Fixed `TransactionTracker` map key which was using `Vec<u8>` when the schema required it to be `Hash`
* Cleaned up parts of `resim` to use the `SystemDatabaseReader` rather than reading directly from the database

## Testing
* All tests which use `TestRunner` may run the system database checker by simply enabling the `post_run_db_check` feature


radix-engine/src/system/system_db_reader.rs
```diff
@@ -140,6 +285,254 @@ impl<'a, S: SubstateDatabase> SystemDatabaseReader<'a, S> {
...
+        let (substs, allow_ownership, allow_non_global_refs) = match key_or_value {
+            KeyOrValue::Key => (
+                &target.kv_store_type.key_generic_substitutions,
+                false,
+                false,
+            ),
+            KeyOrValue::Value => (
+                &target.kv_store_type.value_generic_substitutions,
+                target.kv_store_type.allow_ownership,
+                false,
```
iamyulong, line 368: Should be depending on if the blueprint is transient?

radix-engine/src/system/system_db_checker.rs
```diff
@@ -0,0 +1,746 @@
+use crate::system::node_modules::type_info::TypeInfoSubstate;
```
iamyulong, line 1: Howe we move this file to `scrypto-tests`?

radix-engine/src/system/system_db_checker.rs
```diff
@@ -0,0 +1,746 @@
...
+    Object(BlueprintId),
+    KeyValueStore,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum SystemDatabaseCheckError {
+    NodeError(SystemNodeCheckError),
+    PartitionError(NodeInfo, SystemPartitionCheckError),
+}
+
```
iamyulong, line 137: Nice errors above!!!

radix-engine/src/system/system_db_reader.rs
```diff
@@ -140,6 +285,254 @@ impl<'a, S: SubstateDatabase> SystemDatabaseReader<'a, S> {
...
+    pub fn get_kv_store_type_target(&self, node_id: &NodeId) -> Option<KVStoreTypeTarget> {
+        let type_info = self.fetch_substate::<SpreadPrefixKeyMapper, TypeInfoSubstate>(
+            node_id,
+            TYPE_INFO_FIELD_PARTITION,
+            &TypeInfoField::TypeInfo.into(),
+        )?;
+
+        let kv_store_info = match type_info {
+            TypeInfoSubstate::KeyValueStore(kv_store_info) => kv_store_info,
+            _ => return None,
```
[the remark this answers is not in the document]
talekhinezh: No invariant is broken here (except for if the client passes in an erroneous node_id) so would rather keep this. More long term, this should be a `Result::Err` of some kind

radix-engine/src/system/system_db_reader.rs
```diff
@@ -140,6 +285,254 @@ impl<'a, S: SubstateDatabase> SystemDatabaseReader<'a, S> {
...
+    ) -> Option<BlueprintTypeTarget> {
+        let type_info = self.fetch_substate::<SpreadPrefixKeyMapper, TypeInfoSubstate>(
+            node_id,
+            TYPE_INFO_FIELD_PARTITION,
+            &TypeInfoField::TypeInfo.into(),
+        )?;
+
+        let object_info = match type_info {
+            TypeInfoSubstate::Object(object_info) => object_info,
+            _ => return None,
```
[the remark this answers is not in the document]
talekhinezh: ditto here

radix-engine/src/system/system_db_reader.rs
```diff
@@ -140,6 +285,254 @@ impl<'a, S: SubstateDatabase> SystemDatabaseReader<'a, S> {
...
+        } else {
+            None
+        }
+    }
+
+    pub fn get_kv_store_payload_schema(
+        &self,
+        target: &KVStoreTypeTarget,
+        key_or_value: KeyOrValue,
+    ) -> Option<ResolvedPayloadSchema> {
```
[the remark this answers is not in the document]
talekhinezh: This also depends on if client is passing in a valid `target` or not.

radix-engine/src/system/system_db_reader.rs
talekhinezh: Not sure what you mean here?

radix-engine/src/system/system_db_reader.rs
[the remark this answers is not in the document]
iamyulong: I think it's more about the contract of the interface.

The method says `get_kv_store_type_target` so the callers expect a kv store, so it can't tell if it's missing OR it's of wrong type.

This is similar to `get_blueprint_id` which will panic if it's not an object (of blueprint).

radix-engine/src/system/system_db_reader.rs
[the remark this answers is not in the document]
iamyulong: You're right. I missed the `?` in `self.get_schema(&target.meta, &type_identifier.0)?;`.

radix-engine/src/system/system_db_reader.rs
iamyulong: The third element of the tuple `allow_non_global_ref` should depend on if the owning object is transient or not, rather than the fact this is a key value store?

radix-engine/src/system/system_db_checker.rs
iamyulong: I was slightly worried about this logic being used in the `src`, so it oversimplifies the system, although it's sufficient for testing purpose.

How about we add `#[cfg(test)]` to this file, if we want it in `radix-engine` instead of a tests folder?

radix-engine/src/system/system_db_reader.rs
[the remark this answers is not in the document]
talekhinezh: I've updated interface to return more specific errors now

radix-engine/src/system/system_db_reader.rs
talekhinezh: Ah, for the case of key value stores this is hardcoded (especially since no current transient object has key value stores). We could formalize this by saying "non global refs are only allowed in the schema defined parts of a transient object"

iamyulong (approved):

talekhinezh:
Things to fix:
* Add to CI
* Add conditional compilation for `system_db_checker`

PR #1355, opened 2023-08-19
Architecture: Cleanup Scrypto API

talekhinezh:
## Summary
Cleanup Scrypto API

## Details
* Remove `Result` from `ScryptoVmV1Api`
* Remove `drop_object` from `ScryptoVmV1Api`
* Remove `actor_call_module` api from `ScryptoVmV1Api`
* Remove `get_auth_zone` api from `ScryptoVmV1Api`
* Add `key_value_entry_remove` api to `ScryptoVmV1Api`
* Add `get_usd_price` api to `ScryptoVmV1Api`
* Cleanup stub useage in `GenesisHelper`


radix-engine/src/vm/wasm/constants.rs
```diff
@@ -1,31 +1,32 @@
...
-pub const FEE_BALANCE_FUNCTION_NAME: &str = "fee_balance";
+// Object
+//=================
+pub const OBJECT_NEW_FUNCTION_NAME: &str = "object_new";
+pub const OBJECT_GLOBALIZE_FUNCTION_NAME: &str = "object_globalize";
+pub const OBJECT_GET_BLUEPRINT_ID_FUNCTION_NAME: &str = "object_get_blueprint_id";
+pub const OBJECT_GET_OUTER_OBJECT_FUNCTION_NAME: &str = "object_get_outer_object";
+pub const OBJECT_CALL_FUNCTION_NAME: &str = "object_call";
+pub const OBJECT_CALL_DIRECT_FUNCTION_NAME: &str = "object_call_direct";
+pub const OBJECT_CALL_MODULE_FUNCTION_NAME: &str = "object_call_module";
```
iamyulong, line 21: While I like most of "entity" prefix to the APIs, I prefer to have `invocation` as a dedicated category. 

Object is more the receiver of an invocation, rather than the definer. "calls" are defined on the system.

It could be something like:
```
call_method(node_id, ident)
call_module_method(node_id, module_id, ident)
call_direct_access_method(node_id, ident)
call_function(blueprint_id, ident)
```

radix-engine/src/vm/wasm/wasmi.rs
```diff
@@ -109,30 +109,51 @@ fn consume_buffer(
...
+    runtime
+        .object_call_direct(receiver, ident, args)
+        .map(|buffer| buffer.0)
+}
+
+fn call_module_method(
     mut caller: Caller<'_, HostState>,
     receiver_ptr: u32,
     receiver_len: u32,
     direct_access: u32,
```
iamyulong, line 156: direct_access?

radix-engine/src/vm/wasm/constants.rs
talekhinezh: I guess this is mostly matter of preference/frame of reference. I had it that way originally but found this version to have nice properties such as "if I have an object, what can I do with it", "if i have a blueprint, what can i do with it". 

That is, all of the APIs use objects as frames of reference rather than functions (ie object oriented).

Lets chat around this in our 1:1

radix-engine/src/vm/wasm/wasmi.rs
talekhinezh: good catch! surprised we don't have a test catching this. Will add one if missing

radix-engine/src/vm/wasm/constants.rs
iamyulong: Yep!

iamyulong (approved):

PR #1361, opened 2023-08-21
Returning substate system structures in system flash receipt

jakrawcz-rdx:
This properly fixes the 🛑 described at https://github.com/radixdlt/babylon-node/pull/620#issuecomment-1685974132.

### DIFF
--- a/radix-engine-tests/tests/bootstrap.rs
+++ b/radix-engine-tests/tests/bootstrap.rs
@@ -141,6 +141,7 @@ fn test_bootstrap_receipts_should_have_complete_system_structure() {
     assert_complete_system_structure(wrap_up_receipt.expect_commit_success());
 }
 
+// TODO(after RCnet-V3): this assertion could be re-used for other tests of non-standard receipts.
 fn assert_complete_system_structure(result: &CommitResult) {
     let SystemStructure {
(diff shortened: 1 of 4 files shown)


radix-engine-tests/tests/bootstrap.rs
```diff
@@ -94,6 +94,85 @@ fn test_bootstrap_receipt_should_match_constants() {
     assert_eq!(wrap_up_epoch_change.epoch, genesis_epoch.next());
 }
 
+#[test]
+fn test_bootstrap_receipts_should_have_complete_system_structure() {
```
dhedey, line 98: Nice!

You've written some useful code here, so I wonder if we should move it out to a separate `system_structure.rs` test file, like we've done with the `typed_substate_layout.rs` tests - so we can also use this helper code to validate (eg) the scenario receipts.

dhedey (approved):

radix-engine-tests/tests/bootstrap.rs
jakrawcz-rdx: Yeah, but it seems like a lower-prio refactor/improvement for later.
Let me add a TODO here and merge after CI's 🟢 .

PR #1367, opened 2023-08-22
Feature/add metadata validation

iamyulong:
## Summary

Add validation for URL and origin, from system at runtime.

## Testing

Additional tests added.

### For dApp Developers

All URLs and Origins are now checked by the system.

Method `get_metadata<T>(key: &str)` now returns a `Result<Option<T>, MetadataConversionError>`.


iamyulong:
If we're doing metadata validation in application space, using crate `url`.

```
validate_origin(x.as_str()).unwrap();
```

The `assets/blueprints/metadata` wasm file size increases from `187065` bytes to `477487` bytes, so the overhead is about 300KB.

iamyulong:
If we use crate `regex` with regular expression (from [here](https://uibakery.io/regex-library/url-regex-java)):
```
"^https?:\\/\\/(?:www\\.)?[-a-zA-Z0-9@:%._\\+~#=]{1,256}\\.[a-zA-Z0-9()]{1,6}\\b(?:[-a-zA-Z0-9()@:%_\\+.~#?&\\/=]*)$"
```

The assets/blueprints/metadata wasm file size increases from 187065 bytes to 225329 bytes, so the overhead is about 38KB.

iamyulong:
With regular expression, the file size increased to `1152602`. The size increases by `960 KB`.

iamyulong:
Either way, we shouldn't do the validation in the wasm, given the size overhead.

examples/no-std/Cargo.lock
```diff
@@ -292,13 +292,39 @@ dependencies = [
...
+ "regex",
  "sbor",
  "scrypto-schema",
  "serde_json",
  "strum",
  "utils",
 ]
 
+[[package]]
+name = "regex"
```
talekhinezh, line 304: Do these need to be included now?

talekhinezh (approved):

examples/no-std/Cargo.lock
iamyulong: Yes, it's part of `radix-engine-interface` for wider reusability.

PR #1372, opened 2023-08-23
Feature: new "complete vault balance changes" receipt field to replace the 2 obsolete ones

jakrawcz-rdx:
This addresses the TODOs:
https://github.com/radixdlt/radixdlt-scrypto/blob/5657d3dacf81597bf5e86af07c338c145f7d0b3e/radix-engine/src/transaction/state_update_summary.rs#L24-L28

As discussed with Yulong:
The complete vault-based view of balance changes will allow Node to easily map it into account-based view (using the parent-map that we maintain in Node's DB).
In theory, Node could compute the vault-based balance changes itself, but they are still useful on the Engine's side (in tests).
In fact, for test purposes I had to implement the calculation of account-based changes too (to retro-fit the asserts which are based on accounts).


scrypto-unit/src/test_runner.rs
```diff
@@ -2166,6 +2174,44 @@ impl<E: NativeVmExtension, D: TestDatabase> TestRunner<E, D> {
...
+
+impl<'d, D: SubstateDatabase> DescendantVaults<'d, D> {
+    pub fn new(database: &'d D) -> Self {
+        Self { database }
+    }
+
+    pub fn get_all(&self, node_id: &NodeId) -> IndexMap<ResourceAddress, Vec<NodeId>> {
+        let mut vault_finder = VaultFinder::new();
+        let mut traverser = StateTreeTraverser::new(self.database, &mut vault_finder, 100);
+        traverser.traverse_all_descendents(None, *node_id);
```
iamyulong, line 2189: It seems that `traverse_all_descendents` visit the "root" node as well. 

Maybe we should give it a better name? Also impacts `DescendantVaults`.

iamyulong: I'm terrible at naming. Can only think of `traverse_node_tree` and `VaultsInNodeTree`.

jakrawcz-rdx: `traverse_subtree()`?

iamyulong (approved):
LGTM

jakrawcz-rdx:
thanks for the timely review 👍

PR #1378, opened 2023-08-24
Feature/clean up scrypto api

iamyulong:
## Summary

Some clean-ups to Scrypto Lib.

## Details

- Fixed missed return of role assignment methods
- Updated `AuthZone::pop` to return an `Option<Proof>`

### For dApp Developers

`AuthZone` pop interface is slightly changed. See above.


scrypto/src/component/component.rs
```diff
@@ -372,13 +372,13 @@ impl<O: HasStub> Global<O> {
         }
     }
 
-    fn metadata(&self) -> Attached<Metadata> {
+    pub fn metadata(&self) -> Attached<Metadata> {
```
talekhinezh, line 375: I've tried to keep these private for the most part as `Attached<Metadata>` exposes quite some nastiness underneath. Is there a reason to make these public?

talekhinezh (approved):

PR #1389, opened 2023-08-26
Bugfix: Fix Validator get_redemption_value()

talekhinezh:
## Summary
Fix Decimal computation bug in `get_redemption_value` of Validator Component

### DIFF
--- a/radix-engine/src/blueprints/consensus_manager/validator.rs
+++ b/radix-engine/src/blueprints/consensus_manager/validator.rs
@@ -541,40 +541,24 @@ impl ValidatorBlueprint {
     where
         Y: ClientApi<RuntimeError>,
     {
-        Self::stake_internal(xrd_bucket, api)
+        Self::stake_internal(xrd_bucket, true, api)
     }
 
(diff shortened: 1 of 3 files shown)


radix-engine/src/blueprints/consensus_manager/validator.rs
```diff
@@ -155,6 +155,8 @@ impl NonFungibleData for UnstakeData {
 #[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
 pub enum ValidatorError {
     InvalidClaimResource,
+    InvalidGetRedemptionAmount,
+    UnexpectedDecimalComputationError,
```
lrubasze, line 159: It would be great to have a test for this error.

lrubasze (approved):

radix-engine/src/blueprints/consensus_manager/validator.rs
talekhinezh: I believe with the current constraints on the resource system that this error would be impossible to get to (hence the "Unexpected" part of the error name)

PR #1390, opened 2023-08-26
Bugfix: Fix Pool Decimal Handling

talekhinezh:
## Summary
Add checks and errors in case of `Decimal` overflows to the Pool package

### DIFF
--- a/radix-engine-tests/tests/pool_two_resource.rs
+++ b/radix-engine-tests/tests/pool_two_resource.rs
@@ -936,8 +936,18 @@ fn get_redemption_value_should_not_panic_on_large_values() {
     );
     receipt.expect_commit_success();
 
-    // Act / Assert
-    test_runner.call_get_redemption_value(Decimal::MAX, true);
+    // Act
+    let receipt = test_runner.call_get_redemption_value(Decimal::MAX, true);
+
+    // Assert
+    receipt.expect_specific_failure(|e| {
+        matches!(
+            e,
+            RuntimeError::ApplicationError(ApplicationError::TwoResourcePoolError(
+                TwoResourcePoolError::InvalidGetRedemptionAmount
+            ))
+        )
+    })
 }
 
 fn is_pool_emitter(event_type_identifier: &EventTypeIdentifier) -> bool {
--- a/radix-engine-tests/tests/pool_two_resource.rs
+++ b/radix-engine-tests/tests/pool_two_resource.rs
@@ -968,6 +968,52 @@ fn get_redemption_value_should_not_panic_on_large_values() {
     })
 }
 
+#[test]
+fn contributing_to_a_pool_with_very_large_difference_in_reserves_succeeds() {
+    // Arrange
+    let max_mint_amount = Decimal(I192::from(2).pow(152));
+    let mut test_runner = TestEnvironment::new((18, 18));
+
+    let manifest = {
+        let mut manifest_builder = ManifestBuilder::new();
+        for _ in 0..10 {
+            manifest_builder =
+                manifest_builder.mint_fungible(test_runner.pool_resource1, max_mint_amount)
+        }
+
+        manifest_builder
+            .mint_fungible(test_runner.pool_resource2, dec!("0.000000000000000001"))
+            .take_all_from_worktop(test_runner.pool_resource1, "resource_1")
+            .take_all_from_worktop(test_runner.pool_resource2, "resource_2")
+            .with_name_lookup(|builder, lookup| {
+                let bucket1 = lookup.bucket("resource_1");
+                let bucket2 = lookup.bucket("resource_2");
+                builder.call_method(
+                    test_runner.pool_component_address,
+                    TWO_RESOURCE_POOL_CONTRIBUTE_IDENT,
+                    TwoResourcePoolContributeManifestInput {
+                        buckets: (bucket1, bucket2),
+                    },
+                )
+            })
+            .try_deposit_batch_or_abort(test_runner.account_component_address, None)
+            .build()
+    };
+    test_runner
+        .execute_manifest(manifest, true)
+        .expect_commit_success();
+
+    // Act
+    let receipt = test_runner.contribute(
+        (test_runner.pool_resource1, dec!("1")),
+        (test_runner.pool_resource2, dec!("1")),
+        true,
+    );
+
+    // Assert
+    receipt.expect_commit_success();
+}
+
 fn is_pool_emitter(event_type_identifier: &EventTypeIdentifier) -> bool {
     match event_type_identifier.0 {
         Emitter::Method(node_id, ObjectModuleId::Main) => match node_id.entity_type() {
--- a/radix-engine-tests/tests/pool_two_resource.rs
+++ b/radix-engine-tests/tests/pool_two_resource.rs
@@ -5,6 +5,7 @@ use radix_engine::types::*;
 use radix_engine_interface::api::node_modules::metadata::MetadataValue;
 use radix_engine_interface::api::ObjectModuleId;
 use radix_engine_interface::blueprints::pool::*;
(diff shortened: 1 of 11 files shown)


radix-engine/src/blueprints/pool/one_resource_pool/blueprint.rs
```diff
@@ -569,18 +577,20 @@ impl OneResourcePoolBlueprint {
...
         pool_resource_divisibility: u8,
-    ) -> Decimal {
+    ) -> Result<Decimal, RuntimeError> {
         let amount_owed = pool_units_to_redeem
             .safe_mul(pool_resource_reserves)
-            .unwrap()
+            .ok_or_else(|| OneResourcePoolError::DecimalOverflowError)?
             .safe_div(pool_units_total_supply)
-            .unwrap();
+            .ok_or_else(|| OneResourcePoolError::DecimalOverflowError)?;
```
0xOmarA, line 585: Two minor comments about this: It might be better to use `.and_then` here + `.ok_or`. This will transform this code from:

```rust
let amount_owed = pool_units_to_redeem
    .safe_mul(pool_resource_reserves)
    .ok_or_else(|| OneResourcePoolError::DecimalOverflowError)?
    .safe_div(pool_units_total_supply)
    .ok_or_else(|| OneResourcePoolError::DecimalOverflowError)?;
```

To be:

```rust
let amount_owed = pool_units_to_redeem
    .safe_mul(pool_resource_reserves)
    .and_then(|value| value.safe_div(pool_units_total_supply))
    .ok_or(OneResourcePoolError::DecimalOverflowError)?
```

talekhinezh: good point!

radix-engine/src/blueprints/pool/two_resource_pool/blueprint.rs
```diff
@@ -436,20 +436,17 @@ impl TwoResourcePoolBlueprint {
                 (false, false, false) => Ok((
```
0xOmarA, line 436: (I'm writing out some of my reasoning as I go through the review, apologies if the comments are somewhat lengthy).

I will be referring to `Decimal::Max` as $`\textrm{Max}_{\textrm{Decimal}}`$. I am operating with the assumption that the amount of resources in fungible buckets has a theoretical upper bound of $\textrm{Max}_{\textrm{Decimal}}$ with the understanding that the theoretical and practical limits are different. 

Currently, this bit of code is doing the following:

```math
mint\_amount = \sqrt{ contribution_1 \times contribution_2 }
```

The expression inside the square root has an upper bound of:

```math
contribution_1 \times contribution_2 = \textrm{Max}_{\textrm{Decimal}} \times \textrm{Max}_{\textrm{Decimal}} = \textrm{Max}_{\textrm{Decimal}}^2
```

Meaning that it can indeed overflow. We can refactor this into something that's guaranteed to not overflow if we express it as:

```math
mint\_amount = \sqrt{ contribution_1 \times contribution_2 } = \sqrt{contribution_1} \times \sqrt{contribution_2}
```

Expressing it in the way seen on the RHS of the equation above is safer for two reasons:

* It's impossible to overflow in this case and an unwrap is safe to do (we can choose not to unwrap and return an error instead, that's fine)
* Contributing anything less than or equal to $\textrm{Max}_{\textrm{Decimal}}$ works with no issues.
* If loss of precision is a concern then we can do all of the math with `PreciseDecimal` and then convert back to `Decimal` at the end.

radix-engine/src/blueprints/pool/two_resource_pool/blueprint.rs
```diff
@@ -436,20 +436,17 @@ impl TwoResourcePoolBlueprint {
...
                         .safe_mul(contribution2)
-                        .unwrap()
-                        .sqrt()
-                        .unwrap(),
+                        .and_then(|d| d.sqrt())
+                        .ok_or_else(|| TwoResourcePoolError::DecimalOverflowError)?,
                     contribution1,
                     contribution2,
                 )),
                 (false, _, _) => Ok((
```
0xOmarA, line 444: In a similar manner to the case above, I believe that we should expand out the roots for this to be:

```math
mint\_amount = \sqrt{(contribution_1 + reserves_1) * (contribution_2 + reserves_2)} = \sqrt{(contribution_1 + reserves_1)} * \sqrt{(contribution_2 + reserves_2)}
```

Where $`contribution_n + reserves_n`$ could overflow.

radix-engine/src/blueprints/pool/one_resource_pool/blueprint.rs
```diff
@@ -569,18 +578,20 @@ impl OneResourcePoolBlueprint {
         pool_units_total_supply: Decimal,
         pool_resource_reserves: Decimal,
         pool_resource_divisibility: u8,
-    ) -> Decimal {
+    ) -> Result<Decimal, RuntimeError> {
         let amount_owed = pool_units_to_redeem
```
0xOmarA, line 582: I believe that we should do the division first here since `pool_units_to_redeem` and `pool_resource_reserves` both have an upper bound of `Decimal::MAX` and thus multiplying them first would lead to an overflow if they're large even if the pool units can be claimed.

radix-engine/src/blueprints/pool/one_resource_pool/blueprint.rs
```diff
@@ -363,13 +362,15 @@ impl OneResourcePoolBlueprint {
             reserves > Decimal::ZERO,
         ) {
             (false, false) => Ok(amount_of_contributed_resources),
-            (false, true) => Ok(amount_of_contributed_resources.safe_add(reserves).unwrap()),
+            (false, true) => Ok(amount_of_contributed_resources
+                .safe_add(reserves)
+                .ok_or_else(|| OneResourcePoolError::DecimalOverflowError)?),
             (true, false) => Err(OneResourcePoolError::NonZeroPoolUnitSupplyButZeroReserves),
             (true, true) => Ok(amount_of_contributed_resources
                 .safe_mul(pool_unit_total_supply)
```
0xOmarA, line 370: We should be doing the division first here and then the multiplication so that we do not overflow.

radix-engine-tests/tests/pool_two_resource.rs
```diff
@@ -899,6 +900,56 @@ pub fn contribute_fails_without_proper_authority_present() {
     receipt.expect_specific_failure(is_auth_error)
 }
 
+#[test]
+fn contribution_of_large_values_should_not_cause_panic() {
```
0xOmarA, line 904: This should actually succeed and not fail. 

Contributing two buckets of max mint amount should mean that the mint the following amount of pool units:

```math
mint\_amount = \sqrt{2^{125}} \times \sqrt{2^{125}} = 2^{125}
```

So it should actually work. I will refactor this test so we're contributing _more_ than the max mint amount.

radix-engine-tests/tests/pool_two_resource.rs
```diff
@@ -899,6 +900,56 @@ pub fn contribute_fails_without_proper_authority_present() {
...
+
+    // Act
+    let receipt = test_runner.contribute(
+        (test_runner.pool_resource1, max_mint_amount),
+        (test_runner.pool_resource2, max_mint_amount),
+        true,
+    );
+
+    // Assert
+    receipt.expect_specific_failure(|e| {
```
0xOmarA, line 917: This should actually succeed and not fail. 

Contributing two buckets of max mint amount should mean that the mint the following amount of pool units:

```math
mint\_amount = \sqrt{2^{125}} \times \sqrt{2^{125}} = 2^{125}
```

So it should actually work. I will refactor this test so we're contributing _more_ than the max mint amount.

0xOmarA (approved):

PR #1391, opened 2023-08-26
Feature/scrypto unit testing

0xOmarA:
## Summary

The implementation of Scrypto-Test as seen REP-97.

## Details

This PR implements scrypto-test, the unit testing framework described in REP-97 that follows a different approach to testing and makes testing more accessible to developers.

There isn't much else to have here in the details since everything is as described in the REP.

## Update Recommendations

It's recommended that you use this testing framework for unit tests or smaller tests in general when you want to have a large degree of control over the environment while the `TestRunner` is preserved for larger tests or integration tests when you want to make sure that the dApp behaves as expected in transactions.


assets/template/Cargo.toml_template
```diff
@@ -2,6 +2,7 @@
 name = "${package_name}"
 version = "0.1.0"
 edition = "2021"
+resolver = "2"
```
0xOmarA, line 5: I'm putting this comment here to explain why I made those changes to the Cargo.toml file template.

With this PR, the blueprint macro generates the test bindings for us. Those test bindings are feature gated behind the `test` feature. This allows us to:

* Be able to use the test bindings in tests.
* Not increase the size of the WASM since the bindings are feature gated.

So, we would like to test this crate with the `test` feature enabled to get the test bindings. This requires us to use the new feature resolver and to add this crate as a dev-dependency with the test feature enabled.

radix-engine/src/kernel/call_frame.rs
```diff
@@ -224,8 +224,8 @@ pub enum StableReferenceType {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub struct TransientReference {
-    ref_count: usize,
-    ref_origin: ReferenceOrigin,
+    pub ref_count: usize,
```
talekhinezh, line 227: Are these `pub` still required?

talekhinezh (approved):

radix-engine/src/kernel/call_frame.rs
0xOmarA: Yup, no longer needed. I've reverted the changes.

PR #1393, opened 2023-08-27
Bugfix: Fix take_advanced method in Bucket/Vault

talekhinezh:
## Summary
Fix take_advanced method in Bucket/Vault

### DIFF
--- a/radix-engine-tests/tests/resource.rs
+++ b/radix-engine-tests/tests/resource.rs
@@ -429,6 +429,28 @@ fn test_fungible_resource_take_advanced() {
 }
 
 #[test]
+fn fungible_bucket_take_advanced_max_should_not_panic() {
+    // Arrange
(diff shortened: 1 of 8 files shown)


lrubasze (approved):

radix-engine-tests/tests/resource.rs
```diff
@@ -428,6 +428,94 @@ fn test_fungible_resource_take_advanced() {
...
+            package_address,
+            "RoundingTest",
+            "fungible_bucket_take_advanced_max",
+            manifest_args!(),
+        )
+        .build();
+    let receipt = test_runner.execute_manifest(manifest, vec![]);
+
+    // Assert
+    receipt.expect_failure();
```
lrubasze, line 450: Really minor, but maybe we could expect specific failure here.

PR #1394, opened 2023-08-27
Test: Fuzz Pools

talekhinezh:
## Summary
Implement Pool fuzzers


lrubasze:
I am a huge fun of adding more specialized fuzzers, this is exactly what RE needs!
In our RE there is a fuzzing framework in `fuzz-tests` folder, which I think could be a nice place to implement all of those fuzzers. Fuzzers could benefit from utilizing `AFL` or `libfuzzer`.
What do you think about moving it to `fuzz-tests`?

PR #1395, opened 2023-08-27
Architecture: Ability to recover from panics in native blueprints.

0xOmarA:
## Summary
```
INSTRUCTIONS:
Add summary - one or two sentences explaining the purpose of this PR.
```

## Details
```
INSTRUCTIONS:
Provide further details about the changes, or how they fit into the roadmap.
You can delete this section if it's not useful.
```

## Testing
```
INSTRUCTIONS:
Further details about the tests you've added or manually performed.
You can delete this section if it's not useful.
```

## Update Recommendations
```
INSTRUCTIONS:
This section is to provide recommendations to consumers of this repo about how they
should handle breaking changes, or integrate new features. The two key stakeholder
groups are dApp Developers and Internal Integrators, and there are separate sections
for each.

In order to allow us to compile aggregated update instructions across PRs, please tag the PR
with 0+ of the relevant labels:
* scrypto-lib - Any change to the scrypto library
* sbor - Any breaking change to SBOR encoding/decoding
* manifest - Any change to manifest display, compilation/decompilation
* transaction - Any change which affects the compiled manifest, or transaction semantics
* substate - Any change to substates, the state model, or what's stored in the DB
* native-blueprint-interface - Any change to the interface of native blueprints
* files-moved - Any change to where engine types have moved, which will require
  internal integrators to update their imports

If you have a breaking change which doesn't fix into a category above, then tag it with
the closest label, and discuss in slack/discord.

If your PR contains no breaking changes or new features or hasn't been tagged, this whole
section can be deleted.
```

### For dApp Developers
```
INSTRUCTIONS:
Migration recommendations for a dApp developer to update their dApp/integrations
due to to the change/s in this PR.

These will be aggregated by the Developer Ecosystem team and included in the next Scrypto migration guide
(eg https://docs-babylon.radixdlt.com/main/scrypto/release_notes/migrating_from_0.7_to_0.8.html )
```

### For Internal Integrators
```
INSTRUCTIONS:
Instructions to any internal integrations (eg Node, Toolkit, Gateway, Ledger App) for how the changes may affect
them and recommendations for how they should update to support this change.
```

### DIFF
--- a/radix-engine/src/vm/native_vm.rs
+++ b/radix-engine/src/vm/native_vm.rs
@@ -7,7 +7,7 @@ use crate::blueprints::pool::PoolNativePackage;
 use crate::blueprints::resource::ResourceNativePackage;
 use crate::blueprints::transaction_processor::TransactionProcessorNativePackage;
 use crate::blueprints::transaction_tracker::TransactionTrackerNativePackage;
-use crate::errors::{ApplicationError, NativeRuntimeError, RuntimeError, VmError};
+use crate::errors::{NativeRuntimeError, RuntimeError, VmError};
 use crate::kernel::kernel_api::{KernelNodeApi, KernelSubstateApi};
 use crate::system::node_modules::metadata::MetadataNativePackage;
 use crate::system::node_modules::role_assignment::RoleAssignmentNativePackage;
@@ -90,7 +90,8 @@ impl<I: VmInvoke> VmInvoke for NativeVmInstance<I> {
     where
         Y: ClientApi<RuntimeError> + KernelNodeApi + KernelSubstateApi<SystemLockData>,
     {
-        let func = || match self {
+        #[allow(unused_mut)]
+        let mut func = || match self {
             NativeVmInstance::Extension(e) => e.invoke(export_name, input, api),
             NativeVmInstance::Native {
                 native_package_code_id,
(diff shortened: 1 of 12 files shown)


radix-engine/src/vm/native_vm.rs
```diff
@@ -216,3 +235,19 @@ impl<C: VmInvoke + Clone> NativeVmExtension for OverridePackageCode<C> {
...
     }
 }
+
+/// This function catches panic unwinds in the callbacks passed to it. If the crate is build with
+/// no-std then this function will not be able to catch the panics since [`catch_unwind`] is in the
+/// Rust standard library and not available to no_std:
+/// https://github.com/rust-lang/rfcs/issues/2810
+///
+/// [`catch_unwind`]: std::panic::catch_unwind
+fn catch_unwind<F: FnOnce() -> R, R>(f: F) -> Result<R, Box<dyn core::any::Any + Send + 'static>> {
```
0xOmarA, line 245: Note: we can't catch panics when building the engine with no-std since `std::panic::catch_unwind` is defined in the standard library. I'm mentioning this since one of our CI jobs runs the radix-engine-tests with no-std.

talekhinezh: Would prefer for this to just be more explicit. Rather than a call to `catch_unwind` which may or may not catch panics better to call `catch_unwind` if `std` and a direct call if `no_std`

0xOmarA: Agree that it's nicer. I've made that change.

PR #1396, opened 2023-08-27
Feature: Integrate `wasm-opt` into the `scrypto build` command.

0xOmarA:
## Summary

This PR integrates `wasm-opt` into `scrypto build` to help minimize the size of the built WASM.

## Details

This PR adds a new feature flag to the simulator crate which is enabled by default: the `wasm-opt` feature.

When this feature is enabled, the `scrypto build` command will minimize the size of the built WASM by using wasm-opt with aggressive size optimization.

### For dApp Developers

* When you build your packages after this PR you will notice that they're smaller and that you no longer need to apply wasm-opt yourself to minimize their size.
* The release profile has changed a little bit. You should update your Cargo release profile to minimize the size of your WASM builds.

### For Internal Integrators

None


lrubasze:
Asset sizes shrunk by 80% - it is so cool, never expected such gain 👏 
There is a small performance drop , eg.
- costing::instantiate_radiswap `+13%`
- costing::validate_wasm `+16.7%`
but I think this is acceptable.

update-assets.sh
```diff
@@ -10,61 +10,16 @@ scrypto="cargo run --manifest-path $PWD/simulator/Cargo.toml --bin scrypto $@ --
...
-  ./target/wasm32-unknown-unknown/release/metadata.rpd \
-  ../metadata.rpd
-
-echo "Done!"
+for crate_name in "faucet" "radiswap" "flash_loan" "genesis_helper" "metadata"
+do
+  echo "Building $crate_name..."
+  (cd $crate_name; $scrypto build)
+
+  for extension in "wasm" "rpd"
```
lrubasze, line 18: To avoid loop one could also do this:

`cp ./target/wasm32-unknown-unknown/release/$crate_name.{wasm,rpd} ..`

lrubasze (approved):

update-assets.sh
```diff
@@ -10,61 +10,16 @@ scrypto="cargo run --manifest-path $PWD/simulator/Cargo.toml --bin scrypto $@ --
...
-  -Os -g \
-  --strip-debug --strip-dwarf --strip-producers \
-  -o ../metadata.wasm \
-  ./target/wasm32-unknown-unknown/release/metadata.wasm
-cp \
-  ./target/wasm32-unknown-unknown/release/metadata.rpd \
-  ../metadata.rpd
-
-echo "Done!"
+for crate_name in "faucet" "radiswap" "flash_loan" "genesis_helper" "metadata"
```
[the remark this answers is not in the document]
0xOmarA: hah, finally!

update-assets.sh
0xOmarA: Good point! I've made that change.

PR #1398, opened 2023-08-27
Test/fuzz consensus

talekhinezh:
## Summary
Add ConsensusManager/Validator fuzzing test

### DIFF
--- a/radix-engine-tests/tests/validator.rs
+++ b/radix-engine-tests/tests/validator.rs
@@ -5,6 +5,7 @@ use radix_engine::types::blueprints::transaction_processor::InstructionOutput;
 use radix_engine::types::*;
 use radix_engine_interface::blueprints::consensus_manager::{
     ValidatorAcceptsDelegatedStakeInput, VALIDATOR_ACCEPTS_DELEGATED_STAKE_IDENT,
+    VALIDATOR_GET_REDEMPTION_VALUE_IDENT,
 };
 use radix_engine_interface::blueprints::resource::FromPublicKey;
 use scrypto_unit::*;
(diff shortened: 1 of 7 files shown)


radix-engine/src/blueprints/consensus_manager/validator.rs
```diff
@@ -1572,9 +1590,17 @@ impl ValidatorBlueprint {
         } else {
             amount_of_stake_units
```
0xOmarA, line 1591: We're multiplying first and then dividing which makes this more easily prone to overflows.

I would recommend we refactor this to:

```rust
amount_of_stake_units.
    .safe_div(total_stake_unit_supply)
    .and_then(|d| d.safe_mul(active_stake_amount))
    .ok_or(ValidatorError::UnexpectedDecimalComputationError)?
```

0xOmarA: We also need to change `calculate_stake_unit_amount`.

radix-engine/src/blueprints/consensus_manager/validator.rs
```diff
@@ -1214,7 +1231,7 @@ impl ValidatorBlueprint {
             .and_modify(|pending_amount| {
                 *pending_amount = pending_amount
                     .safe_add(requested_stake_unit_amount)
-                    .unwrap()
+                    .unwrap_or(Decimal::MAX)
```
0xOmarA, line 1234: Not sure if I agree with this silently going to `Decimal::MAX` when we overflow instead of erroring out with something like "you have too much pending, claim some of them and then try again"

radix-engine-tests/tests/validator.rs
```diff
@@ -125,3 +126,120 @@ fn check_if_validator_accepts_delegated_stake() {
...
+    let validator_address = test_runner.new_staked_validator_with_pub_key(pub_key, account);
+
+    // Act
+    let receipt = test_runner.execute_manifest(
+        ManifestBuilder::new()
+            .lock_fee_from_faucet()
+            .call_method(
+                validator_address,
+                VALIDATOR_GET_REDEMPTION_VALUE_IDENT,
+                manifest_args!(Decimal::MIN),
```
0xOmarA, line 230: I think this should be `Decimal::ZERO`? :)

radix-engine/src/blueprints/consensus_manager/validator.rs
talekhinezh: Yup, I added this to our discussion list. Whatever approach we take in pools should be the same here.

radix-engine/src/blueprints/consensus_manager/validator.rs
talekhinezh: Right, this should always fail later when we actually try to move the amount from the Locked stake vault to  Unlocking stake vault. I think it's more of a question of whether to early exit or not in the function due to an overflow.

radix-engine-tests/tests/validator.rs
talekhinezh: good catch!

PR #1399, opened 2023-08-28
Adding code coverage for unit tests

balda-rdx:


shambupujar (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -81,21 +81,21 @@ jobs:
         os: [k8s-linux-runner]
     steps:
     - uses: actions/checkout@v3
-    - uses: actions-rs/toolchain@v1
+    - uses: radixdlt/public-iac-resuable-artifacts/rust-setup@feature/rust-coverage
```
lrubasze, line 84: reusable?

.github/workflows/ci.yml
```diff
@@ -33,46 +32,47 @@ jobs:
...
     name: Run SBOR unit tests
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: [k8s-linux-runner]
     steps:
-    - uses: actions/checkout@v3
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: 1.70.0
```
lrubasze, line 46: How is the rust version configured for code coverage tests?

.github/workflows/ci.yml
balda-rdx: yeah there's a typo in the repo error. nothing we can do I'm afraid

.github/workflows/ci.yml
balda-rdx: good question. so if you take a look at the current pipeline you'll see that there are usually 2 steps
1. setup rust
2. run tests with coverage
In step 1 you can set the toolchain version, which currently defaults to 1.70.0: https://github.com/radixdlt/public-iac-resuable-artifacts/blob/main/rust-setup/action.yml#L4-L7

PR #1400, opened 2023-08-28
Architecture: Ability to recover from panics in native blueprints & system

0xOmarA:
## Summary

This PR adds the logic to recover from panics that may occur in:

* Native blueprints.
* The system layer or below.

## Details

This PR adds two isolation boundary for panics where they are caught at:

1. When a native blueprint is invoked, there is an isolation boundary that exists in the native VM to catch all of the panics that may occur and turn them into results. Panics caught from native blueprints are turned into `RuntimeError::ApplicationError::NativeBlueprintExecutionPanic`
2. There is an isolation boundary at the `SystemService` so that any code in the system layer or below panics the panic will be caught and turned into a `RuntimeError::SystemError::SystemPanic`.

## Testing

I would appreciate a review of the tests for this to make sure that they're actually sound. They're two:

* `panics_can_be_caught_in_the_native_vm_and_converted_into_results`
* `panics_at_the_system_layer_or_below_can_be_caught`.

### For dApp Developers

No impact on dApp developers.

### For Internal Integrators

Instructions here mainly apply to the network team. 

* Make sure that you're using `panic = "unwind"` in the node release builds. 
* Since the panics are turned into Results, you may choose to do different things if a panic is encountered. E.g., a transaction with `RuntimeError::ApplicationError::NativeBlueprintExecutionPanic` may fail and still be comitted whereas it seems to be like a transaction with `RuntimeError::SystemError::SystemPanic` should be rejected. 
* An e2e test to make sure that panics can be caught in the node would be great.

## Limitation

Panic catching is limited to `std` only. 

All nodes must build with `std` enabled. Otherwise, the network will fork when there is a transaction that introduce panics in the native blueprints
- `std` variant will accept such transactions
- `no-std` variant will abort


radix-engine/src/errors.rs
```diff
@@ -411,7 +415,15 @@ pub enum ApplicationError {
...
-    Panic(String),
+    /// A panic was encountered in Scrypto code.
+    ScryptoBlueprintExecutionPanic(String),
+
+    /// A panic was encountered in Native code.
+    NativeBlueprintExecutionPanic {
+        export_name: String,
+        input: ScryptoValue,
+        error: String,
+    },
```
iamyulong, line 426: Can we revert the changes to the `Panic(String)`, which is intended for emitting message **before abortion**.

Instead, suggest to have `NativeRuntimeError::Trap(String)`

0xOmarA:
Note: benchmark appears to be failing because we added a new crate (`radix-engine-macros`) which the benchmark job can't find in the `release/rcnet-v3` branch to bench mark against.

radix-engine/src/errors.rs
```diff
@@ -411,6 +422,7 @@ pub enum ApplicationError {
     // TODO: this should never happen because of schema check?
     InputDecodeError(DecodeError),
 
+    /// A panic was encountered in Scrypto code.
```
iamyulong, line 425: Minor: It can be used for by both native and Scrypto. Maybe we should rename `Panic` to `PanicMessage`?

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -303,6 +303,17 @@ where
...
 
+                // Abort the process if an error has occurred in the system layer or below. The
+                // following code is only enabled when compiling with the standard library. It's
+                // impossible for us to come across a `SystemError::SystemPanic` in no_std since
+                // we can't catch those panics in no_std.
+                #[cfg(feature = "std")]
+                if let Err(RuntimeError::SystemError(SystemError::SystemPanic(..))) =
+                    interpretation_result
+                {
+                    std::process::abort()
```
iamyulong, line 314: Nice!

iamyulong (approved):
Maybe worth noting in the PR description.

---
## Limitation

Panic catching is limited to `std` only. 

All nodes must build with `std` enabled. Otherwise, the network will fork when there is a transaction that introduce panics in the native blueprints
- `std` variant will accept such transactions
- `no-std` variant will abort

---
Also, can we fix the benchmark?

0xOmarA:
@iamyulong Thanks for the review, you're right about the limitation that catching panics is limited only to std. In fact, I've been meaning to discuss this with you. With this, we're seeing differences in behavior between std and no_std which is not ideal in my opinion. Do we wish co continue supporting no_std? I believe that it's not currently being used.

radix-engine/src/errors.rs
0xOmarA: Sounds good!

PR #1412, opened 2023-08-30
Decimal arithmetic operation inspection

lrubasze:
## Summary
Inspect Decimal operations unwraps.
- replace (if necessary) them with proper error handling
- annotate those safe ones
- annotate those to be done later

NOTE: Unwraps in tests were left as is.

### DIFF
--- a/radix-engine/src/blueprints/consensus_manager/validator.rs
+++ b/radix-engine/src/blueprints/consensus_manager/validator.rs
@@ -826,7 +826,7 @@ impl ValidatorBlueprint {
         let validator_address: ComponentAddress =
             ComponentAddress::new_or_panic(api.actor_get_node_id(ACTOR_REF_GLOBAL)?.into());
         let new_sorted_key =
-            Self::to_sorted_key(new_registered, new_stake_amount, validator_address);
+            Self::to_sorted_key(new_registered, new_stake_amount, validator_address)?;
 
         let update = if let Some(cur_index_key) = &validator.sorted_key {
             if let Some(new_index_key) = &new_sorted_key {
@@ -904,7 +904,11 @@ impl ValidatorBlueprint {
                     ApplicationError::ValidatorError(ValidatorError::EpochUnlockHasNotOccurredYet),
                 ));
             }
-            unstake_amount = unstake_amount.safe_add(data.claim_amount).unwrap();
+            unstake_amount = unstake_amount.safe_add(data.claim_amount).ok_or(
+                RuntimeError::ApplicationError(ApplicationError::ValidatorError(
+                    ValidatorError::UnexpectedDecimalComputationError,
+                )),
+            )?;
         }
         nft_resman.burn(bucket, api)?;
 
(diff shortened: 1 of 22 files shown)


radix-engine/src/blueprints/consensus_manager/validator.rs
```diff
@@ -1589,13 +1612,13 @@ impl ValidatorBlueprint {
             Decimal::zero()
         } else {
             amount_of_stake_units
-                .safe_div(total_stake_unit_supply)
-                .and_then(|d| d.safe_mul(active_stake_amount))
-                .ok_or_else(|| {
-                    RuntimeError::ApplicationError(ApplicationError::ValidatorError(
+                .safe_mul(active_stake_amount)
```
talekhinezh, line 1615: We should do division before multiplication here to reduce chance of overflow as well as be more consistent with pools

radix-engine/src/blueprints/consensus_manager/validator.rs
```diff
@@ -1607,20 +1630,18 @@ impl ValidatorBlueprint {
...
-            xrd_amount
+        if total_stake_xrd_amount.is_zero() {
+            Ok(xrd_amount)
         } else {
             xrd_amount
-                .safe_div(total_stake_xrd_amount)
-                .and_then(|d| d.safe_mul(total_stake_unit_supply))
-                .ok_or_else(|| {
-                    RuntimeError::ApplicationError(ApplicationError::ValidatorError(
+                .safe_mul(total_stake_unit_supply)
```
talekhinezh, line 1637: We should do division before multiplication here to reduce chance of overflow as well as be more consistent with pools

radix-engine/src/blueprints/consensus_manager/validator.rs
lrubasze: good catch!

0xOmarA:
I found a `safe_sub().unwrap` here as well: https://github.com/radixdlt/radixdlt-scrypto/blob/fcd1b2d235a79bdb13790d2854ee6f54df4f3141/radix-engine/src/blueprints/resource/non_fungible/non_fungible_vault.rs#L419

radix-engine/src/blueprints/consensus_manager/validator.rs
lrubasze: Done

lrubasze:
@0xOmarA @talekhinezh found also these ones
- `put()` function
https://github.com/radixdlt/radixdlt-scrypto/blob/fcd1b2d235a79bdb13790d2854ee6f54df4f3141/radix-engine-interface/src/blueprints/resource/resource.rs#L53
- `take_by_amount()` function
https://github.com/radixdlt/radixdlt-scrypto/blob/fcd1b2d235a79bdb13790d2854ee6f54df4f3141/radix-engine-interface/src/blueprints/resource/resource.rs#L64
Is it fine to leave it as is?
`take_by_amount()` already returns an error, so this one is easy.
But `put()` does not, additionally it is used in many places (eg. `transaction_executor()`)

talekhinezh (dismissed):

talekhinezh (approved):

PR #1418, opened 2023-08-30
Add more rules to drop_object

talekhinezh:
Fix fungible bucket put bug


bbarwik (approved):

iamyulong (approved):
LGTM.

TODO: Review all function signatures that accept `Bucket`/`Proof`/`FungibleBucket`/`FungibleProof`/`NonFungibleBucket`/`NonFungibleProof`/`ResourceManager` to make sure the logic does not depends on the resource address; if so, add additional check in application space.

PR #1429, opened 2023-09-01
Fix for JMT losing track of lower-tier version after (specific) deletes

jakrawcz-rdx:
This PR fixes a problem discovered while implementing the "partition deletion support in JMT" (https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HioteqmFu91DtjNfFVJXfM4N) and is not its prerequisite.

The problem is exposed by a newly-added test https://github.com/radixdlt/radixdlt-scrypto/commit/90a2824a54543a73d8f3f9f4393dfb339af21a3a#diff-1af1904ec25c16634db8427a89ec05376f0d9d4973fe831f9578663c4ff9c2feR98 and described at https://github.com/radixdlt/radixdlt-scrypto/commit/b686ddacdfe0b403f1a47ad9574b64638e88d789#diff-38bf4cc229d9101c3b6f8ecb5d5d9707dcc784070c2d9c14f4cabbb3c6905031R126-R132

⚠️ The problem is **not** specific to _batch_ partition deletion.
Rather, it occurs every time _a last-but-one higher-tier leaf (e.g. partition or a ReNode) is entirely deleted (by multiple "single" deletes)_. Basically: this leaves the last higher-tier leaf without siblings, which makes JMT compact its node (promote it up, with longer prefix). This "infra" JMT operation does not change the hash of that leaf, **but it changes its version number** - and that's the source of the problem, since the logic before this PR assumed that "lower-tier root version == higher-tier leaf version" which I now see is false.

### DIFF
--- a/radix-engine-stores/src/hash_tree/hash_tree_facade.rs
+++ b/radix-engine-stores/src/hash_tree/hash_tree_facade.rs
@@ -1,7 +1,7 @@
 use sbor::rust::collections::HashMap;
 use sbor::rust::vec::Vec;
 
-use crate::hash_tree::types::LeafKey;
+use crate::hash_tree::types::{LeafKey, Version};
 
 use super::tree_store::{
     ReadableTreeStore, TreeChildEntry, TreeInternalNode, TreeLeafNode, TreeNode,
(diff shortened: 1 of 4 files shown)


jakrawcz-rdx:
Alright, it seems that the "wrong assumption" described above can be encountered much easier, even without deletes, but with a simple series of inserts :(
(the test added in the newest commit fails before this PR)

I am worried that these situations can also happen in regular deployment. The result is that state root hash is computed incorrectly (i.e. as if some previous parts of the tree are ignored by next updates).
That incorrect state root hash is consistent between Validators (since grouping of JMT updates is the same) and thus, luckily, we don't see non-determinism / hash mismatches (on BFT). However, we would not be able to construct valid Merkle proofs (of many substates) based on our malformed JMTs.

talekhinezh (approved):

PR #1433, opened 2023-09-01
Test WASM buffer boundaries

lrubasze:
## Summary
Added tests to verify size of buffers passed over WASM boundary

### DIFF
--- a/radix-engine-tests/tests/wasm_buffers.rs
+++ b/radix-engine-tests/tests/wasm_buffers.rs
@@ -1,5 +1,6 @@
 use radix_engine::errors::*;
 use radix_engine::system::system_modules::costing::SystemLoanFeeReserve;
+use radix_engine::system::system_modules::limits::TransactionLimitsError;
 use radix_engine::transaction::{CostingParameters, TransactionReceipt};
 use radix_engine::types::*;
 use radix_engine::vm::wasm::*;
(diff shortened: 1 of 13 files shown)


radix-engine-tests/tests/wasm_buffers.rs
```diff
@@ -0,0 +1,486 @@
...
+    );
+
+    // Act
+    let receipt = test_wasm_buffer_consume!(
+        test_runner,
+        component_address,
+        buffer_ptr = 0 // Invalid pointer
+    );
+    // Assert
+    receipt.expect_commit_success(); // This is actually success but should be failure
```
lrubasze, line 362: We've got success when invalid pointer is provided. Not really sure if it should be commit succes.
From the WASM perspective the pointer is fine, it is within memory limits: `0 - memory.max`

radix-engine-tests/tests/wasm_buffers.rs
```diff
@@ -0,0 +1,486 @@
...
+        test_runner,
+        component_address,
+        read = (4 * MB, 0, get_memory_len(4 * MB))
+    );
+
+    // Assert
+    receipt.expect_specific_failure(|e| match e {
+        // This error is really nasty and we should somehow prevent it from occuring.
+        // Especially that we know that transaction will fail for smaller sizes...
+        RuntimeError::VmError(VmError::Wasm(WasmRuntimeError::ExecutionError(message))) => {
```
lrubasze, line 251: I really don't like this error and I think it would good to prevent it somehow from occuring.

radix-engine-tests/tests/wasm_buffers.rs
lrubasze: Agreed that this is acceptable for WASM.

iamyulong (approved):

PR #1434, opened 2023-09-01
Add events to the Radiswap blueprint

0xOmarA:
## Summary

This PR adds events to the Radiswap blueprint & it's scenario.


talekhinezh:
We've reserved the `Outer` name for "Outer" objects (e.g. `ResourceManager` is an outer object of `Vault`). Can we replace the use of `outer` in this PR to be either "parent" or "global" or anything else :)?

0xOmarA:
@talekhinezh Makes sense, I've renamed it and should be ready for another review now.

talekhinezh (approved):

PR #1449, opened 2023-09-04
Implement automatic generation of blueprint stubs

0xOmarA:
## Summary

This PP adds automatic code generation for the native stubs exposed through scrypto lib.

## Details

* When you need to update the stubs run `update-bindings.sh`.
* Logic for bindings generation is in a new binary in the simulator crate called `scrypto-bindgen`.


0xOmarA:
I will sadly need to close this PR. This approach works for the most part, but there are some small things where it fails that make it an annoying approach to use.

1. Currently, all of the instantiation functions of native blueprints do not return typed objects. Meaning that they return a `ComponentAddress` that is seen in the schema validation as a `IsGlobalComponent` and not a `IsGlobalTyped` which we typically see with `Global<T>` objects. Thus, if we wish to have "idiomatic stubs" that use `Global<T>` and `Owned<T>` we need to make manual changes to the generated stubs. This need to make manual changes to the generated bindings defeats the purpose of automatically generating them.
2. The current approach produces some really unreadable stubs since it's using the machinery designed for macros. So, the entire stubs file ends up being a big mess of unreadable code. Code readability for stubs is not a must but is certainly nice to have. 

Some ideas on how to move forward post RCNet v3.1:

* Perhaps the most important out them all, I believe that native blueprints should return typed components and not just `ComponentAddress`es. We have decided that this approach is good for Scrypto and it should also be good for native. Seems like most of these typed structs would be hand-written which is fine.
* If we can move the package definitions to the `radix-engine-interface` crate, this would allow us to use a macro to generate those stubs instead of relying on a script that we may or may not run and need to have a check in CI that it ran.

radix-engine-common/src/data/scrypto/custom_well_known_types.rs
```diff
@@ -443,7 +443,7 @@ create_well_known_lookup!(
             ROLE_ASSIGNMENT_TYPES_START + 0,
             named_enum(
                 // This is a clearer name for UIs, but not time to change in the code right now
-                "Rule",
+                "AccessRule",
```
0xOmarA, line 446: @dhedey Needed to make changes to this file since code generation relies on the name in the metadata to match the name of the Rust type.

simulator/src/scrypto_bindgen/translation.rs
```diff
@@ -0,0 +1,310 @@
...
+                &ScopedTypeId(type_identifier.0, element_type),
+                blueprint_name,
+                schema_resolver
+            )?
+        )),
+        TypeKind::Tuple { field_types } => type_ident.unwrap_or(if field_types.is_empty() {
+            "()".to_owned()
+        } else {
+            format!(
+                "({},)",
```
iamyulong, line 135: The extra trailing `,` is to deal with single element tuple?

simulator/src/scrypto_bindgen/translation.rs
```diff
@@ -0,0 +1,310 @@
...
+                            schema_resolver
+                        )?,
+                        type_name(
+                            &ScopedTypeId(type_identifier.0, *err_type_index),
+                            blueprint_name,
+                            schema_resolver
+                        )?
+                    ))
+                }
+                (Some(name), ..) => Ok(name.to_owned()),
```
iamyulong, line 181: Question: This is the reason why the schema type name must match the Rust type name?

simulator/src/scrypto_bindgen/translation.rs
```diff
@@ -0,0 +1,310 @@
...
+                    OwnValidation::IsGlobalAddressReservation,
+                )) =>
+        {
+            "GlobalAddressReservation".to_owned()
+        }
+        TypeKind::Custom(ScryptoCustomTypeKind::Own) => match type_validation {
+            TypeValidation::Custom(ScryptoCustomTypeValidation::Own(
+                OwnValidation::IsTypedObject(package_address, bp_name),
+            )) if package_address.is_none()
+                || package_address == Some(schema_resolver.package_address()) =>
```
iamyulong, line 251: We should look up the address from database?

Also, minor, the `None` is unexpected (again, hopefully covered by tests).

simulator/src/scrypto_bindgen/translation.rs
```diff
@@ -0,0 +1,310 @@
...
+            is_mutable_receiver: false,
+        },
+        Some(ReceiverInfo {
+            ref_types: RefTypes::NORMAL | RefTypes::DIRECT_ACCESS,
+            receiver: Receiver::SelfRefMut,
+        }) => ast::FnType::Method {
+            is_mutable_receiver: true,
+        },
+        None => ast::FnType::Function,
+        _ => panic!("Invalid BitFlags for RefTypes"),
```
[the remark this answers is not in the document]
iamyulong: This is indeed checked [here](https://github.com/radixdlt/radixdlt-scrypto/blob/c347fccc1f0a44b643ad544210362d1a2591e1e1/radix-engine/src/vm/vm.rs#L275-L285) for Scrypto.

simulator/src/scrypto_bindgen/ast.rs
```diff
@@ -0,0 +1,99 @@
+use proc_macro2::{Ident, Span, TokenStream};
+use quote::{quote, ToTokens};
+use radix_engine_interface::types::PackageAddress;
+
+/// Objects of this struct are generated as part of the generation process. This struct can then be
+/// used inside the quote macro and printed out.
+pub struct BlueprintStub {
+    pub blueprint_name: String,
+    pub invocation_fn_signature_items: Vec<InvocationFn>,
```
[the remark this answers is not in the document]
0xOmarA: Sounds good, I've made that change.

simulator/src/scrypto_bindgen/schema.rs
```diff
@@ -0,0 +1,130 @@
...
+        type_identifier: &ScopedTypeId,
+    ) -> Result<TypeValidation<ScryptoCustomTypeValidation>, SchemaError>;
+
+    fn package_address(&self) -> PackageAddress;
+}
+
+pub fn generate_blueprint_stubs<S>(
+    package_definition: BTreeMap<BlueprintVersionKey, BlueprintDefinition>,
+    schema_resolver: &S,
+) -> Result<Vec<BlueprintStub>, SchemaError>
```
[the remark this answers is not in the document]
0xOmarA: Yup, I've renamed this and have begun calling them "interface"s instead of "stub"s in the implementation and it's much nicer!

simulator/src/scrypto_bindgen/schema.rs
```diff
@@ -0,0 +1,130 @@
...
+                    })
+            };
+            let arg_type_indices = {
+                let args_type_kind = schema_resolver.resolve_type_kind(args_type_identifier)?;
+                if let TypeKind::Tuple { field_types } = args_type_kind {
+                    Ok(field_types)
+                } else {
+                    Err(SchemaError::FunctionInputIsNotATuple(*args_type_identifier))
+                }
+            }?;
```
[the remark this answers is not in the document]
0xOmarA: Good point! I've made that change.

simulator/src/scrypto_bindgen/translation.rs
0xOmarA: Yup exactly. To the compiler a `(T)` is just `T`. But `(T,)` is a single element tuple. So, we force them to be tuples with the extra trailing comma.

simulator/src/scrypto_bindgen/translation.rs
0xOmarA: Yup, exactly! It's this bit and also bits like:

```rust
type_ident.unwrap_or(if field_types.is_empty() {
    "()".to_owned()
} else {
    format!(
        "({},)",
        field_types
            .iter()
            .map(|local_type_index| type_name(
                &ScopedTypeId(type_identifier.0, *local_type_index),
                schema_resolver
            ))
            .collect::<Result<Vec<String>, _>>()?
            .join(", ")
    )
})
```

where we prefer returning the `type_ident` that we got from the schema metadata over the types that we construct. So, there is a name in the metadata schema which is not to a rust type we end up returning that and getting a compile error in the final stubs since that type doesn't exist.

simulator/src/scrypto_bindgen/translation.rs
```diff
@@ -0,0 +1,310 @@
...
+        {
+            "Vault".to_owned()
+        }
+        TypeKind::Custom(ScryptoCustomTypeKind::Own)
+            if type_validation
+                == TypeValidation::Custom(ScryptoCustomTypeValidation::Own(
+                    OwnValidation::IsKeyValueStore,
+                )) =>
+        {
+            "NodeId".to_owned()
```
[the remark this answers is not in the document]
0xOmarA: `Own` :D

simulator/src/scrypto_bindgen/translation.rs
0xOmarA: Goot to know that the `None` case is not expected. I thought that a package address of `None` meant "this package".

About looking up the address, why would we want to look it up from the database? Is the check that the passed `PackageAddress` is equal to the one in the `TypeValidation` not sufficient?

iamyulong: I was wrong, despite implementing it myself. :(

`None` means the source package address **IF** the schema originates from a blueprint.

Can you revert the last change about this?

iamyulong (approved):

simulator/src/scrypto_bindgen/translation.rs
0xOmarA: No worries! Will revert now

radix-engine-common/src/data/scrypto/custom_well_known_types.rs
dhedey: Then we should change codegen IMO to have an override for native type names

iamyulong:
There are suggested improvements from @dhedey. Please do not merge before confirming.

update-bindings.sh
```diff
@@ -0,0 +1,44 @@
...
+use radix_engine_interface::blueprints::consensus_manager::*;
+
+use crate::prelude::*;
+
+//==================================================================================================
+// This file has been autogenerated by the ./update-bindings.sh script and none of the contents here
+// are hand-written. If you make any changes to the interface of native blueprints and need to regen
+// the bindings then run the update-bindings.sh script at the root of the repo.
+//
+// Note: there is currently no nice way to format this file since rustfmt doesn't format invocations
```
dhedey, line 21: This can be removed now ✨

scrypto/src/component/stubs.rs
```diff
@@ -27,12 +28,16 @@ extern_blueprint_internal! {
     },
     {
         fn free(&mut self) -> Bucket;
-        fn lock_fee(&mut self, amount: Decimal);
+        fn lock_fee(&mut self, amount: Decimal) -> ();
```
dhedey, line 31: Ideally we'd remove the `-> ()` in this case

dhedey: But we can do that later :)

assets/faucet.wasm
dhedey, line 1: Can we remove all these re-building + re-metering? I don't think they're needed changes

radix-engine-common/src/data/scrypto/custom_well_known_types.rs
dhedey: Just reminder to remove these as part of final markups :)

dhedey (approved):

update-bindings.sh
0xOmarA: Indeed, removing now.

scrypto/src/component/stubs.rs
0xOmarA: I believe our python script can do it, doing it now.

PR #1451, opened 2023-09-04
Minor clean-up of transaction receipt struct

jakrawcz-rdx:
As discussed on Slack, this PR:
- drops `database_updates` from receipt struct,
- fixes doc on batch vs individual changes.

PR #1454, opened 2023-09-04
Applying the "ordered single and batch changes" approach to the `StateUpdates`

jakrawcz-rdx:
This is a next step in the refactoring (started by https://github.com/radixdlt/radixdlt-scrypto/pull/1451):
- the changes inside `StateUpdates` are expressed as `Vec<StateUpdate>` (which unifies single vs batch changes)

I am afraid we cannot make yet another step (i.e. the canonicalization) at `StateUpdates` for rcnet-v3.1.
But AFAIU we should be able to make `DatabaseUpdates` canonical (it is now outside the receipt 😈) and nicely pass it to JMT, substate store, etc.


radix-engine/src/track/track.rs
```diff
@@ -17,50 +17,97 @@ use sbor::rust::mem;
...
 #[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor, Default)]
 pub struct StateUpdates {
-    /// A set of partitions that were entirely deleted.
-    /// Note: this batch changes should be applied *before* the [`system_updates`] below (i.e.
-    /// allowing a latter individual substate creation to be applied).
-    pub partition_deletions: IndexSet<(NodeId, PartitionNumber)>,
-
-    /// A set of individual substate updates (indexed by partition and substate key).
-    pub system_updates: SystemUpdates,
+    pub elements: Vec<StateUpdate>,
```
jakrawcz-rdx, line 23: _(note from author)_

Naming suggestions welcome here.
_(I'd just like to avoid `result.state_updates.state_updates` :p)_

radix-engine/src/track/track.rs
```diff
@@ -17,50 +17,97 @@ use sbor::rust::mem;
...
+                .map(|(node_id, partition_num)| {
+                    StateUpdate::Batch(BatchStateUpdate::DeletePartition(node_id, partition_num))
+                })
+                .chain(system_updates.into_iter().flat_map(
+                    |((node_id, partition_num), by_substate_key)| {
+                        by_substate_key
+                            .into_iter()
+                            .map(move |(substate_key, update)| {
+                                StateUpdate::Single(SingleSubstateUpdate {
+                                    node_id: node_id.clone(),
```
jakrawcz-rdx, line 46: _(note from author)_

This line shows one implication:
If a million substates from the same Node is changed, than this new `Vec<>` representation wastes 33MB on repeating the same `NodeId` million times.
If this is a problem, then we can migrate from `Vec<StateUpdate>` to `IndexMap<NodeId, PartitionStateUpdate>` (I assume the order of applying changes between Nodes will _never_ matter?).
_But my default vote is "do not drive API design by performance aspects before performance problems occur or become obvious"._

radix-engine/src/track/track.rs
talekhinezh: Maybe just `updates`?

radix-engine/src/track/track.rs
talekhinezh: I think that logic makes sense (I wouldn't consider 33MB to be THAT bad). I am a little more skeptical though about the lost information regarding the shared node/partition updates. It looks like we re-index these when converting to `DatabaseUpdates`. Is there a reason to remove the index?

radix-engine/src/track/track.rs
```diff
@@ -17,50 +17,97 @@ use sbor::rust::mem;
...
-                        .iter()
-                        .map(|(substate_key, update)| {
-                            (M::to_db_sort_key(substate_key), update.clone())
-                        })
-                        .collect::<IndexMap<_, _>>(),
-                )
-            })
-            .collect()
+        let mut database_updates = DatabaseUpdates::default();
+        for update in &self.elements {
```
talekhinezh, line 66: `self.updates` I think would feel much better :)

radix-engine/src/track/track.rs
jakrawcz-rdx: > It looks like we re-index these when converting to `DatabaseUpdates`.

Yes, this is true at the moment: we lose the indexing effort and need to re-index into basically-the-same-structure-that-Track-already-had == `DatabaseUpdates`.

However, we are now free to turn the `DatabaseUpdates` into a "canonical change" (i.e. something like `TreeChange` that I needed at https://github.com/radixdlt/radixdlt-scrypto/pull/1430/files#diff-8081b9309cd2a88224cb5012df79c25edd11493cc545b59c3fb772db191a1c0fR11). I actually want to do it as part of "support partition delete in all storages". 
And this will need slightly different indexing anyways (i.e. a canonicalization process, which does `Vec<ChangeOp> -> TreeChange`). So I effectively do not need that previous indexing.

lrubasze:
> I see some CI failing with `Error: Can't use 'tar -xzf' extract archive file: /Users/runner/work/_actions/_temp_bdac7c72-9d89-468b-813e-e8522b4c60a4/7ff6b918-906c-48a6-a969-82ead1c1b0a6.tar.gz. return code: 1.` and I wonder whether it's my fault... 🤔

The same issue observed yesterday in other PRs. Supposedly github issue...

radix-engine/src/track/track.rs
dhedey: I guess the indexing doesn't work any more due to the partition deletes.

But I still wonder if we can actually encode the canonicalization into the form, and still get to keep the indexing, eg, as:
```
StateUpdates {
   initial_partition_deletions: IndexMap<PartitionDeletes>,
   substate_changes: IndexMap<(..), SubstateChange>,
}
```
Which I guess is kinda what we started with :')

dhedey: I'm not worried about the state size, but I do think we'll end up recreating that index (key) => change in a lot of places if we don't have it yet.

jakrawcz-rdx: The https://github.com/radixdlt/radixdlt-scrypto/pull/1454#discussion_r1315712082 suggests that we don't want to rush and merge this mid-step into rcnet-v3.1 - and I kinda agree.

We can schedule a slower migration towards a true canonical form (which can also include indexing, if we find it useful!). Something like:
```
struct StateUpdates {
  by_node: IndexMap<NodeId, NodeTierUpdates>,
}

enum NodeTierUpdates {
  Delta { by_partition: IndexMap<PartitionNumber, PartitionTierUpdates> },
  // possible place to add Batch(DeleteNode) or whatever we need later
}

enum PartitionTierUpdates {
  Delta { by_substate: IndexMap<SubstateKey, SubstateUpdate> }, // individual creations/updates/deletes
  Batch(BatchPartitionUpdate),
}

enum BatchPartitionUpdate {
  Reset { new_contents: IndexMap<SubstateKey, SubstateValue> }, // note: Reset >= Delete
}
```

dhedey: The proposed canonical form loses details about the timeline-order of substate updates inside a transaction. Not respecting timeline-ordering in our API responses is something that integrators kept nagging us about because they found it unexpected.

They mostly care inside a Node/Partition though, so this might be okay.

But perhaps it's more straightforward to create the following from the canonical StateUpdates ourselves later, and leave the transaction receipt almost as-is?
```rust
struct MergableStateUpdates {
  by_node: IndexMap<NodeId, NodeTierUpdates>,
}
```

dhedey: Oh wait I'm sorry looks like we already don't have that:
```
pub type SystemUpdates = IndexMap<(NodeId, PartitionNumber), IndexMap<SubstateKey, DatabaseUpdate>>;
```

So perhaps it's okay to go a step further and just capture it by Node, Partition, and then Substate in the partition.

jakrawcz-rdx:
@dhedey the newest commit implements what we discussed (https://github.com/radixdlt/radixdlt-scrypto/pull/1454#discussion_r1315750951), but it most likely contains a bug (e.g. `minting_of_recovery_badges_succeeds_for_recovery_role` fails with panic `Invalid genesis creation: Transactions overwriting initial flash substates`).

dhedey (approved):

PR #1456, opened 2023-09-04
Feature/typed native component addresses

0xOmarA:
## Summary

This PR changes all of our native blueprints to use typed addresses instead of just a `ComponentAddress` in their interfaces.


radix-engine-interface/src/blueprints/component.rs
```diff
@@ -0,0 +1,157 @@
+use core::marker::PhantomData;
+
+use radix_engine_common::prelude::*;
+
+pub trait TypeInfoMarker {
```
dhedey, line 5: I wonder if we should just move the following trait from `scrypto/component` to here?
```
pub trait HasTypeInfo {
    const PACKAGE_ADDRESS: Option<PackageAddress>;
    const BLUEPRINT_NAME: &'static str;
    const OWNED_TYPE_NAME: &'static str;
    const GLOBAL_TYPE_NAME: &'static str;

    fn blueprint_id() -> BlueprintId {
        let package_address = Self::PACKAGE_ADDRESS.unwrap_or(Runtime::package_address());
        BlueprintId::new(&package_address, Self::BLUEPRINT_NAME)
    }
}
```

And then it could also implement eg `get_own_type_data` / `get_reference_type_data` to also commonise that between engine and scrypto?

radix-engine-interface/src/blueprints/component.rs
```diff
@@ -0,0 +1,157 @@
...
+    const PACKAGE_ADDRESS: Option<PackageAddress>;
+    const BLUEPRINT_NAME: &'static str;
+    const OWNED_TYPE_NAME: &'static str;
+    const GLOBAL_TYPE_NAME: &'static str;
+}
+
+pub struct Global<T>(pub ComponentAddress, PhantomData<T>)
```
dhedey, line 12: I feel like these should maybe be defined in `radix-engine` or `native-sdk` to save them conflicting with the scrypto types?

dhedey (approved):
Nice! A few thoughts for optional neatenings

radix-engine-interface/src/blueprints/component.rs
0xOmarA: I agree, will make that change.

radix-engine-interface/src/blueprints/component.rs
0xOmarA: The thing is that we need to use `Global<T>` and `Owned<T>` in the `radix-engine-interface` crate where all of the invocation structs and types are defined and the `radix-engine-interface` does not depend on the `radix-engine` or the `native-sdk`, quite the opposite: they depend on it. So, defining this type there would result in a circular dependency.

radix-engine-interface/src/blueprints/component.rs
0xOmarA: Explored this a little bit.

In `HasTypeInfo`, the `PACKAGE_ADDRESS` is optional. Thus, in cases when the package address is `None`, the `BlueprintId` is obtained through `Runtime::package_address()` which is of course not available in the engine. So, I'd rather keep them separate for the time being.

PR #1472, opened 2023-09-06
Optimize Scrypto Library (Part 4)

iamyulong:
## Summary

This PR updates `KeyValueStore` and `ResourceBuilder` to fully support registered types.

It helps reduce WASM code size and runtime transaction cost.

**This change is fully backward-compatible.**

---

Example 1:
```rust
use scrypto::prelude::*;

#[blueprint]
#[types(Hash, Epoch)] // First, register the key type and value type here
mod example {
    struct Example {
        kv_store: KeyValueStore<Hash, Epoch>,
    }

    impl Example {
        pub fn new() -> Global<Example> {
            Self {
                // Then, use `new_with_registered_type` instead of `new` for creating a key value store
                kv_store: KeyValueStore::new_with_registered_type(),
            }
            .instantiate()
            .prepare_to_globalize(OwnerRole::None)
            .globalize()
        }
    }
}
```

---

Example 2:
```rust
use scrypto::prelude::*;

#[derive(ScryptoSbor, NonFungibleData)]
pub struct TestData {
    pub amount_due: Decimal,
}

#[blueprint]
#[types(TestData)] // First, register the non-fungible data type here
mod example {
    struct Example {}

    impl Example {
        pub fn create_non_fungible_resource() {
            // Then, use the `_with_registered_type` interface
            ResourceBuilder::new_ruid_non_fungible_with_registered_type::<TestData>(OwnerRole::None)
                .create_with_no_initial_supply();
        }
    }
}
```

## Testing

New tests added

### For dApp Developers

See the above examples for the new features.


talekhinezh (commented):
Not too big of a fan of the triple generic of KeyValueStore and double generic of NonFungibles. Could we instead require Scrypto to use the same blueprint for types?

iamyulong:
> Not too big of a fan of the triple generic of KeyValueStore and double generic of NonFungibles. 

Tried different approach, but found this to be the less intrusive, fully flexible way of handling additional parameters.

It is also how `HashMap`/`IndexMap` are implemented, when dealing with different `Hasher`.

> Could we instead require Scrypto to use the same blueprint for types?

Personally prefer explicit blueprint over implicit "this blueprint". 

I can potentially autogen wrapper types such as `pub struct RadiswapKeyValueStore<K, V>`, but not sure how much gain we get from this.

iamyulong:
@talekhinezh I've updated the PR to use the new syntax. :-)

assets/blueprints/faucet/src/lib.rs
```diff
@@ -15,7 +16,7 @@ mod faucet {
         ) -> Global<Faucet> {
             Self {
                 vault: Vault::with_bucket(bucket),
-                transactions: KeyValueStore::new(),
+                transactions: KeyValueStore::new_with_registered(),
```
talekhinezh, line 19: Minor: not too big of a fan of the name, would like it to be either shorter (e.g. `new_bp_typed`) or more descriptive (`new_with_registered_type`). Middle of the road feels like worse of both worlds?

radix-engine-interface/src/api/key_value_store_api.rs
```diff
@@ -9,6 +9,9 @@ use sbor::rust::prelude::*;
 use sbor::LocalTypeId;
 use sbor::{generate_full_schema, TypeAggregator};
 
+pub const KV_STORE_DATA_SCHEMA_VARIANT_LOCAL: u8 = 0;
+pub const KV_STORE_DATA_SCHEMA_VARIANT_REMOTE: u8 = 1;
```
talekhinezh, line 13: Dead code?

scrypto-derive/src/blueprint.rs
```diff
@@ -222,6 +222,38 @@ pub fn replace_macros(expr: &mut Expr, dependency_exprs: &mut Vec<Expr>) -> Resu
     Ok(())
 }
 
+/// Derive a sensible type identifier from a path, so to pass Radix Engine's constraints.
+pub fn derive_sensible_identifier_from_path(path: &Path) -> Result<String> {
```
talekhinezh, line 226: Perhaps show an example of the transformation in the above documentation just for clarity sake.

assets/blueprints/faucet/src/lib.rs
iamyulong: Slightly worried about `_bp_`. Will go `new_with_registered_type`.

radix-engine-interface/src/api/key_value_store_api.rs
iamyulong: `KV_STORE_DATA_SCHEMA_VARIANT_LOCAL` is and the other is not.

It kind of reminds me that I can optimise the local variant as well.

So, both will be used.

scrypto-derive/src/blueprint.rs
talekhinezh: I meant just in the code doc above :)

talekhinezh (approved):

PR #1475, opened 2023-09-06
Test: Transaction Processor

0xOmarA:
## Summary

This PR makes some minor clean ups to the transaction processor, fixes a bug in the transaction validator that caused some transactions to be rejected, and adds some tests for the transaction processor. 

## Details

Cleaned up the invocation logic in the transaction processor so that it all goes through a single `handle_invocation` function instead of multiple different macros that can go out of sync.

Fixed a bug in the transaction validator whereby instructions like `DROP_NAMED_PROOFS` and `DROP_ALL_PROOFS` dropped the proofs without reducing the ref count in the bucket, this lead the bucket to remain locked and for the transaction to be rejected (this is a rejection and not a failure because this was a bug in the transaction validator, not the engine itself).

The following is an example manifest that can be used to reproduce this bug.
```ruby
CALL_METHOD
    Address("account_sim168v2uak8e622vre9g3j3vxup7wdk32k75u2p73vepvyre7c0s839xs")
    "withdraw"
    Address("resource_sim1tknxxxxxxxxxradxrdxxxxxxxxx009923554798xxxxxxxxxakj8n3")
    Decimal("73")
;
TAKE_FROM_WORKTOP
    Address("resource_sim1tknxxxxxxxxxradxrdxxxxxxxxx009923554798xxxxxxxxxakj8n3")
    Decimal("73")
    Bucket("bucket1")
;
CREATE_PROOF_FROM_BUCKET_OF_AMOUNT
    Bucket("bucket1")
    Decimal("73")
    Proof("proof1")
;
DROP_ALL_PROOFS;

# Even though we have called the `DROP_ALL_PROOFS` instruction above, `bucket1`
# is still locked from the prespective of the transaction validator. This is 
# because the validator didn't reduce the ref-count on the bucket when the proof
# was dropped and therefore it doesn't permit the following instruction to take
# the bucket.
CALL_METHOD
    Address("account_sim168v2uak8e622vre9g3j3vxup7wdk32k75u2p73vepvyre7c0s839xs")
    "try_deposit_or_abort"
    Bucket("bucket1")
    Enum<0u8>()
;
```

The following are two tests that would fail without this fix:
* `creating_proof_and_then_dropping_it_should_not_keep_bucket_locked`
* `creating_proof_and_then_dropping_it_should_not_keep_bucket_locked2`

### DIFF
--- a/transaction/src/validation/id_validator.rs
+++ b/transaction/src/validation/id_validator.rs
@@ -26,7 +26,7 @@ pub struct ManifestValidator {
     /// Bucket id -> lock count
     bucket_ids: NonIterMap<ManifestBucket, usize>,
     /// Proof id to proof info
-    proof_ids: NonIterMap<ManifestProof, ProofKind>,
+    proof_ids: BTreeMap<ManifestProof, ProofKind>,
     /// Set of active allocated global address reservation ids
     address_reservation_ids: IndexSet<ManifestAddressReservation>,
     /// Set of named global address ids
@@ -119,7 +119,10 @@ impl ManifestValidator {
     }
 
     pub fn drop_all_named_proofs(&mut self) -> Result<(), ManifestIdValidationError> {
-        self.proof_ids.clear();
+        let proof_ids = self.proof_ids.keys().copied().collect::<Vec<_>>();
+        for proof_id in proof_ids {
+            self.drop_proof(&proof_id)?
+        }
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


transaction/src/validation/id_validator.rs
```diff
@@ -119,7 +119,10 @@ impl ManifestValidator {
     }
 
     pub fn drop_all_named_proofs(&mut self) -> Result<(), ManifestIdValidationError> {
-        self.proof_ids.clear();
```
0xOmarA, line 122: We had a bug here where calling `self.proof_ids.clear();` never reduced the bucket's ref count. Thus, creating a proof from a bucket and then calling `drop_all_named_proofs` would produce manifests that fail static validation and be rejected.

iamyulong: Nice spot!

iamyulong (approved):

PR #1476, opened 2023-09-07
Add `PackageLoader` to the `radix-engine-tests` crate

0xOmarA:
## Summary

This PR adds a `PackageLoader` struct to the `radix-engine-tests` crate.

## Description

This PR is made to fix some issues we have faced when running tests locally. These issues have been made more apparent now that we're integrating code coverage in our development pipeline.

A very common issue that myself and others run into quite often when running tests locally is the following error:
```
'Failed to read built WASM from path "path/some.wasm" - Os { code: 2, kind: NotFound, message: "No such file or directory" }'
```

We believe that this issue is caused by multiple tests trying to read the same WASM file and some contention happening after the compilation finishes. 

The solution to this issue is implemented in this PR. A `compile-blueprints-at-build-time` feature flag has been added to the `radix-engine-tests` crate which when enabled, will build all of the blueprints at the crates compilation time and make the compiled packages available in tests through the `PackageLoader` struct. The build script compiles all of the blueprints, writes them to the `OUT_DIR` and then they're read by the `PackageLoader` _at compile time_ from the `OUT_DIR`.

If `compile-blueprints-at-build-time` feature is not enabled then the package will be compiled on the fly when you try to load it.

## Note for Reviewers

This diff is quite big and full of various renames and small items that might not be the best use of your time. The following are the files I recommend you review:

* `radix-engine-tests/build.rs`
* `radix-engine-tests/tests/package_loader.rs`
* `radix-engine-tests/Cargo.toml`
* `.github/workflows/ci.yml`
* `scrypto_unit/src/test_runner.rs`
* `run-code-coverage.sh`

### For Internal Integrators

This is not quite for internal integrators, but for the Builder's Team:

* Do not set the `compile-blueprints-at-build-time` as a default feature in the `radix-engine-tests` crate. It will lead to a painfully slow rust analyzer.
* If you're running a single test or even a single module you do not need `compile-blueprints-at-build-time`. 
* Pass `--features compile-blueprints-at-build-time` to `cargo test` or `cargo nextest run` if you want to run all of the tests without the flakyness of reading files from the file system.
* `PackageLoader::get` takes the name of the Cargo package and **NOT** the name of the directory.

### DIFF
--- a/radix-engine-tests/tests/package_loader.rs
+++ b/radix-engine-tests/tests/package_loader.rs
@@ -1,0 +1,44 @@
+#[cfg(feature = "compile-blueprints-at-build-time")]
+mod package_loader {
+    use radix_engine_common::prelude::*;
+    use radix_engine_queries::typed_substate_layout::*;
+    use std::collections::HashMap;
+
+    const PACKAGES_BINARY: &[u8] =
+        include_bytes!(concat!(env!("OUT_DIR"), "/compiled_packages.bin"));
+
+    lazy_static::lazy_static! {
+        static ref PACKAGES: HashMap<String, (Vec<u8>, PackageDefinition)> = {
+            scrypto_decode(PACKAGES_BINARY).unwrap()
+        };
+    }
+
+    pub struct PackageLoader;
+    impl PackageLoader {
+        pub fn get(name: &str) -> (Vec<u8>, PackageDefinition) {
+            if let Some(rtn) = PACKAGES.get(name) {
+                rtn.clone()
+            } else {
+                panic!("Package \"{}\" not found. Are you sure that this package is: a) in the blueprints folder, b) that this is the same as the package name in the Cargo.toml file?", name)
+            }
+        }
+    }
+}
+
+#[cfg(not(feature = "compile-blueprints-at-build-time"))]
+mod package_loader {
+    use radix_engine_common::prelude::*;
+    use radix_engine_queries::typed_substate_layout::*;
+    use std::path::PathBuf;
+
+    pub struct PackageLoader;
+    impl PackageLoader {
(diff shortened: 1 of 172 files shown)


radix-engine-tests/tests/package_loader.rs
```diff
@@ -0,0 +1,44 @@
...
+        };
+    }
+
+    pub struct PackageLoader;
+    impl PackageLoader {
+        pub fn get(name: &str) -> (Vec<u8>, PackageDefinition) {
+            if let Some(rtn) = PACKAGES.get(name) {
+                rtn.clone()
+            } else {
+                panic!("Package \"{}\" not found. Are you sure that this package is: a) in the blueprints folder, b) that this is the same as the package name in the Cargo.toml file?", name)
```
lrubasze, line 22: Just wondering if it make sense to compile the package here instead of panicking.
Not really sure.
On the one hand we would just go further without stopping.
On the other we could hide some problem by silently rebuilding the missing package.

0xOmarA: I personally think that it might be better to have it fail instead of succeed since the main objective of this is to eliminate any contention that might exist on the files. So, this would alert the test author and they'd make the changes that need to be made as per the provided suggestions.

radix-engine-tests/tests/package_loader.rs
```diff
@@ -0,0 +1,44 @@
+#[cfg(feature = "compile-blueprints-at-build-time")]
+mod package_loader {
+    use radix_engine_common::prelude::*;
+    use radix_engine_queries::typed_substate_layout::*;
+    use sbor::prelude::*;
+
+    const PACKAGES_BINARY: &[u8] =
```
0xOmarA, line 7: compile time constant which is really cool!

iamyulong (approved):

PR #1483, opened 2023-09-08
Decimal various optimizations

lrubasze:
## Summary
- Various decimal optimizations 
- refactor `truncate()` to `checked_truncate()` with rounding mode as parameter

## Details
Optimizations list:
- Add ONE_TENTH constant
- Let `dec!` and `pdec!` macros use decimal constants for elementary values
- Remove `to_vec()` from Integer types
- Remove rotate methods from Integer types 
- Remove By trait from Integer types
- Add conversion to primitive integer types for `Decimal` and `PreciseDecimal`

## Testing
- Add missing tests for conversion from primitive types
- Add tests for conversion to primitive types

### For dApp Developers
Refactor scrypto sources to  `checked_truncate()`

### DIFF
--- a/radix-engine-common/src/macros.rs
+++ b/radix-engine-common/src/macros.rs
@@ -2,6 +2,30 @@
 ///
 #[macro_export]
 macro_rules! dec {
+    (0) => {
+        $crate::math::Decimal::ZERO
+    };
+    ("0") => {
+        $crate::math::Decimal::ZERO
+    };
+    ("0.1") => {
+        $crate::math::Decimal::ONE_TENTH
+    };
+    ("1" | 1) => {
+        $crate::math::Decimal::ONE
+    };
+    (10) => {
+        $crate::math::Decimal::TEN
+    };
+    ("10") => {
+        $crate::math::Decimal::TEN
+    };
+    (100) => {
+        $crate::math::Decimal::ONE_HUNDRED
+    };
+    ("100") => {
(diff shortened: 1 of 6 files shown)


radix-engine-common/src/macros.rs
```diff
@@ -2,6 +2,30 @@
...
+    };
+    ("10") => {
+        $crate::math::Decimal::TEN
+    };
+    (100) => {
+        $crate::math::Decimal::ONE_HUNDRED
+    };
+    ("100") => {
+        $crate::math::Decimal::ONE_HUNDRED
+    };
```
iamyulong, line 28: Are we still planing to move "any string decimal parsing" to compile-time?

radix-engine-common/src/math/decimal.rs
```diff
@@ -62,10 +64,11 @@ impl Decimal {
 
     pub const ZERO: Self = Self(I192::ZERO);
 
+    pub const ONE_HUNDREDTH: Self = Self(I192::from_digits([10_u64.pow(Decimal::SCALE - 2), 0, 0]));
```
iamyulong, line 67: Can we test all these constants by converting it to string? I saw you've got a few but wanted to make sure we cover them all.

iamyulong (approved):
LGTM overall

radix-engine-common/src/math/decimal.rs
lrubasze: Done

PR #1486, opened 2023-09-08
Fix event size counting

iamyulong:
## Summary

Fix event size counting

### DIFF
--- a/radix-engine/src/system/system_modules/transaction_runtime/module.rs
+++ b/radix-engine/src/system/system_modules/transaction_runtime/module.rs
@@ -14,7 +14,7 @@ pub struct Event {
 
 impl Event {
     pub fn len(&self) -> usize {
-        self.type_identifier.0.get_length()
+        self.type_identifier.len() + self.payload.len() + 4
     }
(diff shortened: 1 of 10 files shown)


talekhinezh (approved):

radix-engine/src/system/system_modules/transaction_runtime/module.rs
```diff
@@ -14,7 +14,7 @@ pub struct Event {
 
 impl Event {
     pub fn len(&self) -> usize {
-        self.type_identifier.0.get_length()
+        self.type_identifier.len() + self.payload.len() + 4
```
talekhinezh, line 17: What's the +4 for?

iamyulong: This is to account for event flags.

jakrawcz-rdx: please consider a well-named const
_(which also creates a perfect place for putting a rustdoc on, with more detailed explanation)_

PR #1491, opened 2023-09-10
Test/royalty module

0xOmarA:
## Summary

Additional testing for the Royalty module and package royalties. This PR also adds checks to the royalty module and costing module to check for negative amounts which it didn't have before. Luckily, we were saved by the fact that transmutating a negative `Decimal` to a `u128` results in an overflow (which we caught). 

## Testing

* Added tests around the various edge cases of the `RoyaltyAmount`. Specifically: `Decimal::MIN`, -1, 0, 1, max XRD and USD, and `Decimal::MAX` for both components and packages. 
* Added tests that component and package royalties work together as expected. 
* Added tests for the exhaustiveness of the specification of royalties in the package definition.

## Note:

This PR will **LOWER** the code coverage of the royalty module even though all of the possible paths that we can go through have been tested. The lines in code coverage that remain untested are: comments (yeah idk why the tool counts those?), errors that we know are impossible to get to, errors that are too small to test (e.g., the `invocation` method of the package erroring out for an incorrect method name). If you want to quickly check the code coverage of this module before approving the PR then here is a gist of it: https://gist.github.com/0xOmarA/a0a060e15d815d593143f99b4c0ad0a1

### DIFF
--- a/radix-engine-tests/tests/royalty_edge_cases.rs
+++ b/radix-engine-tests/tests/royalty_edge_cases.rs
@@ -1,0 +1,334 @@
+mod package_loader;
+
+use package_loader::PackageLoader;
+use radix_engine::errors::*;
+use radix_engine::transaction::*;
+use radix_engine_queries::typed_substate_layout::*;
+use scrypto_unit::*;
+use transaction::prelude::*;
+
+#[test]
+fn test_component_royalty_edge_cases_at_instantiation() {
+    // Arrange
+    let mut test_runner = TestRunnerBuilder::new().without_trace().build();
+    let package_address =
+        test_runner.publish_package_simple(PackageLoader::get("royalty-edge-cases"));
+
+    for (royalty_amount, error_checking_fn, _) in test_cases().into_iter() {
+        // Act
+        let manifest = ManifestBuilder::new()
+            .lock_fee_from_faucet()
+            .call_function(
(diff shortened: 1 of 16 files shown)


radix-engine-tests/tests/royalty_edge_cases.rs
```diff
@@ -0,0 +1,448 @@
...
+use transaction::prelude::*;
+
+#[test]
+fn test_component_royalty_edge_cases_at_instantiation() {
+    // Arrange
+    let mut test_runner = TestRunnerBuilder::new().without_trace().build();
+    let package_address =
+        test_runner.publish_package_simple(PackageLoader::get("royalty-edge-cases"));
+
+    for (royalty_amount, error_checking_fn, _) in test_cases().into_iter() {
```
talekhinezh, line 17: I think cleaner to have these tests cases in separate tests. Also to enforce the property of "isolation" in unit tests. But overall, good enough for now :)

talekhinezh (approved):

radix-engine-tests/tests/royalty_edge_cases.rs
0xOmarA: Agreed, will refactor into separate tests before merging. I honestly wish Rust had parameterized tests like most other languages!

PR #1493, opened 2023-09-11
Add scenarios that use remote types

iamyulong:
## Summary

Add two scenarios:
- Non-fungible resource using remote type
- `KeyValueStore` using remote type


assets/blueprints/kv_store/src/lib.rs
```diff
@@ -0,0 +1,53 @@
+use scrypto::prelude::*;
+
+#[derive(ScryptoSbor)]
+struct TestType {
+    b: u32,
+}
+
+#[blueprint]
+#[experimental_types(TestType)]
```
0xOmarA, line 9: I take it that this isn't the final name of the attribute and we want to get the scenario out first and then change the attribute name since it doesn't impact anything on ledger?

iamyulong: That's right!

0xOmarA (approved):
LGTM. Will let you coordinate with David as I see that he has some comments on genesis compatibility.

dhedey (approved):
Looks fantastic - thanks @iamyulong ! :)

PR #1498, opened 2023-09-11
running all unit tests using nextest

balda-rdx:


talekhinezh (approved):

lrubasze:
@balda-rdx 
1. I see that `Run Radix Engine tests` finally does not skip tests - what was the reason?
2. I see that `Run Radix Engine tests (release)` still skips tests. Could we have it fixed as well?

balda-rdx:
> @balda-rdx
> 
> 1. I see that `Run Radix Engine tests` finally does not skip tests - what was the reason?
> 2. I see that `Run Radix Engine tests (release)` still skips tests. Could we have it fixed as well?



> @balda-rdx
> 
> 1. I see that `Run Radix Engine tests` finally does not skip tests - what was the reason?
> 2. I see that `Run Radix Engine tests (release)` still skips tests. Could we have it fixed as well?

@lrubasze for 1 I'm leaning towards nextest behaving different in arm64 than in amd64 when using feature `cerde`. 
on 2, yes, I'll fix them on a later PR. Want to get this merged so that people can get unblocked

PR #1499, opened 2023-09-11
Misc/Package Checks

0xOmarA:
## Summary

This package adds the following checks and adds tests for them:

* Added a length limit to the blueprint name which is checked at package instantiation. A blueprint name can at most be 100 characters long.
* Added additional validation for inner and outer blueprints. A blueprint can't be itself's outer blueprint and a blueprint's outer blueprint can't itself be an inner blueprint. 
* Remove support for transient blueprints for WASM (not supported by the blueprint macro, thus not user facing)
* Make the rules around names in package definition more strict and more inline with the SBOR validation.


radix-engine-tests/tests/reference.rs
```diff
@@ -457,3 +404,61 @@ fn test_add_local_ref_to_stored_substate() {
         _ => false,
     });
 }
+
+fn flash_package() -> (DefaultTestRunner, PackageAddress) {
```
0xOmarA, line 408: Must flash the package since we've now added checks for transient blueprints in the package package.

radix-engine-tests/tests/package.rs
```diff
@@ -425,36 +425,32 @@ fn name_validation_feature_set() {
...
+    let method_definition = definition
         .blueprints
         .values_mut()
         .next()
         .unwrap()
         .schema
         .functions
         .functions
-        .insert(
-            String::from("self"),
```
0xOmarA, line 450: `self` used to not be permitted but since we're no longer checking for `Ident` it's now permitted.

talekhinezh (approved):

PR #1507, opened 2023-09-12
enabling release tests and adding post_run_db_check features

balda-rdx:


iamyulong:
Outdated. We will consider adding coverage back after Babylon release, hopefully in a single PR.

PR #1514, opened 2023-09-12
Adding a versioning wrapper to JMT persistent node

jakrawcz-rdx:
Addresses https://whimsical.com/babylon-work-tracker-Am2s7hSa7xHd5yCxgub67a@2bsEvpTYSt1HjL3bk7R35tDxNT3xDRTfhTp.


radix-engine-stores/src/rocks_db_with_merkle_tree/mod.rs
```diff
@@ -216,7 +220,8 @@ impl ReadableTreeStore for RocksDBWithMerkleTreeSubstateStore {
         self.db
             .get_cf(self.cf(MERKLE_NODES_CF), &encode_key(key))
             .unwrap()
-            .map(|bytes| scrypto_decode(&bytes).unwrap())
+            .map(|bytes| scrypto_decode::<VersionedTreeNode>(&bytes).unwrap())
+            .map(|versioned| versioned.into_latest())
```
jakrawcz-rdx, line 224: Technically, the test-only rocksdb impl in Engine repo does not need this, but I wanted the reviewers to see how I plan to support that in the Node repo.

radix-engine-stores/Cargo.toml
```diff
@@ -9,6 +9,7 @@ utils = { path = "../utils", default-features = false }
 radix-engine-common = { path = "../radix-engine-common", default-features = false }
 radix-engine-derive = { path = "../radix-engine-derive", default-features = false }
 radix-engine-store-interface = { path = "../radix-engine-store-interface", default-features = false }
+paste = { version = "1.0.13" }
```
dhedey, line 12: Oh? I'm surprised this is needed?

My bad! I think perhaps the crate with the macro definition (sbor?) needs changing to externalize the paste crate, by eg adding the following to the `sbor` crate's `lib.rs`:

`pub extern crate paste;`

And then we _may_ need to update the `paste!` invocations inside the versioned macro definitions with `paste::paste!` or `crate::paste::paste!`... I can't quite remember how this works.

Anyway, if we do it right, the macro should be self-consistent, and we should be able to remove `paste` from the above.

radix-engine-stores/src/rocks_db_with_merkle_tree/mod.rs
```diff
@@ -10,7 +12,9 @@ use rocksdb::{
     SingleThreaded, WriteBatch, DB,
 };
 use sbor::rust::prelude::*;
+use sbor::HasLatestVersion;
```
dhedey, line 15: Even easier is `use sbor::prelude::*`, which also subsumes the previous line.

radix-engine-stores/Cargo.toml
jakrawcz-rdx: I now tried a few combinations, and the compiler always complained about something not found.

I agree it would be cool to have the macro self-contained, but probably I - a macro noob - am not up to this task :(
(BTW I see `paste = { version = "1.0.13" }` in more places, so maybe this/all macros need a wider fix anyway)

radix-engine-stores/src/rocks_db_with_merkle_tree/mod.rs
jakrawcz-rdx: whoops,
```
error[E0599]: no method named `into_latest` found for enum `VersionedTreeNode` in the current scope
   --> /home/runner/work/radixdlt-scrypto/radixdlt-scrypto/radix-engine-stores/src/rocks_db_with_merkle_tree/mod.rs:223:40
    |
223 |               .map(|versioned| versioned.into_latest())
    |                                          ^^^^^^^^^^^ method not found in `VersionedTreeNode`
```
(https://github.com/radixdlt/radixdlt-scrypto/actions/runs/6161819412/job/16721912689?pr=1514)

Although I thought it compiled locally in intellij o_O

dhedey: Ohh IntoLatest should definitely be in the `sbor` prelude. If it's missing, it should be added :)

radix-engine-stores/Cargo.toml
jakrawcz-rdx: alright, I finally did that 🎉

radix-engine-stores/src/rocks_db_with_merkle_tree/mod.rs
jakrawcz-rdx: ok, fixed 👍

jakrawcz-rdx:
I addressed the comments pretty accurately and the CI is 🟢 , so I'll allow myself to merge this.

PR #1519, opened 2023-09-13
Add tests around unexecuted error paths

0xOmarA:
## Summary

* Add tests for some of the package error paths.
* Add name checks to types.

### DIFF
--- a/radix-engine-tests/tests/package.rs
+++ b/radix-engine-tests/tests/package.rs
@@ -7,6 +7,7 @@ use radix_engine::system::system_modules::auth::*;
 use radix_engine::types::*;
 use radix_engine::vm::wasm::PrepareError;
 use radix_engine::vm::wasm::*;
+use radix_engine_common::constants::MAX_NUMBER_OF_BLUEPRINT_FIELDS;
 use radix_engine_interface::*;
 use sbor::basic_well_known_types::*;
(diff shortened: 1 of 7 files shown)


radix-engine-tests/tests/package.rs
```diff
@@ -643,6 +635,860 @@ fn publishing_of_package_with_a_lookalike_character_fails() {
...
 
+#[test]
+fn test_error_path_when_package_definition_has_too_many_fields() {
+    // Arrange
+    let mut test_runner = TestRunnerBuilder::new().build();
+    let (code, mut definition) = PackageLoader::get("address");
+
+    definition.blueprints.values_mut().for_each(|bp_def| {
+        if let Some(field) = bp_def.schema.state.fields.first() {
+            bp_def.schema.state.fields = (0..300).map(|_| field.clone()).collect::<Vec<_>>();
```
mstrug-rdx, line 646: Maybe it would be worth to add a new constant in validation of the schema substates count in `package.rs` (now there is value 0xff), and then use that constant + 1 here.

mstrug-rdx (approved):

radix-engine-tests/tests/package.rs
0xOmarA: Good point! Will do that.

PR #1526, opened 2023-09-13
remove code coverage

balda-rdx:


.github/workflows/ci.yml
```diff
@@ -223,14 +208,24 @@ jobs:
     runs-on: k8s-linux-runner
     steps:
```
lrubasze, line 209: below is missing
```
env:
      CARGO_LOG: info
```

lrubasze:
Nice work!

As agreed, we would like to align environments for both linux runners (switch to containers or get rid of them). But this might be done separately.

lrubasze (approved):

PR #1530, opened 2023-09-14
Add random tests to verify tracked substate state transition

iamyulong:
## Summary

Add random tests to verify tracked substate state transition

### Notes
Identified a potential problematic transition path `new -> garbage -> write only`. This is currently un-exploitable because we never "take" anything "new" in track.


mstrug-rdx (approved):
Looks ok.

PR #1531, opened 2023-09-14
Bump function locals limit

iamyulong:
## Summary

Bump `MAX_NUMBER_OF_FUNCTION_LOCALS` from 128 to 256.

### For dApp Developers

If you're still running into `TooManyFunctionLocals` error, try breaking functions that take many parameters into smaller ones and try changing the `opt-level` within `Cargo.toml`.


radix-engine-tests/tests/wasm_prepare.rs
iamyulong, line 1: Removed as the binary file is hard to update. The limit check is covered by `test_function_limits`.

lrubasze (approved):

PR #1535, opened 2023-09-14
CI fixes for windows

lrubasze:


.github/workflows/ci.yml
```diff
@@ -430,6 +431,7 @@ jobs:
       with:
         toolchain: 1.70.0
     - name: reset registry
+      if: runner.os != 'Windows'
       run: rm -fr ~/.cargo/registry
```
iamyulong, line 435: I think we can remove the `rm` line? It was added to solve the `syn` issue, which is now fixed in another way.

lrubasze: Good point!

PR #1542, opened 2023-09-14
Fix compare_current_time overflows

talekhinezh:
Fix Epoch Manager compare current time overflows


0xOmarA (approved):
Lol was working on the same thing in #1540.

PR #1545, opened 2023-09-15
Cleanup: Remove dead code

talekhinezh:
Made a pass on removal of dead code


dhedey (changes requested):
Scrypto is a library we shouldn't just be removing code like this. 

Please revert all changes in `versioned.rs` and `models.rs` as the node uses at least a half of the removed code and I don't want to have to fight this so close to release.

PR #1550, opened 2023-09-15
Test/fix tests

talekhinezh:
Fix kernel_fuzzer test by only committing to database if end of transaction transient substates don't contain owned nodes


radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -382,7 +382,9 @@ where
                             execution_trace_module.finalize(&paying_vaults, is_success);
 
                         // Finalize track
-                        let (tracked_nodes, deleted_partitions) = track.finalize();
+                        let (tracked_nodes, deleted_partitions) = track.finalize().expect(
+                            "System invariants should prevent transient substate from owning nodes",
+                        );
```
iamyulong, line 387: Can we match the error code and then panic?

radix-engine/src/track/track.rs
```diff
@@ -188,12 +196,19 @@ impl<'s, S: SubstateDatabase, M: DatabaseKeyMapper + 'static> Track<'s, S, M> {
                     .and_then(|tracked_node| tracked_node.tracked_partitions.get_mut(&partition))
                 {
                     let db_sort_key = M::to_db_sort_key(&substate_key);
-                    tracked_partition.substates.remove(&db_sort_key);
+                    let tracked_substate = tracked_partition.substates.remove(&db_sort_key);
+                    if let Some(substate) =
+                        tracked_substate.and_then(|s| s.substate_value.into_value())
```
iamyulong, line 201: Wonder under what scenario the substate is not tracked but marked as transient.

transaction-scenarios/generated-examples/non_fungible_resource/005--non-fungible-resource-mint-32-nfts.rtm
```diff
@@ -4,7 +4,7 @@ CALL_METHOD
     Decimal("5000")
 ;
 MINT_NON_FUNGIBLE
-    Address("resource_sim1ntjqlurzucu3lxnuxu4e7v57qmfna3gxuka5ll2we0k2zz5ptnlmy0")
+    Address("resource_sim1ngx46l2l7t93rpwchzkv2ywdxpjdxvk8lplrqfte4d8pg92szl8vuz")
```
iamyulong, line 7: Interesting that the scenario outputs are changed.

talekhinezh: I think it just wasn't updated on last commit

radix-engine/src/track/track.rs
talekhinezh: There is no scenario unless the system layer is doing something wonky

radix-engine/src/transaction/transaction_executor.rs
talekhinezh: Yup good point

iamyulong (approved):

PR #1559, opened 2023-09-18
Clean up validator rewards distribution

iamyulong:
## Summary

- Use total effective stake, instead of total stake, when distributing "validator set rewards", so to hand them out ASAP.
- Reset the `proposer_rewards` at epoch end, for the extreme case where `zero` stake validator is in the validator set

### DIFF
--- a/radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
+++ b/radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
@@ -1362,50 +1362,76 @@ impl ConsensusManagerBlueprint {
         //===========================
         // Distribute rewards (fees)
         //===========================
-        let total_individual_amount: Decimal = {
-            let mut sum = Decimal::ZERO;
+        let mut total_effective_stake = Decimal::ZERO;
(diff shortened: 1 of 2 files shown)


radix-engine/src/blueprints/consensus_manager/consensus_manager.rs
```diff
@@ -1414,6 +1446,10 @@ impl ConsensusManagerBlueprint {
             )?;
         }
 
+        // For any reason, if a validator isn't included in the `validator_infos` but has accumulated
+        // proposer rewards, we reset the counter as the rewards has been distributed to other validators.
+        validator_rewards.proposer_rewards.clear();
```
jakrawcz-rdx, line 1451: (optional) I would somehow feel safer if we captured the "not applicable validators" (e.g. from line 1289 `// Excluded due to slashing ?` I guess?) and explicitly removed them from `validator_rewards` before the `for` loop (at 1416). We could even assert that the map is empty at the method end.

jakrawcz-rdx (approved):

PR #1562, opened 2023-09-18
simulator docker image

balda-rdx:
This PR introduce a `Dockerfile` that can be used to build Scrypto package in a reproducible way.

It's primarily used for building the `assets` for now.

## Warning

We've noticed that the output WASM file depends on the host CPU arch.


simulator/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM rust:1.70.0-bullseye
+
+RUN apt update && apt install -y cmake clang build-essential llvm
```
dhedey, line 3: We should specify versions for all of these to make the local build more deterministic.

simulator/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ADD radix-engine-common /app/radix-engine-common
+ADD radix-engine-derive /app/radix-engine-derive
+ADD native-sdk /app/native-sdk
+ADD radix-engine-macros /app/radix-engine-macros
+ADD radix-engine-profiling /app/radix-engine-profiling
+ADD radix-engine-store-interface /app/radix-engine-store-interface
+ADD radix-engine-stores /app/radix-engine-stores
+ADD transaction /app/transaction
+ADD radix-engine-queries /app/radix-engine-queries
+ADD assets /app/assets
```
dhedey, line 23: Can we instead (or as well as) use a docker ignore file? For us developers, lots of these paths will have massive target folders (like, 200GB of target folders :D) which will need ignoring.

simulator/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+WORKDIR /app
+
+RUN cargo install --path ./simulator \
+    && cp /app/simulator/target/release/scrypto /usr/local/bin/ \
+    && rm -rf /app
+
+WORKDIR /src
```
dhedey, line 31: Can we have separate named base layers for the build and the app itself? I imagine that will make the image much lighter, if we can start from a light base image for the app?

We could even use a sub-dependency of the rust image to avoid rust being there, eg bullseye itself https://github.com/rust-lang/docker-rust/blob/5ede3cc7db4695645dfb1a9923cbafa519fb5c49/1.70.0/bullseye/Dockerfile

dhedey: See eg https://hub.docker.com/_/debian for bullseye

simulator/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM rust:1.70.0-bullseye
```
dhedey, line 1: Out of interest - any reason to use `bullseye` rather than `bullseye-slim` here? Alignment across the stack is a fair reason.

simulator/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+WORKDIR /app
+
+RUN cargo install --path ./simulator \
+    && cp /app/simulator/target/release/scrypto /usr/local/bin/ \
+    && rm -rf /app
+
+WORKDIR /src
+
+ENTRYPOINT ["scrypto", "build", "--path", "/src"]
```
dhedey, line 33: Nice. Could we also have a script to demonstrate using this? EG update the `./update-assets.sh` script to use this image rather than calling `scrypto build` directly?

dhedey (commented):
A few early comments / suggestions

simulator/Dockerfile
balda-rdx: no reason at all. will try slim

simulator/Dockerfile
balda-rdx: yes ofc

simulator/Dockerfile
balda-rdx: done

simulator/Dockerfile
balda-rdx: added the dockerignore file with the target/ folder. please do let me know if there's anything else that should go in there

simulator/Dockerfile
balda-rdx: slim seems to working just fine. I would rather use slim to keep the image smaller. let me know if you prefer to go back to bullseye (no slim) to keep consistency across the stack

simulator/Dockerfile
balda-rdx: added yet another script called `update-assets-docker.sh` just in case people want to keep using the existing one.

simulator/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM rust:1.70.0-slim-bullseye as builder
+
+RUN apt update && apt install -y \
+    cmake=3.18.4-2+deb11u1 \
+    clang=1:11.0-51+nmu5 \
+    build-essential=12.9 \
+    llvm=1:11.0-51+nmu5
+#RUN rustup target add wasm32-unknown-unknown
```
iamyulong, line 8: Does the `apt update` affect the version of the dependencies of the software you're installing? Is update really necessary?

simulator/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ADD radix-engine-common /app/radix-engine-common
+ADD radix-engine-derive /app/radix-engine-derive
+ADD native-sdk /app/native-sdk
+ADD radix-engine-macros /app/radix-engine-macros
+ADD radix-engine-profiling /app/radix-engine-profiling
+ADD radix-engine-store-interface /app/radix-engine-store-interface
+ADD radix-engine-stores /app/radix-engine-stores
+ADD transaction /app/transaction
+ADD radix-engine-queries /app/radix-engine-queries
+ADD assets /app/assets
```
iamyulong, line 27: I think we have a recursively dependency here? The assets will be built by the docker image and the docker image depends on the "previous" assets. 

As long as the final WASM files doesn't depend on the assets, we're good. I think this is the case, but maybe worth double checking?

simulator/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+WORKDIR /app
+
+RUN cargo install --path ./simulator
+
+FROM rust:1.70.0-slim-bullseye
```
dhedey, line 33: Do we even need to use the rust bullseye base? Or should we use the [corresponding Debian base](https://hub.docker.com/layers/library/rust/1.70.0-slim-bullseye/images/sha256-1de20372759e9341c5b6e8cc7b18239e91947cf5ba9ddc20c0e5903b2b9fde24?context=repo&tab=vulnerabilities): which is `bullseye-20230703-slim`? 

Happy with either tbh. I don't imagine the Rust image is much on top of bullseye slim, so maybe it's easier to be consistent.

dhedey: Oh hmm I guess we need the rust base image for the scrypto build anyway :D ignore me.

simulator/Dockerfile
dhedey, line 8: This can be removed I guess?

update-assets-docker.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# Check if the Docker image exists locally
+if ! "$REUSE_IMAGE" || ! docker image inspect "$IMAGE_NAME" &> /dev/null; then
+  echo "Docker image $IMAGE_NAME does not exist or --reuse-image flag not provided. Building..."
+  docker build -t $IMAGE_NAME -f simulator/Dockerfile .
+else
+  echo "Docker image $IMAGE_NAME exists, and --reuse-image flag is set. Skipping build."
+fi
+
+for crate_name in "faucet" "radiswap" "flash_loan" "genesis_helper" "metadata" "test_environment" "global_n_owned" "kv_store"
```
iamyulong, line 30: We're missing `max_transaction`, which is added recently.

update-assets-docker.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
dhedey, line 1: It's probably worth just changing `update-assets.sh` to use this script? and then we can all try running it as an additional test on our machines, and check it's deterministic?

update-assets-docker.sh
```diff
@@ -0,0 +1,35 @@
...
+  echo "Docker image $IMAGE_NAME does not exist or --reuse-image flag not provided. Building..."
+  docker build -t $IMAGE_NAME -f simulator/Dockerfile .
+else
+  echo "Docker image $IMAGE_NAME exists, and --reuse-image flag is set. Skipping build."
+fi
+
+for crate_name in "faucet" "radiswap" "flash_loan" "genesis_helper" "metadata" "test_environment" "global_n_owned" "kv_store"
+do
+  echo "Building $crate_name..."
+  docker run -it --entrypoint=scrypto -v $PWD:/src simulator build --path assets/blueprints/$crate_name
```
dhedey, line 33: I've no idea if this is possible, but can we make it so that we can provide parameters to the image instead of needing to override the endpoint?

Or maybe we map $PWD/assets/blueprints/$crate_name:/src and it works without overriding the entrypoint?

simulator/Dockerfile
balda-rdx: right on spot :)

simulator/Dockerfile
balda-rdx: yes

update-assets-docker.sh
balda-rdx: actually....I can update `update-assets.sh` so that it supports both modes. local and docker. will go that way

update-assets-docker.sh
balda-rdx: noted. will add

simulator/Dockerfile
balda-rdx: I'll check but I think there's a dependency between simulator and assets. without assets simulator was failing to build. will check again

update-assets-docker.sh
balda-rdx: oh, the idea is that people won't need to override the entrypoint. they should just run 
`docker run -v $PWD:/src simulator`
the problem here is that assets have local dependencies that use relative paths so I can't just mount the asset to be built, I need the dependent dirs as well, that's why I ended up bind mounting the $PWD and overriding the entrypoint

balda-rdx: will add a README with the steps people should run

simulator/Dockerfile
balda-rdx: yeah I'm afraid it is, otherwise apt won't find the dependencies. that first apt update is how apt updates the local database  with the packages available and versions

simulator/Dockerfile
balda-rdx: @iamyulong so the problem when removing assets is that I can't compile the radix-engine. 
Getting this error
```
62.41 error: could not compile `radix-engine` (lib) due to 5 previous errors
62.41 warning: build failed, waiting for other jobs to finish...
129.5 error: failed to compile `simulator v0.13.0 (/app/simulator)`, intermediate artifacts can be found at `/app/simulator/target`
```
is that expected?

update-assets-docker.sh
balda-rdx: done. merged both scripts into one. please do let me know if you wanna go that way or if you just prefer the docker approach

simulator/Dockerfile
iamyulong: Let's ignore this for now. The simulator does depend on the assets.

README.md
```diff
@@ -93,6 +93,31 @@ scrypto test
 
 **Note:** The commands use the default account as transaction sender.
 
+## Playing with the simulator
```
iamyulong, line 96: Can we change it `Compile blueprints with dockerized simulator`?

README.md
```diff
@@ -93,6 +93,31 @@ scrypto test
...
+We provide a docker image so that you can build your blueprints
```
iamyulong, line 97: `Follow this guide to build reproducible WASM and RDP files for your Scrypto blueprints.`

README.md
```diff
@@ -93,6 +93,31 @@ scrypto test
...
 
+## Playing with the simulator
+We provide a docker image so that you can build your blueprints 
+
+### Using local docker image
+The Dockerfile in the root of the repo should be work to build a docker image which will contain all the dependencies needed to be able build a blueprint using scrypto. 
+
+Build the docker image like. From the repo root
+```
+docker build -t radixdlt/simulator .
```
iamyulong, line 104: `sudo docker`?

README.md
```diff
@@ -93,6 +93,31 @@ scrypto test
...
+```
+
+Then build your package by just running
+```
+docker run -v <your-blueprint-directory>:/src radixdlt/simulator
```
iamyulong, line 109: `<path-to-your-scrypto-crate>`

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM rust:1.70.0-slim-bullseye as builder
+
+RUN apt update && apt install -y \
```
iamyulong, line 3: Replace `apt` with `apt-get`?

README.md
```diff
@@ -93,6 +93,31 @@ scrypto test
...
+```
+
+### Using published docker image
+If you would like to avoid building the docker image, you can skip the build step and do the second step directly, docker will automatically download the docker image we publish
+
+Build your blueprints directly with
+```
+docker run -v <your-blueprint-directory>:/src radixdlt/simulator
```
iamyulong, line 117: Ditto

README.md
iamyulong: and `sudo docker`?

README.md
balda-rdx: yes

README.md
balda-rdx: the use of sudo to run docker really depends on each installation. docker actually suggest to not run it with sudo :)

README.md
balda-rdx: updated it

balda-rdx:
> Nice work!
> 
> Can we possibly add a test, asserting the hash of "everything" example WASM file, run by CI for every commit and all OS types?

@iamyulong would you mind if we added it in a later PR?

iamyulong (approved):

PR #1569, opened 2023-09-19
Implement compile-time `dec!` and `pdec!` macros

lrubasze:
## Summary
Macros `dec!()` and `pdec!()` are capable to define a compile-time constant.

Additional changes:
- Remove `dec!()`and `pdec!()` macro variants with shifting
 
## Testing
Respective tests have been implemented 

### For dApp Developers
It is possible to declare any `Decimal`/`PreciseDecimal` as a compile-time constant.
```
  const a: Decimal = dec!(222);
  const b: Decimal = dec!("0.000001");
``` 
The same applies for `PreciseDecimal`

### DIFF
--- a/radix-engine-common/src/macros.rs
+++ b/radix-engine-common/src/macros.rs
@@ -32,32 +32,6 @@ macro_rules! dec {
     ($x:literal) => {
         $crate::math::Decimal::try_from($x).unwrap()
     };
-    ($base:literal, $shift:literal) => {
-        // Base can be any type that converts into a Decimal, and shift must support
-        // comparison and `-` unary operation, enforced by rustc.
-        {
-            let base = $crate::math::Decimal::try_from($base).unwrap();
-            if $shift >= 0 {
-                base.checked_mul(
-                    $crate::math::Decimal::try_from(
-                        $crate::math::I192::from(10u8)
-                            .pow(u32::try_from($shift).expect("Shift overflow")),
-                    )
-                    .expect("Shift overflow"),
-                )
-                .expect("Overflow")
-            } else {
-                base.checked_div(
-                    $crate::math::Decimal::try_from(
-                        $crate::math::I192::from(10u8)
-                            .pow(u32::try_from(-$shift).expect("Shift overflow")),
-                    )
-                    .expect("Shift overflow"),
-                )
-                .expect("Overflow")
-            }
-        }
-    };
 }
 
 /// Creates a safe integer from literals.
@@ -107,33 +81,6 @@ macro_rules! pdec {
     // If not then something is definitely wrong and panic is fine.
     ($x:literal) => {
         $crate::math::PreciseDecimal::try_from($x).unwrap()
-    };
-
-    ($base:literal, $shift:literal) => {
-        // Base can be any type that converts into a PreciseDecimal, and shift must support
-        // comparison and `-` unary operation, enforced by rustc.
-        {
-            let base = $crate::math::PreciseDecimal::try_from($base).unwrap();
-            if $shift >= 0 {
-                base.checked_mul(
-                    $crate::math::PreciseDecimal::try_from(
-                        $crate::math::I256::from(10u8)
-                            .pow(u32::try_from($shift).expect("Shift overflow")),
-                    )
-                    .expect("Shift overflow"),
-                )
-                .expect("Overflow")
-            } else {
-                base.checked_div(
-                    $crate::math::PreciseDecimal::try_from(
-                        $crate::math::I256::from(10u8)
-                            .pow(u32::try_from(-$shift).expect("Shift overflow")),
-                    )
-                    .expect("Shift overflow"),
-                )
-                .expect("Overflow")
-            }
-        }
     };
 }
 
--- a/radix-engine-common/src/macros.rs
+++ b/radix-engine-common/src/macros.rs
@@ -1,39 +1,3 @@
-/// Creates a `Decimal` from literals.
-///
-#[macro_export]
-macro_rules! dec {
-    (0) => {
-        $crate::math::Decimal::ZERO
-    };
-    ("0") => {
-        $crate::math::Decimal::ZERO
-    };
-    ("0.1") => {
-        $crate::math::Decimal::ONE_TENTH
-    };
-    ("1" | 1) => {
-        $crate::math::Decimal::ONE
-    };
-    (10) => {
-        $crate::math::Decimal::TEN
-    };
-    ("10") => {
-        $crate::math::Decimal::TEN
-    };
-    (100) => {
-        $crate::math::Decimal::ONE_HUNDRED
-    };
-    ("100") => {
-        $crate::math::Decimal::ONE_HUNDRED
-    };
-    // NOTE: Decimal arithmetic operation safe unwrap.
-    // In general, it is assumed that reasonable literals are provided.
-    // If not then something is definitely wrong and panic is fine.
-    ($x:literal) => {
-        $crate::math::Decimal::try_from($x).unwrap()
-    };
-}
-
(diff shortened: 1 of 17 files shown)


radix-engine-macros/src/lib.rs
```diff
@@ -1,57 +1,37 @@
...
-        }
-        Ok(quote! { #item_impl })
-    } else {
-        Err(Error::new(
-            Span::call_site(),
-            "Only functions and impls are supported by `catch_unwind`.",
-        ))
-    }
+#[proc_macro]
+pub fn decimal_string(input: TokenStream) -> TokenStream {
```
lrubasze, line 30: Would prefer to keep all decimal related stuff within one crate `radix-engine-common`.
Unfortunately procedural macros require separate crate, therefore decided to reuse this crate.

radix-engine-common/src/math/decimal.rs
```diff
@@ -80,6 +80,14 @@ impl Decimal {
         Self::ONE
     }
 
+    pub const fn const_mul(self, other: Self) -> Self {
```
lrubasze, line 83: I know, it doesn't look good. But decided to apply minimal changes for now, due to oncoming release.
Ultimate target is to rework existing functions to `const`, which would allow to have compile-time arith ops for Decimal types! Believe it would be really cool feature!

radix-engine-common/src/macros.rs
```diff
@@ -29,32 +29,28 @@ macro_rules! dec {
...
-    };
+    ($x:literal) => {{
+        let (dec_string, dec_places) = radix_engine_macros::decimal_string!($x);
+        if dec_places > $crate::math::Decimal::SCALE {
+            panic!("Unsupported decimal places");
+        }
+        let multiplier: $crate::math::I192 =
+            $crate::math::I192::TEN.const_pow($crate::math::Decimal::SCALE - dec_places);
+        let b = $crate::math::I192::parse_str(dec_string);
+        $crate::math::Decimal(b.const_mul(multiplier))
```
iamyulong, line 40: Why not moving all the computation to the macro?

radix-engine-common/src/macros.rs
```diff
@@ -29,32 +29,28 @@ macro_rules! dec {
...
+    }};
     ($base:literal, $shift:literal) => {
```
iamyulong, line 42: I wonder if we should remove the `($base, $shift)` variant. There is overflow panics as well.

radix-engine-common/src/math/bnum_integer.rs
```diff
@@ -59,6 +59,41 @@ macro_rules! types {
                     pub const BITS: u32 = $wrap::BITS as u32;
                     pub const BYTES: u32 = $wrap::BYTES as u32;
                     pub const N: usize = ($wrap::BYTES / 8) as usize;
+
+                    pub const fn const_from_str(from: &str) -> Result<Self, [<Parse $t Error>] > {
```
iamyulong, line 63: Just wanted to make sure that you're aware of this. Not all `const fn` are pre-computed per rust documentation.

radix-engine-common/src/math/bnum_integer.rs
```diff
@@ -59,6 +59,41 @@ macro_rules! types {
...
+                            Some(val) => Self(val),
+                            None => panic!("Overflow"),
+                        }
+                    }
+
+
+                    pub const fn const_pow(self, pow: u32) -> Self {
+                        match self.0.checked_pow(pow) {
+                            Some(val) => Self(val),
+                            None => panic!("Overflow"),
```
iamyulong, line 94: Suggest we do not add all these overflow variants, at least not close to release.

radix-engine-macros/src/decimal.rs
```diff
@@ -0,0 +1,64 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, Expr};
+
```
iamyulong, line 4: You can return a `Result`, which will be presented to users as a compile-time error.

radix-engine-common/src/macros.rs
lrubasze: Happy to do this. Don't think it is really useful.

radix-engine-common/src/math/bnum_integer.rs
lrubasze: Refactored whole implementation. Hope it looks better :)

radix-engine-macros/src/decimal.rs
lrubasze: Sure, good point!

radix-engine-common/src/macros.rs
lrubasze: Removed

radix-engine-common/src/macros.rs
lrubasze: Done

radix-engine-macros/src/lib.rs
lrubasze: Comment no longer valid. Whole implementation refactored.

radix-engine-macros/src/decimal.rs
```diff
@@ -0,0 +1,83 @@
...
+    match expr {
+        Expr::Lit(lit) => match &lit.lit {
+            syn::Lit::Str(lit_str) => {
+                radix_engine_common::math::Decimal::try_from(lit_str.value()).unwrap()
+            }
+            syn::Lit::Int(lit_int) => {
+                radix_engine_common::math::Decimal::try_from(lit_int.base10_digits()).unwrap()
+            }
+            syn::Lit::Bool(lit_bool) => radix_engine_common::math::Decimal::from(lit_bool.value),
+            _ => panic!("Unsupported literal type!"),
```
dhedey, line 17: In all of these panics, we shouldn't panic, we should return an error with the span of the literal, in order to get the right behaviour in the IDE.

EG:
```rust
other_literal => return Err(Error::new(other_literal.span, "This macro only supports string and integer literals."));
```

dhedey (commented):
Nice stuff! A few additional thoughts

radix-engine-macros/src/decimal.rs
lrubasze: Just working on this after chat with Yulong :)

radix-engine-macros/src/decimal.rs
```diff
@@ -0,0 +1,83 @@
...
+fn get_decimal_from_expr(expr: &Expr) -> radix_engine_common::math::Decimal {
+    match expr {
+        Expr::Lit(lit) => match &lit.lit {
+            syn::Lit::Str(lit_str) => {
+                radix_engine_common::math::Decimal::try_from(lit_str.value()).unwrap()
+            }
+            syn::Lit::Int(lit_int) => {
+                radix_engine_common::math::Decimal::try_from(lit_int.base10_digits()).unwrap()
+            }
+            syn::Lit::Bool(lit_bool) => radix_engine_common::math::Decimal::from(lit_bool.value),
```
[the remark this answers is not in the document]
lrubasze: It is pretty easy. I have a working PoC.
The thing is that there are rounding issues when parsing floats.
I think we would need more research and some discussion how to handle it.

radix-engine-macros/src/decimal.rs
```diff
@@ -0,0 +1,83 @@
...
+            _ => panic!("Unsupported literal type!"),
+        },
+        Expr::Group(group) => get_decimal_from_expr(&group.expr),
```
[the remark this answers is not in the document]
lrubasze: No need. It was needed in previous implementation.

iamyulong (approved):

iamyulong:
Looks good but we need to fix the no_std :)

PR #1573, opened 2023-09-20
Misc: Rename SubstateMutability to LockStatus

iamyulong:
## Summary

Rename `SubstateMutability` to `LockStatus` for fields and key value entries.


radix-engine/src/blueprints/models/native_blueprint_state_macro.rs
```diff
@@ -853,9 +853,9 @@ mod helper_macros {
             paste::paste! {
                 KVEntry {
                     value: $entry_substate.value.map(|v| scrypto_encode(&v).unwrap()),
-                    locked: match $entry_substate.mutability {
-                        SubstateMutability::Mutable => true,
-                        SubstateMutability::Immutable => false,
+                    locked: match $entry_substate.lock_status {
+                        LockStatus::Locked => true,
+                        LockStatus::Unlocked => false,
                     },
```
mstrug-rdx, line 859: Previously if `KVEntry` was mutable that `match` was settings `locked` to `true`, so probably this was a bug.
In updated code this logic has changed to proper value (mutable (unlocked) -> locked=false).

mstrug-rdx (approved):

PR #1587, opened 2023-09-21
publish the scrypto builder docker image

balda-rdx:


.github/workflows/publish-scrypto-builder.yml
```diff
@@ -0,0 +1,32 @@
+name: scrypto builder publish
```
dhedey, line 1: `Publish scrypto-builder image`?

.github/workflows/ci-scrypto-builder.yml
```diff
@@ -1,4 +1,4 @@
-name: Simulator
+name: scrypto builder ci
```
dhedey, line 1: What's the purpose of this workflow out of interest? To show the build works? 

I'm not sure we necessarily need this if it'll be expensive to run on ever PR/Push

But happy if you want it

.github/workflows/ci-scrypto-builder.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     needs: tags
     uses: radixdlt/public-iac-resuable-artifacts/.github/workflows/docker-build.yml@main
     with:
-      runs_on: ubuntu-latest
+      runs_on: ubuntu-latest-16-cores
       image_registry: "docker.io"
       image_organization: "radixdlt"
       image_name: "simulator"
```
dhedey, line 31: The image name is wrong -- also this appears to publish to docker hub - is this intended?

.github/workflows/ci-scrypto-builder.yml
balda-rdx: hmm good point. what if we run on every merge to develop/main? the purpose is just to make sure that the docker image can still be built and it can be used to update the assets

balda-rdx: we can also build if and only if there were changes to the assets or the simulator folder

.github/workflows/publish-scrypto-builder.yml
balda-rdx: yes

.github/workflows/ci-scrypto-builder.yml
balda-rdx: nope. it should not published. fixing

iamyulong (approved):

PR #1588, opened 2023-09-21
Test: Fuzz decimal

talekhinezh:
Fuzz decimal using libfuzzer


fuzz-tests/src/bin/decimal.rs
```diff
@@ -0,0 +1,40 @@
...
+    let _ = decimal.0.checked_abs();
+    let _ = decimal.0.checked_cbrt();
+    let _ = decimal.0.checked_ceiling();
+    let _ = decimal.0.checked_floor();
+    let _ = decimal.0.checked_neg();
+    if decimal.4 >= 0 && decimal.4 <= Decimal::SCALE as i32 {
+        let _ = decimal.0.checked_round(decimal.4, decimal.5);
+    }
+    let _ = decimal.0.checked_add(decimal.1);
+    let _ = decimal.0.checked_sub(decimal.1);
```
lrubasze, line 28: would also try 
```
decimal.1.checked_sub(decimal.0)
decimal.1.checked_div(decimal.0)
```

fuzz-tests/src/bin/decimal.rs
```diff
@@ -0,0 +1,40 @@
...
+    let _ = decimal.0.checked_mul(decimal.1);
+    let _ = decimal.0.checked_div(decimal.1);
+
+    let string = decimal.0.to_string();
```
lrubasze, line 32: 1. How about testing `decimal.1` as well?

2. How about converting to primitive types as well?
```
u8::try_from(decimal.0).unwrap()
u16::try_from(decimal.0).unwrap()
...
```

lrubasze:
Nice work!
Adding AFL will be trivial now. The only thing is that AFL needs some initial input, but that should not be a problem :)

fuzz-tests/src/bin/decimal.rs
```diff
@@ -0,0 +1,40 @@
...
+use libfuzzer_sys::fuzz_target;
+
+use radix_engine_common::math::*;
+
+#[derive(Debug, Arbitrary)]
+struct OneDecimal(Decimal, Decimal, i64, u32, i32, RoundingMode);
+
+// Fuzzer entry points
+#[cfg(feature = "libfuzzer-sys")]
+fuzz_target!(|decimal: OneDecimal| {
```
lrubasze, line 14: Thinking if it would make sense to increase somehow a chance to hit corner cases, eg. Decimal::MAX, Decimal::MIN

fuzz-tests/src/bin/decimal.rs
talekhinezh: Hrm, not sure if testing this ourselves is useful since it's the fuzzer's job to find interesting cases?

lrubasze: true. makes sense

PR #1595, opened 2023-09-25
Test: RoleAssignment fuzzer

0xOmarA:
## Summary

This PR adds a role-assignment state fuzzer which checks that the invariants of the role-assignment module are up-held with random inputs. This fuzzer performs some invocations before the attachment of the module and some invocations after the attachment of the module. 

## Description

* Added a role-assignment fuzzer that checks for the following invariants:
    * There can never be role assignments for reserved roles (role names that start with an `_`).
    * There can never be any role assignments for roles in reserved spaces (such as roles of the RoleAssignment module itself)
    * All of the role names must respect the name rules we have (e.g., none of them can have undisplayable characters)
    * All roles must have names that do not exceed the maximum allowed length.
    * All `AccessRule`s respect the depth and width constraints on access rules.
    * No transaction that involves role-assignment invocations ends with a native vm panic.
    * No additional roles can be added after the module is created.
* Added a `RoleAssignmentDatabaseChecker` struct which implements all of the above checks (except for the last one which is implemented in the fuzzer's `check_invariants` function). This type implements `ApplicationChecker` and thus can be used with the system database checker. 
* `SystemDatabaseChecker::new` now takes an argument of the application checker instead of defaulting to the default value of the application checker used. This allows us to have checkers with some state (the `RoleAssignmentDatabaseChecker` is an example of a database checker that makes use of state). `SystemDatabaseChecker` now implements `Default` which behaves in the same way as `SystemDatabaseChecker::new` did before this change.
* Added a `define_composite_checker` macro which is used to define composite database checkers. This is now used in the `TestRunner`'s implementation of the `Drop` trait so that we can have more than one checker.
* Modified the implementation of `Drop` for the `TestRunner` so that we run the both the `ResourceDatabaseChecker` and the `RoleAssignmentDatabaseChecker` through the composite checker added with `define_composite_checker`.
* Added an additional argument to `ApplicationChecker::on_field` and `ApplicationChecker::on_collection_entry` of the `ModuleId`.
* Following from the above point, the implementation of `SystemDatabaseChecker::check_db` used to only invoke the `on_field` and `on_collection_entry` methods if the module was the main module. This has been changed so that those methods are invoked regardless of the module and the `ModuleId` is now passed to them so that individual `ApplicationChecker`s can determine if they wish to ignore this field or collection entry or not.


fuzz-tests/src/bin/attached_modules/role_assignment.rs
```diff
@@ -0,0 +1,380 @@
+//! This binary has fuzz tests for the role-assignment node module. This fuzzer creates random
```
lrubasze, line 1: Is there particular purpose of putting the test target into a separate folder?

radix-engine/src/system/checkers/role_assignment_db_checker.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// not a valid [`RoleAssignmentCollection`].
+    InvalidCollectionIndex(CollectionIndex),
+
+    /// Attempted to decode the data as the FieldEntry associated with that field but failed to do
+    /// so.
+    FailedToDecodeFieldValue(RoleAssignmentField, Vec<u8>),
+
+    /// Attempted to decode the data as the KeyPayload or EntryPayload associated with a collection
+    /// entry but failed.
+    FailedToDecodeCollectionKeyOrValue(RoleAssignmentCollection, Vec<u8>),
```
talekhinezh, line 274: The four errors above should be checked by the system db checker as it verifies the payloads against the blueprint schema. Perhaps these errors should be wrapped in an additional `UnexpectedError` enum as hitting the above errors would indicate a mismatch between the checker and the package schema stored.

scrypto-unit/src/test_runner.rs
```diff
@@ -477,11 +477,20 @@ impl<E: NativeVmExtension, D: TestDatabase> Drop for TestRunner<E, D> {
             .check_db(&self.database)
             .expect("Database should be consistent");
 
+        // Defining a composite checker of all of the application db checkers we have.
+        radix_engine::define_composite_checker! {
```
talekhinezh, line 481: Would be better I think if the name of the composite checker was one of the arguments into the macro.

fuzz-tests/Cargo.toml
```diff
@@ -9,7 +9,7 @@ cargo-fuzz = true
 
 [dependencies]
 libfuzzer-sys = { version = "0.4", optional = true }
-afl = { version = "0.12.16", features = ["reset_lazy_static"], optional = true }
+afl = { version = "0.12.16", optional = true }
```
lrubasze, line 12: You can also remove the `[patch.crates-io]` section. It is no longer needed.

lrubasze:
This code does not compile. 
Just thinking that maybe we should extend the CI to check other fuzzers? 
ATM we build only transaction fuzzer as a default one

radix-engine/src/system/checkers/role_assignment_db_checker.rs
0xOmarA: Sounds good on the `UnexpectedError`, will add that enum.

scrypto-unit/src/test_runner.rs
0xOmarA: Sounds good, making that change.

fuzz-tests/Cargo.toml
0xOmarA: Sounds good, I've removed that.

0xOmarA:
> This code does not compile.
Just thinking that maybe we should extend the CI to check other fuzzers?
ATM we build only transaction fuzzer as a default one.
We can do it in separate PR, this is just a loud thinking :)

@lrubasze Agreed that should do that, a `cargo check --all` on the crate should be able to detect those errors in CI. This can be in a next PR.

talekhinezh (approved):

PR #1596, opened 2023-09-25
Support AFL and simple_fuzzer for decimal fuzz tests 

lrubasze:
## Summary
- Add AFL support for decimal tests
- Add 'simple_fuzzer' support for decimal tests
- Let 'simple_fuzzer' work with Arbitrary types

### DIFF
--- a/fuzz-tests/src/bin/decimal.rs
+++ b/fuzz-tests/src/bin/decimal.rs
@@ -4,14 +4,15 @@ use arbitrary::Arbitrary;
 #[cfg(feature = "libfuzzer-sys")]
 use libfuzzer_sys::fuzz_target;
 
+#[cfg(feature = "afl")]
+use afl::fuzz;
+
 use radix_engine_common::math::*;
 
 #[derive(Debug, Arbitrary)]
 struct OneDecimal(Decimal, Decimal, i64, u32, i32, RoundingMode);
 
-// Fuzzer entry points
-#[cfg(feature = "libfuzzer-sys")]
-fuzz_target!(|decimal: OneDecimal| {
+fn fuzz_decimal(decimal: OneDecimal) {
     let _ = decimal.0.checked_sqrt();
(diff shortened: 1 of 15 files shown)


fuzz-tests/src/bin/decimal.rs
```diff
@@ -50,4 +55,49 @@ fuzz_target!(|decimal: OneDecimal| {
     let _ = decimal.0.checked_powi(decimal.2);
     let _ = decimal.0.checked_nth_root(decimal.3);
      */
-});
\ No newline at end of file
+}
+
+// Fuzzer entry points
+#[cfg(feature = "libfuzzer-sys")]
```
0xOmarA, line 61: Since we will be focusing more on fuzzing as an automated testing method I think that it might make more sense for us to invest sometime into a `fuzz` macro to lower the amount of boilerplate code required to write fuzz tests that utilize libFuzzer and afl. I imagine something like the following:

```rs
use fuzz_tests::fuzz;

fuzz!(|input: OneDecimal| {
    /* Fuzzing */
})
```

Would expand to:

```rs
#[cfg(feature = "libfuzzer-sys")]
libfuzzer_sys::fuzz_target!(|decimal: OneDecimal| {
    /* Fuzzing */
});

#[cfg(feature = "afl")]
fn main() {
    afl::fuzz!(|decimal: OneDecimal| {
        /* Fuzzing */
    });
}

#[cfg(feature = "simple-fuzzer")]
fn main() {
    fuzz_tests::fuzz!(|decimal: OneDecimal| {
        /* Fuzzing */
    });
}
```

This is nice because it removes the conditional `use` statements, conditional `main` functions and most `cfg` and `cfg_attr`s from the code we write and off-loads that to the macro.

lrubasze: I love it!

0xOmarA (approved):

PR #1600, opened 2023-09-26
Test: Better system fuzzer

talekhinezh:
## Summary
Use LibFuzzer + AFL to implement better System layer fuzzer


fuzz-tests/fuzz.sh
```diff
@@ -107,7 +107,7 @@ function fuzzer_libfuzzer() {
     set -x
     cargo +nightly fuzz $cmd \
         --release \
-        --no-default-features --features std,libfuzzer-sys \
+        --no-default-features --features std,libfuzzer-sys,post_run_db_check\
```
lrubasze, line 110: no need to add `post_run_db_check` when building AFL (line 135)?

talekhinezh: ah good point

lrubasze (approved):

PR #1602, opened 2023-09-27
Support for getting code coverage after fuzzing

lrubasze:
## Summary
This PR allows getting code coverage after fuzzing session is over.

NOTE!
It includes changes also from https://github.com/radixdlt/radixdlt-scrypto/pull/1608

### Code coverage data
`get-coverage-data.sh`
Usage:
`get-coverage-data.sh <optional_target> <corpus_mode>`
If no target is provided then script tries to get coverage for all available targets.
It is possible to retrieve coverage data only if fuzzing corpus data exists for given target:
- AFL - `afl/<target>/<afl_instance_name>/queue`
- libfuzzer  - `corpus/<target>`

Corpus mode tells what to do withe to corpus data before getting code coverage.
Available options:
- `full` - use full corpus generated during fuzzing session. Might last couple of hours and use hundreds of GBs.
- `cmin` - minimise the corpus. Speeds up code coverage generation. Might produce slightly less coverage than `full`.
If no corpus mode is provided then `cmin` is taken by default 
  
### Code coverage report
`get-coverage-report.sh`
Usage:
`get-coverage-report.sh`
Generate coverage report in HTML format using `grcov` command


fuzz-tests/fuzz.sh
```diff
@@ -170,8 +170,8 @@ function fuzzer_simple() {
     fi
 
     set -x
-    cargo $cmd --release \
```
0xOmarA, line 173: Why was the `--release` flag removed here?

lrubasze: To get more debug information when using `simple_fuzzer` to reproduce some crashes.
Just thinking that maybe I could allow to specify it. 
Would be useful for producing code coverage - should perform faster 🤔

lrubasze: Restored `--release` flag. Getting code coverage is way faster when in release mode.

talekhinezh (approved):

PR #1607, opened 2023-10-02
QEMU workflow update

mstrug-rdx:
## Summary
Updated QEMU github workflow.

## Details
All tests from `monkey-tests` takes 5~6 hours of QEMU execution, which is very close to github job time limit (6h). Because of that these tests has been divided into two separate jobs (automatic extraction of half of these tests to separate job).
Also updated `convert.pl` script to remove not needed data output.
Successful action run: https://github.com/radixdlt/radixdlt-scrypto/actions/runs/6377443462


lrubasze (approved):
Looks good :)

PR #1609, opened 2023-10-04
[ON HOLD] Metadata module refactoring

mstrug-rdx:
## Summary
Added verification of metadata key length in calls to functions: `get`, `lock` and `remove`, so the flow of the execution will be stopped in case of too long key and no internal calls to system and kernel layers will be invoked.

Improved memory handling when creating metadata with initial data.

## Testing
Added new test to validate key length check.

## Update Recommendations
Before this PR, querying with too long key will return `None` (as a metadata for such a key cannot be found in db), after this PR error: `MetadataError::KeyStringExceedsMaxLength` will be returned.


talekhinezh (changes requested):
This change breaks the protocol since previous transactions which used a longer key would now fail in a different way. Such a change would require atleast a blueprint version update.

PR #1617, opened 2023-10-10
Migrate fuzzer to another repo

lrubasze:
Moving fuzzer to: https://github.com/radixdlt/radixdlt-fuzz/pull/1


.github/workflows/ci.yml
```diff
@@ -410,26 +410,38 @@ jobs:
...
+    - uses: actions/checkout@v4
+      with:
+        repository: radixdlt/fuzzer
+        ref: develop
+        path: fuzzer
+        token: ${{ secrets.GH_PAT }}
+    - name: setup repos
+      run: |
+        rm -rf radixdlt-scrypto
+        ln -s $GITHUB_WORKSPACE/radixdlt-scrypto radixdlt-scrypto
```
0xOmarA, line 425: I believe that this might result in issues where the CI for `radixdlt/radixdlt-scrypto` fails if the fuzzer doesn't compile, which I do not think should be the case. 

As an example, say we remove `fn foo()` from `radixdlt/radixdlt-scrypto` but the fuzzer depends on this function. This would now mean that to make such a change we need to:

1. Open a PR with it in the scrypto repo.
2. Update the fuzzer in the fuzzer repo to use the PR branch/tag and make necessary code changes.
3. Run CI in Scrypto.
4. Merge Scrypto PR.
5. Update the fuzzer so that it no longer depends on a particular PR's branch/tag and depends on a branch/tag off of main or develop. 

This means that each breaking change to the interface now requires 2 PRs. If we have separated the fuzzer into a separate repo, then I believe that this bit of the CI should be in the fuzzer repo and not here as to not interfere with the scrypto CI.

0xOmarA:
Overall, I'm fine with the fuzzer being moved to a separate repo. Only issue I see is that making any small changes to `radixdlt/radixdlt-scrypto` that are required by the fuzzer are now a separate PR. As an example, additional `Arbitrary`, `Serialize` or `Deserialize` derives are now separate PRs into the Scrypto repo. This might be fine though as I believe that it can be mitigated by doing a pass over all of the types we have and adding the required derives.

.github/workflows/ci.yml
lrubasze: Thanks for the feedback.
Totally agree that the flow will be more complicated. And yes, 2 PRs will be required for breaking change.
This is an inevitable consequence of having more than one repo.

I was also thinking about removing `Build fuzzer` job, since similar check already exists in `fuzzer` repo.
But decided to leave it to get the information if the change is breaking or not :)
Otherwise people would simply forget about making a change `fuzzer` repo.

Sadly GH does not offer a feature to create dependencies between PRs. But we can investigate if there are 3rd party GH actions plugins. Until that time we can agree that if the CI fails for `Build fuzzer` job, we merge it anyway and make sure to create a respective PR for `fuzzer`.

0xOmarA: > I was also thinking about removing Build fuzzer job, since similar check already exists in fuzzer repo.
But decided to leave it to get the information if the change is breaking or not :)
Otherwise people would simply forget about making a change fuzzer repo.

I think that it might be okay if breaking interface changes are introduced to the Scrypto repo without the PR author updating the fuzzer. We can update the fuzzer later on or update it periodically (e.g., update it on each Monday). This would be a similar approach to things we have that depend on the Scrypto repo such as the Radix Engine Toolkit and the Babylon Node Rust core.

lrubasze: This is basically what I suggested. Merge a breaking change without PR in `fuzzer`. The only thing is that CI would fail on `Build fuzzer` job.
Let's  discuss it with the rest of the team.

lrubasze: There is also another option.
`fuzzer` uses `radixdlt-scrypto` as a submodule and by default it needs a specific git sha of the `radixdlt-scrypto` repo.
So, even if some breaking change is merged to the `radixdlt-scrypto` , `fuzzer` will be bound to the working git sha of the `radixdlt-scrypto`. And when someone needs to fuzz something more recent than git sha, he will have to bump the git sha first and fix update the `fuzzer` accordingly.

This way we could skip the `Build fuzzer` job in `radixdlt-scrypto`. How about that?

lrubasze: Greed to get rid of `Build fuzzer` job.

0xOmarA (approved):

PR #1618, opened 2023-10-11
Improve schema cache

iamyulong:
## Summary

Avoid scrypto schema clone, which improves genesis transaction processing by ~20%.


mstrug-rdx (approved):
Looks ok.

PR #1621, opened 2023-10-12
Test - moving proofs and buckets

mstrug-rdx:
## Summary
Added test on moving `Bucket` and `Proof` between components.


radix-engine-tests/tests/component.rs
```diff
@@ -113,3 +113,169 @@ fn blueprint_name_can_be_obtained_from_a_method() {
     // Assert
     assert_eq!(blueprint_name, "ComponentTest")
 }
+
+#[test]
+fn pass_bucket_to_other_component() {
```
talekhinezh, line 118: I think we can remove the pass_bucket and pass_proof tests as we have plenty of those. The pass bucket and proof though is the one we really need. Could we also add a pass vault and proof

radix-engine-tests/tests/blueprints/component/src/component.rs
```diff
@@ -53,3 +61,88 @@ mod component_test {
...
+                .globalize()
+        }
+
+        pub fn check_proof(&self, proof: Proof) {
+            proof.check(self.resource_id).drop();
+        }
+
+        pub fn check_proof_and_burn_bucket(&self, bucket: Bucket, proof: Proof) {
+            proof.check(self.resource_id).drop();
+            bucket.burn();
```
talekhinezh, line 145: Can we also try a test where the `bucket.burn()` happens before `proof.check()`?

mstrug-rdx: I checked that before and there is an error: `KernelError(CallFrameError(DropNodeError(NodeBorrowed(NodeId("f8eb898e760f64a79e74ac60a8465bb2c6cd7a68aae9bbe7eb70947d59d1")))))`

radix-engine-tests/tests/component.rs
mstrug-rdx: I am wondering how can I pass `Vault` returned by 1st component function using `ManifestBuilder` to 2nd component? I don't see any function like for `Bucket` or `Proofs` to take it and lookup by name. The only way to pass `Vault` between components is by cross-blueprint calls?

talekhinezh: It may not be possible in the manifest but should be possible to do in a function which the manifest calls

radix-engine-tests/tests/blueprints/component/src/component.rs
talekhinezh: Oh okay great, can we atleast add that test (with the error check)

talekhinezh (approved):

PR #1623, opened 2023-10-12
Add replay execute-in-memory

iamyulong:
## Summary

Add new command `replay run-in-memory` 

```
$ cargo run --release --bin replay -- execute  ../assets-lfs/babylon-mainnet-transactions-2023-10-11.tar.gz /tmp/database --max-version=500000
New version: 500000, 7dd4403a757f43f4a885e914b8dc38086fdbaf96082fa90067acf1500075e85d, 00:31:02
Time elapsed: 1862.198134519s
State version: 500000
State root hash: 7dd4403a757f43f4a885e914b8dc38086fdbaf96082fa90067acf1500075e85d

$ cargo run --release --bin replay -- execute-in-memory  ../assets-lfs/babylon-mainnet-transactions-2023-10-11.tar.gz --max-version=500000
Time elapsed: 1358.921315754s
State version: 500000
State root hash: 7dd4403a757f43f4a885e914b8dc38086fdbaf96082fa90067acf1500075e85d
```


radix-engine-stores/src/hash_tree_support.rs
```diff
@@ -12,15 +12,17 @@ pub struct HashTreeUpdatingDatabase<D> {
     tree_store: TypedInMemoryTreeStore,
     current_version: Version,
     current_hash: Hash,
+    jmt_enabled: bool,
```
talekhinezh, line 15: Not sure if this is the right place for the flag. Feels like if we want to disable jmt we should just not use the `HashTreeUpdatingDatabase`.

simulator/src/replay/cmd_run.rs
```diff
@@ -150,3 +90,90 @@ impl Run {
         Ok(())
     }
 }
+
+pub fn print_progress(duration: Duration, new_version: u64, new_root: Hash) {
```
talekhinezh, line 94: Could probably keep this function and the other functions private.

talekhinezh (approved):

radix-engine-stores/src/hash_tree_support.rs
iamyulong: I had the same thought initially, then I realised that I would have to track progress manually (which is already done by `HashTreeUpdatingDatabase`). I then viewed it as `VersionedSubstateDatabase`.

Since executing transaction without JMT is mostly for testing/anayzing (instead of state hash computation), I will remove this flag completely. I will make dedicated command for other purposes.

simulator/src/replay/cmd_run.rs
iamyulong: Will make it mod private.

radix-engine-stores/src/hash_tree_support.rs
iamyulong: Done

talekhinezh (approved):

PR #1624, opened 2023-10-12
Feature: Add "replay sync" command

talekhinezh:
## Summary
Add `replay sync <state_manager_db> <destination_db>` which will sync and execute the transactions from the state_manager_db. Part of this process will be checking that the state hashes match.


iamyulong (approved):
Very cool producer-consumer mulit-threading!

PR #1629, opened 2023-10-17
Switch to rust stable

lrubasze:
## Summary

Switch `radixdlt-scrypto` to Rust `stable`

## Details
- Fix `ambiguous name` warnings
- Switched GH workflows to `stable` Rust
- Switched Dockerfile to `stable` Rust docker image

### DIFF
--- a/radix-engine-common/src/data/scrypto/custom_schema.rs
+++ b/radix-engine-common/src/data/scrypto/custom_schema.rs
@@ -211,56 +211,6 @@ impl HasSchemaHash for VersionedScryptoSchema {
     }
 }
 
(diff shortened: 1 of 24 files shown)


radix-engine-common/src/data/scrypto/custom_schema.rs
```diff
@@ -211,56 +211,6 @@ impl HasSchemaHash for VersionedScryptoSchema {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Copy)]
-pub struct ScryptoCustomExtension {}
```
iamyulong, line 215: Nice spot! Surprised that we got two implementations.

iamyulong (approved):

PR #1635, opened 2023-10-26
Fix proof name reservation

lrubasze:
Use proper method when reserving the proof name.

### DIFF
--- a/transaction/src/builder/manifest_builder.rs
+++ b/transaction/src/builder/manifest_builder.rs
@@ -2052,17 +2052,83 @@ impl Default for ManifestBuilder {
 mod tests {
     use super::*;
 
-    #[test]
-    fn test_clone_proof_with_add_instruction_advanced() {
-        let _builder = ManifestBuilder::new()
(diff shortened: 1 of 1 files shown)


0xOmarA (approved):

transaction/src/validation/transaction_validator.rs
```diff
@@ -888,4 +888,30 @@ mod tests {
...
+            123,
+            vec![55],
+            66,
+            ManifestBuilder::new()
+                .take_from_worktop(XRD, dec!(100), "bucket")
+                .create_proof_from_bucket_of_amount("bucket", dec!(5), "proof")
+                .then(|builder| {
+                    let lookup = builder.name_lookup();
+                    let proof_id = lookup.proof("proof");
+                    builder
```
dhedey, line 906: Given this test name, I'd expect this to use the clone proof command, rather than advanced instructions.

Perhaps it would be a more explicit to create a test file for the manifest builder itself, and explicitly add tests for the "add_instruction_advanced" command, which checks it is collision resistant against existing proofs/buckets named "proof" and "bucket"?

dhedey (approved):
Nice, thanks for the fix 👍

PR #1636, opened 2023-10-26
Const constructors for `NonFungibleLocalId`s

0xOmarA:
## Summary

Constant constructors for `NonFungibleLocalId`.

## Details

This PR has no behavioral changes. It only adds `const` constructors for `NonFungibleLocalId` to allow it to be used in const contexts.

The previous set of constructors is kept in place with very few changes in order to maintain backward compatibility. A new set of const constructors with the `const` prefix in their name have been added: `const_integer`, `const_ruid`, `const_string`, and `const_bytes`.

Feel free to review each commit separately.


iamyulong (approved):
LGTM

PR #1637, opened 2023-10-26
[CLI] replay command: heap memory alloc dump

mstrug-rdx:
## Summary
Added command to execute transactions from archive and dump memory allocated during execution.
Refactored `InfoAlloc` module to support multi-threading and moved to `radix-engine-profiling` project.


radix-engine-profiling/src/info_alloc/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+        self.enabled.load(Ordering::Acquire)
+    }
+
+    /// Resets internal counters. Usually used to start measurement.
+    pub fn reset_counters(&self) {
+        INFO_ALLOC_DATA_TLS.with(|val| {
+            val.sum_counter.store(0, Ordering::Release);
+            val.current_level.store(0, Ordering::Release);
+            val.max_level.store(0, Ordering::Release);
+        });
```
iamyulong, line 66: Why do we need atomics if they're thread-local?

radix-engine-profiling/src/info_alloc/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+        if self.is_enabled() {
+            self.decrease_counters(layout.size());
+        }
+        self.allocator.dealloc(ptr, layout);
+    }
+
+    unsafe fn realloc(&self, ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8 {
+        if self.is_enabled() {
+            self.realloc_decrease_counter(layout.size(), new_size);
```
iamyulong, line 142: Maybe 
```
if new_size > layout.size() {
    self.increase_counters(...);
} else {
    self.decrease_counters(...);
}
```

simulator/src/replay/cmd_alloc_dump.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    /// Ignore dump of user type of transactions
+    #[clap(long)]
+    pub ignore_dump_user_tx: bool,
+    /// Dump genesis type of transactions
+    #[clap(short = 'g', long)]
+    pub dump_genesis_tx: bool,
+    /// Dump round update type of transactions
+    #[clap(short = 'r', long)]
+    pub dump_round_update_tx: bool,
```
iamyulong, line 52: Suggest rename to `--exclude-user-transaction`, `--include-generic-transaction` and `--include-round-update-transaction`

radix-engine-profiling/src/info_alloc/mod.rs
mstrug-rdx: Yes, this is a leftover from a previous implementation (when counter was shared between threads), should be fine to change to ordinary number type.

radix-engine-profiling/src/info_alloc/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+/// Allocation data stored in Thread Local Store (separate data for each thread)
+pub struct InfoAllocData {
+    /// Sum of bytes allocated during measurements (for reallocations only additional memory is counted, no dealocation is counted)
+    sum_counter: AtomicIsize,
+    /// Current level of allocated bytes (allocation and deallocation are counted, incl. reallocation)
+    current_level: AtomicIsize,
+    /// Maximum level (peak) of allocated bytes (the max of this field and current_level)
+    max_level: AtomicIsize,
```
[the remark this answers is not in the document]
mstrug-rdx: Signed number is needed as the counters can be enabled in any time of program execution, in such scenario:
1. Allocate 10 bytes
2. Start counting allocations (set counters to 0)
3. Free allocated 10 bytes in point 1
this will result in counter current level value equals -10, for unsigned counter we need to have 0.
Then if we will allocate 11 bytes signed counter will have value 1 and unsigned would have 11.

radix-engine-profiling/src/info_alloc/mod.rs
mstrug-rdx: Yes, it can be implemented that way.

radix-engine-profiling/src/info_alloc/mod.rs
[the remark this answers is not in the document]
iamyulong: Makes sense. Thanks for explaining it.

radix-engine-profiling/src/info_alloc/mod.rs
mstrug-rdx: I've changed counter types, but it requires usage of  `ReffCell` as thread local data due to needed interior mutability (atomics are implemented using `UnsafeCell` internally).

iamyulong (approved):

PR #1639, opened 2023-10-31
Assert scenario end state

iamyulong:
## Summary

Assert scenario execution end state.

### DIFF
--- a/radix-engine-common/src/crypto/hash_accumulator.rs
+++ b/radix-engine-common/src/crypto/hash_accumulator.rs
@@ -29,6 +29,15 @@ impl HashAccumulator {
         }
     }
 
+    pub fn update_no_chain(&mut self, data: impl AsRef<[u8]>) {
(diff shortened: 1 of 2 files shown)


0xOmarA (approved):
Looks good.

Any thoughts on an approach to make it clearer which scenarios diverged and which ones did not. Perhaps by asserting a hash of the state changes in each tx?

dhedey (approved):
Looks good.

A couple of additional thoughts:
* I wonder if it might be better to run each scenario separately, and have a resultant root hash per scenario? This might then allow us to run scenarios against different engine versions more explicitly, and check they all give the same hash.
* I wonder if we can have some kind of make-do receipt hash accumulator as well as the state hash accumulator? Basically, looking at the `LedgerTransactionReceiptV1`, we basically just need to check events agree as well as state changes. These are probably captured in the state store through differing fee payments anyway, but just incase, we should probably capture that too.

PR #1640, opened 2023-10-31
Code coverage support for tests (scrypto coverage command)

bbarwikowski-hacken:
## Summary
This PR adds new command `scrypto coverage` which allows to create code coverage report from tests. It was implemented based on https://github.com/hknio/code-coverage-for-webassembly findings.

Please let me know what you think, in same cases I had to use ugly approach like:
```rust
/// The maximum invoke payload size.
#[cfg(not(feature = "coverage"))]
pub const MAX_INVOKE_PAYLOAD_SIZE: usize = 1 * 1024 * 1024;
#[cfg(feature = "coverage")]
pub const MAX_INVOKE_PAYLOAD_SIZE: usize = 32 * 1024 * 1024;
```
If you have better idea how it could be implemented, please let me know.

In the next commit I'll add test_coverage.sh script, the only problem is I'll need to install nightly version of rust and the same version of llvm as nightly rustc is using. Should I add it in Dockerfile or somewhere else?

### DIFF
--- a/radix-engine/src/transaction/transaction_executor.rs
+++ b/radix-engine/src/transaction/transaction_executor.rs
@@ -59,6 +59,7 @@ pub struct CostingParameters {
 }
 
 impl Default for CostingParameters {
+    #[cfg(not(feature = "coverage"))]
     fn default() -> Self {
         Self {
             execution_cost_unit_price: EXECUTION_COST_UNIT_PRICE_IN_XRD.try_into().unwrap(),
@@ -69,6 +70,19 @@ impl Default for CostingParameters {
             usd_price: USD_PRICE_IN_XRD.try_into().unwrap(),
             state_storage_price: STATE_STORAGE_PRICE_IN_XRD.try_into().unwrap(),
             archive_storage_price: ARCHIVE_STORAGE_PRICE_IN_XRD.try_into().unwrap(),
+        }
+    }
+    #[cfg(feature = "coverage")]
+    fn default() -> Self {
+        Self {
+            execution_cost_unit_price: Decimal::zero(),
+            execution_cost_unit_limit: u32::MAX,
+            execution_cost_unit_loan: u32::MAX,
+            finalization_cost_unit_price: Decimal::zero(),
+            finalization_cost_unit_limit: u32::MAX,
+            usd_price: USD_PRICE_IN_XRD.try_into().unwrap(),
+            state_storage_price: Decimal::zero(),
+            archive_storage_price: Decimal::zero(),
         }
     }
 }
--- a/radix-engine/src/vm/wasm/prepare.rs
+++ b/radix-engine/src/vm/wasm/prepare.rs
@@ -983,15 +983,18 @@ impl WasmModule {
         mut self,
         rules: &R,
     ) -> Result<Self, PrepareError> {
-        let backend = gas_metering::host_function::Injector::new(
-            MODULE_ENV_NAME,
-            COSTING_CONSUME_WASM_EXECUTION_UNITS_FUNCTION_NAME,
-        );
-        gas_metering::inject(&mut self.module, backend, rules).map_err(|err| {
-            PrepareError::RejectedByInstructionMetering {
-                reason: err.to_string(),
-            }
-        })?;
+        #[cfg(not(feature = "coverage"))]
+        {
+            let backend = gas_metering::host_function::Injector::new(
+                MODULE_ENV_NAME,
+                COSTING_CONSUME_WASM_EXECUTION_UNITS_FUNCTION_NAME,
+            );
+            gas_metering::inject(&mut self.module, backend, rules).map_err(|err| {
+                PrepareError::RejectedByInstructionMetering {
+                    reason: err.to_string(),
+                }
+            })?;
+        }
 
         Ok(self)
     }
--- a/scrypto-unit/src/test_runner.rs
+++ b/scrypto-unit/src/test_runner.rs
@@ -125,13 +125,13 @@ impl Compile {
                 coverage_path.set_extension("rpd");
                 let definition = fs::read(&coverage_path).unwrap_or_else(|err| {
                     panic!(
-                        "Failed to read manifest definition from path {:?} - {:?}",
+                        "Failed to read package definition from path {:?} - {:?}",
                         &coverage_path, err
                     )
                 });
                 let definition = manifest_decode(&definition).unwrap_or_else(|err| {
                     panic!(
-                        "Failed to parse manifest definition from path {:?} - {:?}",
+                        "Failed to parse package definition from path {:?} - {:?}",
                         &coverage_path, err
                     )
                 });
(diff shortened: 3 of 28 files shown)


radix-engine/src/vm/wasm/prepare.rs
```diff
@@ -983,15 +983,18 @@ impl WasmModule {
...
-        let backend = gas_metering::host_function::Injector::new(
-            MODULE_ENV_NAME,
-            COSTING_CONSUME_WASM_EXECUTION_UNITS_FUNCTION_NAME,
-        );
-        gas_metering::inject(&mut self.module, backend, rules).map_err(|err| {
-            PrepareError::RejectedByInstructionMetering {
-                reason: err.to_string(),
-            }
-        })?;
+        #[cfg(not(feature = "coverage"))]
```
iamyulong, line 986: Out of curiosity, did you run into any issue with gas metering ingestion for the coverage-enabled wasm code?

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -71,6 +72,19 @@ impl Default for CostingParameters {
...
+        Self {
+            execution_cost_unit_price: Decimal::zero(),
+            execution_cost_unit_limit: u32::MAX,
+            execution_cost_unit_loan: u32::MAX,
+            finalization_cost_unit_price: Decimal::zero(),
+            finalization_cost_unit_limit: u32::MAX,
+            usd_price: USD_PRICE_IN_XRD.try_into().unwrap(),
+            state_storage_price: Decimal::zero(),
+            archive_storage_price: Decimal::zero(),
+        }
```
iamyulong, line 86: Due to the difference in configuration, some tests may fail in `coverage` mode but not in regular test mode. I guess it's not a big deal for most cases. We will need to make sure the `--no-fail-fast` flag is provided when running coverage.

scrypto-unit/src/test_runner.rs
```diff
@@ -122,12 +98,79 @@ impl Compile {
...
+                coverage_path.set_extension("rpd");
+                let definition = fs::read(&coverage_path).unwrap_or_else(|err| {
+                    panic!(
+                        "Failed to read manifest definition from path {:?} - {:?}",
+                        &coverage_path, err
+                    )
+                });
+                let definition = manifest_decode(&definition).unwrap_or_else(|err| {
+                    panic!(
+                        "Failed to parse manifest definition from path {:?} - {:?}",
```
iamyulong, line 134: Minor: `manifest definition` -> `package definition`

simulator/src/scrypto/cmd_coverage.rs
```diff
@@ -0,0 +1,204 @@
...
+        let llvm_major_version = Regex::new(r"LLVM version: (\d+)")
+            .unwrap()
+            .captures(&output_str)
+            .and_then(|cap| cap.get(1).map(|m| m.as_str()))
+            .expect("Failed to read LLVM version of rustc");
+
+        if !is_nightly {
+            println!("Coverage tool requries nightly version of rust toolchain");
+            println!("You can install it by using the following commands:");
+            println!("rustup default nightly");
```
iamyulong, line 56: Wonder if it's possible to use `nightly` for coverage only?

iamyulong (commented):
Really nice work!

I managed to run `scrypto new-package hello` and `scrypto coverage`, after following all the hints for dependencies. 

A coverage report was successfully generated, and the total coverage percentage seems to be right on `index.html`. However, the coverage report for the `hello/src/lib.rs` file is completely missed. It shows zero coverage.

simulator/src/scrypto/cmd_coverage.rs
bbarwik: Yes, it could use "cargo +nightly" when building package, but there may be some issues when someone is using multiple nightly versions

radix-engine/src/vm/wasm/prepare.rs
bbarwik: Yes. When transaction fails for example due to out of gas then dump_coverage cannot be called because it will fail due to  out of gas. When this is commented, the function dump_coverage is not using any build-in native functions so it's going to work even in case of transaction failure.

bbarwikowski-hacken:
> I managed to run `scrypto new-package hello` and `scrypto coverage`, after following all the hints for dependencies.
> 
> A coverage report was successfully generated, and the total coverage percentage seems to be right on `index.html`. However, the coverage report for the `hello/src/lib.rs` file is completely missed. It shows zero coverage.

@iamyulong This should not happened, in my case helo/src/lib.rs has correct coverage report. Did you update dependencies in cargo.toml to use version with coverage? If you did and it didn't work then please contact me to find out why it happened. 

image

simulator/src/scrypto/cmd_coverage.rs
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        // Verify that all llvm tools required to generate coverage report are installed
+        Self::check_command_availability(format!("clang-{}", llvm_major_version))?;
+        Self::check_command_availability(format!("llvm-cov-{}", llvm_major_version))?;
+        Self::check_command_availability(format!("llvm-profdata-{}", llvm_major_version))?;
+
+        // Build package
+        let path = self.path.clone().unwrap_or(current_dir().unwrap());
+        let (wasm_path, _) = build_package(&path, false, false, true, Level::Trace, true)
```
bbarwikowski-hacken, line 109: @iamyulong This approach, with building package before running tests and then reusing it makes tests way faster. Here's my recommendation regarding `scrypto test`:
1. Build all packages in advance, including dependencies (this may be hard when it comes to dependencies)
2. Store information what has been build in ENV, you can also store checksum and build options
3. In `test_runner.rs` try to use cached version, if it doesn't exist then build it and eventually cache it somehow (this may be hard because requires synchronization)

bbarwikowski-hacken:
> Hi @bbarwikowski-hacken, thanks for sharing the screenshot!
> 
> I actually got the same HTML report as yours. I think I was confused by it only showing the hit counts, rather than highlighting covered lines green and uncovered lines red. Any chance we can improve this a bit? Not a blocker and I think the PR is ready for merging. :)

The problem is that hello package has 100% code coverage so there's nothing to highlight 😆 
image

simulator/src/scrypto/cmd_coverage.rs
iamyulong: This is generally a good idea.  :+1: 

I believe the `cargo build` automatically caches previous builds and avoids re-compiling if not updates have been made, right? However, we can potentially speed up tests by caching package publishing processing and genesis bootstrap. 

This is definitely something that we need to measure and improve.

iamyulong (approved):

radix-engine/src/vm/wasm/prepare.rs
iamyulong: That makes sense. I was hoping it wasn't an error with the instrumentation itself, otherwise we have a problem. All is good.

simulator/src/scrypto/cmd_coverage.rs
bbarwikowski-hacken: Yeah it does, but it's slow and requires other tests to wait when using multiple workers/threads

PR #1645, opened 2023-11-10
Disallow PubKey as PrivKey for `cmd_set_default_account` command - display error if user accidentally pass wrong key.

Sajjon:
## Summary

Validate input to `cmd_set_default_account`, ensuring we don't accidentally pass in PublicKey instead of PrivateKey.

## Details

In `main` it is possible to accidentally pass in the PublicKey to the PrivateKey value for command `cmd_set_default_account`. This results in confusing state and error when trying to e.g. perform a transfer.

Screenshot 2023-11-10 at 15 55 19

N.B: we don't get any information on what went wrong, that is because of the program panics at [unwrap at line 62](https://github.com/radixdlt/radixdlt-scrypto/blob/main/simulator/src/resim/config.rs#L62), so we never reach the next line, with more info: `Error::NoDefaultPrivateKey`. This should also be fix (I can do another PR), but this PR fixes so that should never happen...

I got into this state since I accidentally copied the PublicKey instead of the PrivateKey, resulting in this bad state.
Screenshot 2023-11-10 at 15 55 33

## Testing
Screenshot 2023-11-10 at 16 03 31

Now it is not possible to pass in the PublicKey instead of PrivateKey anymore, and we display a very informative error `GotPublicKeyExpectedPrivateKey`.

Added a unit test.

### DIFF
--- a/simulator/src/resim/cmd_set_default_account.rs
+++ b/simulator/src/resim/cmd_set_default_account.rs
@@ -35,3 +35,31 @@ impl SetDefaultAccount {
         Ok(())
     }
 }
+
+#[cfg(test)]
+#[test]
+fn test_validation() {
+    let mut out = std::io::stdout();
+    let private_key = Secp256k1PrivateKey::from_hex(
+        "6847c11e2d602548dbf38789e0a1f4543c1e7719e4f591d4aa6e5684f5c13d9c",
+    )
+    .unwrap();
+    let public_key = private_key.public_key().to_string();
+
+    let make_cmd = |key_string: String| {
+        return SetDefaultAccount {
+            component_address: SimulatorComponentAddress::from_str(
+                "account_sim1c9yeaya6pehau0fn7vgavuggeev64gahsh05dauae2uu25njk224xz",
+            )
+            .unwrap(),
+            private_key: key_string,
+            owner_badge: SimulatorNonFungibleGlobalId::from_str(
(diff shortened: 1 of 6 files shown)


transaction/src/signing/secp256k1/private_key.rs
```diff
@@ -27,6 +27,16 @@ impl Secp256k1PrivateKey {
         self.0.secret_bytes().to_vec()
     }
 
+    pub fn to_hex(&self) -> String {
```
Sajjon, line 30: We convert between key and hex in a few places, makes sense for the type type to have these methods IMO.

simulator/src/resim/cmd_set_default_account.rs
```diff
@@ -19,12 +19,47 @@ pub struct SetDefaultAccount {
 impl SetDefaultAccount {
     pub fn run<O: std::io::Write>(&self, out: &mut O) -> Result<(), Error> {
         let mut configs = get_configs()?;
+        let private_key = parse_private_key_from_str(&self.private_key).map_err(|e| {
+            if Secp256k1PublicKey::from_str(&self.private_key).is_ok() {
+                Error::GotPublicKeyExpectedPrivateKey
```
Sajjon, line 24: If we fail to create a PrivateKey (32 bytes/64 hex char input) from the passed in string, we try to parse it as a PublicKey to see if the user made the very reasonable error of accidentally copy pasting the public key - like I did!

iamyulong (approved):

iamyulong (approved):

PR #1646, opened 2023-11-10
Make it possible to run only 'resim show' using default account's address

Sajjon:
## Summary
Change so that `resim show` is possible, without specifying an address, which will try to use the default account's address.

## Details
Make `address` optional for `resim show` command, defaulting to address of default account (set with `set-default-account`).

## Testing

Screenshot 2023-11-10 at 17 57 48

If `resim show` is called without a default account set, a new error `NoAddressProvidedAndNotDefaultAccountSet` will be thrown.

I've added unit tests and update README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,48 +16,48 @@ Binaries, including libraries, CLIs and docker images, are licensed under the [R
 ## Installation
 
 1. Install Rust - this requires Rust 1.70+ (if rust is already installed, upgrade with `rustup update`)
-    * Windows:
-        * Download and install [`rustup-init.exe`](https://win.rustup.rs/x86_64)
-        * Install "Desktop development with C++" with [Build Tools for Visual Studio 2019](https://visualstudio.microsoft.com/thank-you-downloading-visual-studio/?sku=BuildTools&rel=16)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -16,48 +16,48 @@ Binaries, including libraries, CLIs and docker images, are licensed under the [R
 ## Installation
 
 1. Install Rust - this requires Rust 1.70+ (if rust is already installed, upgrade with `rustup update`)
```
Sajjon, line 18: Ops I should revert this... cargo fmt ran when ai saves the file and it made this changes..

I did add row in table with info about being able to run just `resim show`

simulator/src/resim/cmd_show.rs
```diff
@@ -19,14 +20,45 @@ impl Show {
...
+                } else if let Ok(a) = SimulatorResourceAddress::from_str(address) {
+                    dump_resource_manager(a.0, &substate_db, out).map_err(Error::LedgerDumpError)
+                } else {
+                    Err(Error::InvalidId(address.clone()))
+                }
+            }
+            None => get_configs()
+                .and_then(|c| {
+                    c.default_account.ok_or(Error::LedgerDumpError(
+                        EntityDumpError::NoAddressProvidedAndNotDefaultAccountSet,
```
Sajjon, line 38: Try to use default accounts address, if not present use specific Error

README.md
iamyulong: It's all right. I can see no semantic change, rather than the addition of `resim show`

iamyulong (approved):

PR #1647, opened 2023-11-10
Fix Typos

Sajjon:
## Summary

Fix some typos using [crate "typos"](https://crates.io/crates/typos-cli) (can be configured and can be run part of PRs! Something we should consider for all our repos 🤩 )

When I ran it, it did give some false positives, I've discarded those. It also found typos in some types, which I guess we *cannot* change, without it being a breaking Scrypto change?

```diff
-        S: CommitableSubstateStore,
+        S: CommittableSubstateStore,
```

```diff
-pub type RoleAssingmentLockOwnerOutput = ();
+pub type RoleAssignmentLockOwnerOutput = ();
```


transaction-scenarios/src/scenarios/radiswap.rs
```diff
@@ -503,7 +503,7 @@ impl ScenarioCreator for RadiswapScenarioCreator {
                                     builder,
                                     pool_2,
                                     "Radiswap 2 - ETH/ETC: Pool",
-                                    "[EXAMPLE] The underyling pool between test tokens \"ETH\" and \"ETC\"",
+                                    "[EXAMPLE] The underlying pool between test tokens \"ETH\" and \"ETC\"",
```
iamyulong, line 506: I think we will have to revert this if not to break stokenet.

iamyulong (commented):
Other than the one commented, typo fixes look good.

PR #1648, opened 2023-11-13
Adding `SubstateDatabase::list_entries_from()` variant

jakrawcz-rdx:
This **non-breaking, internal-only change** PR adds a `list_entries_from(from_sort_key)` next to the existing `list_entries()` in `SubstateDatabase`.

### Why?
Node needs to be capable of starting substate listing from a specific key/prefix, in order to implement pagination in Browse API endpoints (like https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/3189211214/RNP-12+-+Browse+API+sub-API+within+Core+API#5.-Object-Collection-Listing).
Node could reach out directly to its RocksDB (and avoid changes to `SubstateDatabase` interface), but Node tries hard (successfully) to implement Browse API through the Engine's `SystemDatabaseReader` API, and that's why I prefer to propagate this change here.
The exact Node `TODO`: https://github.com/radixdlt/babylon-node/pull/741/files#diff-148a2d8426fe568c7f0cab8adda124b9958ea5f3bc23e41c490d566bc8ffe103R822

### DIFF
--- a/radix-engine-profiling/src/rocks_db_metrics/mod.rs
+++ b/radix-engine-profiling/src/rocks_db_metrics/mod.rs
@@ -102,11 +102,12 @@ impl<S: SubstateDatabase + CommittableSubstateDatabase> SubstateDatabase
         }
     }
 
-    fn list_entries(
+    fn list_entries_from(
         &self,
         partition_key: &DbPartitionKey,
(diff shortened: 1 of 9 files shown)


mstrug-rdx (approved):
For me looks ok.

PR #1653, opened 2023-11-22
Fix `resim` buckets parsing

lrubasze:
## Summary
Add support for missing types: `FungibleBucket` and `NonFungibleBucket`

## Testing
Respective tests added

### DIFF
--- a/simulator/src/scrypto_bindgen/translation.rs
+++ b/simulator/src/scrypto_bindgen/translation.rs
@@ -190,6 +190,7 @@ where
         {
             "Bucket".to_owned()
         }
+
         TypeKind::Custom(ScryptoCustomTypeKind::Own)
             if type_validation
                 == TypeValidation::Custom(ScryptoCustomTypeValidation::Own(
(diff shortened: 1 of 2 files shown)


simulator/src/scrypto_bindgen/translation.rs
```diff
@@ -223,6 +224,20 @@ where
             "GlobalAddressReservation".to_owned()
         }
         TypeKind::Custom(ScryptoCustomTypeKind::Own) => match type_validation {
+            TypeValidation::Custom(ScryptoCustomTypeValidation::Own(
+                OwnValidation::IsTypedObject(package_address, bp_name),
+            )) if package_address == Some(RESOURCE_PACKAGE)
+                && bp_name == FUNGIBLE_BUCKET_BLUEPRINT =>
+            {
+                "FungibleBucket".to_owned()
```
dhedey, line 232: Nice! :). How did you spot this out of interest?

We should probably also handle `FungibleProof`/`NonFungibleProof`/`FungibleVault`/`NonFungibleVault`/`FungibleResourceManager`/`NonFungibleResourceManager` too?

dhedey (approved):
Nice! Thanks

simulator/src/scrypto_bindgen/translation.rs
lrubasze: Thanks :)
I was checking where else `OwnValidation::IsBucket` enum variant is used and found this one.

I will add those missing ones in separate PR if that's ok.

PR #1656, opened 2023-11-27
Radix-engine-tests reorganisation

mstrug-rdx:
## Summary
`radix-engine-tests` project reorganisation.

## Details
`tests` directory has been divided into subfolders basing on the execution layers: application, blueprints, system, vm, kernel, db. Files with tests has been moved to new subfolders basing on their content - but here still may be needed some adjustments.

Changing flat file structure to subdirectory based improved compilation time from ~5min to less than 2min due to greatly reduced executables count.

Added library target with `common.rs` file which contains dedicated module with macros for handling file paths to the assets (workspace blueprints and wasm/rpd files and local blueprint and wat files), and also moved to this file `package_loader` module (as it is commonly used by lots of tests). This library is used in integration tests and benches targets (everywhere where is an access to assets file paths), so changing asset paths in future will require change in only one place (`common.rs`). 

Added [`README.md`](https://github.com/radixdlt/radixdlt-scrypto/blob/test/radix-engine-test-reorganization/radix-engine-tests/README.md) file with some descriptions on how to run the tests.


radix-engine-tests/Cargo.toml
```diff
@@ -37,8 +37,9 @@ walkdir = { version = "2.3.3" }
 paste = { version = "1.0.13" }
 lazy_static = { version = "1.4.0" }
 hex = { version = "0.4.3", default-features = false }
-tuple-return = { path = "./tests/blueprints/tuple-return", features = ["test"], default-features = false }
+tuple-return = { path = "./assets/blueprints/tuple-return", features = ["test"], default-features = false }
```
iamyulong, line 40: It's weird that we have a dependency on the test blueprint. Possible to remove it?

iamyulong (approved):
Overall looks good!

radix-engine-tests/Cargo.toml
mstrug-rdx: This dependency is needed for one test which calls directly method from the `tuple-return` blueprint. It would be possible to move that test to separate project or even add it as a test inside this specific blueprint, but this will require manually to run that one test (or use additional script) and will need CI script update also. So I am not sure if the juice is worth the squeeze however I have working change (with moving test to blueprint). @iamyulong what do you think?

mstrug-rdx: Moved `tuple-return` test to `scrypto-test` project.

PR #1661, opened 2023-12-04
Add BLS signatures

lrubasze:
## Summary
This PR adds support for:
- BLS12-381  signatures
- Keccak256 hashing

It also implements `CryptoUtils` module for Client API, which allows to do some crypto-related stuff natively from Scrypto.
`CryptoUtils` functions:
- `keccak_hash()` - generate Keccak-256 hash of the given message 
- `bls_verify()` - do the given BLS signature verification of the given message hash using given public key

## Details
BLS details:
- Algorithm: BLS12-381
- Public key size: 48 bytes (G1)
- Signature size 96 bytes (G2)

In order to get it working I had to move cryptographic primitives (PrivateKey, Signature definitions) from `transaction` to `radix-engine-common`. What is left in `transaction` are the Radix Engine cryptographics, eg. `IntentSignatureV1` , `NotarySignatureV1`.

Cryptographic dependencies in `radix-engine-common` are not built for WASM.
Reasons:
-  ATM they are not required in WASM in Scrypto. 
- They (eg. `blst`) or their deps (eg. `secp256k1-sys`) are not implemented in Rust. And `clang` used in macOS GH runner does not support WASM target. For now we skip it (maybe eventually we would need the WASM support there anyway)

TODO: Costing

## Testing
- Test `CryptoUtils` module using `CryptoScrypto` test blueprint


radix-engine-common/src/crypto/hash.rs
```diff
@@ -75,7 +76,7 @@ pub fn hash<T: AsRef<[u8]>>(data: T) -> Hash {
 //========
 
 /// Represents an error when parsing hash.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor)]
```
talekhinezh, line 79: Can we change this to `Sbor` rather than `ScryptoSbor` so hash is not dependent on scrypto objects.

radix-engine-common/src/data/scrypto/custom_well_known_types.rs
```diff
@@ -379,10 +379,15 @@ create_well_known_lookup!(
             MISC_TYPES_START + 7,
```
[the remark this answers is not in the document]
lrubasze: AFAIK, well-known types change is not necessary, so theoretically we could skip it. But since we require the protocol update anyway I would leave it as is.

radix-engine/src/system/system.rs
```diff
@@ -2849,6 +2849,33 @@ where
...
+        // TODO: apply execution costs
+        Ok(verify_bls(&message, &public_key, &signature) as u32)
+    }
+
+    #[trace_resources]
+    fn keccak_hash(&mut self, data: Vec<u8>) -> Result<Hash, RuntimeError> {
+        // TODO: apply execution costs
+        Ok(keccak_256_hash(&data))
+    }
+}
```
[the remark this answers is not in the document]
lrubasze: Ok

iamyulong (approved):
Ready to go. The missing part is to handle protocol updates, which should follow.

radix-engine-common/src/crypto/hash.rs
lrubasze: Done

radix-engine-common/src/crypto/ed25519/mod.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg(not(target_arch = "wasm32"))]
```
0xOmarA, line 1: For the purposes of the toolkit, this should be a feature flag and not dependent on the architecture. This is because the toolkit is compiled to WASM and requires all of these cryptographic types.

This also applies to the other curves too.

radix-engine-common/src/crypto/secp256k1/private_key.rs
```diff
@@ -1,20 +1,24 @@
 use crate::internal_prelude::*;
-use ::secp256k1::{Message, PublicKey, SecretKey, SECP256K1};
+use ::secp256k1::{All, Message, PublicKey, Secp256k1, SecretKey};
 
 use super::Secp256k1Signature;
 
+lazy_static::lazy_static! {
+    pub static ref SECP256K1_CTX: Secp256k1<All> = secp256k1::Secp256k1::new();
```
0xOmarA, line 7: Probably would want this to be a `pub(crate)`?

Also, is there any advantage to using this over the provided global context in the `secp256k1` crate?

radix-engine-common/src/crypto/mod.rs
```diff
@@ -1,15 +1,24 @@
 mod blake2b;
+mod bls12381;
+mod ed25519;
 mod hash;
 mod hash_accumulator;
+#[cfg(not(target_arch = "wasm32"))]
```
0xOmarA, line 6: Same comment on feature flagging vs architecture.

radix-engine-common/Cargo.toml
```diff
@@ -25,6 +25,13 @@ arbitrary = { version = "1.3.0", features = ["derive"], optional = true }
 rug = { version = "1.18", optional = true }
 ethnum = {version = "1.3.2", default-features = false, optional = true }
 
+# Do not include crypto-related dependencies when building for WASM.
+[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
```
0xOmarA, line 29: ditto on feature flagging.

0xOmarA (changes requested):
The main change that I believe is necessary is removing any dependence on architecture from the source code and use feature flags instead. We're bound to run into issues with the toolkit is the cryptographic types are not available to us in WASM anymore.

radix-engine-common/src/crypto/ed25519/mod.rs
lrubasze: First of all thanks for valuable feedback!
When introducing the BLS I paid special attention to not narrow down the scope and functionality of the RE.
Before this change the cryptographic features were located in `transaction` crate, which was not built for scrypto (`wasm32` target). 
As the matter of fact ATM it is not possible to build `secp2561k1` for WASM (some deps) - this is the main reason I've decided to not include cryprographic features for `wasm32` and this is why I decided to use target architecture switches. 

Regarding RE cryptographic types - these are being compiled.

So these switches are not excluding any functionality that was enabled before the change.
Maybe I will just put some comment in the code to remind us to remove them once `wasm32` target finally works?

radix-engine-common/src/crypto/secp256k1/private_key.rs
lrubasze: > Also, is there any advantage to using this over the provided global context in the `secp256k1` crate?
global context required `global-context` feature for `secp2561` which enforces 'std'

radix-engine-common/src/crypto/mod.rs
lrubasze: Replied above

radix-engine-common/Cargo.toml
lrubasze: Replied above

radix-engine-common/src/crypto/ed25519/mod.rs
lrubasze: After a discussion with @0xOmarA agreed that it cryptographic primitives will be enabled under some feature gate. Radix Engine Toolkit is needing them for all supported targets (for `wasm32` target too - apparently it is possible to compile `secp2561k1` for `wasm32`).
Above change will be introduced in separate PR.

lrubasze: After second discussion we agreed that no need to feature gate cryptographic primitives. We just need to make sure all compiles for `wasm32` target. If yes then it will be enabled by default.

0xOmarA (approved):

radix-engine/src/system/system.rs
```diff
@@ -2849,6 +2849,33 @@ where
     }
 }
 
+#[cfg_attr(
+    feature = "std",
+    catch_unwind(crate::utils::catch_unwind_system_panic_transformer)
+)]
+impl<'a, Y, V> ClientCryptoUtilsApi<RuntimeError> for SystemService<'a, Y, V>
```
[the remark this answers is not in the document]
lrubasze: Since it requires some additional work it was agreed to proceed with this in separate PR.

0xOmarA (approved):

radix-engine-common/src/data/scrypto/custom_well_known_types.rs
```diff
@@ -379,10 +379,15 @@ create_well_known_lookup!(
...
             named_transparent("Origin", string_type_data(),)
         ),
-        // Public key-related types from KEY_TYPES_START
+        (
+            HASH,
+            MISC_TYPES_START + 8,
+            named_transparent("Hash", bytes_fixed_length_type_data(Hash::LENGTH),)
+        ),
```
talekhinezh, line 386: Can we remove this? As this is still a protocol change

PR #1664, opened 2023-12-18
Add costing to `CryptoUtils` API

lrubasze:
## Summary
Costing is based on instruction count measured by [QEMU costing job](https://github.com/radixdlt/radixdlt-scrypto/actions/runs/7410771396)

## Details

For methods that rely on one input [this regression calculator](https://www.socscistatistics.com/tests/regression/default.aspx) was used
Methods:
- `bls12381_v1_verify_cost()` - costing depends on message size
- `bls12381_g2_signature_aggregate_cost()` - depends on signatures count
- `keccak256_hash_cost()` - depends on message size

For methods that rely on two input [this multiple regression calculator](https://www.socscistatistics.com/tests/multipleregression/default.aspx) was used
Methods:
- `bls12381_v1_aggregate_verify_cost()` - costing depends on messages sizes (used average message size for simplicity) and keys count
- `bls12381_v1_fast_aggregate_verify_cost()` - costing depends on message size and keys count

Used following [spreadsheet](https://docs.google.com/spreadsheets/d/1rV0KyB7UQrg2tOenbh2MQ1fo9MXwrwPFW4l0_CVO-6o/edit?usp=sharing) to visualize the results.

## Testing
Added costing tests to easily reproduce the costing process procedure if needed.

### DIFF
--- a/radix-engine/src/system/system.rs
+++ b/radix-engine/src/system/system.rs
@@ -2881,13 +2881,17 @@ where
         public_keys: &[Bls12381G1PublicKey],
         signature: &Bls12381G2Signature,
     ) -> Result<u32, RuntimeError> {
-        self.api.kernel_get_system().modules.apply_execution_cost(
-            ExecutionCostingEntry::Bls12381V1AggregateVerify {
-                size: messages.iter().map(|m| m.len()).sum::<usize>() / messages.len(),
-                keys_cnt: public_keys.len(),
-            },
-        )?;
-        Ok(aggregate_verify_bls12381_v1(messages, public_keys, signature) as u32)
+        if !messages.is_empty() && !public_keys.is_empty() {
+            self.api.kernel_get_system().modules.apply_execution_cost(
+                ExecutionCostingEntry::Bls12381V1AggregateVerify {
+                    size: messages.iter().map(|m| m.len()).sum::<usize>() / messages.len(),
+                    keys_cnt: public_keys.len(),
+                },
+            )?;
+            Ok(aggregate_verify_bls12381_v1(messages, public_keys, signature) as u32)
+        } else {
+            Err(RuntimeError::SystemError(SystemError::InputDataEmpty))
+        }
     }
 
     #[trace_resources(log=message.len(), log=public_keys.len())]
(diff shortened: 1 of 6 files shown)


radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -381,6 +381,40 @@ impl FeeTable {
         500 + Self::data_processing_cost(size)
     }
 
+    #[inline]
+    pub fn bls12381_v1_verify_cost(&self, size: usize) -> u32 {
+        // Based on benchmark `bench_bls_verify`
```
lrubasze, line 386: Not sure if it is fine to rely on benchmark only.

lrubasze: Now relying on instructions count - measured in this [QEMU costing job](https://github.com/radixdlt/radixdlt-scrypto/actions/runs/7410771396)

radix-engine/src/system/system.rs
```diff
@@ -2877,8 +2881,17 @@ where
         public_keys: &[Bls12381G1PublicKey],
         signature: &Bls12381G2Signature,
     ) -> Result<u32, RuntimeError> {
-        // TODO costing
-        Ok(aggregate_verify_bls12381_v1(messages, public_keys, signature) as u32)
+        if !messages.is_empty() && !public_keys.is_empty() {
+            self.api.kernel_get_system().modules.apply_execution_cost(
+                ExecutionCostingEntry::Bls12381V1AggregateVerify {
+                    size: messages.iter().map(|m| m.len()).sum::<usize>() / messages.len(),
```
mstrug-rdx, line 2887: Alternatively it could be: `messages.iter().flat_map(|x| *x).count()`.

0xOmarA, line 2887: Instead of using the average length of messages, would it be better to cost for each message and public key pair individually? 

This comes from a fear that somebody might have something like 9999x 1 byte long messages and 1x 1,000,000 byte long message which averages down to 100 bytes.

lrubasze: ok, will change it

lrubasze: Of course it would be better. But not sure if it adds up this way in case of aggregated verify. Let me check.

lrubasze: Done

lrubasze: Took a deeper look into this and discovered that used approach is not accurate enough in many cases, which could result in costing for less instructions count.
Additionally, if the messages sizes differ significantly the dispersion is even greater.
So, assuming that aggregated verify is a nice-to-have feature, I've decided to do not take aggregation into account and just sum up the costs per all message sizes. This returns higher instructions count. This way we will never charge for less instructions count.

This is a temporary approach until no better solution is found.

0xOmarA (approved):

PR #1668, opened 2024-01-02
Architecture: Remove Hash Tree mut Requirement

talekhinezh:
Remove mut requirement on hash tree methods. Replaced with interior mutability implementation.

### DIFF
--- a/radix-engine-stores/src/hash_tree/test.rs
+++ b/radix-engine-stores/src/hash_tree/test.rs
@@ -260,6 +260,7 @@ fn records_stale_tree_node_keys() {
     let stale_versions = tester
         .tree_store
         .stale_part_buffer
+        .borrow()
         .iter()
         .map(|stale_part| {
             let StaleTreePart::Node(key) = stale_part else {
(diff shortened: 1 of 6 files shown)


radix-engine-stores/src/hash_tree/test.rs
```diff
@@ -403,13 +404,13 @@ fn serialized_keys_are_strictly_increasing() {
...
         .memory
+        .borrow()
         .keys()
         .cloned()
         .collect::<HashSet<_>>();
     tester.put_substate_changes(vec![change(1, 7, 2, Some(80))]);
-    let min_next_key = tester
-        .tree_store
-        .memory
+    let binding = tester.tree_store.memory.borrow();
```
jakrawcz-rdx, line 412: (nitpick) I know these are just tests, but how about something more specific than `binding` :D ?

(alternatively - some helper method on a mem tree store?)

jakrawcz-rdx (approved):
totally good idea 👍

radix-engine-stores/src/hash_tree/test.rs
talekhinezh: good point!

PR #1671, opened 2024-01-06
feature: Add type_name override to Describe proc macro

dhedey:
## Summary
Adds functionality for type name override via `#[sbor(type_name = "NewTypeName")]`.

This is an addition, not a breaking change.

This would allow e.g. us to rename types without affecting their scrypto schema. For example, we could create e.g. `TimePrecisionV2` and rename `TimePrecision` to `TimePrecisionV1`, whilst allowing them both to have an SBOR type name of `TimePrecision`. The latter is required for backwards compatibility.

## Testing
Tests are in `sbor-tests`.

### DIFF
--- a/sbor-derive-common/src/describe.rs
+++ b/sbor-derive-common/src/describe.rs
@@ -317,14 +317,14 @@ mod tests {
             quote! {
                 impl <C: ::sbor::CustomTypeKind<::sbor::RustTypeId> > ::sbor::Describe<C> for Test {
                     const TYPE_ID: ::sbor::RustTypeId = ::sbor::RustTypeId::novel_with_code(
-                        stringify!(Test),
+                        "Test",
                         &[],
                         &#code_hash
                     );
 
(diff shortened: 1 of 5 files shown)


talekhinezh (approved):

sbor-derive-common/src/describe.rs
```diff
@@ -126,6 +128,27 @@ fn handle_normal_describe(
     let (impl_generics, ty_generics, where_clause, child_types, custom_type_kind_generic) =
         build_describe_generics(&generics, &attrs, context_custom_type_kind)?;
 
+    let type_name =
+        get_sbor_attribute_string_value(&attrs, "type_name")?.unwrap_or(ident.to_string());
+
+    let type_id = quote! {
+        ::sbor::RustTypeId::novel_with_code(
```
dhedey, line 135: Note for the future - we can consider just using `core::any::TypeId::of::<Self>` here, if we add where `Self: 'static` to the Describe trait (which should be fine) https://github.com/rust-lang/rust/pull/72488

PR #1675, opened 2024-01-12
Feature: Add protocol update for crypto utils

talekhinezh:


0xOmarA (approved):
Looks good, it might be worth checking with gateway team if they need the booloader to have a `TypeInfo` substate, or at least to let them know that it does not, in case they rely on all nodes having a type info substate.

0xOmarA (approved):
Looks good, it might be worth checking with gateway team if they need the booloader to have a `TypeInfo` substate, or at least to let them know that it does not, in case they rely on all nodes having a type info substate.

radix-engine/src/vm/wasm/wasm_validator.rs
```diff
@@ -60,7 +60,7 @@ impl ScryptoV1WasmValidator {
     ) -> Result<(Vec<u8>, Vec<String>), PrepareError> {
         WasmModule::init(code)?
             .enforce_no_start_function()?
-            .enforce_import_limit()?
+            .enforce_import_limit(self.minor_version)?
```
lrubasze, line 63: minor thing, but the name is no longer adequate :)

lrubasze (approved):

PR #1677, opened 2024-01-15
Feature: Use txn tracker for boot

talekhinezh:
Replace use of `BOOT_LOADER_STATE` with `TRANSACTION_TRACKER`


radix-engine/src/system/checkers/kernel_db_checker.rs
```diff
@@ -14,7 +14,6 @@ pub enum KernelDatabaseCheckError {
     NoOwnerForNonGlobalNode(NodeId),
     ZeroPartitionCount(NodeId),
     CannotOwnBootLoaderNode(NodeId),
```
iamyulong, line 16: Should remove `CannotOwnBootLoaderNode` as well?

talekhinezh: Yup good catch

iamyulong (approved):

PR #1680, opened 2024-01-16
Add generate_validator_fee_fix_state_updates

talekhinezh:
Add state updates for the validator fee fix


0xOmarA (approved):

radix-engine/src/utils/flash_state_generator.rs
```diff
@@ -224,3 +231,37 @@ pub fn generate_seconds_precision_state_updates<S: SubstateDatabase>(db: &S) ->
         ),
     }
 }
+
+pub fn generate_validator_fee_fix_state_updates<S: SubstateDatabase>(db: &S) -> StateUpdates {
```
LukasGasior1, line 235: This is a specific change for a specific mainnet protocol update, doesn't seem quite right to keep it in `utils`. Looks like we'll need some notion of the actual protocol updates here

PR #1684, opened 2024-01-17
Enable BLS aggregate verify

lrubasze:
## Summary
Enable BLS aggregate verify for Radix Engine and Scrypto.

## Details
For `wasm32` and `no_std` added our local implementation of `aggregate_verify_bls12381_v1_no_threads()` which does not rely on threads.
It follows [CoreAggregateVerify](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#name-coreaggregateverify) description.
And is inspired with [`aggregate_verify()` implementation](https://github.com/supranational/blst/blob/master/bindings/rust/src/lib.rs#L1095) from `blst` crate.

## Testing
Added a few positive and negative tests.

### DIFF
--- a/radix-engine-common/src/crypto/signature_validator.rs
+++ b/radix-engine-common/src/crypto/signature_validator.rs
@@ -75,36 +75,92 @@ pub fn verify_bls12381_v1(
     false
 }
 
+#[cfg(any(target_arch = "wasm32", feature = "alloc"))]
+/// Local implementation of aggregated verify for no_std and WASM32 variants (no threads)
+/// see: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#name-coreaggregateverify
+/// Inspired with blst::min_pk::Signature::aggregate_verify
+fn aggregate_verify_bls12381_v1_no_threads(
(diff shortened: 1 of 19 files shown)


radix-engine-common/src/crypto/signature_validator.rs
```diff
@@ -75,36 +75,92 @@ pub fn verify_bls12381_v1(
     false
 }
 
+#[cfg(any(target_arch = "wasm32", feature = "alloc"))]
+/// Local implementation of aggregated verify for no_std and WASM32 variants (no threads)
+/// see: https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#name-coreaggregateverify
+/// Inspired with blst::min_pk::Signature::aggregate_verify
+fn aggregate_verify_bls12381_v1_no_threads(
```
lrubasze, line 82: Alternatively I could fork `blst` crate and modify it accordingly.
But I decided not to fork for easier maintenance in the future (eg. bumping versions).

iamyulong (approved):
LGTM

PR #1695, opened 2024-01-20
Feature: Scrypto Test Builder

0xOmarA:
## Summary

This PR includes a number of changes and improvements to the scrypto-test framework. The following are the changes made:

* Created a new `TestEnvironmentBuilder` that allows for configuring the environment prior to creation. Additionally the `TestEnvironment` can now operate over any database and not just an in-memory database.
* Added a better way for reading state in the testing environment. This new method ensures that the states are opened and closed when they need to be. It also ensures that the references provided as a result of the state reads do not escape outside of the closure.
* Added logging support to the testing environment. It is currently enabled by default and quite rudimentary.

### DIFF
--- a/scrypto-test/src/environment/builder.rs
+++ b/scrypto-test/src/environment/builder.rs
@@ -125,19 +125,20 @@ where
                     .flat_map(|NodeDatabaseUpdates { partition_updates }| {
                         partition_updates.values()
                     })
-                    .filter_map(|item| {
-                        if let PartitionDatabaseUpdates::Delta { substate_updates } = item {
-                            Some(substate_updates.values())
-                        } else {
-                            None
-                        }
-                    })
-                    .flatten()
-                    .filter_map(|item| {
-                        if let DatabaseUpdate::Set(item) = item {
-                            Some(item)
-                        } else {
-                            None
+                    .flat_map(|item| -> Box<dyn Iterator<Item = &Vec<u8>>> {
+                        match item {
+                            PartitionDatabaseUpdates::Delta { substate_updates } => {
+                                Box::new(substate_updates.values().filter_map(|item| {
+                                    if let DatabaseUpdate::Set(value) = item {
+                                        Some(value)
+                                    } else {
+                                        None
+                                    }
+                                }))
+                            }
+                            PartitionDatabaseUpdates::Reset {
+                                new_substate_values,
+                            } => Box::new(new_substate_values.values()),
                         }
                     })
                     .flat_map(|value| {
--- a/scrypto-test/src/environment/builder.rs
+++ b/scrypto-test/src/environment/builder.rs
@@ -65,11 +65,7 @@ impl TestEnvironmentBuilder<InMemorySubstateDatabase> {
             flash_database: FlashSubstateDatabase::standard(),
             added_global_references: Default::default(),
             bootstrap: true,
-            protocol_updates: ProtocolUpdateOptIns {
-                consensus_manager_seconds_precision: true,
-                crypto_utils: true,
-                updated_pools: true,
-            },
+            protocol_updates: ProtocolUpdateOptIns::default(),
         }
     }
 }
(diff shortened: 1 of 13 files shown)


scrypto-test/src/environment/env.rs
```diff
@@ -719,31 +641,47 @@ impl TestEnvironment {
     ///
     /// This method panics if the component state can not be decoded as the generic type parameter
     /// [`S`].
-    pub fn read_component_state<S, N>(&mut self, node_id: N) -> Result<S, RuntimeError>
+    pub fn with_component_state<S, N, F, O>(
```
iamyulong, line 644: Looks like we need to update the rustdoc a bit.

scrypto-test/src/environment/internal.rs
```diff
@@ -4,130 +4,66 @@
 use super::*;
 use crate::prelude::*;
 
+// TODO: I would like to remove the reliance on `CommittableSubstateDatabase` and to instead commit
+//       everything to the track. As in, nothing ever gets committed to the database. Even the
+//       initial bootstrapping should be done in this way. This mainly comes from a desire to use
+//       the node's database with scrypto-test, and it does not implement that trait.
```
iamyulong, line 10: This is actually a highly desired feature. From my interaction with Hacken, they wanted to run test transactions on the Node's database.

scrypto-test/src/environment/client_api.rs
```diff
@@ -44,16 +44,56 @@ macro_rules! implement_client_api {
...
+                                .map(|runtime| runtime.logs.len())
+                                .unwrap_or(0);
+
+                            if logs_before != logs_after {
+                                for (level, message) in kernel
+                                    .kernel_get_system_state()
+                                    .system
+                                    .modules
+                                    .transaction_runtime()
+                                    .map(|module| module.logs.clone())
```
iamyulong, line 86: Minor: might be able to avoid cloning all the logs?

scrypto-test/src/environment/builder.rs
```diff
@@ -0,0 +1,532 @@
...
+                    .node_updates
+                    .values()
+                    .flat_map(|NodeDatabaseUpdates { partition_updates }| {
+                        partition_updates.values()
+                    })
+                    .filter_map(|item| {
+                        if let PartitionDatabaseUpdates::Delta { substate_updates } = item {
+                            Some(substate_updates.values())
+                        } else {
+                            None
```
iamyulong, line 132: Should handle `PartitionDatabaseUpdates::Reset` too? Alternatively, we don't use `PartitionDatabaseUpdates` enum since no partition deletion seems to be supported.

scrypto-test/src/environment/builder.rs
```diff
@@ -0,0 +1,532 @@
...
+                &mut self.database,
+                vm,
+                false,
+            );
+            bootstrapper.bootstrap_test_default().unwrap();
+        }
+        let database_updates = self.flash_database.database_updates();
+        if !database_updates.node_updates.is_empty() {
+            self.database.commit(&database_updates);
+        }
```
iamyulong, line 233: We should do custom flashes last?

scrypto-test/src/environment/env.rs
0xOmarA: Yup, updated.

scrypto-test/src/environment/client_api.rs
0xOmarA: Yup, I've removed this clone

scrypto-test/src/environment/builder.rs
0xOmarA: Yup, I've moved that.

scrypto-test/src/environment/builder.rs
0xOmarA: Good point on handling reset too, I missed that. I've handled it and updated the PR.

iamyulong (approved):

PR #1702, opened 2024-01-30
fix: Fix manifest builder complex deposit batch bug

dhedey:
## Summary

Small bug fix to enable the test case which I've added. Thanks to Michal for the report.


mstrug-rdx (approved):
Looks ok.

PR #1712, opened 2024-02-07
Some improvements and tests related to Scrypto versioning 

lrubasze:
## Summary
- small improvements related to Scrypto minor versioning:
  - Constants collected in one place
  - Generate version state updates by given version
- Add `ProtocolVersionMismatch` WASM error for more clarity 
- Get Vm version from boot store for `TestEnvironment`
- `crypto_utils` flash tests with `TestEnvironment`


radix-engine/src/vm/versions.rs
```diff
@@ -0,0 +1,2 @@
+pub const SCRYPTO_V1_LATEST_MINOR_VERSION: u64 = 1u64;
+pub const SCRYPTO_VM_CRYPTO_UTILS_MINOR_VERSION: u64 = 1u64;
```
iamyulong, line 2: How about making this an enum?

```rust
#[repr(u8)]
pub enum ScryptoVMVersion {
    V1_0,
    V1_1,
}

impl ScryptoVMVersion {
    pub fn latest() -> ScryptoVMVersion {
        ScryptoVMVersion::V1_1
    }
}
```

lrubasze: Sounds good :)

lrubasze: Done

simulator/src/resim/cmd_publish.rs
```diff
@@ -140,7 +141,7 @@ impl Publish {
             let mut vm_type_updates = index_map_new();
             let mut original_code_updates = index_map_new();
             let mut instrumented_code_updates = index_map_new();
-            let instrumented_code = ScryptoV1WasmValidator::new(0u64)
+            let instrumented_code = ScryptoV1WasmValidator::new(ScryptoVmVersion::V1_0)
```
lrubasze, line 144: Not sure here. Should we use `V1_0` or latest?

iamyulong (approved):

simulator/src/resim/cmd_publish.rs
lrubasze: Ok, found scenario where it fails for `V1_0`.
- Prepare a new package that includes features added in `V1_1` or later.
- Attempt to update already published package with the new one

Will change to `ScryptoVmVersion::latest()`

PR #1717, opened 2024-02-12
Various Dependency Clean-ups

iamyulong:
## Summary

- Crate `radix-engine`
    * Removed `types` module, which is not supposed to be consumed externally
    * Removed `prelude` module, which re-exports common modules and has been causing unexpected dependencies

- Crate `radix-engine-interface`
    * Moved `api::node_modules` up to align with `blueprints`
    * Merged `api::system_modules::*` to its parent, for a flat folder of System APIs
    * Moved `NonFungibleGlobalId`, `ScryptoEvent`, `BlueprintId` to `radix-engine-common`
    * Moved `constants::*` to `radix-engine-common`
    * Removed all the common module re-exports, which has been causing unexpected dependencies

- Crate `transaction`
    * Removed all the common module re-exports, which has been causing unexpected dependencies

### For Internal Integrators

It's highly likely that your project will fail to compile due to missing symbols. Most of them can be found in `radix-engine-common`.

Again, there is no logic change, so everything should be fine after fixing the symbols.


monkey-tests/src/lib.rs
```diff
@@ -5,6 +5,8 @@ pub mod resource;
 pub mod two_pool;
 pub mod validator;
 
+use std::ops::AddAssign;
```
lrubasze, line 8: Probably not relevant. 
But this test might not work for `alloc`

radix-engine-tests/tests/blueprints/proof_creation.rs
```diff
@@ -1,6 +1,6 @@
 use radix_engine::{
     errors::{ApplicationError, RuntimeError},
-    types::*,
+
```
lrubasze: Looks like our format check does not detect everything.

radix-engine-tests/tests/system/bucket.rs
```diff
@@ -7,8 +7,9 @@ use radix_engine::{
     blueprints::resource::BucketError,
     errors::{ApplicationError, CallFrameError, KernelError, RuntimeError},
     kernel::call_frame::DropNodeError,
-    types::*,
+
```
lrubasze: ditto

simulator/src/rtmc/mod.rs
```diff
@@ -1,5 +1,6 @@
 use clap::Parser;
-use radix_engine::{types::*, utils::*};
+use radix_engine::{  utils::*};
```
lrubasze, line 2: Formatting again (btw I thought there were some issues with formatting within `radix-engine-tests` only)

lrubasze:
Nice work!

PR #1725, opened 2024-02-27
Committable Substate Overlay

0xOmarA:
## Summary

This PR adds the `SingleSubstateDatabaseOverlay` which is a wrapper that can wrap any `SubstateDatabase` and can be thought of as a "multi-transaction track". It allows substate changes to be committed to overlay such that no data is ever committed to the underlying database and the underlying database can never become corrupted. Whenever a substate is read it is first read from the overlay and then read from the underlying database if it is not found in the overlay.

This use case for this is simple: I would like to run multi-transaction tests against mainnet state and I want the following:

* I do not want any of my changes to be written to the actual node's mainnet database and corrupt it.
* I want the changes that my transaction makes to be persisted for the period of the test and then be discarded at the end.


substate-store-impls/src/substate_database_staging.rs
```diff
@@ -0,0 +1,282 @@
...
+                        partition_updates
+                            .keys()
+                            .map(|partition_num| DbPartitionKey {
+                                node_key: node_key.clone(),
+                                partition_num: *partition_num,
+                            })
+                    },
+                ))
+                .unique()
+                .sorted(),
```
dhedey, line 181: We can't use `unique()` and `sorted()` here - this sticks all keys into memory at once from the underlay store.

Instead, we should use an overlayed iterator, like inside of `list_entries_from`

radix-engine-common/src/iterators/overlaying_iterator.rs
```diff
@@ -0,0 +1,88 @@
+use crate::internal_prelude::*;
+use sbor::rust::cmp::*;
+use sbor::rust::iter::*;
+
+/// An iterator overlaying a "change on a value" (coming from the [`overlaying`] iterator) over a
+/// "base value" (coming from the [`underlying`] iterator).
+/// The one is matched to another by a `K` part (of the iterated tuple `(K, V)`), which both
+/// iterators are assumed to be ordered by.
+pub struct OverlayingIterator<U, O>
```
dhedey, line 9: What do you think of moving this to `utils`, as per Yulong's comment?

substate-store-impls/src/substate_database_staging.rs
```diff
@@ -0,0 +1,282 @@
+use itertools::Itertools;
+use radix_engine_common::prelude::*;
+use substate_store_interface::interface::*;
+
+pub struct SubstateDatabaseStaging<'s, S> {
```
dhedey, line 5: What do you think about naming this more specifically a single layer overlay?

substate-store-impls/src/substate_database_staging.rs
```diff
@@ -0,0 +1,282 @@
...
+                }
+            }
+            // Overlay doesn't contain anything for the provided node key. Return an iterator over
+            // the data in the root store.
+            None => self.root.list_entries_from(partition_key, from_sort_key),
+        }
+    }
+}
+
+impl<'s, S> CommittableSubstateDatabase for SubstateDatabaseStaging<'s, S> {
```
dhedey, line 156: What do you think about also implementing some kind of "commit into underlay" functionality (if S is committable?)

dhedey (commented):
Looking a lot better. Still a few minor tweaks to go, so won't approve just yet

substate-store-impls/src/substate_database_staging.rs
```diff
@@ -0,0 +1,282 @@
...
+                            .iter()
+                            .map(|(sort_key, database_update)| match database_update {
+                                DatabaseUpdate::Set(substate_value) => {
+                                    (sort_key.clone(), Some(substate_value.clone()))
+                                }
+                                DatabaseUpdate::Delete => (sort_key.clone(), None),
+                            })
+                            .sorted_by(|(sort_key_a, _), (sort_key_b, _)| {
+                                sort_key_a.cmp(sort_key_b)
+                            });
```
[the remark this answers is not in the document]
0xOmarA: > Functional issue: I believe the overlay is missing a filter on from_sort_key here.

Good point on that. I've updated that and added extra tests for this particular case to verify it.

> Possible performance issue: It would be more efficient if we used a structure which saved the overlaid substates in a BTreeMap rather than an IndexMap, as this requires sorting the map constantly.

Yup, agreed that `BTreeMap`s would be better here so that we're not constantly sorting when we do a `list_entries_from`. I've made that switch by introducing a new set of types that all use `BTreeMap` instead of `IndexMap`: `StagingDatabaseUpdates`, `StagingNodeDatabaseUpdates`, and `StagingPartitionDatabaseUpdates`. These types implement `From` and `Into` their original types to ensure that they're always in sync with them and so that we never expose any of these staging types outside this module.

substate-store-impls/src/substate_database_staging.rs
0xOmarA: Sure, I've got a few in mind, maybe:

* `SingleLayerSubstateDatabaseOverlay`
* `SingleLayerDatabaseOverlay`
* `SingleLayerDatabaseOverlay`
* `SingleOverlayStorage`

Perhaps overly verbose, but my personal pick is `SingleLayerSubstateDatabaseOverlay` :D Any preference on which one to choose?

substate-store-impls/src/substate_database_staging.rs
0xOmarA: Sounds good, I've switched the impl to use `OverlayingIterator`.

substate-store-impls/src/substate_database_staging.rs
0xOmarA: I would slightly prefer to keep the database immutable since it makes it clear in the type system that the root database is never mutated in any way and that all that this type needs is read-only access to the database. If committing to the root database is needed, perhaps it can be done outside of the overlay. As an example, something like:

```rust
let mut root = InMemorySubstateDatabase::standard();
let overlay = SubstateDatabaseStaging::new(&root);
root.commit(&overlay.into_database_updates());
```

radix-engine-common/src/iterators/overlaying_iterator.rs
0xOmarA: Sounds good, moving to `utils`

substate-store-impls/src/substate_database_staging.rs
0xOmarA: Actually, there might be a very interesting way to achieve this in the Rust type system that allows us to operate over `S`, `&S`, and `&mut S`!

Here is some very rough example code I wrote when experimenting on this:
```rust
pub struct Overlay<S> {
    overlay: StagingDatabaseUpdates,
    root: S,
}

impl<S> Overlay<S> {
    fn new(root: S) -> Self {
        Self {
            root,
            overlay: Default::default(),
        }
    }
}

impl<S> Overlay<&mut S>
where
    S: CommittableSubstateDatabase + BorrowMut<S>,
{
    pub fn commit_to_root(&mut self) {
        self.root
            .borrow_mut()
            .commit(&DatabaseUpdates::from(self.overlay.clone()))
    }
}

#[test]
fn example() {
    // This works
    let db = InMemorySubstateDatabase::standard();
    let overlay = Overlay::new(db);

    // This works
    let db = InMemorySubstateDatabase::standard();
    let overlay = Overlay::new(&db);

    // This works
    let db = InMemorySubstateDatabase::standard();
    let overlay = Overlay::new(&mut db);

    // This doesn't work
    let db = InMemorySubstateDatabase::standard();
    let overlay = Overlay::new(&db);
    overlay.commit_to_root();

    // This works
    let db = InMemorySubstateDatabase::standard();
    let overlay = Overlay::new(&mut db);
    overlay.commit_to_root();
}
```

substate-store-impls/src/substate_database_staging.rs
0xOmarA: I went with `SingleSubstateDatabaseOverlay`. Would be happy to change to something else if you want.

substate-store-impls/src/substate_database_staging.rs
dhedey: Right! Something like this would work, exactly! 

Although slight adjustment - the overlay should be consumed when it's committed...

If it helps, I've had my own stab at this here, in case you like it / would like to merge it in: https://github.com/radixdlt/radixdlt-scrypto/pull/1730

substate-store-impls/src/substate_database_staging.rs
dhedey: You know what, I think actually like just `SubstateDatabaseOverlay` :D.

The word overlay (instead of staging) is enough to indicate a single layer to me.

dhedey (approved):
Really nice! A few minor suggestions then I'd be happy for this to merge.

substate-store-impls/src/single_substate_database_overlay.rs
```diff
@@ -0,0 +1,407 @@
...
+                            .list_entries_from(partition_key, from_sort_key.as_ref());
+                        let overlaying = substate_updates
+                            .iter()
+                            .map(|(sort_key, database_update)| match database_update {
+                                DatabaseUpdate::Set(substate_value) => {
+                                    (sort_key.clone(), Some(substate_value.clone()))
+                                }
+                                DatabaseUpdate::Delete => (sort_key.clone(), None),
+                            })
+                            .filter(move |(sort_key, _)| match from_sort_key {
```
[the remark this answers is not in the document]
0xOmarA: Oh fancy! Yup, this is much better than filtering. Applying this change now to the two areas we used to do filtering.

substate-store-impls/src/single_substate_database_overlay.rs
```diff
@@ -0,0 +1,407 @@
...
+                    None => OverlayLookupResult::NotFound,
+                }
+            }
+            // This particular node key does not exist in the overlay. The substate must be read
+            // from the underlying database.
+            None => OverlayLookupResult::NotFound,
+        };
+
+        match overlay_lookup_result {
+            OverlayLookupResult::Found(substate_value) => substate_value.cloned(),
```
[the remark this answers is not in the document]
0xOmarA: Thank you! I find it to be better and easier to understand than a few nested `Option`s :)

substate-store-impls/src/substate_database_staging.rs
0xOmarA: Thank you so much for #1730! I'm happy that `SubstateDatabaseOverlay` now works with either mutable or immutable references! I've merged the PR and just doing some minor renames of the module names.

PR #1730, opened 2024-03-01
feature: The DatabaseOverlay is mergeable into its root store

dhedey:
## Summary
For discussion with Omar.

### DIFF
--- a/substate-store-impls/src/single_substate_database_overlay.rs
+++ b/substate-store-impls/src/single_substate_database_overlay.rs
@@ -2,10 +2,10 @@ use radix_engine_common::prelude::*;
 use substate_store_interface::interface::*;
 use utils::prelude::borrow::*;
 
-pub type UnmergeableDatabaseOverlay<'a, S> = DatabaseOverlay<&'a S, S>;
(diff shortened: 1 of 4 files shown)


substate-store-impls/src/single_substate_database_overlay.rs
```diff
@@ -1,27 +1,54 @@
...
     overlay: StagingDatabaseUpdates,
 
-    /// The root database that this type overlays. There is no restriction on what type this is at
-    /// this level. This is an immutable reference to a substate store as there is no need for this
-    /// type to own the underlying database or to mutate it in any way.
-    root: &'s S,
+    /// A mutable or immutable reference to the root database that this type overlays.
+    /// It only needs to be mutable if you wish to commit to the root store.
+    /// To be useful, `S` should implement at least `Borrow<D>`.
+    root: S,
```
0xOmarA, line 17: Nice! My impl was missing the use of the two different generic parameters and I think that this is why I was facing issues with it. The use of `S` and `D` here is clear and quite nice!

0xOmarA (approved):

PR #1732, opened 2024-03-07
Add Substate key to accompany the Substate value associated with a JMT leaf

jakrawcz-rdx:
## Summary
This is a minor follow-up to https://github.com/radixdlt/radixdlt-scrypto/pull/1728 (i.e. a tweak to impl of https://radixdlt.atlassian.net/browse/LZ-20).
It turns out that Node may optimize its historical state feature impl when it knows the associated Substates key (apart from value).

## Details
We are not sure we will use the Substate key right away, but it is very cheap to just pass it there and thus we prefer to have it available.

Apart from the feature, I squeezed in a slight simplification of the "associate substates" logic (i.e. avoiding some rewrites by reading directly from `&PartitionDatabaseUpdates`).

## Testing
Adjusted the unit test.

## Update Recommendations
Only Node is supposed to integrate and this will be a trivial API adjustment.

### DIFF
--- a/substate-store-impls/src/hash_tree/substate_tier.rs
+++ b/substate-store-impls/src/hash_tree/substate_tier.rs
@@ -140,7 +140,7 @@ pub struct SubstateSummary {
     /// A global tree node key of this Substate's leaf.
     ///
     /// Note: this is a low-level detail, needed e.g. to properly correlate the Substate value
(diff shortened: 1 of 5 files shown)


dhedey (approved):
Really nice 👍

PR #1740, opened 2024-03-14
Bugfix: Associate JMT leaves created due to "tree restructuring" with their unchanged Substates

jakrawcz-rdx:
## Summary
This fixes a bug found in the impl of https://github.com/radixdlt/radixdlt-scrypto/pull/1728:
It turns out that Node needs to be notified about JMT leaves created during tree restructuring. Previous implementation was deliberately skipping them.

## Testing
Added a unit test for precisely this behavior (which fails before the fix: see the first commit).
Added another unit test to cover partition resets (which luckily always worked fine, but is just a related area with some bug potential).

## Update Recommendations
Only internal Node devs need to adjust to some small API change.

### DIFF
--- a/substate-store-impls/src/hash_tree/substate_tier.rs
+++ b/substate-store-impls/src/hash_tree/substate_tier.rs
@@ -213,16 +213,31 @@ impl<'s, S: ReadableTreeStore + WriteableTreeStore> SubstateTier<'s, S> {
         for (key, node) in tree_update_batch.node_batch.iter().flatten() {
             // We promised to associate Substate values; but not all newly-created nodes are leaves:
             if let Node::Leaf(leaf_node) = &node {
-                // And not every newly-created leaf comes from a value change: (sometimes it is just a tree re-structuring!)
                 let sort_key = Self::to_typed_key(leaf_node.leaf_key().clone());
-                if let Some(substate_value) = substate_updates.get_upserted_value(&sort_key) {
-                    self.base_store.associate_substate(
-                        &self.stored_node_key(&key),
-                        &self.partition_key,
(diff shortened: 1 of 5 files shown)


substate-store-impls/src/hash_tree/substate_tier.rs
```diff
@@ -213,16 +213,31 @@ impl<'s, S: ReadableTreeStore + WriteableTreeStore> SubstateTier<'s, S> {
...
                 let sort_key = Self::to_typed_key(leaf_node.leaf_key().clone());
-                if let Some(substate_value) = substate_updates.get_upserted_value(&sort_key) {
-                    self.base_store.associate_substate(
-                        &self.stored_node_key(&key),
-                        &self.partition_key,
-                        &sort_key,
-                        substate_value,
-                    );
-                }
+                let substate_value = match substate_updates {
```
dhedey, line 217: I quite liked the `get_upserted_value` abstraction... what do you think about getting it back, but perhaps having it be slightly more detailed. E.g. be `get_substate_change` and return:
```
enum SubstateChange {
    Upsert(..), // Delta>Set or Reset>Present
    Delete,  // Delta>Delete or Reset>NotPresent
    None, // Delta>NotPresent
}
```
Which we can then handle more tersely here.

dhedey (approved):

substate-store-impls/src/hash_tree/substate_tier.rs
jakrawcz-rdx: I agree, it is useful 👍  (I just allowed myself to express `None` via regular `Option`)

PR #1747, opened 2024-03-25
Refactor protocol update configs

iamyulong:
## Summary

Refactors Radix Engine protocol update configurations.

## Update Recommendations

To generate state updates, follow example:
```rust
let protocol_updates = ProtocolUpdates::none().with_anemone();
let state_updates = protocol_updates.generate_state_updates(&substate_db);
```


radix-engine-tests/tests/application/metering.rs
```diff
@@ -417,11 +418,13 @@ fn run_basic_transfer_to_virtual_account(mode: Mode) {
 }
 
 fn run_radiswap(mode: Mode, use_v1_1_pools: bool) {
```
lrubasze, line 420: Minor, but maybe worth changing the name of this variable? Not sure if `use_v1_1_pools` is clear.
eg.
`use_v1_1_pools` -> `pool_math_precision_fix`

radix-engine/src/updates/protocol_updates.rs
```diff
@@ -0,0 +1,102 @@
...
+        self.updates.extend(btreeset!(
+            ProtocolUpdate::Bls12381AndKeccak256,
+            ProtocolUpdate::SecondPrecisionTimestamp,
+            ProtocolUpdate::PoolMathPrecisionFix,
+            ProtocolUpdate::ValidatorCreationFeeFix,
+        ));
+        self
+    }
+
+    pub fn with_bottlenose(mut self) -> Self {
```
lrubasze, line 57: Not sure but maybe `with_bottlenose()` should include `with_anemone()`? 🤔 
If not then maybe worth clearly stating that given named update does not include previous updates (however it might happen that some updates depend on previous updates).

radix-engine-tests/tests/application/metering.rs
iamyulong: Yep

radix-engine/src/updates/protocol_updates.rs
lrubasze: Fair enough :)

lrubasze (approved):

radix-engine/src/updates/protocol_updates.rs
```diff
@@ -0,0 +1,106 @@
+use super::state_updates::*;
+use crate::{internal_prelude::*, track::StateUpdates};
+use radix_substate_store_interface::interface::SubstateDatabase;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ProtocolUpdate {
```
0xOmarA, line 6: Seems like a flat enum will get quite large as we add more and more protocol updates. Perhaps this should be something like:

```rust
pub enum ProtocolUpdate {
    AnemoneProtocolUpdate(AnemoneProtocolUpdate),
    BottlenoseProtocolUpdate(BottlenoseProtocolUpdate),
}

pub enum AnemoneProtocolUpdate {
    /// Introduces BLS12-381 and Keccak-256 features.
    Bls12381AndKeccak256,

    /// Exposes second-precision timestamp.
    SecondPrecisionTimestamp,

    /// Increases the math precision with native pool implementations.
    PoolMathPrecisionFix,

    /// Changes the cost associated with validator creation.
    ValidatorCreationFeeFix,

    /// Moves various protocol parameters to state.
    ProtocolParamsToState,
}

pub enum BottlenoseProtocolUpdate {
    /// Exposes a getter method for reading owner role rule.
    OwnerRoleGetter,

    /// Various system patches.
    SystemPatches,

    /// Introduces the account locker blueprint.
    AccountLocker,
}
```

radix-engine/src/updates/protocol_updates.rs
```diff
@@ -0,0 +1,106 @@
...
+        self.updates.insert(update);
+        self
+    }
+
+    pub fn without_update(mut self, update: ProtocolUpdate) -> Self {
+        self.updates.remove(&update);
+        self
+    }
+
+    pub fn generate_state_updates<S: SubstateDatabase>(&self, db: &S) -> Vec<StateUpdates> {
```
0xOmarA, line 81: Perhaps the state updates should be provided in a canonical order and not in the order in which the enum variants are provided?

radix-engine/src/updates/protocol_updates.rs
iamyulong: I'm not sure if I see the benefits.

I viewed each of the protocol update as an "EIP". The definition of EIP shouldn't depend on the network updates that will include it.

radix-engine/src/updates/protocol_updates.rs
0xOmarA: Ah yes you're right. In this case we should remember not to change the ordering of the variants on the enum.

radix-engine/src/updates/protocol_updates.rs
0xOmarA: Main benefits is just code organization. However, as you mentioned, if some upgrades will be made without a network upgrade then flat enum would make more sense.

PR #1749, opened 2024-03-26
Fix preview crash

iamyulong:
## Summary

This PR fixes a crash when previewing a manifest with invalid references:
- Introduced a dedicated bootloading error, which maps to rejections explicitly; 
    - This will prevent engine from entering state update summarize logic
- Updated the state summary and structure inference to be based on state updates, instead of Track's internal structures

## Testing

New test added.

### DIFF
--- a/radix-engine/src/system/system_db_reader.rs
+++ b/radix-engine/src/system/system_db_reader.rs
@@ -11,7 +11,7 @@ use radix_substate_store_interface::db_key_mapper::{
     MappedCommittableSubstateDatabase, SubstateKeyContent,
 };
 use radix_substate_store_interface::interface::{
-    CommittableSubstateDatabase, ListableSubstateDatabase,
+    CommittableSubstateDatabase, DatabaseUpdate, ListableSubstateDatabase,
 };
 use radix_substate_store_interface::{
     db_key_mapper::{DatabaseKeyMapper, MappedSubstateDatabase, SpreadPrefixKeyMapper},
@@ -31,7 +31,9 @@ use crate::system::system_type_checker::{
     BlueprintTypeTarget, KVStoreTypeTarget, SchemaValidationMeta,
 };
 use crate::system::type_info::TypeInfoSubstate;
-use crate::track::TrackedNode;
+use crate::track::{
+    BatchPartitionStateUpdate, NodeStateUpdates, PartitionStateUpdates, StateUpdates,
+};
 use crate::transaction::{
     ObjectInstanceTypeReference, ObjectSubstateTypeReference, PackageTypeReference,
 };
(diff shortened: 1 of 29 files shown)


radix-engine/src/transaction/system_structure.rs
```diff
@@ -157,37 +157,20 @@ impl<'a, S: SubstateDatabase> SubstateSchemaMapper<'a, S> {
...
-                            continue;
+    pub fn add_substate_structures(&mut self, state_updates: &StateUpdates) {
+        for (node_id, node_updates) in &state_updates.by_node {
+            let NodeStateUpdates::Delta { by_partition } = &node_updates;
+
+            for (partition_num, partition_update) in by_partition {
+                match partition_update {
+                    PartitionStateUpdates::Delta { by_substate } => {
+                        for substate_key in by_substate.keys() {
+                            self.add_substate_structure(node_id, partition_num, substate_key);
```
dhedey, line 168: :D nice simplification

radix-engine/src/system/system_modules/costing/costing_module.rs
```diff
@@ -288,7 +271,8 @@ impl<V: SystemCallbackObject> SystemModule<SystemConfig<V>> for CostingModule {
             .apply_execution_cost(ExecutionCostingEntry::BeforeInvoke {
                 actor: &invocation.call_frame_data,
                 input_size: invocation.len(),
-            })?;
+            })
+            .map_err(|e| RuntimeError::SystemModuleError(SystemModuleError::CostingError(e)))?;
```
dhedey, line 275: Nice changes in this file error wise

radix-engine/src/system/system_db_reader.rs
```diff
@@ -918,21 +924,41 @@ impl<'a, S: SubstateDatabase> SystemDatabaseReader<'a, S> {
...
+                .and_then(|node_updates| match node_updates {
+                    NodeStateUpdates::Delta { by_partition } => by_partition.get(&partition_num),
                 })
-                .and_then(|tracked_key| {
-                    tracked_key
-                        .substate_value
-                        .get()
-                        .map(|e| e.as_typed().unwrap())
+                .and_then(|partition_updates| match partition_updates {
+                    PartitionStateUpdates::Delta { by_substate } => {
```
dhedey, line 936: I see this logic in loads of places :).

Jakub has built [partition_updates.get_substate_change(..)](https://github.com/radixdlt/radixdlt-scrypto/pull/1740/files#diff-5c6afaf6cf6b87254ae416cb45c7a46656b915c0e49f28c33a7ee2143d93180a) to simplify this logic - and then we can return `Option<SubstateChange>` from this method, which I think is clearer.

I mean we could even add a `get_substate_change(partition_num, substate_key)` method to `NodeStateUpdates` and a `get_substate_change(node_id, partition_num, substate_key)` method to `StateUpdates` too which would simplify this method a lot.

dhedey (commented):
Really nice!

PR #1755, opened 2024-03-30
feat: Add in place update support to versioned macro

dhedey:
## Summary

Pulls in (and productionizes) versioned macro changes from https://github.com/radixdlt/radixdlt-scrypto/pull/1744 ... and adds a lot of other stuff.

Key changes:
* Updates to Versioned macros - any external `define_versioned!` calls need updating.
* Updates to the native blueprint state macros to support `kind: StaticMultiVersioned`
* Updates to SBOR derive to support `as_type` and ignored enum variants, and tweaks to when `categorize_types` is needed to be more intuitive.
* Added a `PermitSborAttributes` derive which simply allows `#[sbor]` attributes to be used even if the type doesn't derive `Sbor`. This means a helper macro can add sbor-specific attributes, but leave it up to the user to decide if they actually need Sbor on the type or not.
* A new `eager_replace!` procedural macro which is a pre-processor style macro intended for use in declarative macros. It is effectively a more powerful version than `paste!` which is also inspired by `quote!` / `parse_quote!` and gets around this issue: https://github.com/dtolnay/paste/issues/40#issuecomment-2062953012 - and generally solves a lot of frustrations I've had building out the more gnarly declarative macros in the code base - to see it in practice, see `sbor-derive/lib.rs` and `sbor/versioned.rs`.

### Versioned
* Most of the methods on the `Versioned_` types have been renamed to be much clearer.
* Versioned types now contain an internal `_Versions` enum. See yak shaving below.
* Added a `fully_update_to_latest_version_mut` method to Versioned enums, which takes a Versioned enum, updates it in place if necessary, and returns a `&mut` to the resultant latest content. This makes for much more streamlined updates of nested state, which was why I added it when doing prototyping in #1744.
* Also added methods for e.g. `as_unique_version_mut` to ease use of single versioned item
* Improve rust docs on the versioned enum interface
* A few other drive-by fixes of compiler warnings

### Yak-shaving background
* To support in-place update of the Versioned enum from a `&mut` reference, we need some temporary state to store into the enum whilst we take out the current value and try to upgrade it (which requires taking ownership of the value, because the upgrade mechanism uses the more flexible From mechanism which needs an owned object).
* After exploring adding a fake enum variant, we ended up deciding to explore having a wrapper struct, and wrapping the versions into an `Option` so that we could temporarily replace it with `None` during the upgrade process.
* But adding that attribute caused Versioned types which didn't have any `#[derive(Sbor)]` annotations to get a compile error (about an unrecognized `sbor` attribute), so I also added a `PermitSborAttributes` derive macro which doesn't actually derive anything, but does enable `#[sbor(..)]` annotations to be present without giving a confusing compile error, to support versioned types which don't use SBOR derives

## Testing

Added tests for the SBOR changes to `sbor-tests`.

The versioned changes are tested in the rust doc tests, and by virtue of being used by other tests.

### For Internal Integrators
Anyone consuming a versioned type may need to change from using a `match` to using one of the many new methods, e.g.:
* `fully_update_into_latest_version` (`self -> Self::Latest`)
* `fully_update_to_latest_version_mut` (`&mut self -> &mut Self::Latest`)
* `as_unique_version_ref` (`&self -> &Self::Latest`) - only on single versioned enums

If you need to match the versions, you can call `.as_versions_ref()` / `.as_versions_mut()` / `.into_versions()` and then match on that.

Anyone creating a versioned type will need to update the declaration to e.g. something like the following. Notable changes are 1. removal of the `enum` type, and addition of the Versions type.
```rust
define_single_versioned!(
    #[derive(Debug, Clone, PartialEq, Eq, Sbor)]
    #[sbor(child_types = "S::CustomTypeKind<LocalTypeId>, S::CustomTypeValidation")]
    pub VersionedSchema(SchemaVersions)<S: CustomSchema> => Schema<S> = SchemaV1::<S>
);
define_versioned!(
    #[derive(Debug, Clone, PartialEq, Eq, Sbor)]
    VersionedExample(ExampleVersions) {
        previous_versions: [
            1 => ExampleV1: { updates_to: 2 },
            2 => ExampleV2: { updates_to: 4 },
            3 => ExampleV3: { updates_to: 4 },
        ],
        latest_version: {
            4 => Example = ExampleV4,
        },
    }
);
```


iamyulong:
Looks good.

One idea to explore is to make the attributes more structural, such as

```rust
#[sbor(as_type("type", as_ref="", from_value="", transparent_name))]
```

iamyulong (approved):

PR #1759, opened 2024-04-03
Added missing `compile_with_env_vars()` function

mstrug-rdx:
## Summary
Added missing `compile_with_env_vars()` function required for `compile-blueprints-at-build-time` feature in `radix-engine-tests` project.

Also there was an issue when using `compile-blueprints-at-build-time` feature with packages specified to `PackageLoader` in form of path, ex.: `oracle_proxies/oracle_proxy_with_global`. This was fixed in `PackageLoader` by extracting last section from the path.

### DIFF
--- a/radix-engine-tests/src/common.rs
+++ b/radix-engine-tests/src/common.rs
@@ -3,6 +3,7 @@
 pub mod package_loader {
     use radix_common::prelude::*;
     use radix_substate_store_queries::typed_substate_layout::*;
+    use std::path::PathBuf;
 
     const PACKAGES_BINARY: &[u8] =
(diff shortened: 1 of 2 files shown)


lrubasze:
LGTM.

Maybe it is worth to add `--features=compile-blueprints-at-build-time` back to `test.sh`, which has been removed in meantime.

mstrug-rdx:
> Maybe it is worth to add `--features=compile-blueprints-at-build-time` back to `test.sh`, which has been removed in meantime.

@lrubasze yes, we can add that back to test.sh. We need separate PR for that now.

PR #1760, opened 2024-04-03
Account Locker Scenarios

0xOmarA:
## Summary

This PR adds a scenario for the account locker blueprint with some accounts that accept and reject deposits and various store, airdrop, claim, and recover invocations to the account locker. Additionally, this PR makes changes to the interface of scenarios and how they are executed providing a standardized builder pattern for executing scenarios with the added support for protocol updates in scenarios.

## Description

The `account_locker` is a new scenario added for the account locker blueprint where there is a single admin of th locker, 2 resources: a fungible and non-fungible, and 3 users. This scenario aims to cover all of the possible interactions users can have with the account locker including, but not limited to, the following:

* Storing fungible and non-fungible resources in the locker.
* Storing fungible and non-fungible resources in the locker and and attempting to deposit them into accounts that may accept or reject the deposits.
* Airdropping fungible and non-fungible resources into the locker.
* Airdropping fungible and non-fungible resources into the locker and and attempting to deposit them into accounts that may accept or reject the deposits.
* Claiming fungible and non-fungible resources from the locker by amount.
* Claiming non-fungible resources from the locker by ids.
* Recovering fungible and non-fungible resources from the locker by amount.
* Recovering non-fungible resources from the locker by ids.

In addition to the newly added `account_locker` scenario this PR also makes some changes to the interface of scenarios and how they are executed which is motivated by the following:

1. **Non-standardized Interface:** We have multiple tests that rely on executing the scenarios and all of these tests executed the scenarios slightly differently from one another. This means that any change to the execution of scenarios needs to be duplicated to all of the tests that execute them.
2. **Protocol Update Support in Scenarios:** All of the scenarios that we have so far run after genesis, however, this will not be the case for new scenarios added as time goes by, especially for scenarios that depend on packages that are only available after some protocol update, like the account locker.

The solution to this is `DefaultTransactionScenarioExecutor` which is a builder pattern for executing scenarios that's generic, customizable, and configurable that handles the execution of scenarios, allows for configuring certain parameters about the execution, and allows the client to listen for events through callbacks. This has greatly simplified the tests that use the scenarios such as the following examples: [example 1](https://github.com/radixdlt/radixdlt-scrypto/pull/1760/files#diff-90e3adeaa1e63df09242dc438aea7096ae92ab57f728c8b5224765991d070bcc), and [example 2](https://github.com/radixdlt/radixdlt-scrypto/pull/1760/files#diff-6d55fdb4822454453567dc7dae8657c646b4beedcc25ccbad9e87265fb7e5579).

### Node

Please note the following:

* We have a new scenario.
* This scenario does not run after genesis but after the bottlenose protocol update.
* For the case of bottlenose, to find all of the scenarios that should be run for it you can find them through `scenario_builders()[ScenarioRequirements::ProtocolUpdateUpTo(ProtocolUpdate::Bottlenose)]` which returns a `Vec<Box<dyn FnOnce(ScenarioCore) -> Box<dyn ScenarioInstance>>>` of all of the scenario builders for bottlenose.


radix-transaction-scenarios/src/scenarios/account_locker.rs
```diff
@@ -0,0 +1,777 @@
...
+                    Ok(())
+                },
+            )
+            .successful_transaction(|core, config, state| {
+                core.next_transaction_with_faucet_lock_fee(
+                    "account-locker-setting-up-account-deposit-rules",
+                    |builder| {
+                        builder
+                            .call_method(
+                                config.user_account1.address,
```
dhedey, line 148: For clarity when it comes to looking at the named addresses, I'd prefer this account to be named, e.g. `account_with_deny_resource_preference`

radix-transaction-scenarios/src/scenarios/account_locker.rs
```diff
@@ -0,0 +1,777 @@
...
+    pub(crate) account_locker_admin_badge: State<ResourceAddress>,
+
+    pub(crate) fungible_resource: State<ResourceAddress>,
+    pub(crate) non_fungible_resource: State<ResourceAddress>,
+}
+
+impl Default for AccountLockerScenarioConfig {
+    fn default() -> Self {
+        Self {
+            account_locker_admin_account: ed25519_account_for_private_key(1),
```
dhedey, line 29: I'm afraid we can't just use virtual accounts for this, if we want to be able to run this on e.g. stokenet as part of bottlenose instantiation.

If this isn't being run at genesis, then it's possible for other transactions in the interim to have already used these pre-allocated accounts, which would interfere with the execution of this scenario, causing it to fail.

Setting of things like deposit rules might also interfere with other scenarios.

Instead, we should:
* Create a key pair for each account in the Config
* In the first transaction, use `CREATE_ACCOUNT_ADVANCED`, setting the keys up for the accounts, and capturing the account addresses into state.

Perhaps we should rename `ed25519_account_for_private_key` to `genesis_only_ed25519_account_for_private_key` or something to highlight this for other scenarios!

radix-engine-tests/tests/db/jmt_consistency.rs
```diff
@@ -1,37 +1,30 @@
...
-                    previous = Some(receipt);
-                }
-                NextAction::Completed(end_state) => {
-                    next_nonce = end_state.next_unused_nonce;
-                    break;
-                }
+    let db = StateTreeUpdatingDatabase::new(InMemorySubstateDatabase::standard());
+    DefaultTransactionScenarioExecutor::new(db)
+        .on_transaction_execution(|_, _, _, db| {
+            let hashes_from_tree = db.list_substate_hashes();
```
dhedey, line 11: Hmm we have duplicated the content of `assert_state_tree_matches_substate_store` here, which is a little unclean.

Probably we still want `LedgerSimulator`'s `assert_state_tree_matches_substate_store` but maybe they should call into some common method or something?

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+
+    /// Sets the callback to call after executing a scenario transaction.
+    pub fn on_transaction_execution<
+        NF1: FnMut(&ScenarioMetadata, &NextTransaction, &TransactionReceiptV1, &D),
+    >(
+        self,
+        callback: NF1,
+    ) -> TransactionScenarioExecutor<D, W, E, NF1, F2> {
+        TransactionScenarioExecutor {
+            /* Environment */
```
dhedey, line 254: For all of these you should be able to return this instead:
```rust
    Self {
        on_transaction_execution: callback,
        ..self,
    }
```

Or perhaps even simpler: 
```rust
pub fn on_transaction_execution<
        NF1: FnMut(&ScenarioMetadata, &NextTransaction, &TransactionReceiptV1, &D),
    >(mut self, callback: NF1) -> Self {
    self.on_transaction_execution = callback;
    self
}
```

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+                    ]
+                ),*
+            }
+        }
+    };
+}
+
+define_scenario_builders! {
+    // The set of scenarios that should be run after genesis.
+    ScenarioRequirements::Genesis => [
```
dhedey, line 75: In my opinion, the requirements of the scenario should live in the scenario itself. That way, if I want to execute scenario X, I can check if it's relevant from the scenario itself, rather than it living out here.

We can do this in the node, for example.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+use transfer_xrd::TransferXrdScenarioCreator;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ScenarioRequirements {
+    /// These scenarios do not require any protocol updates and are expected to be run at genesis
+    /// and be compatible with it.
+    Genesis,
+    /// These scenarios require one or more protocol updates and this enum specifies the protocol
+    /// update that they require.
+    ProtocolUpdateUpTo(ProtocolUpdate),
```
dhedey, line 34: I'd kinda like us to just make genesis a protocol update, but I don't know how feasible that is.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+            starting_nonce: self.starting_nonce,
+            nonce_handling: self.nonce_handling,
+            network_definition: self.network_definition,
+            /* Callbacks */
+            on_transaction_execution: self.on_transaction_execution,
+            on_scenario_start: callback,
+        }
+    }
+
+    pub fn execute(mut self) -> Result<CompletedExecutionReceipt<D>, ScenarioExecutorError> {
```
dhedey, line 292: I find it a little confusing that this just says `execute` when it's called, and not `execute_all` or something.

Actually, I think a much clearer API would be: `execute(self, scenario_filter: ScenarioFilter)`

Where we have e.g.
```rust
enum ScenarioFilter {
    ExactScenarios([String]),
    AllScenariosValidAtCurrentProtocolVersion, // Would need to save this on the LedgerSimulator I guess?
    AllScenariosValidAsOf(ProtocolUpdate),
    AllScenariosFirstValidAt(ProtocolUpdate),
}
```

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+
+impl<D, W, E, F1, F2> TransactionScenarioExecutor<D, W, E, F1, F2>
+where
+    D: SubstateDatabase + CommittableSubstateDatabase,
+    W: WasmEngine,
+    E: NativeVmExtension,
+    F1: FnMut(&ScenarioMetadata, &NextTransaction, &TransactionReceiptV1, &D),
+    F2: FnMut(&ScenarioMetadata),
+{
+    pub fn new(database: D) -> DefaultTransactionScenarioExecutor<D> {
```
dhedey, line 149: Most of the defaults are reasonable, but I'd rather we take an explicit `network_definition` here.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+
+pub type DefaultTransactionScenarioExecutor<D> = TransactionScenarioExecutor<
+    D,
+    DefaultWasmEngine,
+    NoExtension,
+    fn(&ScenarioMetadata, &NextTransaction, &TransactionReceiptV1, &D),
+    fn(&ScenarioMetadata),
+>;
+
+impl<D, W, E, F1, F2> TransactionScenarioExecutor<D, W, E, F1, F2>
```
dhedey, line 141: Really like what you've done with this! Nice one :D.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+                let mut previous = None;
+                loop {
+                    let next = scenario
+                        .next(previous.as_ref())
+                        .map_err(|err| err.into_full(&scenario))
+                        .unwrap();
+                    match next {
+                        NextAction::Transaction(next) => {
+                            let receipt = self.execute_transaction(&next.raw_transaction)?;
+                            (self.on_transaction_execution)(
```
dhedey, line 351: I'd rather call this `on_transaction_executed`, because `on_transaction_execution` suggests to me it happens _before_ the transaction is executed.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+        Ok(CompletedExecutionReceipt {
+            database: self.database,
+        })
+    }
+
+    fn execute_transaction(
+        &mut self,
+        transaction: &RawNotarizedTransaction,
+    ) -> Result<TransactionReceiptV1, ScenarioExecutorError> {
+        let network = NetworkDefinition::simulator();
```
dhedey, line 382: We should get the `network` from `self`.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+        TransactionScenarioExecutor {
+            /* Environment */
+            database,
+            scrypto_vm: ScryptoVm::default(),
+            native_vm: NativeVm::new(),
+            /* Execution */
+            scenarios_to_execute: ScenarioRequirements::all().iter().copied().collect(),
+            bootstrap: true,
+            starting_nonce: 0,
+            nonce_handling: NonceHandling::Increment(1),
```
dhedey, line 159: This should be something like `next_scenario_nonce_handling` and I feel like it should default to `NextScenarioNonceHandling::StartAtNextAvailable` - `NonceHandling::Increment(1)` is too likely to cause collisions imo.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
dhedey: (Unfortunately the node won't be able to use it as it has more complicated commit logic than just `CommittableSubstateDatabase`, but it's good enough for its purpose here for sure 👍 )

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+#[derive(Clone, Debug)]
+pub enum ScenarioExecutorError {
+    BootstrapFailed,
+    FailedToGetEpoch,
+    TransactionValidationError(TransactionValidationError),
+}
+
+#[derive(Clone, Debug)]
+pub enum NonceHandling {
+    Increment(u32),
```
dhedey, line 416: Perhaps a clearer name might be `ScenarioStartNonceHandling`, and then the `Increment` variant might be more clearly named `PreviousScenarioStartNoncePlus(u32)` and `StartAtNextAvailable` could be `PreviousScenarioEndNoncePlusOne`

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
```diff
@@ -1,54 +1,422 @@
...
+                                as Box<dyn FnOnce(ScenarioCore) -> Box<dyn ScenarioInstance>>
+                        ),*
+                    ]
+                ),*
+            }
+        }
+    };
+}
+
+define_scenario_builders! {
```
dhedey, line 73: I think I just prefer ditching the macro and having:
```rust
pub fn get_all_scenarios() -> Vec<Box<dyn FnOnce(ScenarioCore) -> Box<dyn ScenarioInstance>>> {
    vec![
        Box::new(transfer_xrd::TransferXrdScenarioCreator::create),
        Box::new(radiswap::RadiswapScenarioCreator::create),
        // ...
    ]
}
```

Which we can then filter using `scenario.metadata()`

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
dhedey: (This can be on `ScenarioMetadata`)

radix-engine-tests/tests/db/substate_database_overlay.rs
```diff
@@ -484,30 +484,59 @@ fn run_scenarios(
...
+    for requirement in ScenarioRequirements::all() {
+        if let ScenarioRequirements::ProtocolUpdateUpTo(protocol_update) = requirement {
+            protocol_update
+                .generate_state_updates(ledger1.substate_db_mut(), &network)
+                .into_iter()
+                .for_each(|state_updates| {
+                    ledger1
+                        .substate_db_mut()
+                        .commit(&state_updates.create_database_updates::<SpreadPrefixKeyMapper>())
+                });
```
dhedey, line 497: I think we should have in `LedgerSimulatorBuilder` a method to apply a given protocol update, rather than putting it here.

radix-engine-tests/tests/db/substate_database_overlay.rs
```diff
@@ -484,30 +484,59 @@ fn run_scenarios(
...
-                        (ledger2.substate_db(), &receipt2),
-                    );
-
-                    previous = Some(receipt1);
-                }
-                NextAction::Completed(end_state) => {
-                    next_nonce = end_state.next_unused_nonce;
-                    break;
+    let mut scenario_builders = scenario_builders();
+    for requirement in ScenarioRequirements::all() {
```
dhedey, line 488: Is there any reason to apply protocol updates, then scenarios, then protocol updates, etc?

Could we instead just apply all protocol updates via the `LedgerSimulatorBuilder`, and then just run all scenarios?

radix-engine-tests/tests/db/substate_database_overlay.rs
```diff
@@ -484,30 +484,59 @@ fn run_scenarios(
...
+                .for_each(|state_updates| {
+                    ledger2
+                        .substate_db_mut()
+                        .commit(&state_updates.create_database_updates::<SpreadPrefixKeyMapper>())
+                })
+        };
+
+        let scenario_builders = scenario_builders
+            .shift_remove(requirement)
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
0xOmarA: Hey! We're not making that assumption, the requirements can be specified in any order and they will be executed in Genesis -> Anemone -> Bottlenose order regardless of how they're specified. This is because the first for loop in this function iterates over `ScenarioRequirements::all()` which defines the canonical order of the execution of scenarios.

How do you feel about the above? We can also specify them in a vector here an make the order more explicit and clear to clients of the executor.

radix-engine-tests/tests/db/substate_database_overlay.rs
```diff
@@ -484,30 +484,59 @@ fn run_scenarios(
...
-                    previous = Some(receipt1);
-                }
-                NextAction::Completed(end_state) => {
-                    next_nonce = end_state.next_unused_nonce;
-                    break;
+    let mut scenario_builders = scenario_builders();
+    for requirement in ScenarioRequirements::all() {
+        if let ScenarioRequirements::ProtocolUpdateUpTo(protocol_update) = requirement {
+            protocol_update
+                .generate_state_updates(ledger1.substate_db_mut(), &network)
```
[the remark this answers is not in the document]
0xOmarA: Apologies, name here is not clear, this part makes the assumption that a previous requirement seen by the loop was the anemone protocol update and that anemone was applied at that point, this is true since we iterate over `ScenarioRequirements::all()` in this loop. Seems to me like a name change might be better here, or perhaps even switching to the vectors based approach you left in the other comment.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: Yup, good point, I've made that change.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: Sounds good, will rename this.

radix-engine-tests/tests/db/jmt_consistency.rs
0xOmarA: Yeah agreed that it would be better to have just a single method for this check. I've refactored this into a `validate_state_tree_matches_substate_store` method on the `StateTreeUpdatingDatabase<D>` type that returns a `Result`. In tests this result is not used in any meaningful way and is just unwrapped. The main reason for having it be a `Result` and not a `panic` or an `assert` is to avoid having any panicing APIs on DB types.

radix-transaction-scenarios/src/scenarios/account_locker.rs
0xOmarA: Oh, good point!

I will change the scenario with the change you recommended here. 

The rename sounds good to me and I would also argue that perhaps we should make this an `unsafe` API to force users of this API to read the comments, docs, or at least think about is usage.

0xOmarA: Thinking about this more, I would like us to add the `deprecated` attribute to the `VirtualAccount` type as a whole and not just the constructor functions with a comment on why it is deprecated and should not be used in future scenarios.

0xOmarA: Deprecating it also allows us to specify a clear error message to our future selves as seen below:

image


Only disadvantage though is that using a deprecated type can get quite annoying with all of the warnings in the old scenarios (which we probably make very little changes to so maybe its fine?).

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: Sounds good, I've made those changes.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: Yup agreed, seems like `network_definition` is the only one that's out of place there and the default doesn't make sense to all clients. Will move it to be an explicit parameter passed to `new`.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: Thank you! Shame that the node can't use it though! Perhaps at some point when the code-bases merge we can have an imlpementation of the `TransactionScenarioExecutor` that requires the equivalent of the `CommittableSubstateDatabase` for the node. Another idea, if node wishes to use this despite not implementing the required traits then it could use the `SubstateDatabaseOverlay` wrapper which doesn't require the database implement `CommittableSubstateDatabase` and it, itself, implements `CommittableSubstateDatabase` for all databases it wraps.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: Sounds good on both the rename and changing the default behavior.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: Filtering sounds good to me especially with the refactor I made to the executor and how it works I think that filtering would fit in nicely. although I don't think that it should be a parameter passed to `execute` but rather `.filter` and `.remove_filter` methods on the builder.

The `AllScenariosValidAtCurrentProtocolVersion` is problematic since the executor can't tell what the current protocol version is from the substate store alone and so I can't see how we would find this information, unless we make it part of the "store protocol config in substates" ticket.

radix-engine-tests/tests/db/substate_database_overlay.rs
0xOmarA: Mainly so that the scenarios are un against the protocol update that they were intended for to avoid any unexpected failures/successes that may happen if they're not run against their intended protocol update. 

With the refactored executor there is now ways to have finer grain control over these details, take a look at the tests in `dumper.rs`.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: This has been changed in the latest implementation where we now do indeed have the requirements defined on the scenario itself, on the `ScenarioCreator` implementation to be more specific.

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: I think that it's certainly possible, but might be too big of change for this PR :D

radix-transaction-scenarios/src/scenarios/account_locker.rs
```diff
@@ -0,0 +1,777 @@
...
+                            )
+                            .deposit_batch(config.user_account3.address)
+                    },
+                    vec![
+                        &config.account_locker_admin_account.key,
+                        &config.user_account3.key,
+                    ],
+                )
+            })
+            .finalize(|_, config, state| {
```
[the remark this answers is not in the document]
0xOmarA: > account can transfer asset to itself

Not entirely sure what the last bit means?

I'm assuming the scenario that you're referring to is one where the global caller badge of the account locker is an authorized depositor for an account and is able to deposit resources directly to it despite it being in deny all mode?

radix-engine-tests/tests/db/substate_database_overlay.rs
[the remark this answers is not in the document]
0xOmarA: Commenting an update here, I've refactored the implementation quite a bit based on David's feedback, the protocol requirement of the scenarios is now specified on the `ScenarioCreator` of the scenario and the order of the scenarios as a whole is defined in the executor.

radix-engine-tests/tests/db/substate_database_overlay.rs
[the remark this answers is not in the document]
0xOmarA: This has been refactored to make it clear that each protocol update is applied only once and that is when it is encountered.

radix-transaction-scenarios/src/scenarios/account_locker.rs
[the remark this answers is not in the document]
dhedey: Yes, exactly. And then we want to verify that the asset ended up in the account, and wasn't stored.
The `account can transfer asset to itself` was just essentially some way to verify that, but any other way I think is fine! :)

radix-transaction-scenarios/src/scenarios/account_locker.rs
dhedey: Works for me. We can allow deprecations in the old scenarios 👍

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
dhedey: Ahh well then I think it's better to lose the generics and just box the callbacks.

dhedey: Happy to pair with you on the refactor :D

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
dhedey: > although I don't think that it should be a parameter passed to execute but rather .filter and .remove_filter methods on the builder.

I think the default behaviour is quite unclear with respect to which scenarios actually execute. But if you feel strongly about this, can we at least rename `execute` to `execute_all_matching` or something?

Cargo.toml
```diff
@@ -115,7 +115,7 @@ serde = { version = "1.0.144", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.105" }
 sha3 = { version = "0.10.8", default-features = false, optional = false }
 shared_memory = { version = "0.12.4" }
-strum = { version = "0.24", default-features = false, features = ["derive"] }
+strum = { version = "0.26.2", default-features = false, features = ["derive"] }
```
dhedey, line 118: Why do we need to update `strum`?

radix-transaction-scenarios/src/executor.rs
```diff
@@ -0,0 +1,592 @@
...
+    starting_nonce: u32,
+    /// How the executor should handle nonces and how it should get the next nonce.
+    next_scenario_nonce_handling: ScenarioStartNonceHandling,
+    /// The network definition to use in the execution of the scenarios.
+    network_definition: NetworkDefinition,
+
+    /* Callbacks */
+    /// A callback that is called when a new protocol requirement is encountered. This can be useful
+    /// for clients who wish to apply protocol updates they wish.
+    on_new_protocol_requirement_encountered: F1,
```
dhedey, line 72: As mentioned, if we make these `Box<dyn FnMut(...)>` I imagine the code would be neater and easier to work with :)

radix-transaction-scenarios/src/executor.rs
```diff
@@ -0,0 +1,592 @@
...
+                    .for_each(|state_updates| {
+                        db.commit(&state_updates.create_database_updates::<SpreadPrefixKeyMapper>())
+                    })
+            },
+            on_transaction_executed: |_, _, _, _| {},
+            on_scenario_start: |_| {},
+            after_bootstrap: |_, _| {},
+        }
+        // Genesis Scenarios.
+        .register_scenario_with_create_fn::<TransferXrdScenarioCreator>()
```
dhedey, line 132: We filter the list of all scenarios in the node based on their name in their metadata, so this breaks node compatibility.

radix-transaction-scenarios/src/scenarios/account_locker.rs
```diff
@@ -66,7 +121,10 @@ impl ScenarioCreator for AccountLockerScenarioCreator {
                                     },
                                 )
                                 .try_deposit_entire_worktop_or_abort(
-                                    config.account_locker_admin_account.address,
+                                    state
+                                        .account_locker_admin_account
+                                        .get()
```
dhedey, line 126: You can just use `unwrap()`

radix-transaction-scenarios/src/scenarios/account_locker.rs
```diff
@@ -1,17 +1,15 @@
...
+use radix_engine::updates::ProtocolUpdate;
+use radix_engine_interface::blueprints::account::*;
+use radix_engine_interface::*;
 
 pub struct AccountLockerScenarioConfig {
-    pub account_locker_admin_account: VirtualAccount,
-    pub user_account1: VirtualAccount,
-    pub user_account2: VirtualAccount,
-    pub user_account3: VirtualAccount,
+    pub account_locker_admin_account: PrivateKey,
```
dhedey, line 8: Very minor, but I think it would be a tiny bit clearer if we added the `_key` suffix to all of these

dhedey (approved):
Nice! Will leave it to you to do last tweaks.

The only thing I really need for the node is some way to iterate over all scenario creators and filter based on logical name 👍

radix-transaction-scenarios/src/executor.rs
0xOmarA: You can still filter them by name using `ScenarioFilter::ExactScenarios`, the registered scenarios is esentially the list of all scenarios, does this fit the node's requirements?

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: I think `execute` can be renamed to be more descriptive while still allowing for a simple interface, I propose:

* `execute_all` - Runs all of the scenarios without any filter.
* `execute_all_matching` - Acceps a filter and executes all scenarios that match/pass through some filter.

How do you feel about that?

radix-transaction-scenarios/src/executor.rs
0xOmarA: I typically prefer not to use dynamic dispatch when generics can be used but this seems like a perfect use case for dynamic dispatch. As more and more callbacks were added it was getting somewhat difficult to do the struct updates, which dynamic dispatch solved beautifully. Additionally, I do not suspect that the performance loss from this would be noticable since this part of scenario execution is not performance criticial.

It feels like something could also be done here with macros also or perhaps even a more declarative way to do [Builders with Type State](https://www.youtube.com/watch?v=k8kd22jNcps), but perhaps something that we can exlore later.

radix-transaction-scenarios/src/scenarios/account_locker.rs
0xOmarA: Ah nice! I've switched all places to use `State::unwrap` instead of `State::get`

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
0xOmarA: Hey! I've made the change to using dynamic dispatch on this (and commented in another comment/thread, but commenting here :D)

I've tried switching this to use dynamic dispatch and it works really nice in this context, I've typically been quite reserved about using dynamic dispatch especially in contexts where the compiler can tell what the types are with some extra help. However, I think that this builder (and probably others that have tons of generics) benefited well (syntactically) from using dynamic dispatch.

It still feels like using generics might be better, but it's not nearly as maintainable, and is quite hard to write. It seems like this is a perfect area to use macros to define these [builders with type state](https://www.youtube.com/watch?v=k8kd22jNcps).

radix-transaction-scenarios/src/scenarios/account_locker.rs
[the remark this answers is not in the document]
0xOmarA: Ah, got it! I've added this scenario as you described it. I check that the resources have been deposited by looking at the event in the `CommitResult` of the transaction.

radix-transaction-scenarios/src/executor.rs
dhedey: No it doesn't. As mentioned, we can't use this executor on the node -- because we need to be able to commit ourselves.

We do:
```
        let mut scenario = self
            .find_scenario(epoch, nonce, scenario_name)
            .unwrap_or_else(|| {
                panic!(
                    "Could not find scenario with logical name: {}",
                    scenario_name
                )
            });
```

Where:
```
    fn find_scenario(
        &self,
        epoch: Epoch,
        next_nonce: u32,
        scenario_name: &str,
    ) -> Option<Box<dyn ScenarioInstance>> {
        for scenario_builder in get_builder_for_every_scenario() {
            let scenario =
                scenario_builder(ScenarioCore::new(self.network.clone(), epoch, next_nonce));
            if scenario.metadata().logical_name == scenario_name {
                return Some(scenario);
            }
        }
        None
    }
```

radix-transaction-scenarios/src/scenarios/all_scenarios.rs
dhedey: Sure, we could do that. I feel `execute(ScenarioFilter::All)` is nicer though :). On reflection, I don't mind too much. We can keep it as `execute` for now and I can stick in a PR with a few small tweaks and get your reflections on them?

radix-transaction-scenarios/src/executor.rs
dhedey: (btw - for the node-complexity with transaction execution, see `prepare_scenario_transaction` for the gnarly details :P - the way we execute and then commit scenario transactions is just very different to the simple world in scrypto 👍 )

PR #1762, opened 2024-04-04
Various code clean-ups

iamyulong:
## Summary

- Fixed build failure with rust `1.79.0-nightly`
- Updated Scrypto new package template to use published crates
- Added `from the Radix DLT project` to description metadata
- Moved `assets` to individual crate to get rid of soft links


radix-clis/Cargo.lock
dhedey, line 1: I assume we have a separate ticket for a root Cargo.lock, and at that point, the simulators can become part of the workspace and use the same Cargo.lock.

radix-engine/assets/build.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname "$0")"
+
+# The genesis blueprints were compiled with Scrypto v1.0.0. To reproduce the same
+# wasm files, the Scrypto source code must be placed at `/src` and the blueprints workspace
+# source code must be placed at `/src/assets/blueprints`.
```
dhedey, line 8: :+1:

radix-transaction-scenarios/assets/build.sh
```diff
@@ -0,0 +1,45 @@
...
+
+    # Copy artifacts
+    cp \
+        $WORKSPACE_DIR/target/wasm32-unknown-unknown/release/$PACKAGE_NAME.{wasm,rpd} \
+        $DESTINATION_DIR/
+done
+
+sha256sum *.{wasm,rpd}
+
+# The script should generate the following hashes:
```
dhedey, line 28: It's a little unclear to me what these hashes represent -- maybe we don't need them?

radix-transaction-scenarios/assets/build.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname "$0")"
+
+IMAGE_NAME="radixdlt/scrypto-builder"
+IMAGE_TAG="v1.1.1"
+WORKSPACE_DIR="blueprints"
+DESTINATION_DIR="."
+for PACKAGE_NAME in "flash_loan" "global_n_owned" "kv_store" "max_transaction" "metadata" "radiswap" "royalties"
```
dhedey, line 10: I think it would be simpler and a little more accurate if this was a `build_package.sh` script, for use in development... and as part of running it you have to provide a package name and an image tag, and then the script outputs the `X.rpd` and `X.wasm`.

If we care about verification, the `build_package.sh` script could also output an artifact such as `X.build_info.yaml` which explains which version of the scrypto builder was used to build the file. Separately, we could have a `verify_package_builds.sh` script which greps for `X.wasm` files, loads the `X.build_info.yaml`, runs the build, with the right version of scrypto builder, and checks for equality. (But to be honest, I don't think this is needed for transaction scenarios so much).

Of course, we hit the problem of the existing files not having reproducible builds - in which case, I guess we could have a `build_info.yaml` which instructs us not to verify the build or something.

Anyway, as mentioned, I don't think it's critical to support verification on these builds for now -- but I think possibly it's worth us changing this script to just build a single package using a specific builder version?

radix-transaction-scenarios/generated-examples/royalties/001--royalties--publish-package.rtm
dhedey, line 1: Everything in this folder needs to be reverted, along with any change to `royalties.wasm`

radix-transaction-scenarios/src/runners/dumper.rs
```diff
@@ -78,11 +78,11 @@ pub fn run_all_in_memory_and_dump_examples(
 
     assert_eq!(
         substate_db.get_current_root_hash().to_string(),
-        "43be4cce2d4f2ed2eb519d77dfa770697244e843b2a0f7fd86bdf773d9b6f278"
```
dhedey, line 81: As mentioned, this shouldn't have changed - I imagine we need to revert `royalties.wasm` etc

scrypto-test/assets/blueprints/test_environment/Cargo.toml
dhedey, line 1: What's the need for this? Is it needed in operations?

If not, we should probably exclude it (and the whole `test/` directory in fact) in a `[package] exclude` in the `scrypto-test` Cargo.toml: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

dhedey (approved):
Really nice changes!

Most of my comments are small suggestions, the only one that I think does need fixing is with the modified scenario hashes.

Also, I think we could do with looking through each crate and seeing if we should be using exclude (https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) to cut down on the tarball size?

.gitignore
```diff
@@ -20,34 +20,22 @@ Cargo.lock
...
 # - Various test scrypto packages. We don't care too much about either way - but we do want CI to run without a lock, so let's git ignore these.
 #   Also make sure to blanket ignore all the old Cargo.locks before they moved to blueprints, to avoid developer git issues
 radix-engine-tests/**/Cargo.lock
-radix-clis/tests/blueprints/Cargo.lock
-
-# TYPE 5:
-# - Examples should have their own Cargo.lock files to show good practice
-!examples/**/Cargo.lock
-
-# TYPE 6:
```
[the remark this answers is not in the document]
iamyulong: Sure. We've been constantly changing folder structure and there doesn't seem to be complaints. I guess team is handling this well. :)

Cargo.toml
```diff
@@ -75,7 +75,7 @@ bincode = { version = "2.0.0-rc.1", default-features = false, features = ["deriv
 bitflags = { version = "1.3" }
 blake2 = { version = "0.10.6", default-features = false }
 blst = { version = "0.3.11", default-features = false, optional = false }
-bnum = { version = "0.7.0", default-features = false, features = ["numtraits"] }
+bnum = { version = "0.11.0", default-features = false, features = ["numtraits"] }
```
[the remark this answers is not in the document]
iamyulong: This is to the fix the nightly rust compilation error.

radix-clis/src/scrypto/cmd_new_package.rs
```diff
@@ -44,24 +36,11 @@ impl NewPackage {
             (
                 format!("{{ path = \"{}/sbor\" }}", scrypto_dir),
```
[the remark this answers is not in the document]
iamyulong: I think some of macros are assuming the `sbor` symbol, so I'd rather not change it for now.

radix-clis/Cargo.lock
iamyulong: Definitely. I'd defer this to when we merge the code bases.

radix-transaction-scenarios/generated-examples/royalties/001--royalties--publish-package.rtm
iamyulong: I've reverted everything except royalties, as I assume it's not used in the anemone stokenet.

radix-transaction-scenarios/assets/build.sh
iamyulong: Will update to the script to take an argument for a single package.

PR #1765, opened 2024-04-05
tweak: The JSON `variant_id` is now a u8, in line with Gateway typings

dhedey:
## Summary

https://radixdlt.atlassian.net/browse/SCRY-632

It turns out the Gateway typing for `ProgrammaticScryptoSborValueEnum` was:
```yml
    ProgrammaticScryptoSborValueEnum:
      allOf:
        - $ref: "#/components/schemas/ProgrammaticScryptoSborValue"
        - type: object
          required:
            - variant_id
            - fields
          properties:
            variant_id:
              type: integer
              minimum: 0
              maximum: 255
            variant_name:
              type: string
            fields:
              type: array
              items:
                $ref: "#/components/schemas/ProgrammaticScryptoSborValue"
```
And indeed Gateway was returning `variant_id` as a JSON number for its non-state responses. Core API was returning it as a string though, and it has been untyped in the Gateway API.

As the Gateway sees the most integrations, I think it needs to win here and force the underlying spec/impl to change to u8, and the Core API along with it. (Personally I think I also prefer `variant_id` being a JSON number anyway).

## Testing
Updated existing tests

### For dApp Developers
Anyone using programmatic SBOR JSON should be aware of this inconsistency, and that it's being fixed.

### For Internal Integrators
We should record to add to the node's release notes.

### DIFF
--- a/sbor/src/representations/serde_serialization/serde_serializer.rs
+++ b/sbor/src/representations/serde_serialization/serde_serializer.rs
@@ -925,7 +925,7 @@ mod tests {
                             "key": {
                                 "kind": "Enum",
                                 "type_name": "TestEnum",
-                                "variant_id": "0",
+                                "variant_id": 0,
                                 "variant_name": "UnitVariant",
                                 "fields": []
                             },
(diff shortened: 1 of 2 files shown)


dhedey:
Closing this PR as we're going to keep it as `String` and fix it in the Gateway instead.

PR #1770, opened 2024-04-12
Add `Zeroize` to private key

mstrug-rdx:
## Summary
Added derive of `Zeroize` and `ZeroizeOnDrop` to `Ed25519PrivateKey`. This is required by Sargon project (will be used thorught RET).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -125,6 +125,7 @@ wasmer = { version = "2.2.1" }
 wasmer-compiler-singlepass = { version = "2.2.1" }
 wasmparser = { version = "0.107.0", default-features = false }
 extend = { version = "1.2.0" }
+zeroize = { version = "1.7.0" }
 
 # Both the release and test profiles use `panic = "unwind"` to allow certain parts of the Radix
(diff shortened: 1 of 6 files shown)


CyonAlexRDX (changes requested):
Thx! But I need it on Secp256k1PrivateKey as well :)

mstrug-rdx:
@CyonAlexRDX I did some implementation which uses `SecretKey::non_secure_erase()` as default value but it fills slice with 1's.

CyonAlexRDX:
> @CyonAlexRDX I did some implementation which uses `SecretKey::non_secure_erase()` as default value but it fills slice with 1's.

All ones is not a valid Secp256k1PrivateKey. It is larger than the order of the curve. So best replace the key with the key 31 zeroes followed by 1, i.e the key `0d1`

CyonAlexRDX (approved):
LGTM!

Maybe add a unit tests that calls `zeroize()` on a non `1` private key and assert that it becomes `SecretKeyWrapper::default()`?

PR #1771, opened 2024-04-12
Refactoring: Exposing customization methods for `Executable`

jakrawcz-rdx:
## Summary
The Node needs to be able to customize the `abort_when_loan_repaid: true` for running a transaction pending in the mempool.
This PR standardizes the way to do this (and similar, already-existing) customizations.

## Testing
A refactoring-only: just the regression tests need to pass.

## Update Recommendations
Only the Node developers need to integrate.

### DIFF
--- a/radix-transactions/src/model/executable.rs
+++ b/radix-transactions/src/model/executable.rs
@@ -121,25 +121,48 @@ impl<'a> Executable<'a> {
         }
     }
 
+    // Consuming builder-like customization methods:
+
+    pub fn overwrite_intent_hash(self, hash: Hash) -> Self {
+        let mut updated_self = self;
+        match &mut updated_self.context.intent_hash {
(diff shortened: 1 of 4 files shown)


radix-transactions/src/model/executable.rs
```diff
@@ -121,25 +121,48 @@ impl<'a> Executable<'a> {
...
+        match &mut updated_self.context.intent_hash {
             TransactionIntentHash::ToCheck { intent_hash, .. }
             | TransactionIntentHash::NotToCheck { intent_hash } => {
                 *intent_hash = hash;
             }
         }
+        updated_self
+    }
+
+    pub fn skip_epoch_range_check(self) -> Self {
```
iamyulong, line 137: Minor: can use `mut self`

iamyulong (approved):

PR #1781, opened 2024-04-22
Improve Rust Analyzer performance on `develop`

0xOmarA:
## Summary

A follow up PR from #1779 that does the following:

* Improves the performance of Rust Analyzer autocomplete by removing the cycles causes by `pub extern crate self as some_name` which lead to paths such as `scrypto::scrypto::scrypto::prelude::*` to be valid.
* Removes the need for the `sbor` crate to be imported by Scrypto packages. The only required dependency now is `scrypto` alone. 

### For dApp Developers

You no longer need the `sbor` dependency and it can be safely removed from your `Cargo.toml` file.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,20 @@
 {
     "rust-analyzer.linkedProjects": [
+        // Root Workspace & CLIs.
         "Cargo.toml",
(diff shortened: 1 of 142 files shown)


iamyulong (approved):
LGTM

PR #1782, opened 2024-04-23
Switch to the typed variants in `scrypto` lib

lrubasze:
## Summary
Use typed variants in `scrypto` where possible.
Eg. 
- let `claim_royalties()` return `FungibleBucket` instead of `Bucket`
- let `FungibleVault` methods use `FungibleBucket`
etc.

## Testing
Tests have been updated.

### For dApp Developers
This is breaking change. Compiler might complain about mismatched types. But in most cases it will also nicely provides a suggested change.
eg. 
```bash
error[E0308]: mismatched types
  --> radiswap/src/lib.rs:60:13
   |
51 |         ) -> (Bucket, Option<Bucket>) {
   |              ------------------------ expected `(scrypto::prelude::Bucket, std::option::Option<scrypto::prelude::Bucket>)` because of return type
...
60 |             self.pool_component.contribute((resource1, resource2))
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `(Bucket, Option<Bucket>)`, found `(FungibleBucket, Option<FungibleBucket>)`
   |
   = note: expected tuple `(scrypto::prelude::Bucket, std::option::Option<scrypto::prelude::Bucket>)`
              found tuple `(scrypto::prelude::FungibleBucket, std::option::Option<scrypto::prelude::FungibleBucket>)`

error[E0308]: mismatched types
   --> radiswap/src/lib.rs:115:51
    |
115 |             self.pool_component.protected_deposit(bucket)
    |                                 ----------------- ^^^^^^ expected `FungibleBucket`, found `Bucket`
    |                                 |
    |                                 arguments to this method are incorrect
    |
note: method defined here
   --> /Users/lukaszrubaszewski/work/radixdlt/radixdlt-scrypto/scrypto/src/component/stubs.rs:315:12
    |
315 |         fn protected_deposit(&mut self, bucket: FungibleBucket);
    |            ^^^^^^^^^^^^^^^^^
help: try wrapping the expression in `scrypto::prelude::FungibleBucket`
    |
115 |             self.pool_component.protected_deposit(scrypto::prelude::FungibleBucket(bucket))
    |                                                   +++++++++++++++++++++++++++++++++      +

error[E0308]: mismatched types
   --> radiswap/src/lib.rs:119:13
    |
118 |           fn withdraw(&mut self, resource_address: ResourceAddress, amount: Decimal) -> Bucket {
    |                                                                                         ------ expected `scrypto::prelude::Bucket` because of return type
119 | /             self.pool_component.protected_withdraw(
120 | |                 resource_address,
121 | |                 amount,
122 | |                 WithdrawStrategy::Rounded(RoundingMode::ToZero),
123 | |             )
    | |_____________^ expected `Bucket`, found `FungibleBucket`
    |
help: call `Into::into` on this expression to convert `scrypto::prelude::FungibleBucket` into `scrypto::prelude::Bucket`
    |
123 |             ).into()
    |              +++++++

### CODE REVIEW

File: scrypto/src/component/stubs.rs [hunk 1]
```diff
@@ -89,10 +90,10 @@ extern_blueprint_internal! {
     {
         fn register(&mut self);
         fn unregister(&mut self);
-        fn stake_as_owner(&mut self, stake: Bucket) -> Bucket;
-        fn stake(&mut self, stake: Bucket) -> Bucket;
-        fn unstake(&mut self, stake_unit_bucket: Bucket) -> Bucket;
-        fn claim_xrd(&mut self, bucket: Bucket) -> Bucket;
+        fn stake_as_owner(&mut self, stake: FungibleBucket) -> FungibleBucket;
+        fn stake(&mut self, stake: FungibleBucket) -> FungibleBucket;
+        fn unstake(&mut self, stake_unit_bucket: FungibleBucket) -> FungibleBucket;
```
0xOmarA, line 95: I believe that `unstake` and `claim_xrd` are not correct.

* Input to `unstake` is indeed fungible but the output is a non-fungible claim NFT.
* Input to `claim_xrd` is a non-fungible claim NFT and the output is fungible XRD.

scrypto/src/component/stubs.rs
```diff
@@ -102,12 +103,12 @@ extern_blueprint_internal! {
         fn get_redemption_value(&self, amount_of_stake_units: Decimal) -> Decimal;
         fn signal_protocol_update_readiness(&mut self, vote: String);
         fn get_protocol_update_readiness(&mut self) -> Option<String>;
-        fn lock_owner_stake_units(&mut self, stake_unit_bucket: Bucket);
+        fn lock_owner_stake_units(&mut self, stake_unit_bucket: FungibleBucket);
         fn start_unlock_owner_stake_units(&mut self, requested_stake_unit_amount: Decimal);
-        fn finish_unlock_owner_stake_units(&mut self) -> Bucket;
+        fn finish_unlock_owner_stake_units(&mut self) -> FungibleBucket;
         fn apply_emission(
```
0xOmarA, line 109: Not quite user facing but should `apply_reward`'s bucket be a `FungibleBucket`?

scrypto/src/component/stubs.rs
```diff
@@ -150,18 +151,18 @@ extern_blueprint_internal! {
...
+        fn securify(&mut self) -> NonFungibleBucket;
         fn lock_fee(&mut self, amount: Decimal);
         fn lock_contingent_fee(&mut self, amount: Decimal);
         fn deposit(&mut self, bucket: Bucket);
         fn deposit_batch(&mut self, buckets: Vec<Bucket>);
         fn withdraw(&mut self, resource_address: ResourceAddress, amount: Decimal) -> Bucket;
         fn withdraw_non_fungibles(
             &mut self,
-            resource_address: ResourceAddress,
+            resource_address: NonFungibleBucket,
```
0xOmarA, line 165: Probably an accidental mistake, but this should remain to be a `ResourceAddress`.

scrypto/src/component/stubs.rs
```diff
@@ -178,14 +179,14 @@ extern_blueprint_internal! {
         ) -> Bucket;
         fn lock_fee_and_withdraw_non_fungibles(
             &mut self,
-            amount_to_lock: Decimal,
+            amount_to_lock: NonFungibleBucket,
```
0xOmarA, line 182: Ditto: seems to have completely changed the type here.

scrypto/src/component/stubs.rs
```diff
@@ -178,14 +179,14 @@ extern_blueprint_internal! {
...
             resource_address: ResourceAddress,
             ids: Vec<NonFungibleLocalId>,
         ) -> Bucket;
-        fn create_proof_of_amount(&self, resource_address: ResourceAddress, amount: Decimal) -> Proof;
+        fn create_proof_of_amount(&self, resource_address: FungibleProof, amount: Decimal) -> Proof;
```
0xOmarA, line 186: Ditto: seems to have completely changed the type here.

scrypto/src/component/stubs.rs
```diff
@@ -178,14 +179,14 @@ extern_blueprint_internal! {
...
         fn create_proof_of_non_fungibles(
             &self,
-            resource_address: ResourceAddress,
+            resource_address: NonFungibleProof,
```
0xOmarA, line 189: Ditto: seems to have completely changed the type here.

scrypto/src/component/stubs.rs
```diff
@@ -239,15 +240,16 @@ extern_blueprint_internal! {
         ) -> Global<MultiResourcePool>;
     },
     {
-        fn contribute(&mut self, buckets: Vec<Bucket>) -> (Bucket, Vec<Bucket>);
+        fn contribute(&mut self, buckets: Vec<FungibleBucket>)
+            -> (FungibleBucket, Vec<FungibleBucket>);
         fn redeem(&mut self, bucket: Bucket) -> Vec<Bucket>;
```
0xOmarA, line 245: I think redeem should be a `(FungibleBucket) -> Vec<FungibleBucket>` since the input is the pool units (fungible) and the output is fungible resources.

scrypto/src/component/stubs.rs
```diff
@@ -239,15 +240,16 @@ extern_blueprint_internal! {
...
-        fn protected_deposit(&mut self, bucket: Bucket);
+        fn protected_deposit(&mut self, bucket: FungibleBucket);
         fn protected_withdraw(
             &mut self,
             resource_address: ResourceAddress,
```
0xOmarA, line 249: Do we want to double down on the whole fungible and non-fungible bit and make this a `FungibleResourceManager`?

scrypto/src/component/stubs.rs
```diff
@@ -274,14 +276,14 @@ extern_blueprint_internal! {
         ) -> Global<OneResourcePool>;
     },
     {
-        fn contribute(&mut self, bucket: Bucket) -> Bucket;
+        fn contribute(&mut self, bucket: FungibleBucket) -> FungibleBucket;
         fn redeem(&mut self, bucket: Bucket) -> Bucket;
```
0xOmarA, line 280: Ditto on redeem.

scrypto/src/component/stubs.rs
```diff
@@ -305,15 +307,18 @@ extern_blueprint_internal! {
         ) -> Global<TwoResourcePool>;
     },
     {
-        fn contribute(&mut self, buckets: (Bucket, Bucket)) -> (Bucket, Option<Bucket>);
+        fn contribute(
+            &mut self,
+            buckets: (FungibleBucket, FungibleBucket),
+        ) -> (FungibleBucket, Option<FungibleBucket>);
         fn redeem(&mut self, bucket: Bucket) -> (Bucket, Bucket);
```
0xOmarA, line 314: Ditto on redeem.

scrypto/src/component/stubs.rs
```diff
@@ -382,10 +387,13 @@ extern_blueprint_internal! {
         fn quick_confirm_recovery_role_badge_withdraw_attempt(&mut self) -> Bucket;
         fn cancel_primary_role_badge_withdraw_attempt(&mut self);
         fn cancel_recovery_role_badge_withdraw_attempt(&mut self);
-        fn mint_recovery_badges(&mut self, non_fungible_local_ids: Vec<NonFungibleLocalId>) -> Bucket;
+        fn mint_recovery_badges(
```
0xOmarA, line 390: This should return a `NonFungibleBucket` and not a fungible bucket since the recovery badge is non-fungible.

scrypto/src/component/stubs.rs
```diff
@@ -150,18 +151,18 @@ extern_blueprint_internal! {
...
         fn lock_contingent_fee(&mut self, amount: Decimal);
         fn deposit(&mut self, bucket: Bucket);
         fn deposit_batch(&mut self, buckets: Vec<Bucket>);
         fn withdraw(&mut self, resource_address: ResourceAddress, amount: Decimal) -> Bucket;
         fn withdraw_non_fungibles(
             &mut self,
-            resource_address: ResourceAddress,
+            resource_address: NonFungibleBucket,
             ids: Vec<NonFungibleLocalId>,
         ) -> Bucket;
```
0xOmarA, line 167: I think that this should return `NonFungibleBucket`.

scrypto/src/component/stubs.rs
0xOmarA: It looks like this is an issue with how the function signatures were overriden in `update-bindgen.sh`.

scrypto/src/resource/auth_zone.rs
```diff
@@ -15,13 +15,13 @@ pub trait ScryptoAuthZone {
         &self,
         amount: A,
         resource_address: ResourceAddress,
-    ) -> Proof;
+    ) -> FungibleProof;
```
0xOmarA, line 18: Is this true? Is a proof created by amount always fungible? IIRC I don't think that this is correct.

lrubasze: > According to the sources one can create_proof_of_amount of non-fungibles if amount is None (see [compose_proof_by_amount](https://github.com/radixdlt/radixdlt-scrypto/blob/ea3d7f1f1ef56de3e03c44485f0f21efb113b82f/radix-engine/src/blueprints/resource/auth_zone/auth_zone_composition.rs#L78))

I was wrong, [compose_proof_by_amount](https://github.com/radixdlt/radixdlt-scrypto/blob/ea3d7f1f1ef56de3e03c44485f0f21efb113b82f/radix-engine/src/blueprints/resource/auth_zone/auth_zone_composition.rs#L78) is able to return non-fungible proof if amount is `Some`. Will revert this

scrypto/src/component/stubs.rs
lrubasze: Done. 
As agreed refactored other parts, where `ResourceAddress` appears.
Switched to `FungibleResourceManager`, `NonFungibleResourceManager` or `ResourceManager` depending on context.

scrypto/src/component/stubs.rs
lrubasze: Done

0xOmarA (approved):
Looks good!

scrypto/src/component/stubs.rs
```diff
@@ -150,51 +151,55 @@ extern_blueprint_internal! {
...
     },
     {
-        fn securify(&mut self) -> Bucket;
+        fn securify(&mut self) -> NonFungibleBucket;
         fn lock_fee(&mut self, amount: Decimal);
         fn lock_contingent_fee(&mut self, amount: Decimal);
         fn deposit(&mut self, bucket: Bucket);
         fn deposit_batch(&mut self, buckets: Vec<Bucket>);
-        fn withdraw(&mut self, resource_address: ResourceAddress, amount: Decimal) -> Bucket;
+        fn withdraw(&mut self, resource_address: ResourceManager, amount: Decimal) -> Bucket;
```
iamyulong, line 162: I was a bit shocked by the change, despite knowing that we're introducing ResourceManager, FungibleResourceManager and NonFungibleResourceManager.

Are we feeling comfortable with replacing `resource_address` with resource manager? Curious to learn how other people feel about this.

scrypto/src/component/stubs.rs
```diff
@@ -234,25 +239,26 @@ extern_blueprint_internal! {
...
-            resource_addresses: Vec<ResourceAddress>,
+            resource_addresses: Vec<FungibleResourceManager>,
             address_reservation: Option<GlobalAddressReservation>,
         ) -> Global<MultiResourcePool>;
     },
     {
-        fn contribute(&mut self, buckets: Vec<Bucket>) -> (Bucket, Vec<Bucket>);
-        fn redeem(&mut self, bucket: Bucket) -> Vec<Bucket>;
-        fn protected_deposit(&mut self, bucket: Bucket);
+        fn contribute(&mut self, buckets: Vec<FungibleBucket>)
```
iamyulong, line 247: I think there was some product requirement to support treating non-fungible in a fungible way. @0xOmarA what do you think?

scrypto/src/component/stubs.rs
0xOmarA: It's interesting. If we're saying that a fungible and non fungible resources are `FungibleResourceManager` and `NonFungibleResourceManager` then it makes sense that an undecided resource would be a `ResourceManager`. But I see your point on this being a bit of a big change.

0xOmarA: Perhaps this is one where traits can be useful? As an example, with some changes to the `extern_blueprint` macro I imagine that we can support `resource_address: impl Into<ResourceAddress>` which seems like a more correct argument type for this function.

scrypto/src/component/stubs.rs
0xOmarA: AFAIK pools of non-fungibles was not a product requirement and is actually something that the pools currently reject: https://github.com/radixdlt/radixdlt-scrypto/blob/7dbe606bac3d2661a47c30a43dafb5016074f130/radix-engine/src/blueprints/pool/v1/v1_0/one_resource_pool_blueprint.rs#L32-L34

scrypto/src/component/stubs.rs
iamyulong: I think in my mind there are two different things: addresses (identifiers) and stubs, similar to component address and `Global<T>`.  

The question is more about if we should use address or stubs for the interfaces.

Stubs would make Scrypto-to-Scrypto interaction easier, as there is no more conversion needed and everything is type checked.

On the other side, manifest developers/readers will need to map resource manager stubs to addresses.

iamyulong (approved):

scrypto/src/component/stubs.rs
lrubasze: Agreed to revert these changes (replacing `ResourceAddress` with `ResourceManager`) in this PR.
We will get back to this in separate PR if the team reaches consensus.

PR #1784, opened 2024-04-23
Added `DecimalInternal` type

mstrug-rdx:
## Summary
Added type alias to `Decimal` and `PreciseDecimal` which is needed by Radix Engine Toolkit to have access to `I192` and `I256` types methods in a generic way (from macro).
This is required for Maya integration to convert between `Decimal` and array of bytes.


0xOmarA (commented):
Looks good, naming wise, I prefer `InnerDecimal` (and same for precise decimal too) so that users do not get this type in their autocomplete as easily when trying to do decimals.

PR #1785, opened 2024-04-23
Access Controller Fee Vault.

0xOmarA:
## Summary

This PR adds an XRD vault to the state of access controllers and methods for depositing, withdrawing, and locking XRD from that vault for fees. Anyone is allowed to deposit into the vault. Only the primary role is allowed to withdraw from the vault and any three of the roles are allowed to lock XRD from the vault for fees.

## Details

An `AccessControllerV2` blueprint is added with this PR which has an updates state structure that contains an XRD whose primary usage is for locking fees for recoveries. The addition of this field pushed the state to v2. The update from the v1 state (that doesn't have the vault) to v2 (which has a field for an optional vault) is done lazily, any method call to the access controller updates the state after the protocol update. This means that it is possible to have some access controller on v1 while others are on v2 after the protocol update.

Alongside the above state changes, the following methods have been added to the access controller blueprint:

```rust
// Requires P, C, or R
pub fn lock_recovery_fee(&mut self, amount: Decimal);

// Requires P
pub fn withdraw_recovery_fee(&mut self, amount: Decimal) -> Bucket;

// Anyone can call
pub fn contribute_recovery_fee(&mut self, bucket: Bucket);
```

A `DepositRecoveryXrdEvent` is emitted when XRD is deposited into the vault via the `contribute_recovery_fee` method and a `WithdrawRecoveryXrdEvent` method is emitted when XRD is withdrawn from the vault via the `withdraw_recovery_fee` method.

## Testing

All of the testing done for this PR can be found in the `radix-engine-tests/tests/flash/access_controller.rs` file.

### For Internal Integrators

Be aware that this introduces a protocol update that updates the access controller blueprint from v1 to v2 by doing the following:

* Removes the old schema substate of the package and replacing it with a new one in a different key (different schema hash)
* Removes the old original code substate of the package and replaces it with a new one in a different key (different original code hash)
* Removes the old vm type substate of the package and replaces it with a new one in a different key (different original code hash)
* Updates the blueprint definition with the newly added methods, change in the code hash, change in the schema hash, and added events.
* Updates the blueprint auth configs with the newly added methods.

Additionally, the following changes were made that you should be aware of:

* The schema of the access controller state has changed. 
* It is possible to encounter access controllers with v1 state and others with v2 state after the protocol update.
* The code id of the access controller package has changed.
* The access controller blueprint now has three more methods.
* The set of "native" events has increased by 2.


radix-engine/assets/native_function_base_costs.csv
```diff
@@ -159,6 +159,9 @@
 0d906318c6318c6c4d63f8cc6318c6318cf7bf553d3ca51686318c6318c6,stop_timed_recovery,2001964
 0d906318c6318c6c4d63f8cc6318c6318cf7bf553d3ca51686318c6318c6,timed_confirm_recovery,3300795
 0d906318c6318c6c4d63f8cc6318c6318cf7bf553d3ca51686318c6318c6,unlock_primary_role,1820930
+0d906318c6318c6c4d63f8cc6318c6318cf7bf553d3ca51686318c6318c6,lock_recovery_fee,2407361
```
0xOmarA, line 162: MUST UPDATE BEFORE MERGING, awaiting on CI to finish.

.vscode/settings.json
```diff
@@ -1,20 +1,7 @@
 {
     "rust-analyzer.linkedProjects": [
-        // Root Workspace & CLIs.
         "Cargo.toml",
         "radix-clis/Cargo.toml",
-        // Various blueprints we have.
```
0xOmarA, line 6: These slowed down the IDE performance greatly for me. Every save took minutes to run and each check took a lot longer.

radix-engine/src/blueprints/access_controller/v2/state.rs
```diff
@@ -0,0 +1,103 @@
...
+            controlled_asset,
+            xrd_fee_vault: None,
+            timed_recovery_delay_in_minutes,
+            recovery_badge,
+            state,
+        }
+    }
+}
+
+declare_native_blueprint_state! {
```
0xOmarA, line 80: Must get #1755 in before we merge this PR.

.vscode/settings.json
dhedey: Ooooo is that why my rust analyzer has been stuck on a crash loop since merging in develop yesterday? :D

I've tried `cargo clean`, `rustup update`, changing my rust analyzer version... I'm glad removing this fixes it.

0xOmarA: Yup, I sadly think that this is what's causing it :/ Bit of a shame to be honest since it's nice to get diagnostics in your IDE for these blueprints especially when performing large refactors. Only alternative is the the `cargo_check.py` script that I added but it doesn't give us any diagnostics in the IDE, just command line.

radix-engine/assets/native_function_base_costs.csv
0xOmarA: Updated costing, safe to merge now.

radix-blueprint-schema-init/src/lib.rs
```diff
@@ -113,6 +113,15 @@ pub enum TypeRef<T> {
                // TODO: How to represent a structure containing a generic?
 }
 
+impl<T> TypeRef<T> {
+    pub fn into_static(self) -> T {
```
iamyulong, line 117: Maybe we return an `Option<T>` and do the unwrap at client side?

radix-engine/src/blueprints/native_schema.rs
```diff
@@ -22,9 +23,11 @@ lazy_static! {
...
-    pub static ref ACCESS_CONTROLLER_PACKAGE_DEFINITION: PackageDefinition =
-        AccessControllerNativePackage::definition();
-    pub static ref POOL_PACKAGE_DEFINITION_V1_0: PackageDefinition =
+    pub static ref ACCESS_CONTROLLER_PACKAGE_DEFINITION_V1: PackageDefinition =
+        AccessControllerV1NativePackage::definition();
+    pub static ref ACCESS_CONTROLLER_PACKAGE_DEFINITION_V2: PackageDefinition =
+        AccessControllerV2NativePackage::definition();
+    pub static ref POOL_PACKAGE_DEFINITION_V2: PackageDefinition =
         PoolNativePackage::definition(PoolV1MinorVersion::Zero);
     pub static ref POOL_PACKAGE_DEFINITION_V1_1: PackageDefinition =
```
iamyulong, line 32: I think we'd like consistent names for versions. Will follow up on Slack.

scrypto/src/component/stubs.rs
```diff
@@ -383,6 +383,9 @@ extern_blueprint_internal! {
         fn cancel_primary_role_badge_withdraw_attempt(&mut self);
         fn cancel_recovery_role_badge_withdraw_attempt(&mut self);
         fn mint_recovery_badges(&mut self, non_fungible_local_ids: Vec<NonFungibleLocalId>) -> Bucket;
+        fn lock_recovery_fee(&mut self, amount: Decimal);
+        fn withdraw_recovery_fee(&mut self, amount: Decimal) -> Bucket;
+        fn contribute_recovery_fee(&mut self, bucket: Bucket);
```
iamyulong, line 388: @lrubasze - there are new methods to consider regarding Fungible/NonFungibleBucket. :)

radix-engine/src/blueprints/access_controller/v2/state.rs
0xOmarA: It's been merged.

scrypto/src/component/stubs.rs
0xOmarA: Perhaps it would be good to make these methods accept a `FungibleBucket` now instead of needing a protocol update to change them later :)

radix-blueprint-schema-init/src/lib.rs
0xOmarA: Yeah seems reasonable to me, will do that. I kinda want to add another method that panics, perhaps:

* `into_static`: returns an option.
* `assert_static`: returns T and panics if it's not static.

radix-engine/src/blueprints/native_schema.rs
0xOmarA: Yup my bad, I think that this is a mistake, will correct.

radix-engine-interface/src/blueprints/access_controller/invocations.rs
```diff
@@ -264,3 +264,51 @@ pub struct AccessControllerMintRecoveryBadgesInput {
 }
 
 pub type AccessControllerMintRecoveryBadgesOutput = Bucket;
+
+// region:bottlenose
```
0xOmarA, line 268: Fun fact: RA on VSCode supports c-stye `#region`s which is quite cool!

scrypto/src/component/stubs.rs
iamyulong: For now, rubin is only updating the language layer.

lrubasze: > @lrubasze - there are new methods to consider regarding Fungible/NonFungibleBucket. :)

so happy to see it 😁

radix-engine/src/blueprints/access_controller/v1/package.rs
```diff
@@ -0,0 +1,345 @@
...
+use radix_engine_interface::blueprints::package::*;
+use sbor::rust::prelude::*;
+
+pub struct AccessControllerV1NativePackage;
+
+impl AccessControllerV1NativePackage {
+    pub fn definition() -> PackageDefinition {
+        let mut aggregator = TypeAggregator::<ScryptoCustomTypeKind>::new();
+
+        // Using the feature set from v1. They should be the same between v1 and v1.1.
```
iamyulong, line 15: Minor: need to update comment to `v1.0` and `v2.0`.

radix-engine/src/blueprints/access_controller/v2/package.rs
```diff
@@ -0,0 +1,402 @@
...
+use radix_engine_interface::blueprints::package::*;
+use sbor::rust::prelude::*;
+
+pub struct AccessControllerV2NativePackage;
+
+impl AccessControllerV2NativePackage {
+    pub fn definition() -> PackageDefinition {
+        let mut aggregator = TypeAggregator::<ScryptoCustomTypeKind>::new();
+
+        // Using the feature set from v1. They should be the same between v1 and v1.1.
```
iamyulong, line 15: Ditto

radix-substate-store-queries/src/typed_substate_layout.rs
```diff
@@ -428,7 +428,7 @@ pub enum TypedMainModuleSubstateValue {
     NonFungibleVault(NonFungibleVaultTypedSubstateValue),
     ConsensusManager(ConsensusManagerTypedSubstateValue),
     Validator(ValidatorTypedSubstateValue),
-    AccessController(AccessControllerTypedSubstateValue),
+    AccessController(AccessControllerV2TypedSubstateValue),
```
iamyulong, line 431: We need both `AccessControllerV1` and `AccessControllerV2`?

radix-substate-store-queries/src/typed_substate_layout.rs
```diff
@@ -153,7 +153,7 @@ pub enum TypedMainModuleSubstateKey {
     NonFungibleVault(NonFungibleVaultTypedSubstateKey),
     ConsensusManager(ConsensusManagerTypedSubstateKey),
     ValidatorField(ValidatorTypedSubstateKey),
-    AccessController(AccessControllerTypedSubstateKey),
+    AccessController(AccessControllerV2TypedSubstateKey),
```
iamyulong, line 156: If v1 and v2 are the same, we should keep `AccessControllerTypedSubstateKey`? or we should have `AccessControllerV1` and `AccessControllerV2`?

radix-substate-store-queries/src/typed_substate_layout.rs
0xOmarA: I do not think that v1 is needed here since both v1 and v2 substates can be represented as `AccessControllerV2TypedSubstateValue`.

iamyulong: Can we decode the V1 sbor payload into `AccessControllerV2TypedSubstateValue`?

0xOmarA: Yup should be possible since the main change is that we added a new version (enum variant) so we should be able to decode v1 as v2, I've just added a test for that: `access_controller_state_v1_can_be_decoded_as_v2`.

Also, this is implicitly tested in `test_all_scenario_commit_receipts_should_have_substate_changes_which_can_be_typed` since our scenarios use AccessControllerV1 and their state can be typed.

radix-substate-store-queries/src/typed_substate_layout.rs
0xOmarA: I think that this is similar to the other discussion, I believe that blueprint updates that extend the state (adding fields as an example) we can use latest versions.

iamyulong (approved):

PR #1787, opened 2024-04-25
Faucet blueprint changed magic number to constant

mstrug-rdx:
## Summary
Changed Faucet blueprint `free` method magic number specifying count of XRD returned to a constant. We need to use that constant in `radix-engine-toolkit` project in `trusted-worktop` feature. Also updated test which uses this magic number in assertions.

This PR re-applies PR: https://github.com/radixdlt/radixdlt-scrypto/pull/1706 which now is missing from `develop`.


iamyulong (approved):
LGTM

PR #1791, opened 2024-05-01
Access Controller v2 Scenario

0xOmarA:
## Summary

A small PR that adds an AccessController v2 scenario.

### DIFF
--- a/radix-transaction-scenarios/src/scenarios/access_controller_v2.rs
+++ b/radix-transaction-scenarios/src/scenarios/access_controller_v2.rs
@@ -1,0 +1,154 @@
+use crate::internal_prelude::*;
+use radix_engine::updates::{ProtocolUpdate, ProtocolVersion};
+use radix_engine_interface::blueprints::access_controller::*;
+use radix_engine_interface::*;
+
+pub struct AccessControllerV2ScenarioConfig {
+    primary_role_private_key: PrivateKey,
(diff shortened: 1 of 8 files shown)


radix-transaction-scenarios/src/scenarios/access_controller_v2.rs
```diff
@@ -0,0 +1,154 @@
...
+pub struct AccessControllerV2ScenarioState {
+    pub(crate) access_controller_component_address: State<ComponentAddress>,
+}
+
+impl Default for AccessControllerV2ScenarioConfig {
+    fn default() -> Self {
+        Self {
+            primary_role_private_key: Ed25519PrivateKey::from_u64(1).unwrap().into(),
+            recovery_role_private_key: Ed25519PrivateKey::from_u64(2).unwrap().into(),
+            confirmation_role_private_key: Ed25519PrivateKey::from_u64(3).unwrap().into(),
```
iamyulong, line 38: I remember that we have some methods defining such keys. Maybe we can reuse that?

iamyulong (approved):
LGTM

radix-transaction-scenarios/src/scenarios/access_controller_v2.rs
0xOmarA: Sounds good, will use that.

PR #1794, opened 2024-05-02
Architecture: Kernel Boot Cleanup

talekhinezh:
* Moved KernelBoot read to beginning of boot process
* Moved boot reference type checking into system


radix-engine-tests/tests/kernel/kernel.rs
```diff
@@ -87,6 +84,17 @@ impl KernelCallbackObject for TestCallbackObject {
         TestReceipt
     }
 
+    fn boot_ref_type(&mut self, _node_id: &NodeId, _value: &IndexedScryptoValue) -> Result<StableReferenceType, BootloadingError> {
+        Ok(StableReferenceType::Global)
```
iamyulong, line 88: Ditto, surprised that this isn't breaking tests.

radix-engine-monkey-tests/tests/fuzz_kernel.rs
```diff
@@ -107,6 +107,21 @@ impl KernelCallbackObject for TestCallbackObject {
         TestReceipt
     }
 
+    fn boot_ref_type(
+        &mut self,
+        _node_id: &NodeId,
+        _value: &IndexedScryptoValue,
+    ) -> Result<StableReferenceType, BootloadingError> {
+        Ok(StableReferenceType::Global)
```
iamyulong, line 115: Surprised that this isn't breaking fuzz tests.

radix-engine/src/kernel/kernel.rs
```diff
@@ -98,18 +155,30 @@ impl<'h, M: KernelCallbackObject, S: SubstateDatabase> BootLoader<'h, M, S> {
...
         // Create Kernel
         let mut kernel = {
-            let mut kernel = Kernel::new(&mut self.track, &mut self.id_allocator, &mut callback);
-
             // Check references
-            let engine_references = kernel
-                .check_references(executable.references())
+            let costing_enabled = kernel_boot.is_ref_check_costing_enabled();
+            let engine_references = self
+                .check_references(&mut callback, executable.references(), costing_enabled)
```
iamyulong, line 177: I do like the idea of loading the kernel boot substate upfront, but I'm not sure about the value proposition of moving reference checks to bootloading. It couples a lot of logic, e.g. the direct reference recognition to bootloading. Also, we're directly manipulating the call frame of kernel afterwards anyway.

iamyulong (approved):
Very nice!

PR #1809, opened 2024-05-15
Use two different models for costing params

0xOmarA:


radix-transactions/src/model/executable.rs
```diff
@@ -62,14 +62,14 @@ impl From<(BlueprintId, GlobalAddress)> for PreAllocatedAddress {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor, ManifestSbor)]
-pub struct TransactionCostingParameters {
+pub struct TransactionCostingParametersV2 {
```
jakrawcz-rdx, line 65: These structs do not use our conventional "versioning infra" (the macros, the wrappers, the traits, etc), so maybe we shouldn't use the `...V2` naming here.
Maybe something like "input costing params" vs "output costing params", or "receipt costing params"?
(unless this was already discussed and you are sure we want V1/V2)

radix-transactions/src/model/executable.rs
```diff
@@ -79,6 +79,31 @@ impl Default for TransactionCostingParameters {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, ScryptoSbor, ManifestSbor)]
+pub struct TransactionCostingParametersV1 {
```
jakrawcz-rdx, line 83: Can we add a note (regular comment) explaining why it is introduced?

(also, maybe in the PR description, some link to the investigation)

jakrawcz-rdx (commented):
This should fix the issue indeed 👍

iamyulong (approved):

radix-transactions/src/model/executable.rs
0xOmarA: Sounds good to me, will switch the naming to indicate that our versioning infra is not used here.

PR #1818, opened 2024-05-20
Update SystemFieldKind to be backward compatible

iamyulong:


radix-engine/src/transaction/system_structure.rs
```diff
@@ -34,17 +34,14 @@ pub struct SystemFieldStructure {
 //       for the toolkit which parses it opaquely at the moment when doing transaction preview
 //       in the wallet. This change doesn't apply because BootLoader substates are only edited
 //       during transaction preview.
+//
+// Node depends on this structure, so append only.
 #[derive(Debug, Clone, ScryptoSbor, PartialEq, Eq)]
 pub enum SystemFieldKind {
     TypeInfo,
-    BootLoader(BootLoaderFieldKind),
```
0xOmarA, line 40: The enum used to be defined as:

```rs
#[derive(Debug, Clone, ScryptoSbor, PartialEq, Eq)]
pub enum SystemFieldKind {
    TypeInfo,
    BootLoader,
}
```

If we want to backward compat, shouldn't we revert to that?

iamyulong: Yep, I did check this on `anemone`. 

Node only requires encoding and decoding, so it's backward compatible. :-)

radix-engine/src/transaction/system_structure.rs
```diff
@@ -34,17 +34,14 @@ pub struct SystemFieldStructure {
...
-}
-
-#[derive(Clone, Debug, ScryptoSbor, PartialEq, Eq)]
-pub enum BootLoaderFieldKind {
```
0xOmarA, line 44: The `v1.1.2` tag defines the enum as:

```rs
#[repr(u8)]
#[derive(Debug, Copy, Clone, Sbor, PartialEq, Eq, Hash, PartialOrd, Ord, FromRepr)]
pub enum BootLoaderField {
    Vm = 2u8,
}
```

Is the `2u8` here used as the SBOR Enum discriminator? If it is, then we need to do the same here to be backward compatible

radix-engine/src/transaction/system_structure.rs
0xOmarA: So you're saying that `SystemFieldKind::BootLoader` was never actually persisted and therefore removing that variant is not a breaking change?

radix-engine/src/transaction/system_structure.rs
iamyulong: It's for both. Let me double check this.

0xOmarA: Sounds good, if it's for both then I imagine that the previous ordering was more correct because VmBoot would've had a discriminator of 2?

radix-engine/src/transaction/system_structure.rs
iamyulong: No, I essentially renamed `BootLoader` to `VMBoot` (as now we have more boot loader variants). 

I also appended two new variants.

Both should not affect the encoding and decoding of `BootLoaderField::BootLoader` (now called `BootLoader::VMBoot`).

radix-engine/src/transaction/system_structure.rs
0xOmarA: Expanded macro is:

```rs
impl<X: ::sbor::CustomValueKind> ::sbor::SborEnum<X> for BootLoaderField {
    fn get_discriminator(&self) -> u8 {
        match self {
            Self::Vm => 0u8,
        }
    }
    fn get_length(&self) -> usize {
        match self {
            Self::Vm => 0,
        }
    }
}
```

Looks like the `2u8` there is not used in SBOR and only matters for the `FromRepr`.

iamyulong: Omar mentioned that `repr(u8)` is not used by SBOR unless the structure is annotated with `#[sbor(use_repr_discriminators)]`.

I'm removing the `Sbor` derive as it may introduce different, unexpected discriminators.

0xOmarA (approved):
Discussed in a call, looks good to me

PR #1822, opened 2024-05-23
Add support to just load the build artifacts in `scrypto test` 

lrubasze:
## Summary
Refactor `scrypto new-package` template to just load the build artifacts (WASM and package definition), which should be already built.
Also applied the same change to `hello-world`

## Details
There is no need to compile package in `scrypto` unit test. `scrypto` command builds it for us internally, when `scrypto test` or `scrypto coverage` command is invoked.
Moreover compile it in each unit tests, which are running in parallel, might lead to some indeterministic results, because the build artifacts are being shared between unit tests.

NOTE 1!
This PR does not address the problem for the tests, which are already implemented. It only changes the template and provides the reference how such test should look like.
We could possibly fix it internally by adding a build script, but:
- I think it would make a `scrypto` package look too complicated for a common dApp developer. 
- build script does not share the build progress (not sure, maybe it can be reconfigured). If you add a long build time of a `scrypto` package it coud give a feeling that something hung up.

NOTE 2!
With this change we are also loosing some flexibility when using `cargo` command instead of `scrypto`.
eg. such combination will not work properly 
```
# cargo build just '*.wasm' and will not build the '*.rpd' file
cargo build --target wasm32-unknown-unknown --release
# below expects both '*.wasm' and '*.rpd' to be already built
cargo test --release
# cargo nextest run --release
```
I don't think it is a big deal, because a canonical way to work with **Scrypto** should  be using `scrypto` command.

### For dApp Developers
dApp developers shall refactor their tests by themselves if they are affected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,13 +94,13 @@ jobs:
         uses: ./.github/actions/setup-env
       - name: Build and test hello-world
         run: |
-          cargo build --target wasm32-unknown-unknown --release
-          cargo nextest run --release
+          scrypto build
+          scrypto test
         working-directory: examples/hello-world
       - name: Build and test everything
         run: |
-          cargo build --target wasm32-unknown-unknown --release
-          cargo nextest run --release
+          scrypto build
+          scrypto test
         working-directory: examples/everything
       - name: Build no-std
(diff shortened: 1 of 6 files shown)


lrubasze:
Replaced with #1832

PR #1831, opened 2024-06-03
Documentation/add more book content

talekhinezh:
* Expand on System Layer Documentation


radix-engine/book/src/architecture/application/package/README.md
```diff
@@ -0,0 +1,20 @@
...
+`An Object is of some Blueprint type.`
+
+`A Blueprint is part of a Package.`
+
+`A Package is an Object.`
+
+This circular definition creates the notion of the Package Blueprint and the
+Package Package (similar to Class.class in java). A Package Blueprint is the
+blueprint type of a Package object and Package Package is the package which
+contains the Package Blueprint.
```
iamyulong, line 20: Worth adding something like below?

Due to the circular dependency, the first object (`Package` package object) is created without following standard object creation process. It's directly flashed into the database.

radix-engine/book/src/architecture/application/type_system/README.md
```diff
@@ -6,6 +6,6 @@ the Blueprint. This includes:
 * Function Input/Output 
 * Events
 
-The Type Checking System supports generics. 
+The Schema Checking System supports generics.
```
iamyulong, line 9: Slightly prefer to the previous word. Schema checking gives me the impression that we're checking the validity of schema (which is part of SBOR schema terminology)

iamyulong (approved):

radix-engine/book/src/architecture/application/package/README.md
talekhinezh: Yup, this is a good add.

PR #1844, opened 2024-06-28
Implement `as_fungible()` and `as_non_fungible()` methods only Generic types.

lrubasze:
## Summary

Implement `as_fungible()` and `as_non_fungible()` methods only Generic types.

## Details
In https://github.com/radixdlt/radixdlt-scrypto/pull/1782 these methods were implemented for all three types:
- generic
- fungible
- non-fungible

It allowed to create such combinations as below:
```
let vault = FungibleVault::with_bucket(some_bucket).as_non_fungible();  
```
It is not possible to get `NonFungibleVault` from `FungibleVault`.

## Testing
Tests were updated accordingly.


iamyulong (approved):
LGTM

PR #1846, opened 2024-06-28
Fixed `validate_resource_specifier()` function

mstrug-rdx:
## Summary
`validate_resource_specifier()` function returns wrong result for proper input data: non fungible resource with at least two ids.

## Testing
Added two test to verify `validate_resource_specifier()` function, on current develop 1st test: `verify_validate_resource_specifier_method()` is failing, but should succeed.

### DIFF
--- a/scrypto-test/src/sdk/resource/bucket_factory.rs
+++ b/scrypto-test/src/sdk/resource/bucket_factory.rs
@@ -173,7 +173,7 @@ impl BucketFactory {
                     let mut iter = non_fungibles
                         .keys()
                         .map(|id| id.id_type())
-                        .collect::<HashSet<NonFungibleIdType>>()
+                        .collect::<IndexSet<NonFungibleIdType>>()
                         .into_iter();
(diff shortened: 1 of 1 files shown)


scrypto-test/src/sdk/resource/bucket_factory.rs
```diff
@@ -170,7 +170,11 @@ impl BucketFactory {
                 // 1. All of one type.
                 // 2. This one type is the type of the non-fungible local ids.
                 let id_type = {
-                    let mut iter = non_fungibles.keys().map(|id| id.id_type());
+                    let mut iter = non_fungibles
+                        .keys()
+                        .map(|id| id.id_type())
+                        .collect::<HashSet<NonFungibleIdType>>()
+                        .into_iter();
```
iamyulong, line 177: It doesn't matter in this case, but can we use IndexSet for consistency in the code base?

iamyulong (approved):

PR #1848, opened 2024-07-02
feature: Add `BasicSborAssertion` and `ScryptoSborAssertion` macros

dhedey:
## Summary
These macros output tests which check for a fixed type schema, or a backwards compatible type schema.

See `sbor-assert.rs` in the tests to see how the macros are used, and see `transaction_receipt.rs` and `system_structure.rs` for how these are used in practice.

What's rather cool is that Rust Analyzer / the VS Code extension spots the test and magically offers a "Run Test" button, which makes this much nicer / more magic than I was expecting :).

Screenshot 2024-07-03 at 18 17 57

## Details

You can start with `#[sbor_assert(generate("FILE:<relative_file_path.txt>")]` which will output a test which panics and gives you the current schema.

You can then do this as new versions get added:
```rust
#[sbor_assert(backwards_compatible(
    version1 = "FILE:mytype_schema_v1.txt",
    version2 = "FILE:mytype_schema_v2.txt",
))]
```

I've written rust docs for e.g. `ScryptoSborAssertion`, and the error messages aim to help you get the syntax right. Just start with adding `#[derive(ScryptoSborAssertion)]` and the error messages should guide you.

I also had to add a little extension to the `define_single_versioned!` / `define_multi_versioned!` macros to be able to support adding attributes to only the outer / inner types. It's not the nicest, but it's functional.

## Testing
See `sbor-assert.rs`. Failing tests aren't included - because they're a bit painful to do.

But they've been manually tested, and most of the underlying logic has already been tested as part of the `schema_comparisons.rs` tests.

## Update Recommendations

We can now start using these macros across scrypto and node.


iamyulong (approved):
LGTM

PR #1850, opened 2024-07-17
Update wasmi to 0.35.0

lrubasze:
## Summary
This PR switches `wasmi` from forked version based on `0.23.0` to original latest `0.35.0`.

Fork version was including some modifications (clonable `Store`) that allow caching preinstantiated WASM module, which was simply cloned upon next time it is used (when WasmEngine::instantiate() is called).
When using original `wasmi-0.35.0` we are caching compiled WASM module, which must be instantiated each time it is used (when `WasmEngine::instantiate()` is called). 
According to our measurements instantiation is almost as efficient as cloning (50us vs 40us).

For better performance we use following WASM setup: _LazyTranslation_ + _no WASM validation_.
For details check `wasmi-0.32.0` [release notes](https://github.com/wasmi-labs/wasmi/releases/tag/v0.32.0).

Benchmark indicate:
- significant WASM instantiation speedup - `~80%`
- executution speedup - `~50-60%` for decimal benches

Note!
In order to pass CI following additional changes applied: 
- Merged branch with #1852  (`wasmi-0.35.0` requires minimum Rust version `1.77`) 
- Merged branch with https://github.com/radixdlt/radixdlt-scrypto/pull/1851


radix-engine/src/vm/wasm/wasmi.rs
```diff
@@ -833,42 +797,44 @@ macro_rules! linker_define {
...
-        let module =
-            Module::new(&engine, code).map_err(WasmiInstantiationError::ValidationError)?;
-
-        let instance = Self::host_funcs_set(&module, &mut store)
-            .map_err(WasmiInstantiationError::PreInstantiationError)?
-            .ensure_no_start(store.as_context_mut())
-            .map_err(WasmiInstantiationError::InstantiationError)?;
+        let mut config = wasmi::Config::default();
+
+        // In order to speed compilation we deliberately
```
lrubasze, line 809: Note from the author.

Maybe it would make sense to setup this by extending `WasmiEngineOptions` with respective fields.
And then we could add more setup methods next to `DefaultWasmEngine::default()`.
eg. 
- `DefaultWasmEngine::with_no_validation()` - which would become `DefaultWasmEngine::default()`
- `DefaultWasmEngine::with_validation()`

Just a loose thought. 

We would just need to remember about `wasmer` and find adequate settings for it.

radix-engine/src/vm/wasm/prepare.rs
```diff
@@ -1172,9 +1172,12 @@ impl WasmModule {
         // Because the offset can be an `InitExpr` that requires evaluation against an WASM instance,
         // we're using the `wasmi` logic as a shortcut.
         let code = self.module.bytes();
-        WasmiModule::new(&code[..]).map_err(|e| PrepareError::NotInstantiatable {
-            reason: format!("{:?}", e),
-        })?;
+        WasmiModule::new(&code[..])
+            .map_err(|_| PrepareError::NotCompilable)?
```
lrubasze, line 1176: Note from the author.

Was thinking we could extend `PrepareError::NotCompilable` with a `reason` field similarly to `PrepareError::NotInstantiatable`.
I am just unsure if it safe to modify `PrepareError` enum.

iamyulong (approved):

radix-engine/src/vm/wasm/wasmi.rs
lrubasze: `wasmer` support has just been removed :)

PR #1853, opened 2024-07-19
Farewell `wasmer`

lrubasze:
This PR removes `wasmer` support from Radix Engine.

`wasmer` haven't been really used and it become only a maintenance burden.

This change does not break backward compatibility.
It is also transparent for scrypto developers and integrators.


iamyulong (approved):

0xOmarA:
Some of the benchmarks I'm currently doing show that wasmer is actually faster in some instances than WASMI. I expect that we might bring wasmer back at some point.

PR #1854, opened 2024-07-21
feature: Add `#[sbor(flatten)]` for enum variants

dhedey:
## Summary
Implements the ability to `#[sbor(flatten)]` enum variants from REP-84. Specifically, it can be applied to enum variants with a single unskipped field which is an SBOR tuple. This effectively merges the field's value into the parent enum variant. 

This is useful for enabling a "enum variant as type" refactor where you create a struct for each enum variant, and when you don't wish to add another layer to the SBOR structure of the enum. The "enum variant as type" paradigm can be useful for wanting to handle a specific variant as itself, then roll it up into its parent.

For example, transaction payload encoding effectively works this way, but was done manually by duplicating the field structures until this PR.

A related functionality, `SborEnumVariantFor` from REP-84, has been implemented in #1855. 

## Details
This PR also includes some much needed refactoring to the FieldsData part of the SBOR derive macros.

## Testing
See `enum.rs` under `sbor-tests`.

## Update Recommendations
Anyone using `VersionedTransaction` will have to insert the struct types when matching, e.g. `VersionedTransactionPayload::NotarizedTransactionV1 { signed_intent, .. }` now becomes `VersionedTransactionPayload::NotarizedTransactionV1(NotarizedTransactionV1 { signed_intent })`.


radix-transactions/src/model/versioned.rs
```diff
@@ -36,30 +36,14 @@ const V1_FLASH_TRANSACTION: u8 = 8;
...
-        notary_signature: NotarySignatureV1,
-    },
-    #[sbor(discriminator(V1_SYSTEM_TRANSACTION))]
-    SystemTransactionV1 {
-        instructions: InstructionsV1,
-        blobs: BlobsV1,
-        pre_allocated_addresses: Vec<PreAllocatedAddress>,
-        hash_for_execution: Hash,
-    },
+    #[sbor(flatten, discriminator(V1_INTENT))]
```
lrubasze, line 39: I guess the above TODO is no longer needed.

sbor-tests/tests/enum.rs
```diff
@@ -75,8 +75,52 @@ pub enum Mixed {
     I = 0b11011,
 }
 
+#[derive(Debug, PartialEq, Eq, Sbor)]
+enum FlattenEnum {
+    #[sbor(flatten)]
+    A {
+        #[sbor(skip)]
+        skipped: u32,
```
lrubasze, line 83: Minor.
For more clarity I would use different types for `skipped` and `y` to clearly indicate, which one is skipped in `FlattenedEnum::A`.

radix-transactions/src/model/versioned.rs
dhedey: Good catch! :)

lrubasze (approved):
Nice work!

PR #1857, opened 2024-07-24
refactor: Prefer parameter bounds over where clause

dhedey:
## Summary
We previously had lots of inconsistent syntax for how we capture the API parameters.

As a move to tidy things up, and prepare the way for changes to Error tracking, I have:
* Introduced a `SystemApiError` trait.
  * This replaces various bounds like `Debug` and `ScryptoDecode + ScryptoCategorize` in various places. 
  * Only `RuntimeError` currently implements it. So it looks like maybe it's not even needed(?) but it will prove useful for future refactors I think, so I left it in for now.
* Moved type bounds which can be moved from a `where` expression to the generic parameters list. This has a few advantages in my eyes:
  * Where clauses are very noisy and bulky - this distracts from the intent of the code when reading a lot of code integrating with the system API. Putting the bounds by the generic parameter definitions gets them out of the way.
  * It moves the bound closer to the definition for clarity. You have fewer jumps to understand the generic parameters.
  * It makes the bounds more like parameter bounds
  * I prefer saving them for when bounds _have_ to go in a where clause - that way, it highlights that this bound is unusual, say, for example, the `AccessControllerV2Substate: Transition<I>` bound in `access_controller/v1/mod.rs`
* Re-ordered the type parameters in some places to start with `Y`
* Re-ordered the type bounds on `Y` to start with the `SystemApi` where possible.
* In some places where we expect generics to always be inferred, (typically `Fn` parameters and typed inputs), I used `impl X` to reduce the number of trait bounds.

I would actually also support inlining `api: &mut Y` to be `api: &mut impl SystemApi<...>`, to get rid of generic parameters completely (!), but didn't want to do this without approval, because that would be a lot of refactoring. 

## Testing
Just a syntax refactor, tests pass

## Update Recommendations
A few parameters re-orders, but I don't expect this to affect any users.

### DIFF
--- a/examples/everything/Cargo.lock
+++ b/examples/everything/Cargo.lock
@@ -3,6 +3,19 @@
 version = 3
 
 [[package]]
+name = "ahash"
+version = "0.8.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
+dependencies = [
+ "cfg-if",
+ "getrandom 0.2.12",
+ "once_cell",
+ "version_check",
+ "zerocopy",
+]
+
+[[package]]
 name = "aliasable"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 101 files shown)


jakrawcz-rdx (approved):
I remember when I saw the `<T: ...`> vs `where T: ...` syntax alternatives for the first time, I thought "ohhh yessss there is gonna be so many opinions on that; Good job, Rust language designers!".

I am a fan of "only one way to do one thing". If there are two ways, I am a fan of having a convention which prefers one way.

Normally, I'd vote for `where T: ...`: it can capture some constraints that `<T: ...>` cannot, so it clearly is the only candidate for "one way". However, it also has really cumbersome syntax :(

I figured that so far our rule was "let's do `<T: ...>` when it is kinda short, but `where T: ...` when it is kinda long". Due to "kindism", it is not a good rule (I agree that "kindism" can be replaced by e.g. "an arbitrary limit of 27 characters", but it does not make it much better).

So I actually like the rule of "we do `<T: ...>`, except when it is not possible: `where T: ...` indicates a rare, crazy-complex constraint" 👍 If we commit to making this our convention _(i.e. fixing things in a drive-by, wherever we spot it wrong from this point on)_ then this PR has my 🟢 .

PR #1867, opened 2024-08-06
Abort early when deferred costs and loan are covered

iamyulong:
## Summary

Aborts early when deferred costs and loan are covered, to help reduce overhead of pre-validating transactions.

### DIFF
--- a/radix-engine-interface/src/api/costing_api.rs
+++ b/radix-engine-interface/src/api/costing_api.rs
@@ -7,7 +7,7 @@ pub trait ClientCostingApi<E> {
     fn start_lock_fee(&mut self, amount: Decimal) -> Result<bool, E>;
 
     /// Add cost units to the reserve. This should never fail.
-    fn lock_fee(&mut self, locked_fee: LiquidFungibleResource, contingent: bool);
+    fn lock_fee(&mut self, locked_fee: LiquidFungibleResource, contingent: bool) -> Result<(), E>;
 
     fn consume_cost_units(&mut self, costing_entry: ClientCostingEntry) -> Result<(), E>;
 
--- a/radix-engine/src/system/system_modules/costing/fee_reserve.rs
+++ b/radix-engine/src/system/system_modules/costing/fee_reserve.rs
@@ -572,8 +572,7 @@ impl ExecutionFeeReserve for SystemLoanFeeReserve {
         self.locked_fees
             .push((vault_id, fee.take_all(), contingent));
 
-        if self.abort_when_loan_repaid && self.is_deferred_costs_and_loan_covered().unwrap_or(false)
-        {
+        if self.abort_when_loan_repaid && self.is_deferred_costs_and_loan_covered()? {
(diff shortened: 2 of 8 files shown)


radix-engine/src/system/system_modules/costing/fee_reserve.rs
```diff
@@ -534,6 +571,15 @@ impl ExecutionFeeReserve for SystemLoanFeeReserve {
         // Move resource
         self.locked_fees
             .push((vault_id, fee.take_all(), contingent));
+
+        if self.abort_when_loan_repaid && self.is_deferred_costs_and_loan_covered().unwrap_or(false)
```
lrubasze, line 575: Minor..
Since `is_deferred_costs_and_loan_covered()` returns `FeeReserveError` then maybe no need to map error returned by `is_deferred_costs_and_loan_covered()` to 
`FeeReserveError::Abort(..)`?

lrubasze (approved):
Looks good.
Maybe it would be worth adding a dedicated test for `AbortReason::ConfiguredAbortTriggeredOnFeeLoanRepayment`

radix-engine/src/system/system_modules/costing/fee_reserve.rs
iamyulong: Good point!

0xOmarA (approved):

radix-engine-interface/src/api/costing_api.rs
```diff
@@ -7,7 +7,7 @@ pub trait ClientCostingApi<E> {
     fn start_lock_fee(&mut self, amount: Decimal) -> Result<bool, E>;
 
     /// Add cost units to the reserve. This should never fail.
```
dhedey, line 9: Can we update the comment?

radix-engine/src/system/system_modules/costing/fee_reserve.rs
```diff
@@ -386,6 +391,33 @@ impl SystemLoanFeeReserve {
...
+            .checked_mul(*size)
+            .ok_or(FeeReserveError::Overflow)?;
+            deferred_storage_cost = deferred_storage_cost
+                .checked_add(cost)
+                .ok_or(FeeReserveError::Overflow)?;
+        }
+
+        let sum = deferred_execution_cost
+            .checked_add(deferred_storage_cost)
+            .and_then(|x| x.checked_add(self.xrd_owed))
```
dhedey, line 415: I initially thought this was wrong, but I think this is right actually :D

Can we add a comment here explaining that `xrd_owed` starts as `system_loan_in_xrd` but changes to `0` after the loan is repaid in full :+1:

dhedey (changes requested):
Small request to fix the comment, but otherwise I think this looks good

dhedey (approved):
Sorry, that was meant to be a `Comment` (not to block the review), not a `Request Changes`. Will mark as `Approve` to undo the `Request Changes`

PR #1871, opened 2024-08-10
refactor: Moves all transaction models into `radix-transactions`

dhedey:
## Summary

* Adds `ScryptoDescribe` to the transaction models so that I can assert that the `NotarizedTransactionV1` schema is fixed
  * Technically speaking I should create a `ManifestSchema` and use that; but this is good enough for now, to protect us from accidentally breaking transaction models! Nearly all types map 1:1 between these schemas. We can look to adjust in future.
* Pulls all the transaction models from the node into `radix-transactions`:
  * Because of the existence of the `Flash` transaction, this required moving the `StateUpdates` models into `radix-common` from `radix-engine / track` - which I was initially a bit hesitant about, but I think actually works quite nicely. (`radix-common` is already aware of some of the nuances of nodes and things).
  * Allows us to delete the duplication in `radix-clis`
  * Enables the `VersionedTransactionPayload` to be complete
* Fixes new compiler warnings in rust 1.80.0

## Testing
N/A - just a refactor, existing tests cover it.

## Update Recommendations
* Some types to do with database updates have been moved into the `radix_common::prelude::*`.
* Various transaction models can be deleted from the node as they're now in Scrypto (although a few methods interfacing these transactions with the node will need to be kept on the node side)


radix-common/src/state/state_updates.rs
dhedey, line 1: These were shifted-and-lifted.

radix-engine/src/transaction/transaction_executor.rs
```diff
@@ -264,25 +263,6 @@ impl ExecutionConfig {
     }
 }
 
-pub struct SubstateBootStore<'a, S: SubstateDatabase> {
```
dhedey, line 267: This was unused, so I deleted it as part of the fixing of compiler warnings from 1.80

radix-substate-store-impls/src/state_tree/types.rs
```diff
@@ -1169,7 +1169,7 @@ impl<K: Clone> From<LeafNode<K>> for Node<K> {
 
 impl<P: Clone> Node<P> {
     /// Creates the [`Internal`](Node::Internal) variant.
-    #[cfg(any(test, feature = "fuzzing"))]
+    #[cfg(test)]
```
dhedey, line 1172: The `fuzzing` feature mention was a remanent from when we imported the code originally and as of 1.80 the compiler catches that this isn't a supported feature in the crate.

radix-transactions/src/model/v1/dynamic_addresses.rs
```diff
@@ -0,0 +1,544 @@
+use crate::internal_prelude::*;
```
dhedey, line 1: These were shifted-and-lifted out of the very long `instruction` file

radix-transactions/src/model/v1/flash_transaction.rs
```diff
@@ -0,0 +1,65 @@
+use super::*;
```
dhedey, line 1: This was shifted-and-lifted from node.

radix-transactions/src/model/v1/round_update_transaction.rs
```diff
@@ -0,0 +1,134 @@
+use super::*;
```
dhedey, line 1: This was shifted-and-lifted from node.

radix-clis/src/rtmd/mod.rs
```diff
@@ -98,6 +98,7 @@ pub fn run() -> Result<(), String> {
                             blobs,
                             ..
                         }) => (instructions.0, blobs.blobs),
+                        _ => return Err("Transaction type not currently supported".to_string()),
```
jakrawcz-rdx, line 101: (minor)

```suggestion
                        other_type => return Err(format!("Transaction type {:?} not currently supported", other_type)),
```

radix-substate-store-impls/src/state_tree/tree_store.rs
```diff
@@ -177,7 +170,7 @@ pub trait WriteableTreeStore {
 /// avoid this potential performance implication by having an explicit way of associating an
 /// "unchanged" Substate with its new tree leaf.
 pub enum AssociatedSubstateValue<'v> {
-    Upserted(&'v DbSubstateValue),
+    Upserted(&'v [u8]),
```
jakrawcz-rdx, line 173: I quite like the well-named types instead of arbitrary bytes. Even if it's just an alias, it documents things a bit.

(minor; you can say "but here it literally means arbitrary bytes" and I will be ok)

radix-substate-store-interface/src/interface.rs
```diff
@@ -67,51 +113,63 @@ impl PartitionDatabaseUpdates {
...
-pub enum DatabaseUpdate {
-    Set(DbSubstateValue),
-    Delete,
-}
+impl CreateDatabaseUpdates for BatchPartitionStateUpdate {
+    type DatabaseUpdates = PartitionDatabaseUpdates;
 
-impl DatabaseUpdate {
-    pub fn as_change(&self) -> SubstateChange<'_> {
+    /// Uses the given [`DatabaseKeyMapper`] to express self using database-level key encoding.
```
jakrawcz-rdx, line 154: (nitpick) you can now drop this rustdoc from impls

radix-common/src/state/state_updates.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn as_ref(&self) -> DatabaseUpdateRef<'_> {
+        match self {
+            DatabaseUpdate::Set(update) => DatabaseUpdateRef::Set(update),
+            DatabaseUpdate::Delete => DatabaseUpdateRef::Delete,
+        }
+    }
+}
+
+/// A 1:1 counterpart of [`DatabaseUpdate`], but operating on references.
+pub enum DatabaseUpdateRef<'v> {
```
jakrawcz-rdx, line 201: nice; if we are touching this, should we go for:
```
pub enum DatabaseUpdateWithValueOrRef<V> {
...
}

pub type DatabaseUpdate = DatabaseUpdateWithValueOrRef<DbSubstateValue>;

pub type DatabaseUpdateRef<'v> = DatabaseUpdateWithValueOrRef<&'v [u8]>;
```

(please disregard if it does not deduplicate any "common" handling code and just makes things harder ;p)

jakrawcz-rdx (approved):
I only have minor remarks, please address whatever makes sense and 🟢 !

radix-substate-store-impls/src/state_tree/tree_store.rs
dhedey: Yeah broadly I agree with you! :)

However, here `DbSubstateValue = Vec<u8>` - I wanted to change it to the more natural `&[u8]` rather than using `&Vec<u8>`.

I don't really like the type alias pattern for these sorts of types because they don't protect anything; I prefer new types.

Arguably that's what `AssociatedSubstateValue` is already :) - however, imo really it should wrap a `RawScryptoValue<'a>` - that's essentially what this PR tries to do: https://github.com/radixdlt/radixdlt-scrypto/pull/1860

radix-common/src/state/state_updates.rs
dhedey: Once we have https://github.com/radixdlt/radixdlt-scrypto/pull/1860 I think it's natural to formulate it as:
```
pub enum DatabaseUpdate<'v> {
    Set(RawScryptoPayload<'v>),
    Delete,
}
pub type OwnedDatabaseUpdate = DatabaseUpdate<'static>;
```
(Happy to debate naming) So perhaps we can wait for that?

jakrawcz-rdx: Yes, We Can! ®️

PR #1878, opened 2024-08-15
test: Add extra tests for sending/receiving references to owned nodes

dhedey:


iamyulong (approved):

talekhinezh (approved):
For future reference, it might be better to write these tests as native components rather than scrypto components as it would make it easier to debug and express the test which Scrypto may prevent from expressing easily (e.g. See `system_call_method.rs` test).

PR #1880, opened 2024-08-20
Tweak/protocol update builder tweaks

dhedey:
## Summary
Essentially implements the extensions from here: https://github.com/radixdlt/radixdlt-scrypto/pull/1788

Notably:
* Adds tests / dumps of the protocol update transactions to help catch any divergence of genesis or protocol updates.
* Supports bootstrap in the protocol builder so we don't have to faff with a bootstrapper - and removes the Bootstrapper
* Adds a `VmModules` and `VmInitialize` trait to make it less painful to create a `VmInit`
* Changes the default genesis which is used in tests to be less broken - so e.g. a round update transaction will work directly out of the box
* Adds some extra metadata sanity checks to scenarios to avoid scenarios breaking testnets like stokenet

Possible future extensions:
* Have some kind of static/thread-static LazyCell / EagerCell of `HashOfProtocolSettings` => `PostBootstrapClonableInMemoryDb` in LedgerSimulator, to allow tests to run faster by skipping protocol initialization. Would need to investigate (A) how to do this sanely, and (B) if the perf savings are worth it.

## Testing
Refactor - tests should still pass.

### For dApp Developers

* Replace `CustomGenesis` with `BabylonSettings`
* Note that the `protocol builder` also applies bootstrapping by default


radix-engine/src/vm/vm.rs
```diff
@@ -49,6 +52,68 @@ impl VmApi for VmBoot {
     }
 }
 
+/// The canonical implementation is [`VmModules`].
+pub trait VmInitialize {
```
talekhinezh, line 56: Not sure of the use of this trait? I don't think the we get anything by this abstraction?

radix-engine-tests/tests/application/stake_reconciliation.rs
```diff
@@ -19,7 +19,10 @@ fn test_stake_reconciliation() {
     // Arrange
     let pub_key = Secp256k1PrivateKey::from_u64(1u64).unwrap().public_key();
     let mut ledger = LedgerSimulatorBuilder::new()
-        .with_protocol_version(ProtocolVersion::Babylon)
+        .with_custom_protocol(|builder| builder
+            .with_babylon(BabylonSettings::test_minimal())
+            .only_babylon()
```
talekhinezh, line 24: `with_babylon` + `only_babylon` is confusing to me. Based on what I can tell, this is giving configuration to the babylon protocol update and then specifying to only go up to babylon protocol update. Is that right?

radix-engine-tests/tests/db/substate_database_overlay.rs
```diff
@@ -442,54 +464,89 @@ fn create_database_contents_hash<D: SubstateDatabase + ListableSubstateDatabase>
     accumulator_hash
 }
 
-/// Runs the scenarios on an [`InMemorySubstateDatabase`] and a [`SingleSubstateDatabaseOverlay`] wrapping
+trait DatabaseComparisonScenarioCheck {
+    fn check(
+        &mut self,
```
talekhinezh, line 469: Is self needed?

radix-engine-tests/tests/db/substate_database_overlay.rs
```diff
@@ -442,54 +464,89 @@ fn create_database_contents_hash<D: SubstateDatabase + ListableSubstateDatabase>
     accumulator_hash
 }
 
-/// Runs the scenarios on an [`InMemorySubstateDatabase`] and a [`SingleSubstateDatabaseOverlay`] wrapping
+trait DatabaseComparisonScenarioCheck {
```
talekhinezh, line 467: Minor: Is it worth it using this trait as opposed to the ability to use closures?

talekhinezh (approved):

radix-engine/src/vm/vm.rs
dhedey: So the main use is that it encapsulates the generics (as associated types) to reduce the noise/pollution across the code base.

We can just take a parameter of impl VmInitialize instead of needing a generic parametrization over the native vm extension and scrypto vm.

But as that’s not clear, I’ll add a comment about it as a markup on a later PR 😄

radix-engine-tests/tests/application/stake_reconciliation.rs
dhedey: That’s correct!

But you’re right, it is confusing! Perhaps calling the settings method configure_babylon would be better?

radix-engine-tests/tests/db/substate_database_overlay.rs
dhedey: It is mostly just a way of avoiding needing to copy a massive 4 line fn closure definition across 4 lines.

This is only used for these two tests, tbh I’m tempted to rewrite them to duplicate this logic and it’d probably be clearer!

Thoughts welcome

radix-engine-tests/tests/db/substate_database_overlay.rs
dhedey: Possibly not - this is just a test helper for two tests (see other comment)

Might just rewrite the tests if you’d think it were clearer?

PR #1888, opened 2024-08-29
feat: Create and integrate `InstructionV2`

dhedey:
## Summary

* Creates separate version-independent models for each instruction in `manifest_instruction.rs`, in order to allow them to live under different versioned instructions
* Created some traits (e.g. `ManifestInstruction`, exposing `decompile()` and `side_effect()`) to make it easier to work with different transaction versions, and integrated these into most places where we match on the instructions (in a manner which can be version-independent)
* Added an `InstructionV2` enum along with the `YieldToParent`, `YieldToChild` and `AuthenticateParent` instructions.
* Moved `#[sbor(flatten)]` to live as an attribute on a field, instead of an attribute on the enum variant. In future, this allows a generalization where any tuple could be flattened into a parent tuple or enum variant. (Although this requires being able to read a length and encode / decode the fields separately, so will need some thinking regarding how best to achieve! e.g. a wrapped encoder/decoder to swallow the length; or new traits explicitly for field encoding/decoding)

Next PRs:
* Transaction builder for NotarizedTransactionV2, including putting together signed partial transactions.
* Introduction of validator to turn NotarizedTransactionV2 into ValidatedNotarizedTransactionV2 and that into ExecutableV2 and lots of tests
* Lots of tests of the merkle hashing of the transaction payload

## Testing
Mostly a refactor - existing tests should pass - including e.g. the fixed genesis tests that I added.

Further tests to come in later PRs when the model is in place and unblocks downstream development against interfaces.

### For dApp Developers
The `deposit_batch(account)` method on `ManifestBuilder` is now either `deposit_batch(account, ManifestExpression::EntireWorktop)` or `deposit_entire_worktop()`, in line with the `try_deposit_batch` methods (looks like it got missed when these were generalized).


radix-engine-tests/tests/blueprints/account_locker.rs
```diff
@@ -1418,32 +1418,28 @@ fn state_of_the_account_locker_can_be_reconciled_from_events_alone() {
             )
         });
 
+    fn set_disallowed_preference(builder: ManifestBuilder, account: ComponentAddress, resource: ResourceAddress) -> ManifestBuilder {
```
talekhinezh, line 1421: Could also turn this into an extension trait if you want to retain the builder pattern. For example,
```rust
trait ManifestBuilderExt {
  fn set_disallowed_preference(self, account: ComponentAddress, resource: ResourceAddress) -> Self;
}

impl ManifestBuilderExt for ManifestBuilder {
  fn set_disallowed_preference(self, account: ComponentAddress, resource: ResourceAddress) -> Self {
      self.call_method(
        account,
        ACCOUNT_SET_RESOURCE_PREFERENCE_IDENT,
        AccountSetResourcePreferenceInput {
            resource_address: resource,
            resource_preference: ResourcePreference::Disallowed,
        }
    )
  }
}
```

radix-transactions/src/builder/manifest_builder.rs
```diff
@@ -95,13 +96,21 @@ pub struct NewSymbols {
...
+pub type ManifestV1Builder = ManifestBuilder<InstructionV1>;
+
+impl ManifestBuilder<InstructionV1> {
+    /// Starts a new manifest v1 builder.
     pub fn new() -> Self {
+        Self::new_v1()
+    }
+
+    /// Call as `ManifestBuilder::new_v1()`
+    pub fn new_v1() -> Self {
```
talekhinezh, line 108: Not sure if `new_v1` makes sense since "v1" is already captured by the `InstructionV1` in the type?

talekhinezh (approved):

radix-engine-tests/tests/blueprints/account_locker.rs
dhedey: Yup! I almost did it at the time, but thought for one test it was a bit much.

But I’ll do it because it’s fun 🫡

radix-transactions/src/builder/manifest_builder.rs
dhedey: It’s so that you can do ‘ManifestBuilder::new_v1()’ rather than needing to fully specify ‘ManifestBuilder::<InstructionV1>::X()’ and it also gives parity to new_v2.

Ideally new wouldn’t exist, or wouldn’t exist just on v1, but it needs to map to v1 for backwards compatibility to save potentially breaking lots of dapp code.

radix-engine-tests/tests/blueprints/account_locker.rs
```diff
@@ -1418,32 +1418,28 @@ fn state_of_the_account_locker_can_be_reconciled_from_events_alone() {
...
+            account,
+            ACCOUNT_SET_RESOURCE_PREFERENCE_IDENT,
+            AccountSetResourcePreferenceInput {
+                resource_address: resource,
+                resource_preference: ResourcePreference::Disallowed,
+            }
+        )
+    }
+
+    let builder = ManifestBuilder::new();
```
0xOmarA, line 1432: This is test code so this is just a recommendation, but I really like the use for things like this and I think that we can use `fold` to make this quite neat. Something like the following achieves the same effect:

```rust

let manifest = [
    (user_account1, fungible_resource1),
    (user_account1, non_fungible_resource3),
    (user_account2, fungible_resource2),
    (user_account2, non_fungible_resource2),
    (user_account3, fungible_resource3),
    (user_account3, non_fungible_resource1),
]
.iter()
.fold(ManifestBuilder::new(), |builder, (account, resource)| {
    builder.call_method(
        account,
        ACCOUNT_SET_RESOURCE_PREFERENCE_IDENT,
        AccountSetResourcePreferenceInput {
            resource_address: resource,
            resource_preference: ResourcePreference::Disallowed,
        },
    )
})
.build();
```

0xOmarA: I have found myself using the above fold pattern with the manifest builder very very often lately and it very quickly grew on me.

dhedey: This is a nice idea in principle. That said, I have a semi-irrational dislike for `fold` / `reduce` - basically these guys capture my feelings well: https://www.youtube.com/watch?v=qaGjS7-qWzg

Typically using a loop (or explicit unrolling in this case) is simpler / easier to follow in my opinion. :)

PR #1889, opened 2024-08-30
Update rust to 1.80.1

lrubasze:
- use Rust 1.80.1 in CI
- use Rust 1.80.1 in Docker image
- update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,18 +47,24 @@ Documentation: https://docs.radixdlt.com/docs/scrypto-1
      ```bash
      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
      ```
-2. Enable `cargo` in the current shell:
+2. [Optional] Set Rust version to 1.79, which is currently recommended for Radix Engine
+   ```bash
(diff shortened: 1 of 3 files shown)


dhedey (approved):

README.md
```diff
@@ -47,18 +47,24 @@ Documentation: https://docs.radixdlt.com/docs/scrypto-1
      ```bash
      curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
      ```
-2. Enable `cargo` in the current shell:
+2. [Optional] Set Rust version to 1.79, which is currently recommended for Radix Engine
```
dhedey, line 50: This shouldn't be optional. Perhaps we should say something like:

> Version 1.80+ cannot currently be used to compile scrypto, as it uses WASM features not supported by the current (bottlenose) radix engine. 

Also I think this will need merging with step 1. which currently inaccurately says:
> Install Rust - this requires Rust 1.70+ (if rust is already installed, upgrade with `rustup update`)

iamyulong (approved):

PR #1892, opened 2024-08-31
Feature/allocate fungible address

gguuttss:
## Summary
```
Added the ability to get an GlobalAddressReservation for a FungibleResource.
```

## Testing
```
Have tested functionality to generate an address reservation, and mint a fungible that's its own Owner. Works as intended, but I might be missing something.
```


gguuttss:
Hi, just a pull request to let you know I'd like this feature to be added as I'm using it locally. Perhaps adding it has some consequences I'm unaware of, so feel free to ignore ;)

dhedey:
Thanks @gguuttss - looks like an omission, so including it looks good to me!

Could you move this PR to be a commit from/against `develop`? We should be able to get this into cuttlefish.

iamyulong (approved):
LGTM

PR #1893, opened 2024-08-31
Radix Engine Toolkit transaction receipt bottlenose backporting

0xOmarA:
## Summary

Backporting #1887 to bottlenose


Dockerfile
```diff
@@ -1,16 +1,16 @@
-# Below version includes rust 1.76.0
-ARG RUST_IMAGE_VERSION=@sha256:2a6147ed8a879621f27d2c7a966868de438801084a276498b3a937e326ef7f65
+# Below version includes rust 1.80.1
```
dhedey, line 1: Personally, I'm happy with this, but if we're not doing a scrypto release, @iamyulong  wasn't sure we wanted to do this.

Honestly, given it works OK with 1.80, I'm happy to just update docs and image to suggest 1.80 - then if we want to, we can always consider publishing the new builder image if anyone complains. 👍

radix-engine-tests/tests/system/toolkit_receipt.rs
```diff
@@ -0,0 +1,738 @@
...
+            )
+            .expect("Failed during runtime -> serializable conversion");
+
+        // Serialize the serializable receipt to JSON through serde_json.
+        let serialized_receipt = serde_json::to_string_pretty(&serializable_toolkit_receipt)
+            .expect("Serializing through serde_json failed");
+
+        // Create a file for this transaction receipt.
+        let directory_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("assets")
```
dhedey, line 79: Just a thought - but these aren't really `assets` - perhaps we could call them `generated`?

radix-engine-tests/tests/system/toolkit_receipt.rs
```diff
@@ -0,0 +1,738 @@
...
+use radix_engine::system::system_modules::execution_trace::{ResourceSpecifier, WorktopChange};
+use radix_engine_interface::prelude::MetadataValue;
+use radix_engine_tests::*;
+use radix_engine_toolkit::receipt::*;
+use radix_transaction_scenarios::executor::*;
+use scrypto::*;
+use scrypto_test::prelude::*;
+
+#[test]
+fn test_toolkit_receipt_roundtrip_property() {
```
dhedey, line 14: I wonder if these should be in the `radix-engine-toolkit` crate?

Although happy for them to be here if you'd prefer.

If we move them to the `radix-engine-toolkit`, we'll need to update `ci.yml` to include such a test.

dhedey (approved):

radix-engine-tests/tests/system/toolkit_receipt.rs
dhedey: (maybe not worth doing else we'll get weird merge conflicts upon merging with develop)

radix-engine-tests/tests/system/toolkit_receipt.rs
dhedey: (maybe not worth doing else we'll get weird merge conflicts upon merging with develop)

0xOmarA: I can see them being in either crate. It seems somewhat nice to me to have them in the `radix-engine-tests` crate, but they also don't quite fit in with the engine abstractions. I have a slight preference for keeping them here but could be argued either way. 

If we plan on moving them then I'd say let's move them on `develop` and not on this branch in order to avoid merge conflicts.

radix-engine-tests/tests/system/toolkit_receipt.rs
0xOmarA: Ah, generated is a much better name and you're right, in this case the receipts are not exactly assets. If you feel like we should make this change then we could make it in `develop`.

Dockerfile
0xOmarA: I personally did this change since it was the only way to get CI to pass. Happy to keep it or to revert these changes. I don't see a problem with keeping it if we won't publish a new scrypto version and a new scrypto builder version.

PR #1902, opened 2024-09-03
Architecture/refactor tx processor

talekhinezh:
Separate TransactionProcessor logic into 3 parts in preparation for V2 transactions:
* Blueprint Interface
* Transaction Execution
* Instruction Execution


radix-engine/src/blueprints/transaction_processor/instructions.rs
```diff
@@ -0,0 +1,545 @@
...
+use radix_native_sdk::runtime::LocalAuthZone;
+use radix_rust::prelude::*;
+use radix_transactions::data::transform;
+use radix_transactions::model::manifest_instruction::*;
+use radix_transactions::model::InstructionV1;
+
+pub trait TxnInstruction {
+    fn execute<Y: SystemApi<RuntimeError> + KernelNodeApi + KernelSubstateApi<L>, L: Default>(
+        self,
+        worktop: &mut Worktop,
```
dhedey, line 22: This is a lot of noisy bounds to update / change as we refactor things.

Can I suggest we hide these in a trait? e.g.

```rust
trait ManifestExecutor {
    type Api: SystemApi<RuntimeError> + KernelNodeApi + KernelSubstateApi<Self::SubstateDefault>;
    type SubstateDefault: Default;

    /// Gives access to all the mutable references at once, to avoid lifetime issues.
    fn unpack(&mut self) -> (&mut Worktop, &mut TxnProcessorObjects, &mut Self::Api);

    /// Just access the worktop.
    /// You may need to instead do `unpack` if you need the worktop at the same time as other things.
    fn worktop(&mut self) -> &mut Worktop { self.unpack().0 }

    /// Just access the objects.
    /// You may need to instead do `unpack` if you need the objects at the same time as other things.
    fn objects(&mut self) -> &mut TxnProcessorObjects { self.unpack().1 }

    /// Just access the api.
    /// You may need to instead do `unpack` if you need the api at the same time as other things.
    fn api(&mut self) -> &mut Self::Api { self.unpack().2 }
}
```

And then you can have some type representing the execution kernel for the different processors, or for a temporarily workaround, you could just do:
```rust
impl<'a, Y: SystemApi<RuntimeError> + KernelNodeApi + KernelSubstateApi<L>, L: Default> ManifestExecutor for (&'a mut Worktop, &'a mut TxnProcessorObjects, &'a mut Y) {
    type Api: SystemApi<RuntimeError> + KernelNodeApi + KernelSubstateApi<Self::SubstateDefault> = Y;
    type SubstateDefault: Default = L;

    fn unpack(&mut self) -> (&mut Worktop, &mut TxnProcessorObjects, &mut Self::Api) {
        (&mut self.0, &mut self.1, &mut self.2 )
    }
}
```

Then `TxnInstruction` can look a lot simpler / less noisy:
```rust
pub trait TxnInstruction {
    fn execute(self, executor: impl ManifestExecutor) -> Result<InstructionOutput, RuntimeError>;
}
```

dhedey (approved):
Nice!

I have a quick suggestion to adapt `TxnInstruction` to be less noisy / specific, otherwise looks good :)

radix-engine/src/blueprints/transaction_processor/instructions.rs
talekhinezh: Hrm, could certainly wrap all of these around a single type to keep the interface clean though I'm not sure that exposing trait to access these things would give much benefit as opposed to just accessing fields directly. Will test out different methods in the next PR.

dhedey: Well a concrete type would still need awkward generics.

You need a trait to make use of associated types to encapsulate the generic proliferation.

These instructions shouldn’t care that L: Default. I guess having done lots of cleanup PRs recently I’ve seen the pain all these generics are causing 😅

PR #1906, opened 2024-09-04
fix: Fix format for radix engine tests

dhedey:
## Summary

With Yulong's approval, removes the automod to ensure that `radix-engine-tests` get formatted. (I think I may have broken it accidentally when rehashing the ./format.sh script, because it wasn't clear to me that format didn't work with automod).

Also runs format on all the tests, where this was missed.


0xOmarA:
Just my POV: I'm happy that automod is removed since it didn't interact well with my IDE and Rust Analyzer. Adding a new module to tests required an RA restart when automod was in place which slowed things down.

PR #1908, opened 2024-09-06
Feature/subintent validation part 1

dhedey:
## Summary
A few parts:
* Introduce `ManifestNamedAddress` newtype for alignment with the others.
* Updated `ManifestInstruction` to have a more detailed `effect()` which covers what it does.
* Re-ordered `IntentCoreV2` to be easier to stream
* Polymorphized Manifests, and have them store extra information:
  * All manifests now store known naming, if available
  * SystemTransactionManifestV1 now stores preallocated addresses
  * The TransactionManifestV2 stores child intent hashes
  * There will be a SubintentManifestV2 which also stores child intent hashes
* Added a `StaticManifestInterpreter` which accepts a visitor for extra analysis
* Added an `ASSERT_WORKTOP_IS_EMPTY` instruction

## Testing

Have a long list of tests to write on the confluence doc. Will do these when the features are in to enable writing the tests to be easier.

## Update Recommendations
N/A


radix-transactions/src/manifest/static_manifest_interpreter.rs
```diff
@@ -0,0 +1,760 @@
...
+
+    fn handle_new_address_reservation<V: ManifestInterpretationVisitor>(
+        &mut self,
+        visitor: &mut V,
+        package_address: &PackageAddress,
+        blueprint_name: &str,
+        preallocated_address: Option<&GlobalAddress>,
+    ) -> ControlFlow<V::Error<'a>, ManifestAddressReservation> {
+        let new_address_reservation =
+            ManifestAddressReservation(self.address_reservation_state.len() as u32);
```
0xOmarA, line 436: I think that the logic is currently identical to that of the `ManifestIdAllocator`, so no issues with it being different logic or anything, but why not make use of it here? It seems like it would be better suited for this.

radix-transactions/src/manifest/static_manifest_interpreter.rs
```diff
@@ -0,0 +1,760 @@
...
+        source_amount: BucketSourceAmount<'a>,
+    ) -> ControlFlow<V::Error<'a>> {
+        let new_bucket = ManifestBucket(self.bucket_state.len() as u32);
+        self.bucket_state.push(BucketState {
+            name: self
+                .manifest
+                .get_known_object_names_ref()
+                .known_bucket_name(new_bucket),
+            created_at: self.location,
+            proof_locks: 0,
```
0xOmarA, line 315: This is nice!

radix-transactions/src/manifest/static_manifest_interpreter.rs
```diff
@@ -0,0 +1,760 @@
...
+            source_amount,
+        });
+        visitor.on_new_bucket(new_bucket, resource_address, source_amount)
+    }
+
+    fn get_existing_bucket<V: ManifestInterpretationVisitor>(
+        &mut self,
+        bucket: ManifestBucket,
+    ) -> ControlFlow<V::Error<'a>, &mut BucketState<'a>> {
+        match self.bucket_state.get_mut(bucket.0 as usize) {
```
0xOmarA, line 326: It's quite nice that we're doing lookups here based on the index.

radix-transactions/src/manifest/static_manifest_interpreter.rs
dhedey: Hah well in my head initially I wanted this to replace `ManifestIdAllocator`, but in reality I think id allocator has its niche :).

I'd be comfortable introducing a `ManifestIdAllocator`, but I think on balance I prefer the current code.
By storing in a Vec this implementation requires that the ids be consecutive, so I slightly prefered making it explicit / self-contained in this impl.

radix-common/src/data/manifest/model/manifest_address.rs
```diff
@@ -18,9 +18,30 @@ pub enum ManifestAddress {
     /// TODO: prevent direct construction, as in `NonFungibleLocalId`
     Static(NodeId),
     /// Named address, global only at the moment.
-    Named(u32),
+    Named(ManifestNamedAddress),
```
talekhinezh, line 21: Minor: rather than "named", I think `Ref` would be a better name

talekhinezh (approved):

radix-engine/src/utils/native_blueprint_call_validator.rs
```diff
@@ -15,69 +14,79 @@ use radix_engine_interface::blueprints::resource::*;
 use radix_engine_interface::object_modules::metadata::*;
 use radix_engine_interface::object_modules::role_assignment::*;
 use radix_engine_interface::object_modules::royalty::*;
+use radix_transactions::manifest::*;
 use radix_transactions::prelude::*;
 
-pub fn validate_call_arguments_to_native_components(
-    instructions: &[impl ManifestInstruction],
+pub fn validate_call_arguments_to_native_components_any(
+    manifest: &AnyTransactionManifest,
```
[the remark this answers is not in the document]
dhedey: Ahh with a later PR, it could take `&[impl InstructionSet]` - happy to change it, but I think this is also OK as is.
A manifest allows you to give better error messages, as you can e.g. resolve names for buckets.

radix-common/src/data/manifest/model/manifest_address.rs
dhedey: Hmm, I can see this. But then arguably also `ManifestBucket` => `ManifestBucketRef` etc. So I'm not fully sold.

Happy to do some more brainstorming though :). This also applies to `ManifestNamedIntent`

PR #1910, opened 2024-09-08
Feature/Cuttlefish Transaction Processing

talekhinezh:
* Add "Versioned System Logic" to system which versions portions of system logic given a protocol version
* Add `SystemBoot::V2` which includes a `VersionedSystemLogic`
* Add `SystemBoot` update to Cuttlefish protocol update
* Unified `ExecutableIntentV1` and `ExecutableIntentV2` into `ExecutableIntent`
* Add the ability for the System Layer to update call frame data
* Remove use of `current_depth()` api for absolute value checks and replaced with less fragile checks
* Some modification to transaction processor costing for v2 execution


radix-engine/src/updates/cuttlefish.rs
```diff
@@ -62,9 +75,51 @@ impl ProtocolUpdateBatchGenerator for CuttlefishBatchGenerator {
...
+    if let UpdateSetting::Enabled(_settings) = &system_logic_update {
+        transactions.push(ProtocolUpdateTransactionDetails::flash(
+            "cuttlefish-protocol-system-logic-updates",
+            generate_system_logic_updates(store),
+        ));
+    }
     ProtocolUpdateBatch { transactions }
 }
+
+fn generate_system_logic_updates<S: SubstateDatabase + ?Sized>(db: &S) -> StateUpdates {
```
iamyulong, line 93: Maybe rename to `generate_system_logic_v2_updates`?

radix-transactions/src/model/execution/executable_transaction.rs
```diff
@@ -76,20 +62,41 @@ impl ExecutableTransaction {
...
+                intent
+                    .references
+                    .insert(proof.resource_address().clone().into());
+            }
+            for resource in &intent.auth_zone_init.simulate_every_proof_under_resources {
+                intent.references.insert(resource.clone().into());
+            }
         }
+
+        if let Some(root) = intents.get_mut(0) {
```
iamyulong, line 87: Any reason why we restrict address allocation to intent 0?

talekhinezh: I'll add a comment

iamyulong (approved):

scrypto-test/src/ledger_simulator/ledger_simulator.rs
```diff
@@ -1136,7 +1136,7 @@ impl<E: NativeVmExtension, D: TestDatabase> LedgerSimulator<E, D> {
     {
         let nonce = self.next_transaction_nonce();
         self.execute_transaction(
-            TestTransaction::new_from_nonce(manifest, nonce)
+            TestTransaction::new_v1_from_nonce(manifest, nonce)
```
0xOmarA, line 1139: Are there plans of changing this part of the code to use v2 transactions and v2 instructions for the execution instead of v1?

scrypto-test/src/environment/builder.rs
```diff
@@ -246,6 +246,7 @@ where
                 };
 
                 System {
+                    versioned_system_logic: VersionedSystemLogic::V1,
```
0xOmarA, line 249: I might understand this as I review, but I assume that in here it doesn't matter if we do system v1 or v2 since the test environment doesn't deal with instructions or transactions at all so having it just be set to v1 or v2 should not matter. Is there any other changes in the system logic that could be useful to the transaction environment that aren't just support for new transactions?

radix-engine/src/system/system_callback.rs
```diff
@@ -61,6 +64,158 @@ pub type SystemBootSubstate = SystemBoot;
...
+
+                    output
+                }
+            }
+        };
+
+        Ok(output)
+    }
+
+    pub fn should_consume_cost_units<Y: SystemBasedKernelApi>(&self, api: &mut Y) -> bool {
```
dhedey, line 206: It's not all cost units... Perhaps we should give it a clearer name, e.g. `should_bill_client_costing`?

radix-engine/src/system/system_callback.rs
```diff
@@ -61,6 +64,158 @@ pub type SystemBootSubstate = SystemBoot;
...
+                    let virtual_non_fungibles =
+                        intent.auth_zone_init.initial_non_fungible_id_proofs.clone();
+                    let auth_zone = AuthModule::create_auth_zone(
+                        &mut system_service,
+                        None,
+                        virtual_resources,
+                        virtual_non_fungibles,
+                    )?;
+
+                    api.kernel_set_call_frame_data(Actor::Function(FunctionActor {
```
dhedey, line 153: This doesn't seem to set it for each thread? It just keeps overriding it. Maybe I misread?

radix-engine/src/system/system_callback.rs
```diff
@@ -61,6 +64,158 @@ pub type SystemBootSubstate = SystemBoot;
...
+                        ident: TRANSACTION_PROCESSOR_RUN_IDENT.to_string(),
+                        auth_zone,
+                    }))?;
+                }
+
+                {
+                    let mut system_service = SystemService::new(api);
+                    let intent = intents.get(0).unwrap();
+
+                    let txn_processor = TxnProcessor::<InstructionV1>::init(
```
dhedey, line 167: Looks like we're not actually creating call frames here, and just use the root, is that right?

I've added test cases to the list to check it uses the TransactionProcessor global caller for back-compat.

radix-engine/src/system/system_callback.rs
```diff
@@ -61,6 +64,158 @@ pub type SystemBootSubstate = SystemBoot;
...
+                        intent.encoded_instructions.clone(),
+                        global_address_reservations.clone(),
+                        intent.blobs.clone(),
+                        MAX_TOTAL_BLOB_SIZE_PER_INVOCATION,
+                        &mut system_service,
+                    )?;
+                    let output = txn_processor.execute(&mut system_service)?;
+
+                    let owned_nodes = api.kernel_get_owned_nodes()?;
```
dhedey, line 176: A lot of this code looks common to setting up / disposing of a call frame - perhaps it should be commonised?

radix-engine/src/system/system_callback.rs
```diff
@@ -61,6 +64,158 @@ pub type SystemBootSubstate = SystemBoot;
...
+            }
+        };
+
+        Ok(output)
+    }
+
+    pub fn should_consume_cost_units<Y: SystemBasedKernelApi>(&self, api: &mut Y) -> bool {
+        match self {
+            VersionedSystemLogic::V1 => {
+                // Skip client-side costing requested by TransactionProcessor
```
dhedey, line 209: Out of interest, do you know why we skip this particular costing for the transaction processor?

Maybe it's absorbed by the flat transaction cost?

radix-engine/src/system/system_modules/costing/costing_module.rs
```diff
@@ -380,10 +380,8 @@ impl PrivilegedSystemModule for CostingModule {
         api: &mut impl SystemBasedKernelApi,
         invocation: &KernelInvocation<Actor>,
     ) -> Result<(), RuntimeError> {
-        let depth = api.kernel_get_current_depth();
-
         // Skip invocation costing for transaction processor
-        if depth == 0 {
+        if api.kernel_get_system_state().current_call_frame.is_root() {
```
dhedey, line 384: MAJOR: I don't think this is right now that we've changed the definition of "root" - I think we need to put this into versioned logic.

Previous, we didn't bill for the Root > Processor call. Now we don't bill for the Processor > XX calls - this seems wrong?

I think the above comment also needs changing.

dhedey: In fact, I imagine this is why this PR changed all the costings. I imagine we can revert them all if we fix this.

radix-engine/src/system/system_modules/costing/costing_module.rs
```diff
@@ -472,7 +471,7 @@ impl<ModuleApi: SystemModuleApiFor<Self>> SystemModule<ModuleApi> for CostingMod
         }
 
         // Skip invocation costing for transaction processor
-        if depth == 0 {
+        if is_root {
```
dhedey, line 474: Ditto

radix-engine/src/system/system_modules/costing/costing_module.rs
```diff
@@ -504,7 +503,7 @@ impl<ModuleApi: SystemModuleApiFor<Self>> SystemModule<ModuleApi> for CostingMod
         }
 
         // Skip invocation costing for transaction processor
-        if depth == 0 {
+        if api.system_state().current_call_frame.is_root() {
```
dhedey, line 506: Ditto

radix-engine/src/system/system_modules/auth/auth_module.rs
```diff
@@ -87,7 +87,10 @@ impl AuthModule {
         system: &mut SystemService<Y>,
         blueprint_id: &BlueprintId,
     ) -> (BTreeSet<ResourceAddress>, BTreeSet<NonFungibleGlobalId>) {
-        let is_root_call_frame = system.kernel_get_current_depth() == 0;
+        let is_root_call_frame = system
```
dhedey, line 90: I think this could benefit from being versioned - like the other issues.

radix-engine/src/system/system_callback.rs
```diff
@@ -1381,28 +1554,13 @@ impl<C: SystemCallbackObject> KernelTransactionCallbackObject for System<C> {
...
-                let output: Vec<InstructionOutput> = scrypto_decode(&rtn).unwrap();
-                output
-            }
-            ExecutableIntents::V2(_) => {
-                unimplemented!();
-            }
-        };
+        let system_logic_version = system_service.system().versioned_system_logic;
+
+        let output = system_logic_version.execute_transaction(
```
dhedey, line 1559: I don't disagree with this, but just want to check -- your thinking is that at Cuttlefish, we swap V1 transactions to use the V2 processor too?

dhedey (commented):
A few important comments, and a few minor markups we can do when there's a moment.

PR #1912, opened 2024-09-10
feat: Can compile subintent and system manifests

dhedey:
## Summary
* Added support for compiling all manifests
* Some minor refactoring / fixing / simplifying in the compiler
* Fixed a few bugs

## Testing
* Added one test in e2e.rs - lots more to follow


radix-clis/src/resim/cmd_run.rs
```diff
@@ -54,7 +51,7 @@ impl Run {
                 blobs.push(std::fs::read(path).map_err(Error::IOError)?);
             }
         }
-        let compiled_manifest = compile(
+        let compiled_manifest = compile_manifest_v1(
```
lrubasze, line 54: How about using here newly added `compile_manifest_pretty_error()`?

radix-clis/src/rtmc/mod.rs
```diff
@@ -69,8 +63,8 @@ pub fn run() -> Result<(), String> {
         }
     }
 
-    let manifest =
-        compile(&content, &network, BlobProvider::new_with_blobs(blobs)).map_err(|err| {
+    let manifest = compile_manifest_v1(&content, &network, BlobProvider::new_with_blobs(blobs))
```
lrubasze, line 66: ditto

radix-transactions/src/manifest/compiler.rs
```diff
@@ -25,21 +25,44 @@ pub fn compile_error_diagnostics(
     }
 }
 
-pub fn compile<B>(
+pub use compile_manifest_v1 as compile;
```
lrubasze, line 28: Is this `compile` used anywhere?
If yes then how about renaming to `compile_manifest` for easier grepping (current `compile_manifest` could be  merged with `compile_manifest_v1`)?

radix-transactions/src/manifest/parser.rs
```diff
@@ -282,134 +308,44 @@ pub enum SborValueIdent {
...
     pub fn from_ident(ident: &str) -> Option<Self> {
         let value = match ident {
             // ==============
             // SBOR composite value types
             // ==============
-            "Enum" => SborValueIdent::Enum,
-            "Array" => SborValueIdent::Array,
-            "Tuple" => SborValueIdent::Tuple,
-            "Map" => SborValueIdent::Map,
+            "Enum" => ManifestValueIdent::Enum,
```
lrubasze, line 321: Not sure, but maybe we should use `Self` here and below to spare some work when renaming in the future?

radix-transactions/src/manifest/manifest_instructions.rs
```diff
@@ -32,11 +53,13 @@ pub struct TakeAllFromWorktop {
 }
 
 impl ManifestInstruction for TakeAllFromWorktop {
+    const IDENT: &'static str = "TAKE_ALL_FROM_WORKTOP";
```
lrubasze, line 56: I really like it!

lrubasze:
Nice work :)
And as stated in description would be good to add more tests (eg. to cover new instructions when pretty compiling)

lrubasze (approved):

radix-transactions/src/manifest/compiler.rs
dhedey: It could be used by downstream clients. But perhaps it's okay to remove. Will remove...

PR #1913, opened 2024-09-10
Feature: Subintent Execution Implementation Part 1

talekhinezh:
* Add `YieldToChild` and `YieldToParent` support (though arguments are not yet supported)
* Add Transaction Processor context switching support
* Add multiple stacks to Kernel

This is the first part of Subintent execution implementation. Some TODOs for future PRs:
* Implement other V2 Instructions: `VerifyParent` and `AssertWorktopIsEmpty`
* Disallow ability to call `lock_fee` in subintents
* Pass resources between subintents when yield is called
* Verification of nullifier rules


0xOmarA (approved):

0xOmarA (approved):

radix-engine/src/system/transaction/multithread_txn_processor.rs
```diff
@@ -0,0 +1,138 @@
...
+use radix_common::prelude::{BlueprintId, GlobalAddressReservation};
+use radix_engine_interface::blueprints::transaction_processor::{
+    TRANSACTION_PROCESSOR_BLUEPRINT, TRANSACTION_PROCESSOR_RUN_IDENT,
+};
+use radix_rust::prelude::*;
+use radix_transactions::model::{ExecutableTransaction, InstructionV2};
+use sbor::prelude::ToString;
+
+/// Multi-thread transaction processor for executing multiple subintents
+pub struct MultiThreadedTxnProcessor {
```
dhedey, line 20: If we're looking to drop the "thread" naming, perhaps a `MultiIntentProcessor`?

radix-engine/src/system/transaction/multithread_txn_processor.rs
```diff
@@ -0,0 +1,138 @@
...
+use crate::system::actor::{Actor, FunctionActor};
+use crate::system::system::SystemService;
+use crate::system::system_callback::{System, SystemBasedKernelApi};
+use crate::system::system_modules::auth::AuthModule;
+use radix_common::constants::TRANSACTION_PROCESSOR_PACKAGE;
+use radix_common::prelude::{BlueprintId, GlobalAddressReservation};
+use radix_engine_interface::blueprints::transaction_processor::{
+    TRANSACTION_PROCESSOR_BLUEPRINT, TRANSACTION_PROCESSOR_RUN_IDENT,
+};
+use radix_rust::prelude::*;
```
dhedey, line 15: Import wise, lots of this can be replaced with `radix_engine::internal_prelude::*` to avoid lots of import churn if we move things.

radix-engine/src/system/transaction/multithread_txn_processor.rs
```diff
@@ -0,0 +1,138 @@
...
+    ) -> Result<Self, RuntimeError> {
+        let mut txn_processors = vec![];
+
+        // Setup
+        let intents = executable.intents();
+        for (thread_id, intent) in intents.iter().enumerate() {
+            api.kernel_switch_stack(thread_id)?;
+
+            let mut system_service = SystemService::new(api);
+            let virtual_resources = intent
```
dhedey, line 38: We agreed to not use the name `virtual` for these going forward. Perhaps `initial_proof_resources`, `initial_non_fungible_proofs`?

radix-engine/src/system/transaction/multithread_txn_processor.rs
```diff
@@ -0,0 +1,138 @@
...
+            let virtual_non_fungibles =
+                intent.auth_zone_init.initial_non_fungible_id_proofs.clone();
+            let auth_zone = AuthModule::create_auth_zone(
+                &mut system_service,
+                None,
+                virtual_resources,
+                virtual_non_fungibles,
+            )?;
+
+            api.kernel_set_call_frame_data(Actor::Function(FunctionActor {
```
dhedey, line 51: Hmm this feels weird to me - is this right that we're _replacing_ the root call frame with this?

In principle, this isn't a problem, but I think we need to discuss what the abstraction is around `Actor::Root` and how it interacts with other parts of the codebase (cf comments on https://github.com/radixdlt/radixdlt-scrypto/pull/1919 )

radix-engine/src/system/transaction/multithread_txn_processor.rs
```diff
@@ -0,0 +1,138 @@
...
+                }
+            }
+        }
+
+        assert!(parent_stack.is_empty());
+
+        Ok(())
+    }
+
+    pub fn cleanup<Y: SystemBasedKernelApi>(self, api: &mut Y) -> Result<(), RuntimeError> {
```
dhedey, line 109: I think I'd rather run `cleanup` greedily when each subintent terminates. I think it gives a more intuitive / timely orphaned node error, if a user is thinking of the intents like stack frames.

radix-engine/src/system/transaction/multithread_txn_processor.rs
```diff
@@ -0,0 +1,138 @@
...
+                ResumeResult::YieldToChild(child) => {
+                    let child = *children_mapping.get(child).unwrap();
+                    parent_stack.push(cur_thread);
+                    cur_thread = child;
+                }
+                ResumeResult::YieldToParent => {
+                    cur_thread = parent_stack.pop().unwrap();
+                }
+                ResumeResult::Done => {
+                    if let Some(parent) = parent_stack.pop() {
```
dhedey, line 95: I think I'd like us to be a little stricter here - e.g. set the processor to `ProcesserState::Consumed`, and ensure that when we yield to child, we check that we're in `ProcesserState::Active(processor)`

dhedey (commented):
Nice. Because of the importance of this, I've done a very close review.
Quite a few of the comments are just about naming, but a few are questions about correctness / tightness.

PR #1915, opened 2024-09-14
feat: Subintent Validation - Part 2

dhedey:
## Summary

This is mostly a refactor PR, to achieve the following key goals:
* Allow preparation to have configuration
* Allow both transaction validation and preparation to change based on variable configuration. This can change based on the substate store, and if we recreate the validator after running a protocol update, we should be good.
* Align `LedgerTransaction` with the needs of the node, so lots of code can be deleted / removed.
* Make it simpler / more intuitive to work with transaction building, by
  * Unify the transaction validators into one
  * Make using the `RawX` wrappers mandatory, but also add helper methods to allow for easy preparation / validation / executable creation.

Next main PRs will:
* Add validation of V2 transactions
* Add substate for resolving the validation configuration
* Separate SubintentInstructionV2 from TransactionInstructionV2 
* Finally, add lots of tests / scenarios

## Testing
Existing tests pass. Full tests of subintent processing to follow (I'm making a list).

## Update Recommendations

Code touching transaction building may need to be tweaked. Notably, if you are working using Vecs, you will want to work with e.g. `RawNotarizedTransaction` instead.


radix-transactions/src/validation/transaction_validator.rs
```diff
@@ -1,74 +1,45 @@
...
-    pub fn default(network_id: u8) -> Self {
+    pub fn latest(network_definition: &NetworkDefinition) -> Self {
+        Self::babylon(network_definition.id)
+    }
+
+    pub fn babylon(network_id: u8) -> Self {
         Self {
             network_id,
-            max_notarized_payload_size: MAX_TRANSACTION_SIZE,
+            max_signatures_per_intent: 16,
```
dhedey, line 28: Max payload size has now been moved to preparation, as it makes more sense to be there - otherwise we can miss it by accident (!)

radix-common/src/constants/transaction_validation.rs
```diff
@@ -20,6 +11,3 @@ pub const MAX_TIP_PERCENTAGE: u16 = u16::MAX;
 /// The max epoch range
 /// Should be ~ 1 month. The below is ~30 days given 5 minute epochs.
 pub const MAX_EPOCH_RANGE: u64 = 12 * 24 * 30;
```
LukasGasior1, line 13: why don't we move `MIN/MAX_TIP_PERCENTAGE` and `MAX_EPOCH_RANGE`  to `PreparationSettings` too?

radix-transactions/src/builder/transaction_builder.rs
```diff
@@ -54,15 +54,19 @@ impl TransactionV1Builder {
...
             .push(signer.sign_with_public_key(&prepared.transaction_intent_hash()));
         self
     }
 
     pub fn multi_sign<S: Signer>(mut self, signers: &[&S]) -> Self {
         let intent = self.transaction_intent();
-        let prepared = intent.prepare().expect("Intent could be prepared");
+        let prepared = intent
+            .prepare(PreparationSettings::latest_ref())
+            .expect("Intent could be prepared");
```
LukasGasior1, line 69: "could **not**"  :)

radix-transactions/src/model/ledger_transaction.rs
```diff
@@ -68,71 +107,167 @@ impl PreparedLedgerTransaction {
...
+
+    fn validate(
+        self,
+        validator: &TransactionValidator,
+        accepted_kind: AcceptedLedgerTransactionKind,
+    ) -> Result<ValidatedLedgerTransaction, TransactionValidationError> {
+        let validated_inner = match self.inner {
+            PreparedLedgerTransactionInner::Genesis(t) => {
+                if !accepted_kind.permits_genesis() {
+                    return Err(TransactionValidationError::Other(
```
LukasGasior1, line 137: looks like we could be more concrete than a catch-all "Other", e.g. `TransactionValidationError::DisallowedTransactionKind`, wdyt?

radix-transactions/src/model/ledger_transaction.rs
```diff
@@ -21,7 +21,46 @@ const ROUND_UPDATE_V1_LEDGER_TRANSACTION_DISCRIMINATOR: u8 = 2;
...
+    User,
+    Validator,
+    ProtocolUpdate,
+}
+
+impl LedgerTransactionKind {
+    fn discriminator_for_hash(&self) -> u8 {
+        match self {
+            LedgerTransactionKind::Genesis => GENESIS_LEDGER_TRANSACTION_DISCRIMINATOR,
+            LedgerTransactionKind::User => USER_V1_LEDGER_TRANSACTION_DISCRIMINATOR,
```
LukasGasior1, line 35: Looks like we're reusing the same discriminator for v2? is this intended? If so, should we use a separate, version-agnostic const for that? Or perhaps add a comment.

radix-transactions/src/model/preparation/traits.rs
```diff
@@ -199,17 +203,28 @@ macro_rules! transaction_payload_v2 {
...
     }
 }
 
+pub trait ValidatedTransactionPayload: IntoExecutable {}
+
+#[derive(Debug, Copy, Clone)]
+pub enum TransactionPayloadKind {
+    CompleteUserTransaction,
+    LedgerTransaction,
+    Other,
```
LukasGasior1, line 222: it seems a bit safer if we explicitly listed all kinds here?

LukasGasior1 (approved):
Few minor comments, but looks good overall (as far as I can tell :smile: )

radix-transactions/src/builder/transaction_builder.rs
dhedey: It's actually correct-ish. https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style

The recommendations say you should state what your expectation was that wasn't met. Ideally using the word "should". So according to the rust recommendations I think we should use "Intent should be preparable". So will change to that :+1:

radix-transactions/src/model/ledger_transaction.rs
dhedey: Yeah, this is for backwards compatibility. Will add a comment :).

radix-common/src/constants/transaction_validation.rs
dhedey: Good question... I guess conceptually I want preparation to be simple limits (for the sake of building a valid payload hash, for things that a low-level client such as a ledger might need to worry about) - whereas the validation is for any non-trivial logic, and applied specifically to user transactions (not test transactions).

I'll add some comments in a later PR to explain the difference :D

radix-transactions/src/builder/transaction_builder.rs
LukasGasior1: I didn't know about that recommendation! Sounds good :+1:

radix-transactions/src/model/preparation/traits.rs
LukasGasior1: Hehe I wasn't sure if I missed any other usages or it was really just the length check :) I like both options

PR #1916, opened 2024-09-14
Feature/pass resources between subintents

talekhinezh:
* Add ability to pass resources between subintents
* Allow child subintents to use lock_contingent_fee
* Cleanup each thread when done immediately rather than at the end of the transaction


iamyulong (approved):
LGTM

radix-engine/src/system/transaction/multithread_intent_processor.rs
```diff
@@ -0,0 +1,189 @@
...
+                    parent_stack.push(cur_thread);
+                    Some((child, value, false))
+                }
+                ResumeResult::YieldToParent(value) => {
+                    let parent = parent_stack.pop().unwrap();
+                    Some((parent, value, false))
+                }
+                ResumeResult::ChildIntentDone(value) => {
+                    let parent = parent_stack.pop().unwrap();
+                    Some((parent, value, true))
```
dhedey, line 112: I think it would be great if we stored a state `ThreadState::Active / ThreadState::Complete` against each thread, and give a `RuntimeError` if:
* Child intents are resumed when complete
* Any thread except thread 0 returns `ResumeResult::RootIntentDone`
* Thread 0 returns `ResumeResult::ChildIntentDone`
* At the end, if any threads are not `ThreadState::Complete`

These would all form extra sanity checks - and would be useful to catch issues in tests if we use a `TestTransaction` which purposefully doesn't undergo validation.

radix-engine/src/system/transaction/multithread_intent_processor.rs
```diff
@@ -0,0 +1,189 @@
...
+                }
+                ResumeResult::RootIntentDone => None,
+            };
+
+            if let Some((next_thread, value, intent_done)) = switch_to {
+                // Checked passed values
+                Self::check_yielded_value(&value, api)?;
+                api.kernel_send_to_stack(next_thread, value.clone())?;
```
dhedey, line 120: Why do we both need to `send_to_stack` and send value to `txn_thread.resume(..)`?

I guess `send_to_stack` is about handling moving the owned objects, but we need to tell the other thread what it receives?

Can we avoid the clone if `kernel_send_to_stack` takes a reference to the value? I assume it just needs to extract references from it?

radix-engine/src/system/transaction/intent_processor.rs
```diff
@@ -111,27 +115,37 @@ impl<I: TxnInstruction + ManifestDecode + ManifestCategorize> TxnProcessorThread
...
-                        // intent and subintents and require subintents to end with a yield to parent.
-                        // They can then return a ResumeResult::DoneYieldToParent or something.
-                        if self.instructions.is_empty() {
-                            // Drop worktop and return ResumeResult::Done
-                            break;
+                    Yield::ToChild(child, value) => ResumeResult::YieldToChild(
+                        child,
+                        IndexedScryptoValue::from_scrypto_value(value),
+                    ),
+                    Yield::ToParent(value) => {
```
dhedey, line 138: I'd quite like it if we do some runtime checks here by passing in an `is_subintent` flag, and:
* Checking we only call `Yield::ToParent` in a subintent
* Checking we only return `Ok(ResumeResult::RootIntentDone)` in a non-subintent

Otherwise, returning a RuntimeError. This will catch badly configured `TestTransaction`s which purposefully don't undergo validation.

radix-engine/src/kernel/call_frame.rs
```diff
@@ -651,10 +651,6 @@ impl<C, L: Clone> CallFrame<C, L> {
         }
 
         for node_id in message.copy_stable_transient_references {
-            if from.depth >= to.depth {
```
dhedey, line 654: I remember discussing this before a few months ago - but are we comfortable this check can be removed without issue?

dhedey (commented):
Nice! A few tweak suggestions

radix-transactions/src/builder/manifest_builder.rs
```diff
@@ -2306,7 +2306,7 @@ where
     }
 }
 
-impl<M: BuildableManifestWithParent> ManifestBuilder<M>
+impl<M: BuildableManifest> ManifestBuilder<M>
```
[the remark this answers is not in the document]
talekhinezh: Good catch, not sure why this was changed, will revert.

radix-engine/src/transaction/receipt_schema_cuttlefish.txt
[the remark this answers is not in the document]
talekhinezh: cool will do

radix-engine/src/system/transaction/multithread_intent_processor.rs
talekhinezh: I'll add this as a TODO comment as it's not strictly necessary and probably better implemented when we write such tests.

radix-engine/src/system/transaction/multithread_intent_processor.rs
talekhinezh: Yes exactly. It's similar to our logic of calling a method with arguments, we need to update kernel call frame as well as let higher level understand what the args are. The difference here is that "Threads" is a system construct rather than kernel construct (If we implemented kernel threads the args can come from the kernel via callback).

Yup, good point, will update to use reference.

radix-engine/src/kernel/call_frame.rs
talekhinezh: Yes, this was mostly a sanity check

radix-engine/src/system/transaction/intent_processor.rs
talekhinezh: Would prefer that `intent_processor` is not aware of whether it is a child or root. Will add some of these checks in the multithread processor.

PR #1917, opened 2024-09-15
Tweak/markups and TestTransaction builder

dhedey:
## Summary
* Markups from previous PRs
* Introduce a builder for `TestTransaction::V2` which results in simpler / more readable tests.
* Also tweaked scrypto prelude 

Notes to Josh:
* Doing this, I notice that adding a YIELD_TO_PARENT instruction from the child breaks things.
* Tests should start with `use scrypto_test::prelude::*` to save lots of messy imports.
* Perhaps we should consider adding some scenarios. Will add it to our test list.

## Details
N/A

## Testing
Existing tests pass


radix-transactions/src/model/v1/test_transaction.rs
```diff
@@ -45,6 +120,39 @@ pub struct PreparedTestIntent {
     pub initial_proofs: BTreeSet<NonFungibleGlobalId>,
 }
 
+impl PreparedTestIntent {
+    #[allow(deprecated)]
```
lrubasze, line 124: Should be 
```suggestion
    #[deprecated]
```

lrubasze (approved):
Nice work :)

radix-transactions/src/model/v1/test_transaction.rs
dhedey: Building a V1 test intent is fine, because it's a legacy but supported standard. So I don't view it as being deprecated.

Preparing of instructions in a v1 way is deprecated though because it uses an old, less good, model for what to hash. But obviously it's ok (and required) to use the deprecated method for V1 transactions. We have to add an `#[allow(deprecated)]` here to avoid a compiler warning.

lrubasze: Fair enough :)

PR #1918, opened 2024-09-15
Fixed the name of the role when calling set_burnable()

klembee:
Fixed something by changing something


dhedey (approved):

dhedey:
Thanks @klembee for the fix! 👍

PR #1921, opened 2024-09-16
Refactors/further subintent fixes

dhedey:
## Summary
* Add support for `.bin` schema snapshots
* Enable `TransactionReceipt` to be changed.
* Make `LedgerSimulator` work with any transaction type.
* Add v2 support to the scenario runner

Next PR: validation.

## Testing
Existing tests pass. Will be adding more tests / scenarios in future PRs.


radix-transactions/src/builder/transaction_builder.rs
```diff
@@ -140,8 +140,16 @@ impl TransactionV1Builder {
     }
 }
 
+#[derive(Default)]
 pub struct PartialTransactionV2Builder {
```
lrubasze, line 144: Maybe it was already discussed...
Any reasons for using `Partial` instead of `Sub`?

lrubasze: Ok, had second thought on it. 
Probably because this is not a subtransaction but rather a part of transaction :)
Ignore it

dhedey: Right - exactly.  This is effectively a "subtree" of a transaction. But I think I was worried that calling it a SubTransaction might be confusing. That said, maybe it's better. Will raise it in the original naming thread as your suggestion :).

radix-transactions/src/builder/transaction_builder.rs
```diff
@@ -365,28 +411,38 @@ impl TransactionV2Builder {
...
+
+    pub fn create_intent_and_subintent_info(&mut self) -> &TransactionIntentV2 {
         if self.intent_and_subintent_signatures.is_none() {
-            let (instructions, blobs, child_hashes) = self
-                .manifest
-                .as_ref()
-                .expect("Manifest must be provided")
-                .clone()
-                .for_intent();
+            let manifest = mem::take(&mut self.manifest)
```
lrubasze, line 421: Minor.
`Option` does have it's own `take()` - maybe it would be more readable here.
```
let manifest = self
    .manifest
    .take()
    .expect("Manifest must be provided before this action is performed");
```

radix-transactions/src/builder/transaction_builder.rs
```diff
@@ -457,7 +517,7 @@ impl TransactionV2Builder {
 
     pub fn create_signed_transaction_intent(&mut self) -> &SignedTransactionIntentV2 {
         if self.signed_intent.is_none() {
-            self.create_intent_and_subintent_signatures();
+            self.create_intent_and_subintent_info();
             let (root_intent, subintent_signatures) =
                 mem::take(&mut self.intent_and_subintent_signatures).unwrap();
```
lrubasze, line 522: 1. `Option::take()` instead of `mem::take()` here and below?

2. How about using `expect()` similarly to other places?
eg.
```
expect("Root intent and subintent signatures must be provided before this action is performed");
```

scrypto-test/src/ledger_simulator/ledger_simulator.rs
```diff
@@ -1126,89 +1125,89 @@ impl<E: NativeVmExtension, D: TestDatabase> LedgerSimulator<E, D> {
         self.publish_package_with_owner(package_dir.as_ref(), owner_badge)
     }
 
-    pub fn execute_manifest<T>(
+    fn resolve_suggested_config(&self, manifest: &impl BuildableManifest) -> ExecutionConfig {
+        match manifest.suggested_execution_config_type() {
```
lrubasze, line 1129: `suggested_execution_config_type()` seems like being used only for tests.
So how about using below approach and ditch the `suggested_execution_config_type()` and `SuggestedExecutionConfigType` (Notarized variant is not set anywhere anyway)?
```
        let manifest = AnyTransactionManifest::from(manifest);
        match manifest {
            AnyTransactionManifest::V1(m)
            | AnyTransactionManifest::V2(m)
            | AnyTransactionManifest::SubintentV2(m) => ExecutionConfig::for_test_transaction(),
            AnyTransactionManifest::SystemV1(m) => {
                ExecutionConfig::for_system_transaction(NetworkDefinition::simulator())
            }
        }
        .with_kernel_trace(self.with_kernel_trace)
```

lrubasze (approved):
Left some minor comments, but overall LGTM!

radix-transactions/src/builder/transaction_builder.rs
dhedey: Ahh this is because it's guaranteed to exist from the line above -- rather than being a user-facing message. But I'll add an expect for it anyway :)

scrypto-test/src/ledger_simulator/ledger_simulator.rs
dhedey: It's an interesting idea... But I only created `AnyTransactionManifest` for the occasions where we can't have a concrete type to work at. It feels weird to me to map into it just to get an execution config and then map out of it in order to execute.

I guess I'd like to understand what you don't like about the current model? I guess that (a) it's not clear that it's used only for tests, and (b) introducing a new enum for this feels a bit over-the-top. I agree with both critiques (a) and (b).

Perhaps a method `default_test_execution_config(network_definition: NetworkDefinition)` might be more clearer / direct? :D. Will proceed with that, but happy to hear more thoughts from you on this! 👍

dhedey: Perhaps `fn default_test_execution_config_type(&self) -> DefaultTestExecutionConfigType;`?

radix-transactions/src/model/v1/test_transaction.rs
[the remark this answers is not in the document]
dhedey: Good idea 👍

PR #1923, opened 2024-09-20
fix: made locker package visible in scrypto-test

gguuttss:
Using TestEnvironment in combination with packages that made use of the AccountLocker wasn't possible. Now it is.

## Summary
```
Packages that made use of the AccountLocker weren't testable using the TestEnvironment. With this fix they are.
```

## Details
```
I reckon it was just an omission, but it could be there was a reason for leaving out the AccountLocker from being used in the TestEnvironment. Feel free to ignore if it messes with the rest of the code base. 

It seems the AccountLocker package wasn't visible, as it hadn't been added to the GLOBAL_VISIBLE_NODES in the environment constants. All I did was add it.
```

## Testing
```
Not a lot, only test I've done is try to publish a package that makes use of the AccountLocker using the TestEnvironment. Before the fix it didn't work:

KernelError(CallFrameError(CreateFrameError(PassMessageError(GlobalRefNotFound(NodeId("0d906318c6318c6fe2d9198c6318c6318cf7bd4f3bf55557c6318c6318c6")))))).

After the fix, it does work.
```

### DIFF
--- a/scrypto-test/src/environment/constants.rs
+++ b/scrypto-test/src/environment/constants.rs
@@ -2,7 +2,7 @@ use crate::prelude::*;
 
 /// Defines the set of Nodes that all test [`CallFrame`]s have visibility to when they're first
 /// created. This contains all of the well-known addresses of nodes.
-pub(super) const GLOBAL_VISIBLE_NODES: [NodeId; 28] = [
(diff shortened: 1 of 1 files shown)


gguuttss:
Will close for now, as David has mentioned Discord that the AccountLocker should not be added to the GLOBAL_VISIBLE_NODES.

PR #1924, opened 2024-09-23
refactor: Streamline engine init

dhedey:
## Summary
This PR explores the idea of moved the reading of the boot substates higher up in the boot process, just before we execute, and storing this onto the `XInit` parameter objects, along with any generics, so that the `XInit` can fully specify `X`.

This simplifies [transaction_executor.rs](https://github.com/radixdlt/radixdlt-scrypto/pull/1924/files#diff-796f5fb2614f1a182ebb4c1d13b67f246cd4265f76d2e9f04624e23fbcd3c192) and anywhere else we need to execute transactions with different set-ups.

I've:
* Created utility methods `XInit::load` for loading these `Boot` methods from the Database, and removed the `BootStore` trait.
* The `XInit` fully capture the generics of their corresponding `X` which allows it to be used to _create_ an `X`. This means we don't have to separately provide generics for e.g. `System<Vm<...>>` which make methods harder to use, and make the code uglier. This has allowed e.g. neatening up the abstractions in `InjectCostingError`.

I've also taken the liberty to fix up a few minor areas as part of this PR:
* Got rid of `TransactionExecutor` struct which after various refactors it was a wrapper with no purpose any more
* Renamed `KernelTransactionCallbackObject` to `KernelTransactionExecutor` - could maybe even call it `TransactionExecutor` now.
* Standardized on `V` (for VM) for the `SystemCallbackObject` generic 
* Removed `IntentHashNullification::System`
* Fixed the two test kernel wrappers (`InjectCostingError` and `MockKernel`) to be generic over the type of System it receives, so we don't have to keep updating it if System changes.
* Improved `SystemCallback`
* Minor tweak to API for creating `StateUpdates`.

## Testing
Existing tests pass

### DIFF
--- a/radix-common/src/state/state_updates.rs
+++ b/radix-common/src/state/state_updates.rs
@@ -16,6 +16,12 @@ pub struct StateUpdates {
 }
 
 impl StateUpdates {
+    pub fn empty() -> Self {
+        Self {
+            by_node: Default::default(),
+        }
+    }
+
     /// Starts a Node-level update.
     pub fn of_node(&mut self, node_id: NodeId) -> &mut NodeStateUpdates {
         self.by_node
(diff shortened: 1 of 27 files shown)


iamyulong (approved):
Looks good!

PR #1927, opened 2024-09-26
Make locker package globally visible

0xOmarA:
## Summary

This PR makes the locker package globally visible by adding it to the set of globally visible nodes. It is only added after the cuttlefish protocol update so the behavior stays the same when cuttlefish node drops.


radix-common/src/constants/always_visible_nodes.rs
```diff
@@ -16,7 +16,7 @@ pub use radix_common::constants::*;
 // - Split bootstrapping into state flushing and transaction execution (the "chicken-and-egg" problem)
 //
 lazy_static! {
-    pub static ref ALWAYS_VISIBLE_GLOBAL_NODES: IndexSet<NodeId> = {
+    pub static ref ALWAYS_VISIBLE_GLOBAL_NODES_V1: IndexSet<NodeId> = {
```
dhedey, line 19: Now that we have the method, I think it's best to just always use the method to get access. So could we make these both not `pub`?

radix-engine/src/kernel/kernel.rs
```diff
@@ -125,11 +141,13 @@ impl<'h, S: SubstateDatabase> BootLoader<'h, S> {
...
-        // Unused for now
-        let _ = kernel_boot;
-
         // Upper Layer Initialization
-        let system_init_result = E::init(&mut self.track, &executable, callback_init);
+        let system_init_result = E::init(
+            &mut self.track,
+            &executable,
+            callback_init,
+            kernel_boot.always_visible_global_nodes(),
```
dhedey, line 149: Side note: Nothing wrong with this, it's needed for now - but it feels weird to me that system needs to care about reference checking of the call frame inits...

We're going to refactor call frame init stuff anyway, so I'm looking forward to moving this reference checking out of system then 👍

radix-engine-monkey-tests/tests/fuzz_kernel.rs
```diff
@@ -471,7 +471,12 @@ fn kernel_fuzz<F: FnMut(&mut KernelFuzzer) -> Vec<KernelFuzzAction>>(
     let mut substate_db = InMemorySubstateDatabase::standard();
     let mut track = Track::new(&substate_db);
     let mut callback = TestCallbackObject;
-    let mut kernel = Kernel::new_no_refs(&mut track, &mut id_allocator, &mut callback);
+    let mut kernel = Kernel::new_no_refs(
+        &mut track,
+        &mut id_allocator,
+        &mut callback,
+        &ALWAYS_VISIBLE_GLOBAL_NODES_V2,
```
dhedey, line 478: I suggest we have a `const fn latest() -> Self` on `AlwaysVisibleGlobalNodesVersion` and use that to resolve the latest, rather than hard coding it to `V2`.

So this would be: `always_visible_global_nodes(AlwaysVisibleGlobalNodesVersion::latest())`.

Honestly, if  `Kernel::new_no_refs` is for tests, we could even just hard-code this into it and avoid changing all these signatures.

radix-engine/src/kernel/call_frame.rs
```diff
@@ -557,11 +560,12 @@ pub enum SubstateDiffError {
     ContainsDuplicateOwns,
 }
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct CallFrameInit<C> {
     pub data: C,
     pub global_addresses: IndexSet<GlobalAddress>,
     pub direct_accesses: IndexSet<InternalAddress>,
+    pub always_visible_global_nodes: &'static IndexSet<NodeId>,
```
dhedey, line 568: I'm a little confused why this is on `CallFrameInit`.

I think in the future, possibly having a flag `uses_always_visible_global_nodes` might make sense here, but for now it'd always be true.

I feel that the `always_visible_global_nodes` is a global property, so should conceptually live in the kernel itself, probably? And then it can be provided when the call to `get_node_visibility` comes in:

```rust
    fn kernel_get_node_visibility(&self, node_id: &NodeId) -> NodeVisibility {
        self.stacks.cur().get_node_visibility(self.always_visible_references, node_id)
    }
```

Then we don't need to have always visible global nodes here, or have system care about it.

radix-engine/src/kernel/kernel.rs
```diff
@@ -52,6 +53,21 @@ impl KernelBoot {
     pub fn babylon() -> Self {
         Self::V1
     }
+
+    pub fn cuttlefish() -> Self {
+        Self::V2(AlwaysVisibleGlobalNodesVersion::V2)
+    }
+
+    pub fn always_visible_global_nodes_version(&self) -> AlwaysVisibleGlobalNodesVersion {
```
dhedey, line 61: I wonder if we can make this not pub?

radix-engine/src/updates/cuttlefish.rs
```diff
@@ -1,12 +1,14 @@
...
 use super::*;
-use crate::system::system_callback::*;
+use crate::{kernel::kernel::KernelBoot, system::system_callback::*};
 
 #[derive(Clone)]
 pub struct CuttlefishSettings {
     /// Add configuration for system logic versioning
     pub system_logic_update: UpdateSetting<NoSettings>,
+    /// Updating the always visible global nodes to include the account locker package.
+    pub always_visible_global_nodes_update: UpdateSetting<NoSettings>,
```
dhedey, line 11: I think probably calling this `kernel_version_update` might be a nicer naming scheme, in case we want to add other things to Kernel changes this update.

scrypto-test/src/environment/builder.rs
```diff
@@ -200,6 +200,13 @@ where
             self.database.commit(&database_updates);
         }
 
+        // Getting the kernel boot to use for the kernel creation.
+        let kernel_boot = self.database.get_existing_substate::<KernelBoot>(
```
dhedey, line 204: I think you want `KernelBoot::load(&self.database)` which has the correct fallbacks when run on pre-bottlenose.

radix-engine/src/updates/cuttlefish.rs
```diff
@@ -89,6 +96,12 @@ fn generate_principal_batch(
             generate_system_logic_v2_updates(store),
         ));
     }
+    if let UpdateSetting::Enabled(_settings) = &always_visible_global_nodes_update {
+        transactions.push(ProtocolUpdateTransactionDetails::flash(
+            "cuttlefish-protocol-update-always-visible-global-nodes",
```
dhedey, line 101: Ditto on the naming scheme. (Unless we choose to separate the two updates into two, to selectively enable them)

dhedey (commented):
Nice! A few tweaks suggested, happy to discuss them :)

radix-common/src/constants/always_visible_nodes.rs
0xOmarA: Sounds good, these two do not need to be public.

radix-transactions/src/builder/transaction_builder.rs
```diff
@@ -359,7 +359,7 @@ impl PartialTransactionV2Builder {
     ) -> (SignedPartialTransactionV2, TransactionObjectNames) {
         let (transaction, names) = self.build_with_names();
         let validator = TransactionValidator::new_with_static_config_network_agnostic(
-            ValidationConfig::latest(),
+            TransactionValidationConfig::latest(),
```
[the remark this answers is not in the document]
0xOmarA: Yeah I could not compile `develop` because of this so I had to make these changes. Figured that perhaps some accidental renaming work got in early in `develop`.

radix-engine/src/updates/cuttlefish.rs
0xOmarA: Sounds good, going with that name.

radix-engine/src/updates/cuttlefish.rs
0xOmarA: Yup, naming sounds good to me. Will go with the name you recommended.

scrypto-test/src/environment/builder.rs
0xOmarA: Neat!

radix-engine/src/kernel/kernel.rs
0xOmarA: Yup, but why not keep it public?

radix-engine/src/kernel/call_frame.rs
0xOmarA: > I think in the future, possibly having a flag uses_always_visible_global_nodes might make sense here, but for now it'd always be true.

It could be modeled as a flag. I feel like a set is better here than using a flag since it feels slightly nicer.

> I feel that the always_visible_global_nodes is a global property, so should conceptually live in the kernel itself, probably? And then it can be provided when the call to get_node_visibility comes in:

We could have the globally visible nodes be stored in the kernel and then passed to `get_node_visibility` inside of `kernel_get_node_visibility`. However, the primary issue I see with this is that `get_node_visibility` gets called from other places that are not `kernel_get_node_visibility` and that do not have access to kernel state such as `CallFrame::pass_message` and `CallFrame::process_input_substate_key`. This would mean that we still have to have the globally visible nodes be stored in the callframe or at least accessible to it.

dhedey: Ahh fair enough! Happy to keep this as-is then :)

radix-engine/src/kernel/kernel.rs
dhedey: Very minor, it just feels like an internal detail that other things shouldn't need to care about :)

dhedey (approved):
Looks good. Scenarios need regenerating, then happy to merge when it goes green.

PR #1928, opened 2024-09-26
Change the min rounds per epoch

0xOmarA:
## Summary

Updates the min rounds per epoch in cuttlefish to be 100


radix-engine-tests/tests/protocol/consensus_manager.rs
```diff
@@ -0,0 +1,46 @@
...
+            CONSENSUS_MANAGER,
+            MAIN_BASE_PARTITION,
+            ConsensusManagerField::Configuration,
+        )
+        .into_payload()
+        .fully_update_and_into_latest_version();
+
+    // Assert
+    assert_eq!(
+        dbg!(config).config.epoch_change_condition.min_round_count,
```
dhedey, line 23: I think the `dbg` was left in by accident?

radix-engine/src/updates/cuttlefish.rs
```diff
@@ -130,3 +144,28 @@ fn generate_cuttlefish_transaction_validation_updates() -> StateUpdates {
...
+            MAIN_BASE_PARTITION,
+            ConsensusManagerField::Configuration,
+        )
+        .into_payload()
+        .fully_update_and_into_latest_version();
+
+    consensus_manager_config
+        .config
+        .epoch_change_condition
+        .min_round_count = 100;
```
dhedey, line 163: So on the node, in `GenesisManagerConfig`, I see two main variants: `productionDefaults()` (used by Mainnet and Stokenet) sets `minRoundsPerEpoch` to 500, but `testDefaults()` already has it on 100, and other networks have it set to `Long.MAX_VALUE`.

Could we be clever here and only update it to `100` if it's currently `500`? This avoids breaking test networks which configure genesis in a certain way and then rely on network lengths being a certain length.

dhedey: In the scenarios I suggest we can then override the `BabylonSettings` to use the mainnet defaults, so we can observe this change there 👍. The mainnet genesis configuration was as follows. This will cause a change in the scenario summary for genesis, but that's fine :)

```
        ConsensusManagerConfig {
            max_validators: 100,
            epoch_change_condition: EpochChangeCondition {
                min_round_count: 500,
                max_round_count: 3000,
                target_duration_millis: 300000,
            },
            num_unstake_epochs: 2016,
            total_emission_xrd_per_epoch: dec!(2853.881278538812785),
            min_validator_reliability: Decimal::one(),
            num_owner_stake_units_unlock_epochs: 8064,
            num_fee_increase_delay_epochs: 4032,
            validator_creation_usd_cost: dec!(1000),
        }
```

dhedey (commented):
Nice! A few suggestions to make it not break testnets

radix-engine-tests/tests/protocol/consensus_manager.rs
0xOmarA: Indeed, thanks for the catch.

radix-transactions/src/builder/transaction_builder.rs
```diff
@@ -359,7 +359,7 @@ impl PartialTransactionV2Builder {
     ) -> (SignedPartialTransactionV2, TransactionObjectNames) {
         let (transaction, names) = self.build_with_names();
         let validator = TransactionValidator::new_with_static_config_network_agnostic(
-            ValidationConfig::latest(),
```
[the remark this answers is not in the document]
0xOmarA: Thanks. Looks good now.

radix-engine/src/updates/cuttlefish.rs
0xOmarA: > Could we be clever here and only update it to 100 if it's currently 500? This avoids breaking test networks which configure genesis in a certain way and then rely on network lengths being a certain length.

Yup, we could do it that way. I think that we could also use the feature you added a while back where protocol updates can have settings to control what the min rounds per epoch gets updated to. How would you feel about that? 

>In the scenarios I suggest we can then override the BabylonSettings to use the mainnet defaults, so we can observe this change there 👍. The mainnet genesis configuration was as follows. This will cause a change in the scenario summary for genesis, but that's fine :)

Sounds good, will do!

dhedey: > I think that we could also use the feature you added a while back where protocol updates can have settings to control what the min rounds per epoch gets updated to. How would you feel about that?

Yep, it was designed for this. Happy to have a setting here, but the setting should default to "automatic" which should follow the logic specified I think :). Basically we just want to be changing the production default.

radix-engine/src/updates/babylon.rs
```diff
@@ -167,7 +181,7 @@ impl BabylonSettings {
         Self {
             genesis_data_chunks,
             genesis_epoch: Epoch::of(1),
-            consensus_manager_config: ConsensusManagerConfig::test_default(),
+            consensus_manager_config: ConsensusManagerConfig::mainnet_genesis(),
```
iamyulong, line 184: Why do we need to change this? This seems to break all scenario assertions.

dhedey: Was a suggestion from me here: https://github.com/radixdlt/radixdlt-scrypto/pull/1928#discussion_r1777619337

But I actually meant change in `dumper.rs` > `run_all_protocol_updates`, change:
```
        let babylon_settings = BabylonSettings::test_complex()
.with_consensus_manager_config(ConsensusManagerConfig::mainnet_genesis());
        let protocol_executor = ProtocolBuilder::for_network(&network_definition)
            .configure_babylon(|_| babylon_settings)
            .from_bootstrap_to_latest();
```

This will then only change it for the protocol update tests, so we can get more realistic protocol change receipts.

radix-engine-tests/assets/serialized_receipts/basic_subintents/create-accounts.json
```diff
@@ -0,0 +1,44 @@
+{
```
dhedey, line 1: Hmm I'm surprised this didn't error because they weren't regenerated... In a separate PR, 

I would suggest in a separate PR that we either:
* Move this into the scenario `dumper.rs` (probably slightly easier)
* Keep it separate, and use a `FolderContentAligner`, and have separate update/validate methods, and add its `update` function to `./update-scenarios.sh` so it will run when someone updates a scenario.

dhedey (approved):
Really nice, thanks! :)

radix-engine-tests/assets/serialized_receipts/basic_subintents/create-accounts.json
0xOmarA: Sounds good to me. Will move this into the scenario in a separate PR.

PR #1929, opened 2024-09-27
Fix the metadata for cuttlefish

0xOmarA:
## Summary

Fix the metadata for cuttlefish


iamyulong (approved):

radix-engine-tests/tests/protocol/metadata.rs
```diff
@@ -0,0 +1,15 @@
+use scrypto_test::prelude::*;
+
+#[test]
+fn metadata_is_changed_after_cuttlefish() {
+    // Arrange
+    let mut ledger = LedgerSimulatorBuilder::new().build();
+
+    // Act
+    let MetadataValue::Url(url) = ledger.get_metadata(XRD.into(), "icon_url").unwrap() else {
```
dhedey, line 9: We could change `get_metadata` to take an `impl ResolvableGlobalAddress`

radix-engine/src/updates/cuttlefish.rs
```diff
@@ -160,3 +174,86 @@ fn generate_cuttlefish_transaction_validation_updates() -> StateUpdates {
         ),
     )
 }
+
+fn generate_cuttlefish_metadata_fix<S: SubstateDatabase + ?Sized>(db: &S) -> StateUpdates {
+    struct MetadataUpdates {
```
dhedey, line 179: MINOR: Instead of this, it might be more standard to use `MetadataInit` and the `metadata_init! { }` macro below - which would let us do `description: "Hello"` instead of `description: Some(MetadataValue::String("Hello".into()))`  :)

For each node, we could then use `MetadataNativePackage::init_system_struct(init)` to validate and extract the kv entries to `IndexMap<Vec<u8>, KVEntry>` (or even take out a `MetadataNativePackage::map_metadata_init(init) -> Result<IndexMap<Vec<u8>, KVEntry>, ...>` method from inside `init_system_struct` to make it cleaner).

And then just map `IndexMap<Vec<u8>, KVEntry>` to `set_substate` calls.

radix-engine/src/updates/cuttlefish.rs
```diff
@@ -160,3 +174,86 @@ fn generate_cuttlefish_transaction_validation_updates() -> StateUpdates {
...
+            .collect()
+        }
+    }
+
+    let reader = SystemDatabaseReader::new(db);
+
+    // The metadata entries that we would like to update and the values that we would like to update
+    // them to be.
+    let metadata_updates = indexmap! {
+        XRD => MetadataUpdates {
```
dhedey, line 204: Still in discussion on these :)

radix-engine/src/updates/cuttlefish.rs
```diff
@@ -160,3 +174,86 @@ fn generate_cuttlefish_transaction_validation_updates() -> StateUpdates {
...
+            description: Some(MetadataValue::String("Badges generated automatically by the Radix system to represent the authority of a global caller. These badges cease to exist at the end of their transaction.".into())),
+            icon_url: Some(MetadataValue::Url(UncheckedUrl("https://assets.radixdlt.com/icons/icon-global_caller_resource.png".into())))
+        },
+        SECP256K1_SIGNATURE_RESOURCE => MetadataUpdates {
+            name: Some(MetadataValue::String("ECDSA Secp256k1 Signature Resource".into())),
+            description: Some(MetadataValue::String("Badges generated automatically by the Radix system to represent ECDSA Secp256k1 signatures applied to transactions. These badges cease to exist at the end of their transaction.".into())),
+            icon_url: Some(MetadataValue::Url(UncheckedUrl("https://assets.radixdlt.com/icons/icon-ecdsa_secp256k1_signature_resource.png".into())))
+        },
+        ED25519_SIGNATURE_RESOURCE => MetadataUpdates {
+            name: Some(MetadataValue::String("EdDSA Ed25519 Resource".into())),
```
dhedey, line 225: This is inconsistent with the above. Should probably be `EdDSA Ed25519 Signature Resource`*

dhedey (commented):
Nice! 👍

radix-engine/src/updates/cuttlefish.rs
dhedey: In my opinion, from a code readability / verification standpoint, checking the keys are right is easier than extracting a name from `Some(MetadataValue::String("Hello".into()))`, and ignoring lots of entries with no descriptions - but fair enough.

dhedey (approved):
Great stuff, thanks!

PR #1933, opened 2024-09-30
Feature/cuttlefish deps updates

lrubasze:
## Summary
Update some dependencies


radix-clis/go.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
dhedey, line 1: What does this do? Is it intended to be committed? Can we add some comments explaining its intention and when it should be run?

radix-common/src/crypto/ed25519/private_key.rs
```diff
@@ -11,37 +11,37 @@ impl Ed25519PrivateKey {
     pub const LENGTH: usize = 32;
 
     pub fn public_key(&self) -> Ed25519PublicKey {
-        Ed25519PublicKey(PublicKey::from(&self.0).to_bytes())
+        Ed25519PublicKey(SigningKey::from(&self.0).verifying_key().to_bytes())
```
dhedey, line 14: I wonder if we should probably change to wrap a `SigningKey` here?

It would save us recreating it a lot when we do lots of signing/public key extraction, and `SigningKey` implements `Zeroize`, but `SecretKey` doesn't - which seems to suggest that `SigningKey` is intended to be the in-memory signing key type they intend us to use.

dhedey (commented):
Nice! Happy to approve post discussions

radix-clis/go.sh
lrubasze: Added by mistake. Will remove.

radix-common/src/crypto/ed25519/private_key.rs
lrubasze: Yeah, that makes sense. I will refactor it.

radix-clis/go.sh
lrubasze: Done

radix-common/src/crypto/ed25519/private_key.rs
lrubasze: Done

radix-common/src/crypto/ed25519/private_key.rs
```diff
@@ -1,28 +1,22 @@
 use super::Ed25519Signature;
 use crate::internal_prelude::*;
-use ed25519_dalek::{Keypair, PublicKey, SecretKey, Signer};
-use zeroize::Zeroize;
+use ed25519_dalek::{Signer, SigningKey};
 
-#[derive(Zeroize)]
-#[zeroize(drop)]
-pub struct Ed25519PrivateKey(SecretKey);
+// No need to derive Zeroize.
```
dhedey, line 5: I believe we still need explicit `zeroize` to allow clients to zeroize without dropping - e.g. I think Sargon needed it?

dhedey: https://docs.rs/zeroize/latest/zeroize/
> This crate has custom derive support for the Zeroize trait, gated under the zeroize crate’s zeroize_derive Cargo feature, which automatically calls zeroize() on all members of a struct or tuple struct.

dhedey: I guess we technically don't need "on drop" if the child is doing it on drop already

iamyulong: It's mainly needed for zeroize composition, especially considering that one can't extract the inner contents.

lrubasze: Unfortunately it is not possible to derive Zeroize for `SigningKey`, because only `secret_key` implements `zeroize()` and `verifying_key` does not.
```
pub struct SigningKey {
    /// The secret half of this signing key.
    pub(crate) secret_key: SecretKey,   
    /// The public half of this signing key.
    pub(crate) verifying_key: VerifyingKey,
}
```
We also does not have access to the `SigningKey` fields, so we could implement zeroize manually.

So I see 2 options:
1. back to the previous implementation, where `Ed25519PrivateKey` was wrapping `SecretKey` and derive Zeroize
```
pub struct Ed25519PrivateKey2(SecretKey)
```
2.  Use unsafe manual zeroing.

I think we should rather avoid unsafe, so'd prefer option 1.

WDYT

lrubasze: Done.
slack discussion [here](https://rdxworks.slack.com/archives/C071C6DB8TG/p1727867648027169?thread_ts=1727706454.898939&cid=C071C6DB8TG)

radix-common/src/crypto/ed25519/private_key.rs
```diff
@@ -1,47 +1,64 @@
...
+        *self.0 = None;
+    }
+}
 
 impl Ed25519PrivateKey {
     pub const LENGTH: usize = 32;
 
+    fn signing_key(&self) -> &SigningKey {
+        let option = &*self.0;
+        option.as_ref().expect("Key expected")
```
dhedey, line 29: VERY MINOR: Perhaps "Cannot access signing key after zeroizing"

dhedey (approved):

PR #1943, opened 2024-10-04
tests: Added scenario and tests for subintent depth

dhedey:
## Summary
Adds a scenario and basic tests to check that a subintent depth of four is allowed, but a depth of 5 errors.

Also adds some helpers to the ledger simulator to help with constructing notarized transactions.


lrubasze (approved):
LGTM

PR #1944, opened 2024-10-04
Limit references and signatures in transaction v1 & v2

iamyulong:
## Summary

This PR introduces the following transaction validation limits for cuttlefish

- Across the whole transaction:
    * Total Signer Signatures: 64
    * Total References (explicitly used in instructions): 512
 - Per intent:
    * Max Signer Signatures: 16
    * Max References (explicitly used in instructions): 512

Also changed `max_subintent_count` to `32`.


radix-transactions/src/validation/transaction_validator.rs
```diff
@@ -76,13 +81,16 @@ impl TransactionValidationConfig {
...
+            max_total_references: usize::MAX,
         }
     }
 
     pub const fn cuttlefish() -> Self {
         Self {
+            max_references_per_intent: 512,
             v2_transactions_allowed: true,
-            max_subintent_count: 128,
+            max_subintent_count: 32,
```
dhedey, line 93: Ahh my bad - it looks like we already have `PreparationSettings.max_subintents_per_transaction` - (my fault for duplicating these) - perhaps these should be merged?

Or we should just read it off `validation_settings.preparation_settings.max_subintents_per_transaction` inside the validator if we want to check it again.

radix-transactions/src/validation/transaction_validator.rs
```diff
@@ -190,14 +200,23 @@ impl TransactionValidator {
     ) -> Result<ValidatedNotarizedTransactionV1, TransactionValidationError> {
         self.validate_intent_v1(&transaction.signed_intent.intent)?;
 
-        let encoded_instructions = Rc::new(manifest_encode(
-            &transaction.signed_intent.intent.instructions.inner.0,
-        )?);
+        self.check_reference_limits(vec![
```
dhedey, line 203: MINOR: For better encapsulation, perhaps we should move `self.check_reference_limits` inside `validate_intent_v1`?

And arguably `check_signature_limits` should be in `validate_signatures_v1`?

radix-transactions/src/validation/transaction_validator.rs
```diff
@@ -500,12 +515,32 @@ impl TransactionValidator {
                 2,
             ));
         }
+
         let transaction_intent = &prepared.signed_intent.transaction_intent;
+        let root_subintent = &prepared.signed_intent.transaction_intent;
```
dhedey, line 520: According to our agreed naming scheme, the term `subintent` is reserved for non-root intents.

Either call it `transaction_intent` (which is already used in the line above) or `root_intent`.

radix-transactions/src/validation/transaction_validator.rs
```diff
@@ -577,6 +629,69 @@ impl TransactionValidator {
         })
     }
 
+    pub fn check_reference_limits(
+        &self,
+        subintent_references: Vec<&IndexSet<Reference>>,
```
dhedey, line 634: `intent_references`*

radix-transactions/src/validation/transaction_validator.rs
```diff
@@ -555,6 +590,23 @@ impl TransactionValidator {
         let non_root_subintents = &prepared.partial_transaction.non_root_subintents;
         let non_root_subintent_signatures = &prepared.non_root_subintent_signatures;
 
+        self.check_reference_limits(
```
dhedey, line 593: MINOR: Ideally this should live in `validate_v2_intent_cores_and_subintent_connection_counts` to prevent duplication between `validate_signed_partial_transaction_v2` and `validate_notarized_v2`

radix-transactions/src/validation/transaction_validator.rs
```diff
@@ -577,6 +629,69 @@ impl TransactionValidator {
...
 
+    pub fn check_reference_limits(
+        &self,
+        subintent_references: Vec<&IndexSet<Reference>>,
+    ) -> Result<(), TransactionValidationError> {
+        let mut total = 0;
+        for (index, refs) in subintent_references.iter().enumerate() {
+            if refs.len() > self.config.max_references_per_intent {
+                return Err(TransactionValidationError::TooManyReferencesForIntent {
+                    index,
```
dhedey, line 640: From the transaction builder perspective, we don't have an "intent index" abstraction at all.

We either have "Transaction Intent" OR `SubintentIndex` into the non_root_subintents array (i.e. 0 = first non-root subintent). Feel free to rename `SubintentIndex` to `NonRootSubintentIndex` if you want, I think it got missed in the name refactor.

dhedey: In terms of this error - I'm happy if we just drop the index :)

dhedey (approved):
Approving. Happy to merge if we fix the naming and tests.

radix-transactions/src/validation/transaction_validator.rs
iamyulong: Will remove `max_subintent_count`.

radix-transactions/src/validation/transaction_validator.rs
iamyulong: After looking into this a bit more, I think it's better to move the limits check up instead of inside `validate_intent_v1`. 

This helps reduce unnecessary, heavy validation work when things are not within limits.

radix-transactions/src/validation/transaction_validator.rs
iamyulong: Will rename

radix-transactions/src/validation/transaction_validator.rs
iamyulong: Will rename

radix-transactions/src/validation/transaction_validator.rs
iamyulong: See comments above - we do limit checks before validation.

radix-transactions/src/validation/transaction_validator.rs
iamyulong: I think it's better to know which intent went wrong, for debugging purpose.

For partial transaction,  there is no "Transaction Intent". The index of intents is probably the easiest way to reveal such information.

radix-transactions/src/validation/transaction_validator.rs
dhedey: I'm comfortable with this - although I'm not entirely sure I agree.

* If we wanted to optimize for avoiding unnecessary work then the intent checks would be at prepare time.
* If we care about code clarity, putting intent checks in an intent validation method feels like we're less likely to miss these if we create other composable validations (e.g. if we add a method to validate a subintent or something).

That said, I agree that over-arching checks of pan-transaction limits (of e.g. references or total signatures) need to be at a higher layer, or passed down as bounds to constrain a lower layer.

So I'm happy to keep with this approach.

radix-transactions/src/validation/transaction_validator.rs
dhedey: > I think it's better to know which intent went wrong, for debugging purpose.

Sure - but then I'd argue that we should map a `ManifestValidationError` into a `LocatedManifestValidationError` inside `TransactionValidationError::ManifestValidationError` with a location of either an `IntentLocation::Root` or `IntentLocation::NonRootSubintent(SubintentIndex)`, and could be done for all these validations.

Happy to do this when I add more tests.

PR #1949, opened 2024-10-07
Feature/nullification costing

iamyulong:
## Summary

Add deferred costing for nullifications.


radix-engine/src/system/system_callback.rs
```diff
@@ -1571,7 +1571,61 @@ impl<V: SystemCallbackObject> KernelTransactionExecutor for System<V> {
...
+                                                TransactionStatusV1::CommittedSuccess,
+                                            )),
+                                            lock_status: LockStatus::Unlocked,
+                                        },
+                                    ))
+                                    .len(),
+                                ),
+                            };
+                            return costing
+                                .apply_deferred_execution_cost(
```
iamyulong, line 1612: Over charging for failed subintents.

radix-engine/src/system/system_callback.rs
```diff
@@ -1571,7 +1571,61 @@ impl<V: SystemCallbackObject> KernelTransactionExecutor for System<V> {
...
+                match hash_nullification {
+                    IntentHashNullification::TransactionIntent { .. } => {
+                        // Transaction intent nullification is historically not costed.
+                    }
+                    IntentHashNullification::Subintent { .. } => {
+                        if let Some(costing) = modules.costing_mut() {
+                            let read_not_found =
+                                IOAccess::ReadFromDbNotFound(CanonicalSubstateKey {
+                                    node_id: TRANSACTION_TRACKER.into_node_id(),
+                                    partition_number: PartitionNumber(0),
```
dhedey, line 1585: This appears a little hacky (given that I believe the original intention of `IOAccess` was to be able to track exactly what was read).

Could we perhaps either:
* Use the actual substate key? Say via `validate_intent_hash_uncosted` returning it?
* Acknowledge the hack with a commented line `let placeholder_key = CanonicalSubstateKey { ... }`, and we can then reuse `placeholder_key` in the two places below.

radix-engine/src/system/system_callback.rs
dhedey: Can I ask we add a comment to the code for this?

radix-engine/src/system/system_modules/costing/costing_entry.rs
```diff
@@ -290,6 +295,9 @@ pub mod owned {
         RefCheck {
             event: RefCheckEventOwned,
         },
+        Nullification {
+            io_access: Vec<IOAccess>,
```
dhedey, line 299: Naively, I'd have expected this to be split between execution cost for the read, and finalization cost for the write?

dhedey: Also I wonder if we should use more intuitive names here, for the purpose of the receipt? (because "Nullification" is not really a common name in crypto/literature, even if we're using it internally)
* `ExecutionCostingEntry::CheckIntentValidity`
* `FinalizationCostingEntry::RecordIntentCommits`

dhedey (approved):
Nice! A few thoughts/suggestions

radix-engine/src/system/system_callback.rs
```diff
@@ -1571,7 +1571,61 @@ impl<V: SystemCallbackObject> KernelTransactionExecutor for System<V> {
                         )
                     }
                 }
-            };
+            }
+            .and_then(|_| {
+                match hash_nullification {
+                    IntentHashNullification::TransactionIntent { .. } => {
+                        // Transaction intent nullification is historically not costed.
```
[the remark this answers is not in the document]
iamyulong: There is no base fee, but `tx_payload_cost` fee which is proportional to payload size.

```
// The size of a typical transfer transaction is 400 bytes, and the cost will be 400 * 40 = 16,000 cost units
// The max size of a transaction is 1 MiB, and the cost will be 1,048,576 * 40 = 41,943,040 cost units
// This is roughly 1/24 of storing data in substate store per current setup.
mul(cast(size), 40)
```

radix-engine/src/system/system_callback.rs
```diff
@@ -1571,7 +1571,61 @@ impl<V: SystemCallbackObject> KernelTransactionExecutor for System<V> {
...
+                    }
+                    IntentHashNullification::Subintent { .. } => {
+                        if let Some(costing) = modules.costing_mut() {
+                            let read_not_found =
```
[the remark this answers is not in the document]
iamyulong: Because we check if the hash already exists.

Will add comment.

radix-engine/src/system/system_callback.rs
iamyulong: Agree with the sentiment, although ultimately this probably should be replaced with a constant (with the current logic as a basis), which is why I didn't bother making it neater.

Will try to clean this up a little bit.

radix-engine/src/system/system_modules/costing/costing_entry.rs
iamyulong: Sounds reasonable to me.

radix-engine/src/system/system_callback.rs
iamyulong: Sure

radix-engine/src/system/system_callback.rs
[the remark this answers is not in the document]
dhedey: Oh! I thought there was a base fee. Interesting. Perhaps we should add one at Cuttlefish to account for the ident? Or just bill for TransactionIntent nullification from Cuttlefish?

radix-engine/src/system/system_modules/costing/costing_entry.rs
iamyulong: I've tried to add the whole deferred finalization cost, and then realized it doesn't fit the finalization cost model. 

Finalization costs are supported to reflect on post-execution costs in Node.

But here, we really want to just bill for execution time as a result of IOAccess. Intent hashes are periodically removed, so little overhead to the state storage.

radix-engine/src/system/system_callback.rs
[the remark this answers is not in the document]
iamyulong: We have various uncosted substate reads (such as systemboot, vmboot, kernelboot).

Per the comment above, the minimum transaction cost unit is about `16,000`, which is just enough to cover one  `ReadFromDBNotFound` (which is much more expensive than `ReadFromDB`).

I think we can live with this for now, OR introduce a larger base.

radix-engine/src/system/system_modules/costing/costing_entry.rs
iamyulong: `FinalizationCostingEntry::RecordIntentCommits` is actually  covered by `FinalizationCostingEntry::CommitStateUpdates`

dhedey: Are you sure? When I looked at the code, finalisation of the state updates was calculated before the intent state changes were written

radix-engine/src/system/system_callback.rs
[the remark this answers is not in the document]
dhedey: What do you think about the system logic tweak to also bill for transaction intent reads? Would be quite straightforward I imagine?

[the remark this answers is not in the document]
iamyulong: > Perhaps we should add one at Cuttlefish to account for the ident

At execution time, we can't tell if it's v1 or v2. 

If we start charging for transaction intent from cuttlefish, fees for all transaction will be increased, which has a user impact. Not too sure about if it's worth to do it now.

radix-engine/src/system/system_modules/costing/costing_entry.rs
iamyulong: You're right! I don't know what I was thinking. Let me add it explicitly. 

And looks like we don't need to defer this cost, but apply at finalization time. If it fails, there is no commit anyway. Even fairer to users.

radix-engine/src/system/system_callback.rs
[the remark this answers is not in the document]
dhedey: We're in system here, right? 

o have access to the version number either from version state (current) or from a generic (if we merge https://github.com/radixdlt/radixdlt-scrypto/pull/1919 ).

> If we start charging for transaction intent from cuttlefish, fees for all transaction will be increased, which has a user impact. Not too sure about if it's worth to do it now.

It will, but not by much I imagine?

radix-engine/src/system/system_modules/costing/costing_entry.rs
dhedey: Sounds good.

And if we add billing for non-contingent intents (currently just transaction intents), these should possibly be billed up front/deferred.

radix-engine/src/system/system_callback.rs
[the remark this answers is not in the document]
iamyulong: > It will, but not by much I imagine?

It's about 16,000.

iamyulong:
Merging this PR. 

The only pending question is "if we want to apply costing for the transaction intent hash status, and whether it's for v2 only OR v1 + v2", which can be addressed separately.

PR #1952, opened 2024-10-10
Add Secp256k1 key recovery to CryptoUtils v2

lrubasze:
## Summary
- Add Secp256k1 key recovery to CryptoUtils v2
- Apply costing for CryptoUtils v2 methods
- Fix error returned in case of Protocol version mismatch
- Add some tests 
- CryptoUtils metering tests


radix-engine-tests/tests/application/metering.rs
```diff
@@ -67,6 +67,11 @@ fn run_all(mode: CostingTaskMode) {
             &run_mint_small_size_nfts_from_manifest,
             "cost_mint_small_size_nfts_from_manifest.csv",
         );
+
+        // Run cost tests for crypto_utils test from Cuttlefish onward
+        if protocol_version >= ProtocolVersion::Cuttlefish {
```
0xOmarA, line 72: Isn't this condition always going to be true from this point onwards?

radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -564,11 +564,23 @@ impl FeeTable {
...
-        //     444420.94242 -> 446000
-        let instructions_cnt = 446000;
+        //     464236 -> 500000
+        let instructions_cnt = 500000;
         // Convert to cost units
         instructions_cnt / CPU_INSTRUCTIONS_TO_COST_UNIT
     }
+
+    #[inline]
+    pub fn secp256k1_ecdsa_key_recover_cost(&self) -> u32 {
```
0xOmarA, line 574: Why did we base our costing on `test_crypto_scrypto_key_recover_secp256k1_ecdsa`? In that test we create a new ledger simulator, call the simulator, build manifests, among many other things which means that the number of instructions we get from that test is for `secp256k1_key_recovery + ledger_simulator_building + manifest_execution + wasm invocations + etc...`. Wouldn't it be more accurate to calculate the number of instructions that it takes to recover a Secp256k1 natively?

0xOmarA: In other words, this isn't the "number of instructions to perform Secp256k1 recovery" but rather the "number of instructions to perform Secp256k1 recovery in wasm and execute it via our `LedgerSimulator`"

scrypto/src/crypto_utils/crypto_utils.rs
```diff
@@ -140,4 +140,20 @@ impl CryptoUtils {
             ) != 0
         }
     }
+
+    /// Performs ECDSA Secp256k1 key recovery.
+    pub fn secp256k1_ecdsa_key_recover(
+        hash: impl AsRef<Hash>,
```
0xOmarA, line 146: Minor: maybe call it `message_hash`?

radix-engine-tests/tests/application/metering.rs
lrubasze: No, because we iterate over versions starting from `Genesis`
see [line 37](https://github.com/radixdlt/radixdlt-scrypto/pull/1952/files/5a97d1350e181d2359eb514bb146f66799d0dccd#diff-9f9d7f5a18faa9e6112ad24f3c3f6cb17a017e88b40e3fb62ac3d5268dc027c3R37)

radix-engine/src/system/system_modules/costing/fee_table.rs
lrubasze: QEMU costing job runs many tests, `test_crypto_scrypto_key_recover_secp256k1_ecdsa` included.

When getting the instruction count we grep QEMU job results for given instruction. eg. `crypto_utils_secp256k1_ecdsa_key_recover`
image
The instruction we want to find must be "traced" (marked with `trace_resources` attribute, see [this](https://github.com/radixdlt/radixdlt-scrypto/pull/1952/files/5a97d1350e181d2359eb514bb146f66799d0dccd#diff-882b9c55afb3d9bed9557b03b333eb2357f3515a16683d19ed91d6f46f1e17f4R749)).

> Wouldn't it be more accurate to calculate the number of instructions that it takes to recover a Secp256k1 natively?

`crypto_utils_secp256k1_ecdsa_key_recover` is traced, which means we do exactly what you suggest :)
So despite there are much more instructions executed for this test, we take into account only those related to `crypto_utils_secp256k1_ecdsa_key_recover`

0xOmarA: Makes sense to me, thank you for the explanation.

0xOmarA (approved):

radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -564,11 +564,23 @@ impl FeeTable {
         // Based on  `test_crypto_scrypto_verify_secp256k1_ecdsa_costing`
         //   instructions_cnt = 464236 (input is always 32 bytes long)
         //   Lets round:
-        //     444420.94242 -> 446000
-        let instructions_cnt = 446000;
+        //     464236 -> 500000
```
dhedey, line 567: I don't mind this at all, but purely out of interest, why are we rounding this one up a lot more than `ed25519_verify_cost` ?

dhedey (approved):
Looks good, although costing isn't my specialist area

scrypto/src/crypto_utils/crypto_utils.rs
lrubasze: Done

radix-engine/src/system/system_modules/costing/fee_table.rs
lrubasze: I think I have overlooked it. Was working on `Secp256k1` key recover and saw that other `Secp256k1`-related methods needs updating, so I've updated it. But forgot about `Ed25519`. I think it might be updated too for consistency.
Thanks for spotting it!

PR #1953, opened 2024-10-10
Optimize gas host function

iamyulong:
## Summary

Optimize the gas host function implementation.

Results on `Macbook Pro 16-inch, 2021 - Apple M1 Max`:

```rust
costing::spin_loop_v2   time:   [1.9165 s 1.9216 s 1.9277 s]                                   
                        change: [-28.394% -27.953% -27.554%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/radix-engine-tests/assets/wasm/loop_v2.wat
+++ b/radix-engine-tests/assets/wasm/loop_v2.wat
@@ -1,0 +1,29 @@
+(module
+
+  ;; Simple function that always returns `()`
+  (func $Test_f (param $0 i64) (result i64)
+    ;; Loop starts!
+    (loop $loop
+      br $loop
+    )
+    ;; Loop ends!
+
+    ;; Encode () in SBOR at address 0x0
+    (i32.const 0)
+    (i32.const 92)  ;; prefix
+    (i32.store8)
+    (i32.const 1)
+    (i32.const 33)  ;; tuple value kind
+    (i32.store8)
+    (i32.const 2)
+    (i32.const 0)  ;; tuple length
+    (i32.store8)
+
+    ;; Return slice (ptr = 0, len = 3)
+    (i64.const 3)
+  )
+
+  (memory $0 1)
+  (export "memory" (memory $0))
+  (export "Test_f" (func $Test_f))
+)
--- a/radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
+++ b/radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
@@ -53,6 +53,34 @@ impl<'y, Y: SystemApi<RuntimeError>> ScryptoRuntime<'y, Y> {
             String::from_utf8(blueprint_name).map_err(|_| WasmRuntimeError::InvalidString)?;
         Ok((package_address, blueprint_name))
     }
+
+    #[cold]
+    pub fn consume_wasm_execution_units_unoptimized(
+        &mut self,
+        mut n: u32,
+    ) -> Result<(), InvokeError<WasmRuntimeError>> {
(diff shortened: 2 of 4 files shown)


radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
```diff
@@ -53,6 +53,34 @@ impl<'y, Y: SystemApi<RuntimeError>> ScryptoRuntime<'y, Y> {
             String::from_utf8(blueprint_name).map_err(|_| WasmRuntimeError::InvalidString)?;
         Ok((package_address, blueprint_name))
     }
+
+    #[cold]
+    pub fn consume_wasm_execution_units_unoptimized(
```
dhedey, line 58: MINOR comments:
* I don't think this needs to be `pub`
* I'm not particularly a fan of the `unoptimized` name, I think I'd slightly prefer the method name was a bit more semantic (e.g `consume_wasm_execution_exceeding_buffer`).
* If we have a semantic method name, we could simplify this method a little, and eek out a tiny bit more performance. I appreciate this is only hit about 100 times less frequently than the hot path, but I think it also helps with readability.
```rust
// We know this is > 0 based on the precondition of the method.
let n_remaining_after_buffer_used = n - self.wasm_execution_units_buffer;
let amount_to_request = n_remaining_after_buffer_used
    .checked_add(WASM_EXECUTION_COST_UNITS_BUFFER)
    .unwrap_or(u32::MAX);
self.api
    .consume_cost_units(ClientCostingEntry::RunWasmCode {
        package_address: &self.package_address,
        export_name: &self.export_name,
        wasm_execution_units: amount_to_request,
    })
    .map_err(InvokeError::downstream)?;
self.wasm_execution_units_buffer += amount_to_request - n_remaining_after_buffer_used;
```

radix-engine-tests/assets/wasm/loop_v2.wat
```diff
@@ -0,0 +1,29 @@
...
+
+  ;; Simple function that always returns `()`
+  (func $Test_f (param $0 i64) (result i64)
+    ;; Loop starts!
+    (loop $loop
+      br $loop
+    )
+    ;; Loop ends!
+
+    ;; Encode () in SBOR at address 0x0
```
dhedey, line 11: All this stuff isn't actually needed because the loop spins forever, right? Or is some return needed for the thing to pass validation?

dhedey (approved):

radix-engine-tests/assets/wasm/loop_v2.wat
iamyulong: It's not required in this case, as we don't care about the execution results. It has little impact on the test.

radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
iamyulong: The last line doesn't seems correct. Should be `=` instead, right?

dhedey: Conceptually it could be `self.wasm_execution_units_buffer = WASM_EXECUTION_COST_UNITS_BUFFER` if we - but can't change that for now in case we ever exceed `u32::MAX` (somehow!) as we'd bill something different.

We've agreed to tweak this going forward when we tweak the instruction costing.

radix-engine-tests/assets/wasm/loop_v2.wat
dhedey: Sounds good. Happy with your choice of leave it / comment it / delete it.

PR #1955, opened 2024-10-10
Chore/fixes for node integration

dhedey:
## Summary

This is a first PR of a lot of minor tweaks to support https://github.com/radixdlt/babylon-node/pull/1007 in the node.

Captures a few changes, quite a few are reversions from changes in the last 6 months which cause pain in the node or make the node less performant.
 
* Converted `UserTransaction` to be able to be the decode target of encoded `NotarizedTransactionV1` and `NotarizedTransactionV2`.
* `Executable` or `ValidatedNotarizedTransactionV2` needs to be Send so it can live in our mempool:
    * I've removed the Rcs.
    * I've also changed it so that we convert `Prepared` => `Validated` => `Executable` as `self` references.
    * The engine can now take a &ExecutableTransaction allowing us to avoid clones in the node
* Reverted the bottlenose change for `abort_on_loan_repaid`  to the `ExecutionConfig` instead of the executable itself. This allows us to re-use the same `&ExecutableTransaction` for both execution modes, and I think it's also just a more correct abstraction.
* Added a variety of helper methods and types - things that were missed when I ported various types from the node to the engine to help decrease churn as I worked on V2 transactions.
* Finally managed to get rid of `LegacyStateUpdates` - including in genesis. This is part of me trying to unify genesis with protocol updates... we're very nearly there!
* Added auto-trigger scenarios to the scenario outputs so we know if we've broken anything.

## Testing
* Existing tests pass


lrubasze (approved):

lrubasze (approved):
LGTM

radix-transactions/src/model/user_transaction.rs
```diff
@@ -69,10 +69,16 @@ impl UserSubintentManifest {
     }
 }
 
-#[derive(Debug, Clone, Eq, PartialEq)]
+const V1_DISCRIMINATOR: u8 = TransactionDiscriminator::V1Notarized as u8;
+const V2_DISCRIMINATOR: u8 = TransactionDiscriminator::V2Notarized as u8;
+
+/// This can be used like [`VersionedTransactionPayload`], but just for notarized tranasctions.
```
lrubasze, line 75: transactions?

radix-common/src/state/state_updates.rs
```diff
@@ -70,8 +70,41 @@ impl StateUpdates {
         }
     }
 
-    pub fn into_legacy(self) -> LegacyStateUpdates {
-        self.into()
+    /// Roughly equivalent to the LegacyStateUpdates (when they existed), ignoring partition resets.
+    /// Should only be used for tests, not for committed to a database.
```
lrubasze, line 74: `committed` sounds weird here
not for committing to a database?

PR #1958, opened 2024-10-14
Feature/assert next call instructions

talekhinezh:
* Implement `ASSERT_NEXT_CALL_RETURNS_ONLY` and `ASSERT_NEXT_CALL_RETURNS_INCLUDE` instructions


radix-engine/src/system/transaction/intent_processor.rs
```diff
@@ -387,10 +425,64 @@ impl IntentProcessorObjects {
...
+                ))
+            })?;
+        }
+
         Ok(())
     }
 }
 
+struct ResourceConstraintChecker {
+    fungible_resources: BTreeMap<ResourceAddress, Decimal>,
```
dhedey, line 441: By default, can we use IndexMap unless we have to use an invariant ordering?
`IndexMap` is more efficient than BTreeMap, and less surprising when things get arbitrarily ordered by byte-ordering of resource address.

radix-engine/src/system/transaction/instructions.rs
```diff
@@ -318,6 +319,38 @@ impl TxnNormalInstruction for AssertWorktopContainsNonFungibles {
...
 }
 
+impl TxnNormalInstruction for AssertNextCallReturnsOnly {
+    fn execute<Y: SystemApi<RuntimeError> + KernelNodeApi + KernelSubstateApi<L>, L: Default>(
+        self,
+        _worktop: &mut Worktop,
+        objects: &mut IntentProcessorObjects,
+        _api: &mut Y,
+    ) -> Result<InstructionOutput, RuntimeError> {
+        objects.next_call_return_constraints = Some(NextCallReturnConstraints {
```
dhedey, line 329: For consistency, perhaps this should be `next_call_returns_constraints = Some(NextCallReturnsConstraints` ?

dhedey: Perhaps we should panic or throw an error if this is already set when we set it?

It shouldn't be possible (the validation should enforce it) but sometimes we run transactions without validation (e.g. system transactions).

radix-common/src/data/manifest/model/manifest_resource_assertion.rs
```diff
@@ -144,6 +148,46 @@ impl ManifestResourceConstraints {
...
+                    .remove(&resource_address)
+                    .unwrap_or_default();
+                constraint
+                    .validate_fungible(amount)
+                    .map_err(ManifestResourceConstraintsError::ResourceConstraint)?;
+            } else {
+                let ids = non_fungible_resources
+                    .remove(&resource_address)
+                    .unwrap_or_default()
+                    .clone();
```
dhedey, line 170: I think the clone is unneeded?

radix-common/src/data/manifest/model/manifest_resource_assertion.rs
```diff
@@ -144,6 +148,46 @@ impl ManifestResourceConstraints {
...
+                constraint
+                    .validate_non_fungible(ids)
+                    .map_err(ManifestResourceConstraintsError::ResourceConstraint)?;
+            }
+        }
+
+        if exact {
+            if !fungible_resources.is_empty() || !non_fungible_resources.is_empty() {
+                return Err(ManifestResourceConstraintsError::UnwantedResourcesExist);
```
dhedey, line 179: It might be nice to give an example resource or even resources?

radix-common/src/data/manifest/model/manifest_resource_assertion.rs
```diff
@@ -144,6 +148,46 @@ impl ManifestResourceConstraints {
...
+                    .unwrap_or_default()
+                    .clone();
+                constraint
+                    .validate_non_fungible(ids)
+                    .map_err(ManifestResourceConstraintsError::ResourceConstraint)?;
+            }
+        }
+
+        if exact {
+            if !fungible_resources.is_empty() || !non_fungible_resources.is_empty() {
```
dhedey, line 178: This is only correct if these are all non-zero balances. If this is required, it should IMO be expressed in the parameter name or rust doc. Currently I think this invariant _is_ correctly upheld inside the instruction processor, but it's a little indirect for my liking.

Instead, perhaps we could just iterate across the remaining `fungible_resources` and `non_fungible_resources` and if any of the balances are non-zero, then return an error with an example resource? That would then remove this pre-condition; and also allow us to give an example resource which was present.

dhedey: And I think we should add a test for this (to avoid potential regressions in future)

e.g.
* `ASSERT_NEXT_CALL_RETURNS_EXACTLY <empty>` and then `withdraw XRD Decimal("0")` should succeed
* `ASSERT_NEXT_CALL_RETURNS_EXACTLY <empty>` and then `withdraw <nfs> <empty>` should succeed

dhedey (approved):
Nice. Happy to merge subject to discussion of comments 👍

radix-common/src/data/manifest/model/manifest_resource_assertion.rs
dhedey: Also, for good regression coverage, would be nice to have a quick test which applies an unmeetable `ASSERT_NEXT_CALL_RETURNS_EXACTLY` before a `CALL_FUNCTION` and both `YIELD_` commands (and e.g. check it gives a failure)

radix-engine/src/system/transaction/instructions.rs
talekhinezh: I think it's fine behavior that the system will just override the last setting.

radix-engine/src/system/transaction/intent_processor.rs
talekhinezh: This was originally `IndexMap` but changed it to use `BTreeMap` since I want the ability to remove from a map without using `swap_remove`. Will use `NonIterMap` instead.

radix-common/src/data/manifest/model/manifest_resource_assertion.rs
talekhinezh: good catch

radix-common/src/data/manifest/model/manifest_resource_assertion.rs
talekhinezh: > Also, for good regression coverage, would be nice to have a quick test which applies an unmeetable `ASSERT_NEXT_CALL_RETURNS_EXACTLY` before a `CALL_FUNCTION` and both `YIELD_` commands (and e.g. check it gives a failure)

This is already covered by tests `when_less_is_returned_assert_next_call_returns_only_should_fail` and `when_more_is_returned_assert_next_call_returns_only_should_fail` though it's missing the `YIELD_TO_CHILD` command. Will add this.

radix-engine/src/system/transaction/intent_processor.rs
dhedey: Nothing wrong with `swap_remove`, it's just remove which explains how it affects the ordering.

But yeah, `NonIterMap` was designed for this purpose. Or if we need to iter, maybe there's some other way to do the algorithm?

talekhinezh: Ah good point, will switch back to `IndexMap`

PR #1961, opened 2024-10-14
update: Use rust 1.81.0 for cuttlefish

dhedey:
## Summary
Updates to use [current stable rust](https://blog.rust-lang.org/2024/09/05/Rust-1.81.0.html) (1.81.0) for Cuttlefish, which we can then update to in the node.

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -13,7 +13,7 @@ runs:
         #   - "rustup default XXXX"
         # * Ensure you update the version in the "radix-clis-scrypto-coverage" step,
         #   or update the comment explaining why a different version is used.
-        toolchain: 1.80.1
+        toolchain: 1.81.0
         default: true
         target: wasm32-unknown-unknown
         components: rustfmt
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -9,13 +9,13 @@ runs:
       with:
         # IMPORTANT:
         # To update the Rust version, you must update it in five total places (including the toolchain below):
-        # => [README.md - part 1] Update "Install Rust - we recommend to use Rust XXXX"
-        # => [README.md - part 2] Update "rustup default XXXX"
-        # => [action.yml - part 1] Update "toolchain: XXXX" below
-        # => [action.yml - part 2] Update the three lines starting "rustup toolchain install nightly-...`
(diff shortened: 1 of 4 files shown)


iamyulong (approved):

lrubasze:
I think we need to update as well a docker image sha in `Dockerfile` and instructions in `README.md`.

Dockerfile
```diff
@@ -1,10 +1,5 @@
-# Below version includes rust 1.80.1
-ARG RUST_IMAGE_VERSION=@sha256:e8e40c50bfb54c0a76218f480cc69783b908430de87b59619c1dca847fdbd753
-# If you want to use latest version then uncomment next line
-# ARG RUST_IMAGE_VERSION=:slim-bookworm
-# Alternatively you can build docker with argument: --build-arg="RUST_IMAGE_VERSION=:slim-bookworm"
-
-FROM rust${RUST_IMAGE_VERSION} AS base-image
+# https://hub.docker.com/_/rust/tags
+FROM rust:1.81.0-slim-bookworm AS base-image
```
lrubasze, line 2: The image sha under this tag is subject to change.
So might happen that our docker images will not be binary the same despite the same tag.

.github/actions/setup-env/action.yml
```diff
@@ -7,13 +7,16 @@ runs:
...
       uses: RDXWorks-actions/toolchain@master
       with:
-        # IMPORTANT: If you update this Rust version:
-        # * Ensure you also update it in the README in two places:
-        #   - "Install Rust - we recommend to use Rust XXXX"
-        #   - "rustup default XXXX"
-        # * Ensure you update the version in the "radix-clis-scrypto-coverage" step,
-        #   or update the comment explaining why a different version is used.
-        toolchain: 1.80.1
+        # IMPORTANT:
```
lrubasze, line 10: I like this instruction!

Dockerfile
dhedey: > The image sha under this tag is subject to change.

Indeed - but the resultant image ( https://hub.docker.com/r/radixdlt/scrypto-builder/tags ) will be fixed / immutable, which is all I think that matters?

lrubasze: The build of the resultant image will be not reproducible. If we can live with it then good to go!

lrubasze (approved):

PR #1962, opened 2024-10-15
fix: Remove empty partitions from state updates

dhedey:
## Summary
Fixes an issue noticed when removing `LegacyStateUpdates` in the node integration PR -- due to how we mapped the `TrackedSubstates`, we end up getting lots of empty partitions in the state updates (about 50% of partitions are empty -- basically any partitions where we only read, but didn't update substates!).

This had the side effect of:
* Wasting time in the node's JMT (it looks like it actually causes wasted database reads 🤦)
* Causing confusion when displaying state changes, and comparing them against actual changes - indeed, I previously added a "rebuild, removing empty partitions" method in the Transaction receipt display to hide this
* Generally, just wasting memory/bandwidth

Important notes regarding backwards compatibility:
* The scenario state change hashes simply did `scrypto_encode(&state_changes)` so have given a different hash now that the empty values are filtered out - but importantly the state changes in the formatted receipts have not changed (because these had my hack to filter out empty substates before). This means that we are not ultimately changing the list of substates getting committed.
* I *do not believe* this impacts node compatibility -- consensus relies on the state root hash which filters out these sorts of oddities... That said, I'll run this through on the node to check it can sync a historic ledger, but I'm not expecting any issues there.

## Testing
* Existing tests pass
* I can sync stokenet node past genesis


radix-engine/src/track/track.rs
```diff
@@ -65,30 +65,37 @@ impl TrackedSubstates {
...
+                                    DatabaseUpdate::Set(substate.value.into())
+                                }
+                            },
+                        };
+                        Some((tracked.substate_key, update))
+                    })
+                    .collect();
+
+                // Filter out empty partition updates, to avoid wasted work downstream (e.g. in the Merkle Tree in the node)
+                if partition_updates.len() > 0 {
```
dhedey, line 93: This is the key change which fixes the issue - the rest is refactors.

iamyulong (approved):

PR #1964, opened 2024-10-17
Feature/subintent execution trace

iamyulong:
## Summary

Update execution trace to not break for transaction v2.


radix-engine/src/kernel/kernel_api.rs
```diff
@@ -202,26 +202,41 @@ pub struct SystemState<'a, M: KernelCallbackObject> {
     pub caller_call_frame: &'a M::CallFrameData,
 }
 
-/// Internal API for kernel modules.
-/// No kernel state changes are expected as of a result of invoking such APIs, except updating returned references.
+/// Internal API for system modules only.
+///
+/// TODO: Do not use uncosted API within protocol
```
dhedey, line 207: Is this TODO meant to be done in this PR? Or in future protocol updates?

radix-engine/src/system/system_modules/execution_trace/module.rs
```diff
@@ -461,13 +461,16 @@ impl<ModuleApi: SystemModuleApiFor<Self> + SystemModuleApiResourceSnapshotExtens
     SystemModule<ModuleApi> for ExecutionTraceModule
 {
     fn on_create_node(api: &mut ModuleApi, event: &CreateNodeEvent) -> Result<(), RuntimeError> {
+        if api.current_stack_id_uncosted() != 0 {
```
dhedey, line 464: MINOR: I'm not a big fan of arbitrary comparisons to `0`.

Perhaps we could have an `is_transaction_intent_thread()` system function?
Or even better, we could have this be `current_stack` which returns some data, e.g. an `is_root` flag, or an `is_contingent` flag or some enum or something?

dhedey (commented):
Nice!

Although I'd argue your summary is a little misleading:
> Update execution trace to support transaction v2 (subintent capability).

Really it doesn't break on subintents, rather than supports the subintent capability 👍 . But it's fine for now :).

There are also a couple of test failures to resolve

dhedey (approved):

radix-engine/src/system/system_modules/execution_trace/module.rs
iamyulong: There is no "transaction intent" concept in the system yet. Not sure that will do any good.

radix-engine/src/kernel/kernel_api.rs
iamyulong: Future protocol updates. They're very cheap (reading a variable from memory), so there is no urgency in fixing them.

PR #1969, opened 2024-10-21
tweak: `Decimal` and `PreciseDecimal` subunits are no longer `pub`

dhedey:
## Summary

* Makes the internal value of `Decimal` / `PreciseDecimal` private - this avoids some gotchas that us and the community have experienced - where they intend to e.g. create 2 from `Decimal(2.into())` but instead get 2 attos.
* Makes some of the functions `const`
* Adds `Decimal::ONE_ATTO`, `PreciseDecimal::ONE_ATTO`, `Decimal::ONE_SUBUNIT` and `PreciseDecimal::ONE_PRECISE_SUBUNIT` constants.

Whilst this is a "breaking change" for people who chose to update to Cuttlefish scrypto, I think it is a low-risk break, in that:
* I imagine it’s very rare for people to actually do this (or actually do this intentionally), so it likely won’t affect too much code in the wild.
* The change is very simple - `Decimal(x)` to `Decimal::from_attos(x)` - and is signposted on the `Decimal` / `PreciseDecimal` rust docs to point users in the right direction

## Testing
Some new tests are added.

## Update Recommendations
* Instead of `Decimal(x)` use `Decimal::from_attos(x)`
* Instead of `PreciseDecimal(x)` use `PreciseDecimal::from_precise_subunits(x)`
* Instead of `decimal.0` use `decimal.attos()`
* Instead of `precise_decimal.0` use `precise_decimal.precise_subunits()`

### DIFF
--- a/radix-common/src/math/precise_decimal.rs
+++ b/radix-common/src/math/precise_decimal.rs
@@ -27,15 +27,20 @@ use crate::*;
 /// The finite set of values are of the form `m / 10^36`, where `m` is
 /// an integer such that `-2^(256 - 1) <= m < 2^(256 - 1)`.
 ///
+/// ```text
 /// Fractional part: ~120 bits / 36 digits
 /// Integer part   : 136 bits / 41 digits
 /// Max            :  57896044618658097711785492504343953926634.992332820282019728792003956564819967
 /// Min            : -57896044618658097711785492504343953926634.992332820282019728792003956564819968
+/// ```
 ///
-/// Unless otherwise specified, all operations will panic if underflow/overflow.
+/// Unless otherwise specified, all operations will panic if there is underflow/overflow.
+///
+/// To create a PreciseDecimal with a certain number of precise `10^(-36)` subunits,
+/// use [`PreciseDecimal::from_precise_subunits`].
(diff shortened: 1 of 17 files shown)


radix-common/src/math/precise_decimal.rs
```diff
@@ -27,15 +27,20 @@ use crate::*;
...
+/// ```
 ///
-/// Unless otherwise specified, all operations will panic if underflow/overflow.
+/// Unless otherwise specified, all operations will panic if there is underflow/overflow.
+///
+/// To create a PreciseDecimal with a certain number of precise `10^(-36)` subunits,
+/// use [`PreciseDecimal::from_precise_subunits`].
 #[cfg_attr(feature = "fuzzing", derive(Arbitrary))]
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct PreciseDecimal(pub InnerPreciseDecimal);
```
lrubasze, line 38: Removing `pub` might be painful for dApp devs. I've seen you asked in the devs channel if ok to do this. They agreed but not sure if every dev got to read it.

PR #1971, opened 2024-10-23
Add some reference related tests

iamyulong:
## Summary

Add some reference related tests.

### DIFF
--- a/radix-engine-tests/tests/system/threading.rs
+++ b/radix-engine-tests/tests/system/threading.rs
@@ -1,6 +1,8 @@
 use radix_engine_tests::common::PackageLoader;
 use scrypto_test::prelude::*;
 
+// Some of the tests in this file are to demonstrate the current behavior.
+
 #[test]
 fn can_transfer_locked_bucket_between_threads() {
     let mut ledger = LedgerSimulatorBuilder::new().build();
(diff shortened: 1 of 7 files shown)


radix-engine-tests/tests/system/threading.rs
```diff
@@ -0,0 +1,268 @@
...
+                    network_id: NetworkDefinition::simulator().id,
+                    start_epoch_inclusive,
+                    end_epoch_exclusive,
+                    min_proposer_timestamp_inclusive: None,
+                    max_proposer_timestamp_exclusive: None,
+                    intent_discriminator: 1,
+                })
+                .manifest_builder(|builder| {
+                    builder
+                        // Unfortunately, there is no way to grab the received references
```
dhedey, line 113: This test doesn't show whether we can use those references? Can we try doing a vault freeze here perhaps?

Or wait I assume this just-works because it's already in the subintent manifest? So... passing references doesn't do anything really?

radix-engine-tests/tests/system/threading.rs
```diff
@@ -0,0 +1,268 @@
+use radix_engine_tests::common::PackageLoader;
+use scrypto_test::prelude::*;
+
```
dhedey, line 3: Might be worth a quick comment at the top explaining that these tests mostly just demonstrate existing behaviour, so we can protect ourselves against future regressions? (rather than being concrete product requirements).

radix-engine-tests/tests/system/threading.rs
```diff
@@ -0,0 +1,268 @@
...
+#[test]
+fn can_transfer_locked_bucket_between_threads() {
```
dhedey, line 5: I assume we have a test elsewhere which Josh added which checks we can't transfer a proof between threads? Maybe that should be moved here?

dhedey (approved):

radix-engine-tests/tests/system/threading.rs
iamyulong: I assumed so. Let me just add one.

radix-engine-tests/tests/system/threading.rs
iamyulong: For sure, to prevent regression and also to make sure relevant code paths are covered.

PR #1972, opened 2024-10-23
Decrease max number of buffers

iamyulong:
## Summary

Decrease max number of buffers from `32` to `4`.


radix-engine-tests/tests/vm/wasm_limits.rs
```diff
@@ -0,0 +1,43 @@
+use radix_engine_tests::common::PackageLoader;
+use scrypto_test::prelude::*;
+
+#[test]
+fn test_crate_buffers_within_limits() {
```
dhedey, line 5: create*?

dhedey (approved):

PR #1973, opened 2024-10-23
Align WASM feature set in scrypto compiler

lrubasze:
## Summary
Since **LLVM 19** below features are enabled by default (see [release notes](https://releases.llvm.org/19.1.0/tools/clang/docs/ReleaseNotes.html#webassembly-support))
- reference-types
- multivalue.

This PR makes sure that Scrypto compiler builds C dependencies with WASM `MVP` + `mutable-global` + `sign-ext` features (exactly as RadixEngine expects and exactly what Rust native libraries are built with) no matter what LLVM version is available on the system.

NOTE!
Important takeaway is the fact that C dependencies are build with system LLVM, while native Rust libraries are built with LLVM built-in into Rust compiler.

## Test
Relevant tests added


scrypto-compiler/src/lib.rs
```diff
@@ -756,6 +769,7 @@ impl ScryptoCompiler {
 
     fn wasm_optimize(&self, wasm_path: &Path) -> Result<(), ScryptoCompilerError> {
         if let Some(wasm_opt_config) = &self.input_params.wasm_optimization {
+            println!("optimizing WASM {:?}", wasm_opt_config);
```
iamyulong, line 772: Can we have a `--verbose` flag for these prints?

iamyulong (approved):
Looks really good!

PR #1974, opened 2024-10-24
Refactor metadata validation

iamyulong:
## Summary

Refactor metadata validation, particularly checking length before validating URLs.


radix-engine-tests/assets/blueprints/metadata2/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+mod metadata2 {
+    struct M {}
+
+    impl M {
+        pub fn f() {
+            let package_address = METADATA_MODULE_PACKAGE;
+            let blueprint_name = METADATA_BLUEPRINT;
+            let function_name = METADATA_CREATE_WITH_DATA_IDENT;
+
+            /*
```
dhedey, line 13: Is this commented-out code intended to be deleted? :)

dhedey: Ahh no it's explaining how the below was generated, my bad! All good.

Perhaps we could have a comment here briefly explaining that? But only if you feel like it, no worries, I was just being slow.

radix-engine-tests/benches/metadata_validation.rs
```diff
@@ -0,0 +1,20 @@
...
+use scrypto::prelude::CheckedUrl;
+use std::hint::black_box;
+
+#[allow(unused_must_use)]
+fn bench_validate_urls(c: &mut Criterion) {
+    c.bench_function("metadata_validation::validate_urls", |b| {
+        b.iter(|| {
+            black_box(
+                CheckedUrl::of("https://www.example.com/test?q=x")
+                    .ok_or(MetadataValueValidationError::InvalidURL("".to_owned())),
```
dhedey, line 13: MINOR: This is always, valid, right? So we should probably just `unwrap()` here - otherwise greedy allocation of a string might slow down the test a little.

radix-engine/src/object_modules/metadata/package.rs
```diff
@@ -16,10 +16,9 @@ use super::{RemoveMetadataEvent, SetMetadataEvent};
 
 #[derive(Debug, Clone, Eq, PartialEq, ScryptoSbor)]
 pub enum MetadataError {
-    KeyStringExceedsMaxLength { max: usize, actual: usize },
-    ValueSborExceedsMaxLength { max: usize, actual: usize },
     ValueDecodeError(DecodeError),
-    MetadataValidationError(MetadataValidationError),
+    MetadataKeyValidationError(MetadataKeyValidationError),
```
dhedey, line 20: This will require temporarily using `, regenerate` to update the `VersionedLocalTransactionExecution`.

radix-engine/src/object_modules/metadata/package.rs
```diff
@@ -16,10 +16,9 @@ use super::{RemoveMetadataEvent, SetMetadataEvent};
 
 #[derive(Debug, Clone, Eq, PartialEq, ScryptoSbor)]
 pub enum MetadataError {
-    KeyStringExceedsMaxLength { max: usize, actual: usize },
```
dhedey, line 19: We can't remove these error variants as it breaks the error deserialization of historic receipts in the node (until we fix that to store historic schemas - which I really want to do one day soon ;))

Even if they're unused, we need to keep them for now I'm afraid.

radix-engine/src/object_modules/metadata/package.rs
```diff
@@ -262,8 +261,9 @@ impl MetadataNativePackage {
         Ok(Own(node_id))
     }
 
+    /// This method assumes that the data has been pre-checked.
     pub fn init_system_struct(
-        data: MetadataInit,
+        data: IndexMap<Vec<u8>, (Option<Vec<u8>>, bool)>,
```
dhedey, line 266: How confident are we that these refactors are 100% backwards-compatible across Cuttlefish in terms of protocol-visible changes? e.g. we're changing which errors we return (which I think is safe I think assuming we only error in the same cases). And e.g. we aren't making any sys-calls/billing in the middle, which seems safe?

I see we've replaced some encode unwraps with e.g. `MetadataKeyValidationError::InvalidValue` - I guess we don't actually assume this to be hit, and indeed if it unwrapped previously, it shouldn't introduce any issues?

radix-engine/src/object_modules/metadata/package.rs
```diff
@@ -262,8 +261,9 @@ impl MetadataNativePackage {
         Ok(Own(node_id))
     }
 
+    /// This method assumes that the data has been pre-checked.
```
dhedey, line 264: Can we perhaps make it not `pub`?

Perhaps we should make it not `pub` and rename it to `init_system_struct_from_pre_validated_data` or something?

And then we can re-expose some higher-level `pub` method which validates and then inits?

dhedey (approved):
Nice! A couple of questions I'd like discussing before merge, but looks good :)

radix-engine-tests/assets/blueprints/metadata2/src/lib.rs
iamyulong: Will some add notes.

radix-engine-tests/benches/metadata_validation.rs
iamyulong: Happy to unwrap.

radix-engine/src/object_modules/metadata/package.rs
iamyulong: Will do.

radix-engine/src/object_modules/metadata/package.rs
iamyulong: You're right. Thought we had that error to string conversion already.

radix-engine/src/object_modules/metadata/package.rs
iamyulong: > How confident are we that these refactors are 100% backwards-compatible across Cuttlefish in terms of protocol-visible changes

I did think about this thoroughly. As far I can tell, there is only error code difference which should not affect consensus, but would appreciate another pair of eyes.

> I see we've replaced some encode unwraps with e.g. MetadataKeyValidationError::InvalidValue - I guess we don't actually assume this to be hit, and indeed if it unwrapped previously, it shouldn't introduce any issues?

This is right, considering `MetadataInit` often comes from an sbor-encoded inputs.

Keeping it in the validation logic so it's self-contained and no need to make assumption about sbor-able.

radix-engine/src/object_modules/metadata/package.rs
iamyulong: Will make it non-public.

radix-engine/src/object_modules/metadata/package.rs
dhedey: We do have error to string conversion, but to save space, it happens on the API at response serving time; and uses the _current_ `RuntimeError`. I'd like to store some historic schemas at each protocol update in the node so we can convert with these old schemas whilst still updating the errors.

radix-engine/src/object_modules/metadata/package.rs
dhedey: Sounds great. And yes, I agree that I think it's safe, so I'm happy with this.

PR #1978, opened 2024-10-25
Add base cost to consume_wasm_execution_units host function

lrubasze:
## Summary
Added WASM execution unit base to assure simple WASM calls are not being undercharged.

## Motivation

Test | Base | PR | %
-- | -- | -- | --
costing::scrypto_malloc | 592.2±0.84ms | 504.3±2.87ms | -14.84%
costing::scrypto_sbor_decode | 569.5±1.32ms | 500.3±2.76ms | -12.15%
costing::scrypto_sha256 | 392.4±1.20ms | 367.5±1.57ms | -6.35%
costing::spin_loop_v1 | 628.2±5.74ms | 474.9±2.19ms | -24.40%
costing::spin_loop_v2 | 2.7±0.00s | 959.4±0.90ms | -64.47%

Based on the benchmarks on the `base`, the `spin_loop_v2` does not finish within 1 second, despite consuming around 100 million execution cost units. 

It's made possible by iterating an empty code block (except a `br` instruction), where `consume_wasm_execution_units` host function becomes the dominant factor. 

This PR fixes it by introducing a base cost to account for the execution of the host function itself.

### DIFF
--- a/radix-engine/src/system/system_modules/costing/fee_table.rs
+++ b/radix-engine/src/system/system_modules/costing/fee_table.rs
@@ -176,9 +176,25 @@ impl FeeTable {
         _export_name: &str,
         wasm_execution_units: u32,
     ) -> u32 {
-        // From `costing::spin_loop`, it takes 5.5391 ms for 1918122691 wasm execution units.
-        // Therefore, cost for single unit: 5.5391 *  1000 / 1918122691 * 100 = 0.00028877714
-
+        // W - WASM execution units
+        // C - cost units
+        // c - single cost unit
+        // T - execution time (1 µs = 100 c => 1 ms = 100,000 c)
+        //
+        // Cost units might be expressed as
+        //  C = T * c
+        //
+        // To convert W to C, we need a d.
+        //   C = W / divider
+        //   divider = W / C
+        //   divider = W / (T * c)
+        //
+        // From `costing::spin_loop_v2`, it takes T=960 ms to consume W=274,517,401,207 wasm execution
+        // units (measured at GH benchmark, git rev 1fd85c47ef, EC2 instance type c6a.4xlarge).
+        //   T = 960 ms = 960 * 100,000
+        //   W = 274,517,401,207
+        // Therefore
+        //   divider = 274,517,401,207 / (960 * 100,000) = 2859.556 ~= 3000
         wasm_execution_units / 3000
     }
 
--- a/radix-engine-tests/assets/metering/cuttlefish/cost_crypto_utils.csv
+++ b/radix-engine-tests/assets/metering/cuttlefish/cost_crypto_utils.csv
@@ -1,18 +1,18 @@
-Total Cost (XRD)                                                           ,            1.04183166606,    100.0%
-- Execution Cost (XRD)                                                     ,                0.6651783,     63.8%
+Total Cost (XRD)                                                           ,            1.05047606606,    100.0%
+- Execution Cost (XRD)                                                     ,                0.6738227,     64.1%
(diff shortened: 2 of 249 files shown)


radix-engine/src/system/system_modules/costing/fee_table.rs
```diff
@@ -176,10 +176,11 @@ impl FeeTable {
         _export_name: &str,
         wasm_execution_units: u32,
     ) -> u32 {
-        // From `costing::spin_loop`, it takes 5.5391 ms for 1918122691 wasm execution units.
-        // Therefore, cost for single unit: 5.5391 *  1000 / 1918122691 * 100 = 0.00028877714
+        // From `costing::spin_loop`, it takes 10.824 ms for 2918134342 wasm execution units
+        // (measured at Apple M1 Max, git rev 680d511cdb)
+        // Therefore, cost for single unit: 10.824 *  1000 / 2918134342 * 100 = 0.0003709 ~= 0.0004
 
-        wasm_execution_units / 3000
```
dhedey, line 182: Something here doesn't seem right -
1 / 0.00028877714 = 3462 which rounded down to 3000
1 / 0.0003709 = 2696 which would round down to 2500, not 4000

That said, using 2500 would make wasm more expensive. Maybe a few more of us should run this to provide a larger sample size?

radix-engine/src/vm/wasm/wasmi.rs
```diff
@@ -563,6 +563,11 @@ fn consume_wasm_execution_units(
     caller: Caller<'_, HostState>,
     n: u64,
 ) -> Result<(), InvokeError<WasmRuntimeError>> {
+    // Add 10,000 base units to make sure the we do not undercharge for WASM execution,
+    // which might lead to system exploitation.
+    // This is especially important in corner-cases such as `costing::spin_loop_v2` benchmark.
+    let n = n + 10_000;
```
dhedey, line 569: It feels to me slightly better to put this in `consume_wasm_execution_units` rather than the `wasmi` adaptor.

Also this would need to be dependent on the `VmVersion`... ideally statically so we don't introduce a branch in every call in this critical code path.

radix-engine/src/system/system_modules/costing/fee_table.rs
lrubasze: I am inventing new maths to reduce WASM costs 😁 
Agree, this calculation is wrong.

However, this `5.5391 ms` duration was wrong too. Current durations at 
- Apple M1 Max is `10.8ms` which gives
`1 / 0.000563 = 1776`
- GH CI `17.8 ms` which gives
`1 / 0.000928 = 1077`

So we were undercharging for WASM.

I suggest we use CI benchmark times for calculations. Which would give us
`17.8 * 1000 / 2918134342 * 100 = 0.000609978; 1 / 0.000609978 = 1639`

radix-engine/src/vm/wasm/wasmi.rs
lrubasze: Agree. 
This is a draft PR. Firstly I wanted to check `spin_loop_v2` benchmark duration on CI machine.

radix-engine/src/system/system_modules/costing/fee_table.rs
iamyulong: As discussed, this legacy `spin_loop` has been using `NoOpWasmRuntime`, which has a different performance profile than the standard runtime. 

We've updated the benchmarks as part of #1982 - no undercharging is observed except for the `spin_loop_v2` case.

iamyulong (approved):

PR #1980, opened 2024-10-27
feature: Added `PreviewTransactionV2`

dhedey:
## Summary
Adds support for a `PreviewTransactionV2`, and the preview API in the node will use its payload bytes (a slight pivot from V1 where the preview model was taken in JSON). It can be created from the `TransactionV2Builder`.

This PR also includes various refactors to the `TransactionValidator` as part of enabling validation of `PreviewTransactionV2`s:
* Moved reference and sig count validation into the right place so it's harder to accidentally miss, and feels more natural. Generally I was trying to architect it so that it's hard to miss a check when introducing new variants (at least in v2).
* Sig validation takes preview mode as well as non-preview mode, so that the cost of sig validation can be taken into account for the cost estimate. This also allows sig validation to be performed last as it's one of the most expensive parts of validation.
* Adds a location to all the validation errors so it tells you which intent the error is in. 

And also took on another small refactor:
* Adds an `assert_matches!(..)` macro so we get better test output on failure for lots of tests.

## Testing
* Additional tests on validation of transactions with invalid signatures 
* Various preview v2 tests in `preview_v2.rs`


radix-transactions/src/validation/transaction_validator.rs
```diff
@@ -1298,6 +1274,322 @@ impl HeaderAggregation {
...
+            return Err(
+                SignatureValidationError::IncorrectNumberOfSubintentSignatureBatches
+                    .located(TransactionValidationErrorLocation::AcrossTransaction),
+            );
+        }
+        for (index, (subintent, signatures)) in non_root_subintents
+            .iter()
+            .zip(non_root_subintent_signatures)
+            .enumerate()
+        {
```
iamyulong, line 1344: We're not validating the length of `signatures`?

iamyulong (approved):

radix-transactions/src/validation/transaction_validator.rs
dhedey: If there aren't explicit tests for subintent signatures, I'll add them 👍

iamyulong: I vaguely remember that I added such a test.

dhedey: We missed a test on signature count on the root manifest, so I've added that.

PR #1981, opened 2024-10-28
Add rust-toolchain.toml to scrypto package template

iamyulong:
## Summary

Add `rust-toolchain.toml` to scrypto package template

## Testing

Existing tests and manual local test.

### DIFF
--- a/radix-clis/assets/template/rust-toolchain.toml_template
+++ b/radix-clis/assets/template/rust-toolchain.toml_template
@@ -1,0 +1,5 @@
+[toolchain]
+channel = "1.81.0"
+components = []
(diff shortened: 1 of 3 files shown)


radix-clis/assets/template/rust-toolchain.toml_template
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "1.81.0"
```
dhedey, line 2: Nice :).

Can we update the instructions in `action.yml` to include updating this when we update the Rust version? (and remove the instructions which still refer to the `README.MD` but no longer need to?)

dhedey (approved):

PR #1984, opened 2024-10-28
Feature/add nullification to receipts

dhedey:
## Summary
* Adds the committed nullifications to the receipt (with extra data which can't be extracted from the state changes - notably whether the intent hashes were Transaction intents or Subintents) - this will be used in the node for subintent status
* Improves the `basic_subintents` scenario in line with the task doc:
  * Adds a `trading_with_subintent` where a child and parent trade
  * The complex subintent now is successful, and includes all the new instructions 
  * Adds a demonstration that a subintent can fail and then be committed
* Improved the transaction builder behaviour:
  * `.build()` methods now give a `DetailedXXX` which is validated and includes hashes. This is generally more useful.
  * There is a `manifest_builder_with_lookup` method which is quite nice to use.
  * The `manifest_builder` method now doesn't validate the manifest - that will happen later in the build process anyway.

## Testing
* Adds nullification tests, including the behaviour of the subintent


iamyulong (approved):
LGTM

PR #1985, opened 2024-10-29
Misc/crypto utils cleanup

lrubasze:
Some renamings around ECDSA public key recovery methods.


radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
```diff
@@ -648,10 +648,6 @@ impl<'y, Y: SystemApi<RuntimeError>> WasmRuntime for ScryptoRuntime<'y, Y> {
             return Err(InvokeError::SelfError(WasmRuntimeError::InputDataEmpty));
         }
 
-        if self.scrypto_vm_version < ScryptoVmVersion::crypto_utils_v1() {
```
lrubasze, line 651: This is checked at WASM preparation stage (`prepare.rs - enforce_import_constraints()`)

radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
```diff
@@ -680,10 +676,6 @@ impl<'y, Y: SystemApi<RuntimeError>> WasmRuntime for ScryptoRuntime<'y, Y> {
...
```
lrubasze, line 683: ditto

radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
dhedey: 👍. You mean in `ScryptoV1WasmValidator` when the package is first uploaded?

Which is okay, because we know that WASM can't be uploaded to invoke these methods until after the right protocol update.

To make this a little more obvious, perhaps we could add a Rustdoc to these methods saying something like?:
```
/// This method is only available to packages uploaded after cuttlefish due to checks in 
/// [`ScryptoV1WasmValidator::validate`].
```

lrubasze: Sure, makes sense :)

lrubasze: Done

PR #1990, opened 2024-10-30
Add cost for transaction intent hash status

iamyulong:
## Summary

Starting from cuttlefish, Radix network will charge every transaction for checking the validity of transaction intent hash and committing its status into substate store.

This effectively increase the cost of every transaction by `0.013` XRD.

### DIFF
--- a/radix-engine/src/system/system_callback.rs
+++ b/radix-engine/src/system/system_callback.rs
@@ -1588,14 +1588,16 @@ impl<V: SystemCallbackObject> KernelTransactionExecutor for System<V> {
                 }
             }
             .and_then(|_| {
-                if match hash_nullification {
+                let charge_for_nullification_check = match hash_nullification {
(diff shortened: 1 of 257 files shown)


radix-engine/src/system/system_callback.rs
```diff
@@ -1582,27 +1588,24 @@ impl<V: SystemCallbackObject> KernelTransactionExecutor for System<V> {
                 }
             }
             .and_then(|_| {
-                match hash_nullification {
+                if match hash_nullification {
```
dhedey, line 1591: NITPICK/OPINION/OPTIONAL: I find `if match` a little hard to read - perhaps having a separate named boolean variable `let charge_for_nullification_check = match ...;` might be a little easier?

dhedey (approved):
Looks good to me! Thanks.

PR #1992, opened 2024-10-31
Use `Dynamic*Address` types in the `ManifestInput` models

0xOmarA:
## Summary

This PR changes the address types used in the `ManifestInput` models to use dynamic addresses. This fixes an issue that we found in the static analyzer (which was propagated in the toolkit) where `add_authorized_depositor` with a `NamedAddress` would fail when we attempt to convert it into a typed invocation.


radix-engine-interface/src/blueprints/resource/proof_rule.rs
```diff
@@ -25,6 +25,25 @@ pub enum ResourceOrNonFungible {
     Resource(ResourceAddress),
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Ord, PartialOrd, ManifestSbor)]
+pub enum ManifestResourceOrNonFungible {
+    NonFungible(NonFungibleGlobalId),
```
dhedey, line 30: This needs a `ManifestNonFungibleGlobalId`

radix-engine-monkey-tests/src/lib.rs
```diff
@@ -417,7 +416,7 @@ impl<T: TxnFuzzer> FuzzTest<T> {
                         TWO_RESOURCE_POOL_BLUEPRINT_IDENT,
                         TWO_RESOURCE_POOL_INSTANTIATE_IDENT,
                         TwoResourcePoolInstantiateManifestInput {
-                            resource_addresses: (pool_resource1, pool_resource2),
+                            resource_addresses: (pool_resource1.into(), pool_resource2.into()),
```
dhedey, line 419: Side note: Will be nice when we can tweak `call_function` to either take a `ManifestTwoResourcePoolInstantiateInput` or a `TwoResourcePoolInstantiateInput` ^_^

dhedey: Or even when we can rename `call_function` to `call_function_manual` and have it be `call_function(TwoResourcePoolInstantiateInput { .. })` or `call_function(ManifestTwoResourcePoolInstantiateInput { .. })`, via the `IsFunctionInput` trait we discussed in a previous PR ✨

radix-engine-tests/tests/dec_macros/Cargo.lock
dhedey, line 1: Where does this change come from? Have I/we missed source controlling this `Cargo.lock` previously? Do we need to add this lock to `./update-cargo-locks-minimally.sh`?

radix-engine/assets/blueprints/Cargo.lock
dhedey, line 1: Out of interest, what prompted this change?

This `Cargo.lock` isn't supposed to be updated because it's for the genesis faucet and this was correct as of scrypto 1 - as per `../build.sh`

radix-transaction-scenarios/src/scenarios/account_locker.rs
```diff
@@ -207,15 +207,15 @@ impl ScenarioCreator for AccountLockerScenarioCreator {
                                 state.account_rejecting_fungible_resource.unwrap(),
                                 ACCOUNT_SET_RESOURCE_PREFERENCE_IDENT,
                                 AccountSetResourcePreferenceInput {
-                                    resource_address: state.fungible_resource.unwrap(),
+                                    resource_address: state.fungible_resource.unwrap().into(),
```
dhedey, line 210: This shouldn't need an `.into()` because we use `AccountSetResourcePreferenceInput` rather than `AccountSetResourcePreferenceManifestInput`.

dhedey: SIDENOTE: In fact this suggests we should really be removing `ManifestSbor` from all the non-manifest models, to avoid this confusion. Then we would need to use `AccountSetResourcePreferenceManifestInput` here...

But happy to delay this until when we have the `HasManifestModel` trait...

radix-transactions/src/builder/manifest_namer.rs
```diff
@@ -978,25 +978,11 @@ pub trait ResolvableComponentAddress {
...
-}
-
-impl<'a> ResolvableComponentAddress for String {
-    fn resolve(self, registrar: &ManifestNameRegistrar) -> DynamicComponentAddress {
-        registrar.name_lookup().named_address_id(self).into()
-    }
-}
-
-impl<A: TryInto<DynamicComponentAddress, Error = E>, E: Debug> ResolvableComponentAddress for A {
+impl<A, E> ResolvableComponentAddress for A
```
dhedey, line 981: Out of interest, why this refactor?

We've been recently moving in the opposite direction, saving `where` bounds for only the complex bounds which require them.

Relevant context: https://github.com/radixdlt/radixdlt-scrypto/pull/1857 and [this slack thread](https://rdxworks.slack.com/archives/C071C6DB8TG/p1722037974593499) where we discussed:

> General guidelines:
> - Prefer impl T over a type parameter for input parameters (for new code, where it’s not a breaking change to remove a type parameter). A benefit is it reduces the number of type parameters where these are expected to just be inferred.
> - Prefer type parameter constraint over a where constraint where possible

radix-transactions/src/builder/manifest_namer.rs
```diff
@@ -1031,24 +1017,6 @@ impl<A: TryInto<DynamicResourceAddress, Error = E>, E: Debug> ResolvableResource
     }
 }
 
-impl<'a> ResolvableResourceAddress for &'a str {
```
dhedey, line 1034: Why have we removed these??

It means that manifest builder methods such as `take_non_fungibles_from_worktop` which opt into a `ResolvableResourceAddress` can't take a named resource address (via the name) any more?

radix-transactions/src/builder/manifest_namer.rs
```diff
@@ -1106,24 +1060,6 @@ impl<A: TryInto<DynamicGlobalAddress, Error = E>, E: Debug> ResolvableGlobalAddr
     }
 }
 
-impl<'a> ResolvableGlobalAddress for &'a str {
```
dhedey, line 1109: Ditto

radix-transactions/src/builder/manifest_namer.rs
dhedey: Maybe we should add a test to demonstrate this so someone else isn't tempted to remove these in future :D

dhedey (changes requested):
Nice! A few things I've noticed - happy to jump on a call if you'd like to talk through any of these 👍

radix-engine-tests/tests/dec_macros/Cargo.lock
0xOmarA: Agreed to move it to the tests assets folder and to add it to the cargo lock update script.

radix-engine-tests/tests/application/static_resource_movements_visitor.rs
```diff
@@ -996,26 +996,6 @@ fn aggregation_balance_change_test_cases() {
     }
 }
 
-#[test]
-fn static_analysis_on_a_vault_direct_method_succeeds() {
```
dhedey, line 1000: Why was this removed out of interest? I guess because it's pointless now, because it couldn't fail to be analyzed?

radix-engine/src/system/system_callback.rs
```diff
@@ -1596,7 +1596,6 @@ impl<V: SystemCallbackObject> KernelTransactionExecutor for System<V> {
                     IntentHashNullification::Subintent { .. }
                     | IntentHashNullification::SimulatedSubintent { .. } => true,
                 };
-
```
dhedey, line 1599: Oops this might have been me, my bad. No need to fix

dhedey (approved):
Looks good! But before merging, can we fix git history (as per slack)?

PR #1995, opened 2024-11-01
Tweak/neatenings to resolvable traits

dhedey:
## Summary
Proposes a tweak to move out the `Resolvable` traits into `radix-common`, separate from the `New` / `Referenced` / `Consumed` traits in the `ManifestBuilder`.

These traits add some functionality on top of just resolution; and can provide code-based documentation to capture additional semantics to the methods on the `ManifestBuilder`.

NOTE: The base branch may need updating to `develop` after it is merged.

## Testing
Existing tests pass.


radix-common/src/data/manifest/model/manifest_address_reservation.rs
```diff
@@ -14,6 +10,109 @@ use crate::*;
...
+    fn resolve_address_reservation(&self, name: &str) -> ManifestAddressReservation;
+}
+
+/// This trait is intended to be used as an `impl` argument in helper methods
+/// operating on manifests, to resolve a [`ManifestAddressReservation`] from a name, an id,
+/// or from itself.
+///
+/// The resolution process relies on a [`NamedManifestAddressReservationResolver`] which can
+/// provide a lookup to/from names.
+pub trait ResolvableManifestAddressReservation {
```
0xOmarA, line 30: Why do we have a `Resolvable` trait for each resolvable type that we have instead of having a single `ResolvableInto` trait? 

I tested this out a little bit and I think that I very much like having a single trait. Here's a code example that compiles:

```rust
use radix_common::prelude::*;
use radix_transactions::prelude::*;

/// An example resolver trait.
pub trait Resolver {
    fn resolve_address_reservation(
        &self,
        name: &str,
    ) -> ManifestAddressReservation;

    fn resolve_named_address(&self, name: &str) -> DynamicGlobalAddress;

    fn resolve_bucket(&self, name: &str) -> ManifestBucket;

    fn resolve_proof(&self, name: &str) -> ManifestProof;
}

// Traits to resolve from and into a type. They're written this way so that they
// mirror the `From` and `Into` traits in the standard library, we even provide
// an `Into` implementation for all `From` implementations in a very similar way
// to the standard library.

pub trait TryResolveFrom<T>
where
    Self: Sized,
{
    type Error;

    fn try_resolve_from(
        value: T,
        resolver: &impl Resolver,
    ) -> Result<Self, Self::Error>;
}

pub trait TryResolveInto<T> {
    type Error;

    fn try_resolve_into(
        self,
        resolver: &impl Resolver,
    ) -> Result<T, Self::Error>;
}

/* Generic impl of Into for all types that impl From */
impl<T, U> TryResolveInto<U> for T
where
    U: TryResolveFrom<T>,
{
    type Error = U::Error;

    fn try_resolve_into(self, resolver: &impl Resolver) -> Result<U, Self::Error> {
        U::try_resolve_from(self, resolver)
    }
}

/* All types that implement ResolveFrom also implement an infallible TryResolveFrom */
impl<T, U> TryResolveFrom<T> for U where U: ResolveFrom<T> {
    type Error = std::convert::Infallible;

    fn try_resolve_from(value: T, resolver: &impl Resolver) -> Result<Self, Self::Error> {
        Ok(value.resolve_into(resolver))
    }
}

pub trait ResolveFrom<T> {
    fn resolve_from(value: T, resolver: &impl Resolver) -> Self;
}

pub trait ResolveInto<T> {
    fn resolve_into(self, resolver: &impl Resolver) -> T;
}

/* Generic impl of Into for all types that impl From */
impl<T, U> ResolveInto<U> for T
where
    U: ResolveFrom<T>,
{
    fn resolve_into(self, resolver: &impl Resolver) -> U {
        U::resolve_from(self, resolver)
    }
}

/* Generic impl of ResolvableFrom for all types that impl From */
impl<T, U> ResolveFrom<T> for U where U: From<T> {
    fn resolve_from(value: T, _: &impl Resolver) -> Self {
        value.into()
    }
}

// An example implementation of the trait and how we can use it in resolutions.
impl<T> ResolveFrom<T> for ManifestBucket
where
    T: AsRef<str>,
{
    fn resolve_from(value: T, resolver: &impl Resolver) -> Self {
        resolver.resolve_bucket(value.as_ref())
    }
}

// We get to implement Option resolution for all `Option` types by default and
// without the need for us to implement it for each type.
impl<T, U> ResolveFrom<Option<T>> for Option<U>
where
    U: ResolveFrom<T>,
{
    fn resolve_from(value: Option<T>, resolver: &impl Resolver) -> Self {
        value.map(|value| value.resolve_into(resolver))
    }
}

// Testing out the implementation of the traits to make sure that they do
// compile.
fn example_fn() {
    fn func_that_takes_resolvable_bucket(_: impl ResolveInto<ManifestBucket>) {}
    fn func_that_takes_an_optional_resolvable_bucket(
        _: impl ResolveInto<Option<ManifestBucket>>,
    ) {
    }

    func_that_takes_resolvable_bucket("hello world!");
    func_that_takes_an_optional_resolvable_bucket(Some("hello world!"));
}
```

What I like about this is:
1. We have a single resolvable trait which seems very elegant.
2. We have an abstraction that's very similar to that of the standard library in the form of `ResolvableFrom` and `ResolvableInto` and even the default `Into` impl for all `From` that we see in the standard library.
3. It allows us to write generic implementations for things like `Option<T>`, `Result<O, E>` and other types (see the implementation for options in the example I provided above).
4. I very very much like how similar this is to the standard library and how the amount of generic implementations we can provide for all of these traits. The interface of the manifest builder would primarily make use of the try traits. As an example `fn consume_bucket(bucket: impl TryResolveInto<ManifestBucket>);`

radix-common/src/data/manifest/model/manifest_dynamic_addresses.rs
0xOmarA, line 1: Perhaps we should rename this module now that we renamed the types.

radix-common/src/data/manifest/model/manifest_expression.rs
```diff
@@ -20,25 +20,72 @@ pub enum ManifestExpression {
...
+}
+
+impl<B: ResolvableManifestBucket> ResolvableBucketBatch for Vec<B> {
+    fn resolve(self, resolver: &impl NamedManifestBucketResolver) -> ManifestBucketBatch {
+        let buckets = self.into_iter().map(|b| b.resolve(resolver)).collect();
+        ManifestBucketBatch::ManifestBuckets(buckets)
+    }
+}
+
+impl<B: ResolvableManifestBucket> ResolvableBucketBatch for BTreeSet<B> {
```
0xOmarA, line 39: We can provide a generic implementation for all types that implement `IntoIterator<Item = B>`. Example:

```rust
impl<T, B> ResolvableBucketBatch for T where T: IntoIterator<Item = B>, B: ExistingManifestBucket {
    fn consume_and_resolve(self, registrar: &ManifestNameRegistrar) -> BucketBatch {
        let buckets = self
            .into_iter()
            .map(|b| b.mark_consumed(registrar))
            .collect();
        BucketBatch::ManifestBuckets(buckets)
    }
}
```

radix-common/src/data/manifest/model/manifest_expression.rs
```diff
@@ -20,25 +20,72 @@ pub enum ManifestExpression {
...
 #[cfg_attr(feature = "fuzzing", derive(Arbitrary))]
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub enum BucketBatch {
+pub enum ManifestBucketBatch {
     ManifestBuckets(Vec<ManifestBucket>),
     EntireWorktop,
 }
 
-impl BucketBatch {
+pub trait ResolvableBucketBatch {
```
0xOmarA, line 28: Let's rename to `ResolvableManifestBucketBatch` for consistency with the other types.

radix-transactions/src/model/v2/child_subintent_hashes_v2.rs
```diff
@@ -66,6 +66,63 @@ impl From<ManifestNamedIntent> for ManifestNamedIntentIndex {
     }
 }
 
+//========
+// resolution
+//========
+
+/// This is for use with the [`ResolvableManifestNamedIntent`] trait.
+/// Implementers should panic if a bucket cannot be found.
```
0xOmarA, line 74: Minor: need to change the comment so it doesn't say bucket.

radix-transactions/src/model/v2/child_subintent_hashes_v2.rs
```diff
@@ -66,6 +66,63 @@ impl From<ManifestNamedIntent> for ManifestNamedIntentIndex {
...
+pub trait NamedManifestIntentResolver {
```
0xOmarA, line 75: We have pretty much moved all of the resolvable traits to `radix-common` except for this one. Why?

0xOmarA (commented):
I'm happy with the separation done in this PR but confused as to why the resolvable intent traits still live in the `radix-transactions` crate.

Also, and maybe this is one for the future, I hope that we consider the resolvable traits that I suggested in one of the comments in this PR.

radix-transactions/src/model/v2/child_subintent_hashes_v2.rs
0xOmarA: I'd like us to align on this first before we merge.

radix-common/src/data/manifest/model/manifest_expression.rs
dhedey: It clashes with the other generic implementation.

radix-transactions/src/model/v2/child_subintent_hashes_v2.rs
dhedey: Because `ManifestIntent` doesn't exist in `radix-common`... because it's not actually a custom type in the  manifest value model - because we didn't want to open that can of worms at Cuttlefish.

radix-common/src/data/manifest/model/manifest_dynamic_addresses.rs
dhedey: Yup! I was tempted by this too - but wanted to avoid possibly breaking the diff 👍.

radix-common/src/data/manifest/model/manifest_address_reservation.rs
dhedey: I like the idea in principle, but I think I'd want it to be more generic, i.e. we have two traits:
* `Resolve<X>` - which doesn't need a resolver.
* `ResolveFrom<T, X>` - which has the resolve method take a `Resolver<T, X>`.

I imagine we _might_ hit some issues though with overlapping blanket implementations and things; but we could try it.

dhedey: What are your thoughts on that ^? 👍

dhedey: Have just tried it, and I get conflicts between these:
```rust
impl<T, K, X, E> ResolveFrom<K, X> for T
where
    T: TryInto<X, Error = E>,
    E: Debug,
{
    fn resolve(self, _resolver: &impl KeyedResolver<K, X>) -> X {
        self
            .try_into()
            .unwrap_or_else(|err| panic!("Value was not a valid {}: {err:?}", type_name::<X>()))
    }
}
```
and:
```rust
impl<'a, X> ResolveFrom<&'a str, X> for &'a str {
    fn resolve(self, resolver: &impl KeyedResolver<&'a str, X>) -> X {
        resolver.resolve_from_key(self)
    }
}

impl<'a, X> ResolveFrom<&'a str, X> for &'a String {
    fn resolve(self, resolver: &impl KeyedResolver<&'a str, X>) -> X {
        resolver.resolve_from_key(self.as_str())
    }
}

impl<'a, X> ResolveFrom<&'a str, X> for String {
    fn resolve(self, resolver: &impl KeyedResolver<&'a str, X>) -> X {
        resolver.resolve_from_key(self.as_str())
    }
}
```

I think we might need to implement the blanket alongside each `X`:
```rust
impl<T, K, X, E> ResolveFrom<K, X> for T
where
    T: TryInto<X, Error = E>,
    E: Debug,
{
    fn resolve(self, _resolver: &impl KeyedResolver<K, X>) -> X {
        self
            .try_into()
            .unwrap_or_else(|err| panic!("Value was not a valid {}: {err:?}", type_name::<X>()))
    }
}
```

dhedey: Ahh even worse, this blanket implementation can't work -- because `KeyedResolve` is "foreign", and therefore a third crate could implement `KeyedResolve` and `TryInto<X>` on its type `T`, breaking coherence.

This explains why I originally put the trait alongside its definitions I guess?
```rust
/// ```rust
/// # struct X;
/// # type K = str;
/// # use radix_rust::prelude::*;
/// impl<T, E> KeyedResolve<X, K> for T
/// where
///     T: TryInto<X, Error = E>,
///     E: Debug,
/// {
///     fn resolve(self, _resolver: &impl KeyedResolver<K, X>) -> X {
///         self
///         .try_into()
///         .unwrap_or_else(|err| panic!("Value was not a valid {}: {err:?}", core::any::type_name::<X>()))
///     }
/// }
/// ```
```

dhedey: But perhaps we can just ignore the blanket implementation from `TryInto<X, Error = E>` in this case, and just implement the definite ones we need? This might work

radix-rust/src/resolve.rs
```diff
@@ -0,0 +1,318 @@
...
+/// because that causes implementation collisions with labels for types which could implement
+/// `TryFrom<&str>`.
+///
+/// ## Implementers
+/// * You should prefer to implement [`LabelledResolveFrom`] as it is easier to implement
+/// due to trait coherence rules. Sometimes you can only implement [`LabelledResolve`]
+/// however.
+/// * If requiring a labelled resolution in your bounds, prefer [`LabelledResolve`]
+/// because slightly more types can implement it.
+pub trait LabelledResolve<Y: LabelledResolvable> {
```
0xOmarA, line 104: It might be useful to provide a `LabelledResolve` implementation for all types that implement `Resolve` which would allow us to use this trait in all of our interfaces.

radix-rust/src/resolve.rs
```diff
@@ -0,0 +1,318 @@
...
+// - An iterator over something that resolves to X, resolves to
+//   the given collection/s of X.
+// Feel free to add more collections here as needed.
+//==============================================================
+
+impl<X: LabelledResolvable> LabelledResolvable for Vec<X> {
+    type ResolverOutput = X;
+}
+
+impl<'a, T: IntoIterator, X: LabelledResolvable<ResolverOutput = X>> LabelledResolveFrom<T>
```
0xOmarA, line 287: Minor: I understand that this is a matter of taste but:

1. To me these generic bounds are too complex to just write inline.
2. We're using a where clause already, we might as well just define the generic bounds there.

radix-rust/src/resolve.rs
```diff
@@ -0,0 +1,318 @@
...
+            .collect()
+    }
+}
+
+impl<X: LabelledResolvable> LabelledResolvable for IndexSet<X> {
+    type ResolverOutput = X;
+}
+
+impl<
+        'a,
```
0xOmarA, line 305: Unused lifetime?

0xOmarA (approved):

radix-rust/src/resolve.rs
dhedey: It's not possible because of clashing blanket implementations

radix-rust/src/resolve.rs
dhedey: I'm happy if you want to bring up this discussion as a team so we can revise our guidance about how to balance using bounds-at-definition, impl arguments and where clauses :)

PR #2003, opened 2024-11-15
Add uncompressed public key recover

iamyulong:


radix-common/src/crypto/signature_validator.rs
```diff
@@ -22,6 +22,28 @@ pub fn verify_and_recover_secp256k1(
     None
 }
 
+#[cfg(feature = "secp256k1_sign_and_validate")]
+pub fn verify_and_recover_secp256k1_uncompressed(
+    signed_hash: &Hash,
+    signature: &Secp256k1Signature,
+) -> Option<[u8; secp256k1::constants::UNCOMPRESSED_PUBLIC_KEY_SIZE]> {
```
dhedey, line 29: Is it worth us adding an explicit new type for this, e.g. `ExpandedSecp256k1PublicKey` - which we can use here and in scrypto? (And can be a semantic transparent wrapper around a `pub [u8; 65]`)

dhedey: And maybe adding to the Rust doc for `Secp256k1PublicKey` that it's the compressed public key, which is the default format used in the Radix stack?

radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
```diff
@@ -794,4 +794,25 @@ impl<'y, Y: SystemApi<RuntimeError>> WasmRuntime for ScryptoRuntime<'y, Y> {
...
+        &mut self,
+        message: Vec<u8>,
+        signature: Vec<u8>,
+    ) -> Result<Buffer, InvokeError<WasmRuntimeError>> {
+        let hash = Hash::try_from(message.as_slice()).map_err(WasmRuntimeError::InvalidHash)?;
+        let signature = Secp256k1Signature::try_from(signature.as_ref())
+            .map_err(WasmRuntimeError::InvalidSecp256k1Signature)?;
+
+        self.api
+            .consume_cost_units(ClientCostingEntry::Secp256k1EcdsaKeyRecover)?;
```
lrubasze, line 811: Is this `Secp256k1EcdsaKeyRecover` variant used here deliberately?
I guess the costs for uncompressed variant are similar, so this probably makes sense.

dhedey (approved):

radix-engine/src/vm/wasm/wasmi.rs
```diff
@@ -910,6 +910,29 @@ fn secp256k1_ecdsa_verify_and_key_recover(
...
+    let signature = read_memory(
+        caller.as_context_mut(),
+        memory,
+        signature_ptr,
+        signature_len,
+    )?;
+
+    runtime
+        .crypto_utils_secp256k1_ecdsa_verify_and_key_recover_uncompressed(message, signature)
+        .map(|buffer| buffer.0)
```
CyonAlexRDX, line 933: what does this method do? returns `u64` and not a public key? why do we call `buffer.0`? should we not return the full buffer and update the return type to be `Result<[u8: 64], _>`?

radix-engine/src/vm/wasm_runtime/scrypto_runtime.rs
iamyulong: I saw this too. The uncompressed version should be cheaper, but didn't bother changing it.

radix-common/src/crypto/signature_validator.rs
iamyulong: I will add the wrapper.

radix-engine/src/vm/wasm/wasmi.rs
iamyulong: We can not pass the array raw to wasm as a function return, and have to wrap it into a buffer (which will be read by wasm later).

The wasmi linker requires a u32/u64 return type, so we had to unwrap the `Buffer`.

CyonAlexRDX: so the caller of this function reads out the 65 bytes how? via `mut caller`? I lack the WASM knowledge to understand this :D but just looked wrong. I assume it is some WASM vodoo magic making this work...

CyonAlexRDX: wrote simultaneously... thx makes sense

PR #2009, opened 2024-11-21
Feature/fix fuzzing build

iamyulong:
## Summary

Fix `fuzzing` feature build.

### DIFF
--- a/radix-engine-interface/src/blueprints/resource/non_fungible/non_fungible_resource_manager.rs
+++ b/radix-engine-interface/src/blueprints/resource/non_fungible/non_fungible_resource_manager.rs
@@ -420,15 +420,15 @@ impl<'a> Arbitrary<'a> for NonFungibleDataSchema {
     // ScryptoSchema, therefore implementing arbitrary by hand.
     // TODO: Introduce a method that genearates NonFungibleDataSchema in a truly random manner
     fn arbitrary(_u: &mut Unstructured<'a>) -> Result<Self> {
-        Ok(Self::Local {
-            schema: VersionedSchema::V1(SchemaV1 {
+        Ok(Self::Local(LocalNonFungibleDataSchema {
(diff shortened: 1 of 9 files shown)


run-code-coverage.sh
```diff
@@ -3,6 +3,12 @@
 set -x
 set -e
 
-CARGO_INCREMENTAL=0 RUSTFLAGS='-Cinstrument-coverage' LLVM_PROFILE_FILE='cargo-test-%p-%m.profraw' cargo test --features compile-blueprints-at-build-time --no-fail-fast
```
dhedey, line 6: Out of interest, we seem to have removed `compile-blueprints-at-build-time` - does it still work? I guess so :D.
Perhaps we just miss some coverage around the scrypto compiler, but we can ignore that.

radix-engine-interface/src/blueprints/resource/non_fungible/non_fungible_resource_manager.rs
```diff
@@ -420,15 +420,15 @@ impl<'a> Arbitrary<'a> for NonFungibleDataSchema {
     // ScryptoSchema, therefore implementing arbitrary by hand.
     // TODO: Introduce a method that genearates NonFungibleDataSchema in a truly random manner
     fn arbitrary(_u: &mut Unstructured<'a>) -> Result<Self> {
-        Ok(Self::Local {
-            schema: VersionedSchema::V1(SchemaV1 {
+        Ok(Self::Local(LocalNonFungibleDataSchema {
```
dhedey, line 423: SIDE NOTE (unimportant): It seems to be that `scrypto_decode(unstructured.bytes()).map_err(...)` might be better than this, but maybe it converges too slowly to a valid value.

dhedey (approved):

run-code-coverage.sh
iamyulong: I ran into some issue with pre-build.

PR #2015, opened 2024-11-25
Add further transaction validation tests

dhedey:
## Summary
Covering:
* Each `HeaderValidationError`
* Each `InvalidMessageError`
* Each untested `SignatureValidationError`
* Each `ManifestValidationError`
* Each `PrepareError`
* Also a test to cover the display of all new compiler errors

Also had a few minor fixes:
* Removed a couple of unused `HeaderValidationError`s
* Fixed trailing comma handling in the `assert_matches!` macro
* Fixed all the other compiler errors to use plain text encoding, so they can be viewed more easily in diffs/editors

### DIFF
--- a/radix-transactions/src/errors.rs
+++ b/radix-transactions/src/errors.rs
@@ -3,11 +3,9 @@ use sbor::*;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum HeaderValidationError {
-    UnknownVersion(u8),
     InvalidEpochRange,
     InvalidTimestampRange,
     InvalidNetwork,
-    InvalidCostUnitLimit,
     InvalidTip,
     NoValidEpochRangeAcrossAllIntents,
     NoValidTimestampRangeAcrossAllIntents,
--- a/radix-transactions/src/model/preparation/decoder.rs
+++ b/radix-transactions/src/model/preparation/decoder.rs
@@ -4,10 +4,10 @@ use sbor::*;
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum ValueType {
     Blob,
-    Message,
     Subintent,
-    ChildIntentConstraint,
-    IntentSignatures,
+    ChildSubintentSpecifier,
+    SubintentSignatureBatches,
+    // Too many signatures is captured at validation time
 }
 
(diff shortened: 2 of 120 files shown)


radix-transactions/src/errors.rs
```diff
@@ -3,11 +3,9 @@ use sbor::*;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum HeaderValidationError {
-    UnknownVersion(u8),
```
dhedey, line 6: These were unused

radix-transactions/src/model/preparation/decoder.rs
```diff
@@ -4,10 +4,10 @@ use sbor::*;
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum ValueType {
     Blob,
-    Message,
```
dhedey, line 7: Message was unused

radix-transactions/tests/test_manifest_compiler_error_diagnostics.rs
```diff
@@ -1,15 +1,19 @@
...
 
-        let err = compile(manifest, &NetworkDefinition::simulator(), $blob_provider).unwrap_err();
-
-        let x = compile_error_diagnostics(manifest, err, $style);
+        let x = compile_any_manifest_with_pretty_error(
+            manifest,
+            $manifest_kind,
+            &NetworkDefinition::simulator(),
+            $blob_provider,
+            CompileErrorDiagnosticsStyle::PlainText,
```
lrubasze, line 14: I very much like the idea of switching to plain text, diagnostic files are much more readable.
But maybe it would make sense to leave 1 or 2 tests with TextTerminalColors just to makes sure it works as expected, when we eg. update `annotate-snippets` crate.

lrubasze (approved):
LGTM

radix-transactions/tests/test_manifest_compiler_error_diagnostics.rs
dhedey: I'd support adding a separate explicit test for this, but that said:
* I'd hope `annotate-snippets` test `Renderer::styled()` themselves, so I don't feel strongly that we need to test it
* I doubt we'll be updating `annotate-snippets` regularly

PR #2016, opened 2024-11-25
Fix manifest parsing

lrubasze:
## Summary
Fix parsing of manifest values within parenthesis with expected length of 1

## Tests
Test added

### DIFF
--- a/radix-transactions/src/manifest/parser.rs
+++ b/radix-transactions/src/manifest/parser.rs
@@ -1013,7 +1013,18 @@ impl Parser {
         open: Token,
         close: Token,
     ) -> Result<Vec<ValueWithSpan>, ParserError> {
-        self.advance_exact(open)?;
+        self.parse_values_any_with_open_close_spans(open, close)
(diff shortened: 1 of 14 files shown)


radix-transactions/src/manifest/parser.rs
```diff
@@ -1013,22 +1013,48 @@ impl Parser {
...
+        match values.len() {
+            1 => Ok(values[0].clone()),
+            0 => Err(ParserError {
+                error_kind: ParserErrorKind::InvalidNumberOfValues {
+                    actual: values.len(),
+                    expected: 1,
+                },
+                // no values returned, use opening and closing spans
+                span: Span {
+                    start: open_span.end,
```
dhedey, line 1053: This is fine as-is, but I also wouldn't mind using `open_span.end` through `close_span.start` in the wildcard `n >= 2` case too, which I guess would let us condense the code a little?

dhedey (approved):

PR #2030, opened 2024-11-27
CI fix + update Rust version comments

lrubasze:
- Warn (instead of return error) when no tests to run for `nextest`
(nextest changed default behaviour since version `0.9.85`)
- Update Rust version comments

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -27,7 +27,9 @@ runs:
         components: rustfmt
 
     - name: Install nextest
-      uses: RDXWorks-actions/install-action@nextest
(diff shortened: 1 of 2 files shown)


.github/actions/setup-env/action.yml
```diff
@@ -8,7 +8,7 @@ runs:
       uses: RDXWorks-actions/toolchain@master
       with:
         # IMPORTANT:
-        # To update the Rust version, you must update it in five total places (including the toolchain below):
+        # To update the Rust version, you must update it in below places (including the toolchain below) in this repo:
```
dhedey, line 11: `in the below places`*

dhedey (approved):

PR #2042, opened 2024-12-18
Feature/support non fungibles in execution trace

lrubasze:
## Summary

- Add support for non-fungible resources in `ExecutionTrace` resource changes 
- Removed support for Recall operations in `ExecutionTrace` resource changes
With `recall` it is not possible to determine the account the resources are recalled from, because `recall` operates with internal vault.

## Testing
Relevant tests added

### DIFF
--- a/radix-engine/src/system/system_modules/execution_trace/module.rs
+++ b/radix-engine/src/system/system_modules/execution_trace/module.rs
@@ -91,8 +91,10 @@ impl From<&BucketSnapshot> for ResourceSpecifier {
 
 #[derive(Debug, Clone)]
 pub enum VaultOp {
-    Put(ResourceAddress, Decimal), // TODO: add non-fungible support
+    Put(ResourceAddress, Decimal),
     Take(ResourceAddress, Decimal),
+    // We intentionally disregard IDs and only use the amount for non-fungibles.
+    // This maintains backward compatibility for users of the `ResourceChange` struct.
     TakeNonFungibles(ResourceAddress, Decimal),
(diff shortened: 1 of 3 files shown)


radix-engine/src/system/system_modules/execution_trace/module.rs
```diff
@@ -91,10 +91,12 @@ impl From<&BucketSnapshot> for ResourceSpecifier {
...
 #[derive(Debug, Clone)]
 pub enum VaultOp {
-    Put(ResourceAddress, Decimal), // TODO: add non-fungible support
+    Put(ResourceAddress, Decimal),
     Take(ResourceAddress, Decimal),
+    // We intentionally disregard IDs and only use the amount for non-fungibles.
+    // This maintains backward compatibility for users of the `ResourceChange` struct.
+    TakeNonFungibles(ResourceAddress, Decimal),
     TakeAdvanced(ResourceAddress, Decimal),
-    Recall(ResourceAddress, Decimal),
```
dhedey, line 97: Is it worth mentioning that Recall is not supported (and why)?
And perhaps (somewhere) that this feature isn't comprehensive/a good abstraction, and will be deprecated in the node's Core API at Dugong, with the intention to have it removed at a later point?

dhedey (approved):
Great work, thanks :)

radix-engine/src/system/system_modules/execution_trace/module.rs
lrubasze: Done :)

PR #2043, opened 2024-12-19
Improve logging macros to work with IDEs better

lrubasze:
## Summary
Improve logging macros to work with IDEs better

## Testing
Respective tests added

## Update Recommendations
Change is backward-compatible. No actions required

### DIFF
--- a/scrypto/src/macros.rs
+++ b/scrypto/src/macros.rs
@@ -23,6 +23,9 @@ macro_rules! error {
         // syntactical errors or provide hints and completions.
         // Using `false` ensures the expression is optimized out in case of expressions
         // that mutate data or have side effects.
+        // As of now, the linter does not complain about unreachable code, but if it does
+        // in the future, we should address it proactively.
+        #[allow(unreachable_code)]
         if false {
             let _ = ($($args),+);
         }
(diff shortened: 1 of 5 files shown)


scrypto/src/macros.rs
```diff
@@ -17,7 +17,16 @@ macro_rules! error {
...
 #[macro_export]
 macro_rules! error {
-    ($($args: expr),+) => {{}};
+    ($($args: expr),+) => {{
+        // The operation below does nothing at runtime but consumes the expressions,
+        // allowing tools (e.g., rust-analyzer) to parse them correctly to identify
+        // syntactical errors or provide hints and completions.
+        // Using `false` ensures the expression is optimized out in case of expressions
+        // that mutate data or have side effects.
+        if false {
```
dhedey, line 26: I'm surprised this doesn't give an unreachable code lint. I wouldn't be surprised if such a lint were added some time. Perhaps we should make it this:
```rust
#[allow(unreachable_code)]
if false {
    let _ = ::scrypto::prelude::sbor::rust::format!($($args),+);
}
```

dhedey (approved):
Nice!

scrypto/src/macros.rs
```diff
@@ -17,7 +17,16 @@ macro_rules! error {
...
+            let _ = ($($args),+);
```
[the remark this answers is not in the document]
lrubasze: Good point! Done

scrypto/src/macros.rs
lrubasze: Yeah, that is pretty surprising :) Done

PR #2058, opened 2025-01-19
Add tests for the typed invocations type

0xOmarA:
## Summary

This PR adds tests to ensure that the `TypedManifestNativeInvocation` type has all of the functions and methods that are defined on blueprints that we support. This is because clients such as the toolkit are heavily relying on this and it must cover all functions

## Tests

* Added a test that does NOT check that all blueprints are supported but that all functions, methods, and direct methods of the blueprints we've supported are supported.

### DIFF
--- a/radix-common/src/lib.rs
+++ b/radix-common/src/lib.rs
@@ -87,7 +87,7 @@ pub mod prelude {
 
 pub(crate) mod internal_prelude {
     pub use super::prelude::*;
-    pub use crate::*;
+    pub use crate::manifest_type;
     pub use sbor::representations::*;
     pub use sbor::traversal::*;
 }
--- a/radix-common/src/lib.rs
+++ b/radix-common/src/lib.rs
@@ -87,6 +87,7 @@ pub mod prelude {
 
 pub(crate) mod internal_prelude {
     pub use super::prelude::*;
+    pub use crate::*;
     pub use sbor::representations::*;
     pub use sbor::traversal::*;
 }
--- a/radix-engine-interface/src/blueprints/component.rs
+++ b/radix-engine-interface/src/blueprints/component.rs
@@ -1,5 +1,5 @@
+use core::hash::{Hash, Hasher};
 use core::marker::PhantomData;
-
 use radix_common::prelude::*;
 
 pub trait TypeInfoMarker {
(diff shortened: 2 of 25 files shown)


radix-common/src/lib.rs
```diff
@@ -87,6 +87,7 @@ pub mod prelude {
 
 pub(crate) mod internal_prelude {
     pub use super::prelude::*;
+    pub use crate::*;
```
dhedey, line 90: I believe this creates an infinite import possibilities loop which slowed rust-analyzer to a crawl before?

Can we instead be more precise about what to import?

radix-transactions/src/manifest/static_resource_movements/typed_invocation.rs
```diff
@@ -313,6 +314,19 @@ macro_rules! define_manifest_typed_invocation {
                 }
             }
             pub use uniform_match_on_manifest_typed_invocation;
+
+            /// This is a function that's there for testing purposes only.
+            pub fn typed_native_invocation_function_table() -> HashMap<&'static str, HashMap<&'static str, HashSet<&'static str>>> {
```
dhedey, line 319: Putting this behind `#[cfg(test)]` or `#[doc(hidden)]` is typical if it's just exposed for testing purposes.

radix-engine-tests/tests/misc/typed_manifest_invocation.rs
```diff
@@ -0,0 +1,189 @@
...
+                            .difference(typed_invocations_blueprint_function_type_functions)
+                            .collect::<HashSet<_>>();
+
+                    if !functions_in_typed_invocation_but_not_in_ledger.is_empty() {
+                        panic!("For blueprint \"{blueprint_name:?}\" and function type \"{function_type:?}\" the following set of functions are defined in the typed invocation enum but are not present on ledger: {functions_in_typed_invocation_but_not_in_ledger:?}")
+                    }
+                    if !functions_in_ledger_but_not_in_typed_invocation.is_empty() {
+                        panic!("For blueprint \"{blueprint_name:?}\" and function type \"{function_type:?}\" the following set of functions are defined in ledger but are not present in the typed invocations enum: {functions_in_ledger_but_not_in_typed_invocation:?}")
+                    }
+                }
```
dhedey, line 76: It feels like we shouldn't stop here, and we should actually check that they also have the same schema if possible? :)

We can do this by storing a `ScryptoSingleTypeSchema` from each, and then:

```rust
compare_single_type_schemas(
    SchemaComparisonSettings::require_equality(),
    blueprint_schema,
    typed_schema,
).assert_valid("blueprint definition", "typing definition");
```

dhedey (changes requested):
Nice work!

Slightly worried about the infinite import loop issue, and have a suggestion for adding an extra layer of use for the tests, so marking as `Request changes` for now.

radix-common/src/lib.rs
0xOmarA: Good point! I've resolved this.

radix-transactions/src/manifest/static_resource_movements/typed_invocation.rs
0xOmarA: This doesn't work since the `radix-transactions` is not compiled with the test flag in tests: https://users.rust-lang.org/t/how-to-import-a-function-from-tests-in-another-file/88981/11?u=0xomara

dhedey: Did you try `#[doc(hidden)]`?

radix-engine-tests/tests/blueprints/account_locker.rs
```diff
@@ -196,7 +196,7 @@ fn store_can_only_be_called_by_storer_role() {
                         ACCOUNT_LOCKER_STORE_IDENT,
                         AccountLockerStoreManifestInput {
                             bucket,
-                            claimant: account.into(),
+                            claimant: account.into_manifest_address().into(),
```
dhedey, line 199: It might be nice to add a `From<ComponentAddress> for Global<ManifestComponentAddress, _>`

radix-engine-interface/src/blueprints/component.rs
```diff
@@ -9,84 +9,73 @@ pub trait TypeInfoMarker {
...
+    Ord,
+    ScryptoEncode,
+    ScryptoDecode,
+    ScryptoCategorize,
+    ManifestEncode,
+    ManifestDecode,
+    ManifestCategorize,
+)]
+#[sbor(transparent, child_types = "A")]
+pub struct GenericGlobal<A, M>(pub A, #[sbor(skip)] PhantomData<M>)
```
dhedey, line 32: :D nice advanced SBOR annotations here 👍

sbor/src/schema/schema_comparison/schema_comparison_settings.rs
```diff
@@ -137,6 +137,36 @@ impl SchemaComparisonCompletenessSettings {
             allow_compared_to_have_more_root_types: false,
         }
     }
+
+    pub const fn with_allow_root_unreachable_types_in_base_schema(mut self) -> Self {
```
dhedey, line 141: Some of these are missing `const` :)

dhedey (approved):
Nice 👍

PR #2090, opened 2026-03-20
Support Dynamic Resources in Static Analyzer

BusyWritingCode:
## Summary

The static resource movements analyzer previously only understood resources with known, pre-existing `ResourceAddress` values. When a manifest created a new resource within the transaction itself (via `allocate_global_address` + one of the `*CreateWithInitialSupply` calls) and then performed operations on it using named addresses (withdrawals, deposits, mints, etc.), the analyzer had no way to track those resources. They fell through to the "unspecified resources" path, which eventually caused a panic: `"Account withdraw output should not have unspecified resources"`.

This was hit in production, the [radix-engine-toolkit](https://github.com/radixdlt/radix-engine-toolkit) uses this analyzer, and when the Gateway API's aggregator encountered one of these manifests on mainnet, it panicked and stopped aggregating.

### What changed

**New `AnalyzerResourceAddress` type**, A new enum that can represent both static (on-ledger, known `ResourceAddress`) and dynamic (created within the manifest being analyzed) resources:

```rust
pub enum AnalyzerResourceAddress {
    Static(ResourceAddress),
    Dynamic {
        blueprint_id: BlueprintId,
        named_address: u32,
    },
}
```

For dynamic resources, the `blueprint_id` identifies the creating blueprint (e.g. `FungibleResourceManager`) and `named_address` ties back to the manifest-level named address. This gives us enough information to determine fungibility and track the resource across instructions without knowing its final on-ledger address.

**All core types updated**, `TrackedResources`, `ResourceBounds`, `AccountWithdraw`, `AccountDeposit`, `NetWithdraws`, `NetDeposits`, `ManifestResourceConstraints`, and related types all changed from `ResourceAddress` to `AnalyzerResourceAddress`. A `From<ResourceAddress>` impl and `impl Into<AnalyzerResourceAddress>` parameters keep the API ergonomic for the common static-address case.

**Address reservation tracking in the visitor**, The `StaticResourceMovementsVisitor` now tracks address reservations (mapping reservation -> (blueprint, named_address)) alongside named addresses. This is populated from `OnNewNamedAddress` events when the address has an associated reservation.

**Effect implementations updated for dynamic resolution**, Previously, any invocation with a `ManifestResourceAddress::Named(...)` would bail out to "unspecified resources". Now, the analyzer resolves named addresses through the tracked named addresses and address reservations maps to produce a proper `AnalyzerResourceAddress::Dynamic`. This covers:

- Account operations: `AccountWithdraw`, `AccountWithdrawNonFungibles`, `AccountLockFeeAndWithdraw`, `AccountLockFeeAndWithdrawNonFungibles`
- Account locker operations: `AccountLockerRecover`, `AccountLockerRecoverNonFungibles`, `AccountLockerClaim`, `AccountLockerClaimNonFungibles`
- Pool operations: `TwoResourcePoolProtectedWithdraw`, `MultiResourcePoolProtectedWithdraw`
- Resource creation: `FungibleResourceManagerCreateWithInitialSupply`, `NonFungibleResourceManagerCreateWithInitialSupply`, `NonFungibleResourceManagerCreateRuidWithInitialSupply`, these were previously lumped in with "unknown output" invocations and now have proper implementations that resolve the address reservation to a dynamic resource address
- Minting on dynamic resources: `FungibleResourceManagerMint`, `NonFungibleResourceManagerMint`, `NonFungibleResourceManagerMintRuid`, `NonFungibleResourceManagerMintSingleRuid`, when called on a named address receiver, these now produce tracked output with the correct dynamic address

If a named address can't be resolved (i.e. it wasn't set up through `allocate_global_address`), the analyzer gracefully falls back to the existing "unspecified resources" behavior.

## Testing

- Added a new test (`dynamic_resource_create_deposit_withdraw_redeposit_is_correctly_classified`) that exercises the full lifecycle: allocate a global address, create a fungible resource with initial supply using that reservation, deposit to an account, withdraw via the named address, and redeposit. The test verifies that all withdrawals, deposits, net withdrawals, and net deposits are correctly tracked using `AnalyzerResourceAddress::Dynamic`.
- All existing static analyzer tests were updated to work with `AnalyzerResourceAddress` (adding `.into()` conversions where needed) and continue to pass.
- Beyond the in-repo tests: all transactions available on mainnet and stokenet were pulled and run through the analyzer to validate the core assumption that static analysis succeeds for all successful transactions. This dataset is too large to include in the repository but confirmed there are no regressions and that the fix resolves the production failures.


GhenadieVP (approved):

LukasGasior1 (approved):
LGTM!

PR #2092, opened 2026-06-03
feat(crypto): BLS12-381 min-sig (G1) signature verification host function

a137x:
## Summary

Adds a native host function for verifying BLS12-381 signatures in the
**minimal-signature-size (min-sig)** variant — 96-byte public keys in **G2** and
48-byte signatures in **G1**, under the ciphersuite
`BLS_SIG_BLS12381G1_XMD:SHA-256_SSWU_RO_NUL_`.

This is the mirror image of the existing min-pk function
(`bls12381_v1_verify` — 48-byte G1 keys, 96-byte G2 signatures, `…G2…_POP_` DST).
The engine already links `blst`, so the implementation reuses the same dependency
via its `min_sig` module; no new dependencies are added.

## Motivation

The min-sig variant is the scheme used by **unchained drand networks** — most
notably **quicknet** (`bls-unchained-g1-rfc9380`), drand's production beacon for
fast (3s), low-overhead, publicly-verifiable randomness, and the beacon underlying
drand timelock encryption (tlock).

Today a Scrypto blueprint cannot verify a drand quicknet beacon on-chain:

- quicknet signatures live in **G1** with a **NUL** (basic) DST, whereas the
  existing `bls12381_v1_verify` only accepts **G2** signatures with the **POP** DST; and
- verifying even a single G1 signature in WASM is prohibitively expensive in
  execution units.

A native host function closes both gaps and lets on-chain protocols consume drand
randomness directly (and verify the beacon-signature component of tlock): verifiable
lotteries and raffles, randomized selection/sortition, sealed-bid and commit-reveal
mechanisms, and anything that wants tamper-evident external randomness without
trusting a centralized oracle.

## What's included

- **`radix-common`** — new SBOR types `Bls12381G2PublicKey` (96B) and
  `Bls12381G1Signature` (48B), and `verify_bls12381_v1_min_sig` (`blst::min_sig`,
  with the public key validated in-group / non-identity since it is untrusted input).
- **Host ABI** — `CryptoUtils::bls12381_v1_verify_min_sig` + `wasm_api` extern, the
  `WasmRuntime` trait method + the scrypto and no-op runtime impls, the `wasmi` host
  function / `Func::wrap` / `linker_define` binding, and the function-name constant.
- **Costing** — `ClientCostingEntry` / `ExecutionCostingEntry` variants, the
  conversion, and a `fee_table` cost function.
- **Protocol gating** — `prepare.rs` import constraint behind `crypto_utils_v3`, a
  matching `ProtocolVersionMismatch` test case, and `ScryptoVmVersion::V1_3` /
  `crypto_utils_v3()`.
- **Dugong** — a disabled-by-default `vm_boot_to_enable_crypto_utils_v3` flash.
- **Tests** — a new `crypto_scrypto_v3` test blueprint plus flash and system
  integration tests.

## Protocol gating

The new import is rejected below `crypto_utils_v3()` in `prepare.rs`, exactly like
the existing v1/v2 crypto host functions. `crypto_utils_v3` maps to a new
`ScryptoVmVersion::V1_3`, introduced by the **Dugong** protocol update as a
**disabled-by-default** VM-boot flash — consistent with Dugong currently shipping
un-enacted.

`ScryptoVmVersion::latest()` advances to **V1_3**. Note this is the **dev-time
validator ceiling** (so packages can be compiled and validated against the new
import); the **runtime** VM version is governed by the `VmBoot` substate and remains
at cuttlefish / V1_2 until the Dugong flash is enabled. Nothing is enacted by
default, and the cost-breakdown snapshots and existing crypto tests are unchanged.

## Testing

- Native unit tests (`radix-common`) and an end-to-end integration test
  (`radix-engine-tests`) verify against **real drand quicknet beacons** (rounds
  21180130 / 21180131, fetched from `api.drand.sh`), both at the `verify_*` level and
  through a published blueprint on a V1_3-booted ledger.
- Gating is proven in both directions: publishing the v3 blueprint is rejected
  without the VM-boot flash and succeeds with it (`flash` tests), and the
  `ProtocolVersionMismatch` unit test confirms the import is rejected below V1_3.
- `cargo check --workspace --tests` is clean; the metering cost-breakdown snapshots
  (`run_cost_tests`) and the existing crypto suites pass unchanged.

## Caveats / open decisions for reviewers

1. **Cost function is a placeholder.** `bls12381_v1_verify_min_sig_cost` currently
   mirrors the min-pk coefficients. min-sig has a cheaper hash-to-curve (G1) but a
   more expensive G2 public-key subgroup check, so the true cost differs. A
   calibration test (`test_crypto_scrypto_verify_bls12381_v1_min_sig_costing`,
   `resource_tracker` feature) is included; the constant should be set from its
   measured output before enactment. The function is deterministic, so this is a
   fee-accuracy matter, not a consensus one.
2. **Verify-only.** No aggregate / fast-aggregate variants — drand quicknet publishes
   a single, already-aggregated threshold signature per round, so single verification
   suffices. Aggregate variants can be added later if a use case needs them.
3. **Where to enact V1_3.** This introduces V1_3 via a disabled Dugong flash;
   maintainers may prefer a dedicated post-Dugong protocol update. That choice is
   independent of the host function, gating, and tests here.

## Notes

- The min-sig public key (`Bls12381G2PublicKey`, 96B) and signature
  (`Bls12381G1Signature`, 48B) are new SBOR types, distinct from the min-pk
  `Bls12381G1PublicKey` / `Bls12381G2Signature`.
- The public key is validated during verification (stricter than libraries that
  pass a trusted, hard-coded beacon key), since in the host-function context it is
  supplied as untrusted blueprint input.

### DIFF
--- a/radix-common/src/crypto/bls12381/mod.rs
+++ b/radix-common/src/crypto/bls12381/mod.rs
@@ -1,7 +1,11 @@
 mod private_key;
 mod public_key;
+mod public_key_g2;
 mod signature;
+mod signature_g1;
 
 pub use private_key::*;
 pub use public_key::*;
+pub use public_key_g2::*;
 pub use signature::*;
+pub use signature_g1::*;
--- a/radix-common/src/crypto/bls12381/public_key_g2.rs
+++ b/radix-common/src/crypto/bls12381/public_key_g2.rs
@@ -1,0 +1,98 @@
+use crate::internal_prelude::*;
+
+/// Represents a BLS12-381 G2 public key (96 bytes).
+///
+/// This is the public key of the BLS12-381 "minimal-signature-size" (min-sig) variant,
+/// where signatures live in G1 (48 bytes) and public keys in G2 (96 bytes).
+/// It is the variant used by unchained drand networks (e.g. quicknet) and is
+/// paired with [`Bls12381G1Signature`] and the
+/// [`crate::crypto::BLS12381_MIN_SIG_CIPHERSITE_V1`] domain separation tag.
+///
+/// Note: this is distinct from [`Bls12381G1PublicKey`], which is the public key of the
+/// "minimal-pubkey-size" (min-pk) variant used by [`verify_bls12381_v1`].
+#[cfg_attr(feature = "serde", derive(::serde::Serialize, ::serde::Deserialize))]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Sbor)]
+#[sbor(transparent)]
+pub struct Bls12381G2PublicKey(
+    #[cfg_attr(feature = "serde", serde(with = "hex::serde"))] pub [u8; Self::LENGTH],
+);
+
+impl Bls12381G2PublicKey {
+    pub const LENGTH: usize = 96;
+
+    pub fn to_vec(&self) -> Vec<u8> {
+        self.0.to_vec()
+    }
+}
+
+impl TryFrom<&[u8]> for Bls12381G2PublicKey {
+    type Error = ParseBlsPublicKeyError;
+
+    fn try_from(slice: &[u8]) -> Result<Self, Self::Error> {
+        if slice.len() != Self::LENGTH {
+            return Err(ParseBlsPublicKeyError::InvalidLength(slice.len()));
+        }
+
+        Ok(Self(copy_u8_array(slice)))
+    }
+}
+
+impl AsRef<Self> for Bls12381G2PublicKey {
+    fn as_ref(&self) -> &Self {
+        self
+    }
+}
+
+impl AsRef<[u8]> for Bls12381G2PublicKey {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+//======
+// text
+//======
+
+impl FromStr for Bls12381G2PublicKey {
+    type Err = ParseBlsPublicKeyError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let bytes = hex::decode(s).map_err(|_| ParseBlsPublicKeyError::InvalidHex(s.to_owned()))?;
+        Self::try_from(bytes.as_slice())
+    }
+}
+
+impl fmt::Display for Bls12381G2PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(f, "{}", hex::encode(self.to_vec()))
+    }
+}
+
+impl fmt::Debug for Bls12381G2PublicKey {
+    fn fmt(&self, f: &mut fmt::Formatter) -> Result<(), fmt::Error> {
+        write!(f, "{}", self)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use sbor::rust::str::FromStr;
+
+    #[test]
+    fn g2_public_key_text_round_trip() {
+        // drand quicknet group public key (G2, 96 bytes)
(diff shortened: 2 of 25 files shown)


a137x:
A fair question was raised: Xi'an will have its own VRF (for shard placement / validator shuffling) — so why verify an external beacon instead of surfacing
  Xi'an's entropy to Scrypto? Having read the [Xi'an proposal](https://hyperscale.rs/xian_proposal.pdf), I think the two are complementary, and there's a strong
  case for landing this now regardless.

  **What Xi'an's beacon is.** Per the proposal, the global consensus layer holds *"the randomness beacon that drives shuffling,"* with validator-to-shard assignment *"determined by publicly verifiable randomness derived from the previous epoch's consensus output"* (the staking model *"draws heavily on MultiversX's Secure Proof of Stake"*). That is a consensus-internal, epoch-grained randomness for *uniform validator shuffling* — exactly what sharding security needs. But:

  - **It isn't scoped to reach Scrypto.** The Radix Engine work in the proposal is sharding adaptation and Babylon feature parity; *"new Scrypto language features"* are explicitly out of scope. Exposing the beacon to contracts would be additional work beyond an already 18-month (mainnet-ready) + 12-month (stabilization) single-developer effort, gated behind a validator/governance enactment vote.
  - **It's internal.** Even if surfaced later, it's Radix-local consensus output — not externally / cross-chain verifiable.
  - **It can't do timelock.** A beacon derived from your own future consensus output can't underpin timelock encryption; tlock requires an external beacon publishing future-round signatures on a known schedule.

  **The present-day gap this fills.** Today there is no verifiable randomness source available to Scrypto, of any kind:

  | Source | Model | Adoption | Verifiable on Radix *today*? | On Radix? |
  |---|---|---|---|---|
  | **drand** (League of Entropy / quicknet) | Push beacon, HTTP, ~3s rounds | De-facto public-randomness standard; verified on ETH mainnet via EIP-2537 | ❌
  quicknet is BLS **min-sig (G1)**; native `bls12381_v1_verify` is **min-pk (G2)/POP** only | Beacon runs externally, free — **this PR adds the missing verify** |
  | **Supra dVRF** | Pull (request/callback) | Several chains | ⚠️  primitive exists (min-pk BLS; the `sign_and_verify_supra` test was its motivating example) | ❌
  **never deployed on Radix** |
  | **Xi'an beacon** | Consensus-internal, epoch-grained (sPoS-style) | Radix-specific | — (not shipped) | ❌ ~2+ yrs out; scoped for shuffling, not
  Scrypto-exposed; no timelock |
  | Chainlink VRF / others | Pull | EVM-wide | ❌ | ❌ |

  until enacted. Nothing changes by default.

  **Why drand specifically:** de-facto standard, externally anchored, and cross-ecosystem-interoperable — [drand quicknet is now verifiable on Ethereum mainnet](https://docs.drand.love/blog/2025/08/26/verifying-bls12-on-ethereum) (Randamu verified round #20986100 in block #23188016 using the [EIP-2537](https://eips.ethereum.org/EIPS/eip-2537) BLS12-381 precompiles). And, uniquely, it enables [**timelock encryption**](https://eprint.iacr.org/2023/189.pdf) — sealed-bid auctions, MEV-resistant ordering, commit-without-reveal. A natural follow-on is a native BLS12-381 **pairing** host fn for full on-chain tlock decryption, which I've assessed as approachable with essentially this PR's engine footprint. None of this competes with Xi'an; it fills a gap Xi'an's shuffling beacon isn't designed to fill.

  **Questions for the maintainers:**

  1. Is the Xi'an beacon BLS-based, and is there any intent to surface it to Scrypto (host fn or readable substate), or is it staying consensus-internal?
  2. Post-Xi'an, is there appetite to verify *external* beacons (drand) for cross-ecosystem interop and timelock — capabilities the shuffling beacon isn't designed for?
  3. Would you consider merging this min-sig verify independently of Xi'an, as the present-day answer for verifiable randomness on Radix? It's small, gated, and disabled-by-default, and it sets up timelock for a later pairing primitive.

